PR #32, opened 2016-01-17
Fix some grammatical issues

wopian:
Updated strings to use better-suited words and modified a few that didn't sound right.


inorichi:
Thanks.

PR #51, opened 2016-01-23
Load the last page when switching to the previous chapter

icewind1991:
When navigating to the previous chapter load the last page instead of the first


Syborg49:
I'd advice against it. Maybe it can be optional, because when I usually revisit the last chapter I want to read it from the start .

j2ghz:
I think making it a setting is the best solution

inorichi:
I think a setting for this is not really needed. The user can always open the menu and slide to the first page. Adding a lot of options is not a good idea either.

For now, it won't be a setting. If it's asked a lot I will consider it.

PR #53, opened 2016-01-23
save per-manga sort order

icewind1991:
Save the sort order for a manga in the sqlite manga table

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.java
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.java
@@ -68,6 +68,10 @@ public class Manga implements Serializable {
     public static final int COMPLETED = 2;
     public static final int LICENSED = 3;
 
+    public static final int SORT_AZ = 0;
+    public static final int SORT_ZA = 1;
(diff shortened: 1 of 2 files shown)


inorichi:
Wait, please use the field `chapter_flags` for that. That was my intention for that field. Don't alter the database scheme.

That field will also store the filters and whatever is added later.

To change a flag, we can use a method like this (taken from the `View` class from Android):

    private void setFlags(int flag, int mask) {
        chapter_flags = (chapter_flags & ~mask) | (flag & mask);
    }

Then we create static fields with the value and the mask for something (again, from the `View` class).

    public static final int VISIBLE = 0x00000000;
    public static final int INVISIBLE = 0x00000004;
    public static final int GONE = 0x00000008;

    static final int VISIBILITY_MASK = 0x0000000C;

And then methods like this would set a flag for a given mask:

    public void setChapterOrder(int order) {
        setFlags(order, ORDER_MASK);
    }

icewind1991:
Ok, I was confusing the chapter_flags with the status field so I thought the chapter flags field wasn't free to use for this.

Adjusted it to use the chapter_flags

inorichi:
Yup, it looks good. Thanks!

PR #74, opened 2016-01-26
Add comments

NoodleMage:
Added comments for cache classes. 
Added TODO to cache classes.
Renamed some methods


app/src/main/java/eu/kanade/tachiyomi/data/cache/CoverCache.java
```diff
@@ -20,33 +20,70 @@
 
 import eu.kanade.tachiyomi.util.DiskUtils;
 
+/**
+ * Class used to create cover cache
+ * Makes us of Glide(which can avoid repeating requests) for saving the file.
```
inorichi: Makes use of

app/src/main/java/eu/kanade/tachiyomi/data/cache/CoverGlideModule.java
```diff
@@ -16,6 +24,6 @@ public void applyOptions(Context context, GlideBuilder builder) {
 
     @Override
     public void registerComponents(Context context, Glide glide) {
-
+        //TODO empty class?
```
inorichi: Yes. There's nothing to add there

app/src/main/java/eu/kanade/tachiyomi/data/cache/ChapterCache.java
```diff
@@ -45,43 +69,67 @@ public ChapterCache(Context context) {
...
         } catch (IOException e) {
-            // Do Nothing.
+            // Do Nothing. TODO error handling?
         }
     }
 
-    public boolean remove(String file) {
+    /**
+     * Remove file from cache.
+     * @param file name of chapter file md5.o.
```
inorichi: .0, not .o

app/src/main/java/eu/kanade/tachiyomi/data/cache/CoverCache.java
```diff
@@ -20,33 +20,70 @@
...
+     * @return true if cache dir exist or is created.
+     */
     private boolean createCacheDir() {
+        // TODO return value never used.
         return !cacheDir.exists() && cacheDir.mkdirs();
     }
 
+    /**
+     * Download the cover with Glide (it can avoid repeating requests) and save the file.
+     * TODO maybe remove and update the call made to this method with ,null?
```
inorichi: Nah, it's fine like this :)

app/src/main/java/eu/kanade/tachiyomi/data/cache/CoverCache.java
```diff
@@ -134,8 +215,13 @@ private void loadFromCache(ImageView imageView, File file) {
...
-    // Helper method to load the cover from network into the specified image view.
-    // It does NOT save the image in cache
+    /**
+     * Helper method to load the cover from network into the specified image view.
+     * It does NOT save the image in cache!
+     * @param imageView imageView where picture should be displayed.
+     * @param thumbnailUrl url of thumbnail.
+     * @param headers headers included in Glide request.
+     */
     public void loadFromNetwork(ImageView imageView, String thumbnailUrl, LazyHeaders headers) {
```
inorichi: While you are at it check for nullability on the thumbnailUrl like in the methods above. It may crash the app

NoodleMage:
What did you mean by nullability then? Because I checked every function and everything is caught as far as I can see. So I presumed you ment the null check before checking an empty string (which apparently online they think is the case. Not gonna check stackoverflow instead of the documentation again xD https://stackoverflow.com/questions/27086808/android-check-null-or-empty-string).

inorichi:
public static boolean isEmpty(@Nullable CharSequence str) {
        if (str == null || str.length() == 0)
            return true;
        else
            return false;
    }

This is the method `TextUtils.isEmpty`. Calling it nullability is  already checked. I meant to call it in the function `loadFromNetwork` 

Also, please squash your commits so that everything is seen together :)

NoodleMage:
Lol. Something went wrong xD. But have to go. Will fix this later today :(

NoodleMage:
Will do squash later :)

PR #82, opened 2016-01-28
Changed the wording on some strings

cyalins:
Fixed grammar issues and reworded some strings for clarity

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -66,7 +66,7 @@
     <string name="landscape">Landscape</string>
     <string name="default_columns">Default</string>
     <string name="pref_library_update_interval">Library update period</string>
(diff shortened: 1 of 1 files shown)


wopian:
Uncomplete is the grammatically correct form.

Incomplete refers to something that isn't completed, with the intention of **never** getting completed.

Uncomplete refers to something that isn't completed, with the intention of getting completed.

JohnFlowerful:
Uncomplete is not a word. Uncompleted is, though.

cyalins:
Oxford and Merriam-Webster just define 'incomplete' as not being finished, so I guess it works in this case

[source](http://www.oxforddictionaries.com/definition/english/incomplete)

wopian:
Better yet, use "Ongoing" to match what all the sources use.

"Only update ongoing manga"

PR #86, opened 2016-01-28
CONTRIBUTING.md

j2ghz:
More ideas welcome


inorichi:
I would also mention that the debug version is equal to the number of commits seen in the main page (its' the easier way to know if you're on the latest build available).

M2ck:
Maybe add something about translation (e.g. link to the strings.xml file).

Issue #79, opened 2016-01-27
Feature request: option to manually set manga covers


PorchettaMan:
With the exception of Batoto, the various manga sources often use wrong or very low quality covers. It's a purely aesthetic thing, but it would be nice to have an option to manually select manga covers from local storage, like MangaWatcher allows to.

Bloodwork:
I noticed this too. Kissmanga has terrible covers even though it's the next best HQ source after Bato.to.

NoodleMage:
http://mcd.iosphe.re/. Maybe we can look at this API.Even if it looks a bit outdated for some manga :).

inorichi:
I forgot to close this. It's already available.

PR #87, opened 2016-01-28
Can now manually set cover pictures. #79

NoodleMage:
ezgif-2841955649


app/src/main/java/eu/kanade/tachiyomi/data/cache/CoverCache.java
```diff
@@ -220,14 +222,15 @@ public void loadFromCacheOrNetwork(ImageView imageView, String thumbnailUrl, Laz
      * @param imageView imageView where picture should be displayed.
      * @param file      file to load. Must exist!.
      */
-    private void loadFromCache(ImageView imageView, File file) {
+    public void loadFromCache(ImageView imageView, File file) {
         Glide.with(context)
                 .load(file)
-                .diskCacheStrategy(DiskCacheStrategy.NONE)
                 .centerCrop()
+                .signature(new StringSignature(String.valueOf(System.currentTimeMillis())))
```
inorichi: Using this, the cache will be invalidated on every call. We should avoid it if possible. I would use a string signature based on the modification date of the file.

PR #90, opened 2016-01-29
Reworded and shortened some strings

cyalins:


wopian:
Maximum simultaneous downloads feels OTT, as simultaneous is already telling you how many it'll download at once.

inorichi:
I think so, too.

PR #98, opened 2016-01-31
Download updates

NoodleMage:
Code that checks for update when version in settings > about is clicked

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/injection/component/AppComponent.java
+++ b/app/src/main/java/eu/kanade/tachiyomi/injection/component/AppComponent.java
@@ -10,6 +10,7 @@ import eu.kanade.tachiyomi.data.mangasync.services.MyAnimeList;
 import eu.kanade.tachiyomi.data.source.base.Source;
 import eu.kanade.tachiyomi.data.sync.LibraryUpdateService;
 import eu.kanade.tachiyomi.data.sync.UpdateMangaSyncService;
+import eu.kanade.tachiyomi.data.updater.UpdateDownloader;
 import eu.kanade.tachiyomi.injection.module.AppModule;
 import eu.kanade.tachiyomi.injection.module.DataModule;
 import eu.kanade.tachiyomi.ui.catalogue.CataloguePresenter;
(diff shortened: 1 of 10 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAboutFragment.java
```diff
@@ -54,4 +68,41 @@ private String getFormattedBuildTime() {
         }
         return "";
     }
+
+    @SuppressWarnings("SameReturnValue") private boolean checkVersion() {
+        //Check for update
+        UpdateChecker updateChecker = new UpdateChecker(getActivity());
```
inorichi: Better use onCreate

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAboutFragment.java
```diff
@@ -54,4 +68,41 @@ private String getFormattedBuildTime() {
...
+        updateChecker.checkForApplicationUpdate()
```
inorichi: This subscription is leaking the fragment. It should be unsubscribed during `onDestroyView` (store a reference to it and call `unsubscribe` if it's not null

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAboutFragment.java
```diff
@@ -54,4 +68,41 @@ private String getFormattedBuildTime() {
...
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(next -> {
+                            //Get version of latest release
```
inorichi: After removing the try/catch this is incorrectly indented

app/src/main/java/eu/kanade/tachiyomi/injection/component/AppComponent.java
```diff
@@ -50,6 +52,8 @@
     void inject(ReaderActivity readerActivity);
     void inject(MangaActivity mangaActivity);
     void inject(SettingsAccountsFragment settingsAccountsFragment);
+
+    void inject(SettingsAboutFragment settingsAboutFragment);
```
inorichi: Not needed anymore

Issue #92, opened 2016-01-29
Zoom start position


inorichi:
Following #85, when the image is cropped in the screen, it's centered.

I think the best alternative is the following:

A configurable start position (maybe a 5th icon in the bottom menu), with multiple options:
 - Automatic (starts on the right for RTL, left for LTR and center for vertical pager)
 - Left
 - Center
 - Right

We also have top, center and bottom, but I think starting always from the top is enough.

Issue #85, opened 2016-01-28
One more zoom type


j2ghz:
I was just thinking there could be one more zoom type
* If width > height  -> fit height
* If height > width -> fit width

Good for not having to pan regular pages while zooming double pages enough to read.

inorichi:
What should it be called? Smart fit?

j2ghz:
Sound good. I don't think there is a name that would be this short and more descriptive

j2ghz:
@inorichi It's great, except it starts in the middle, it should start on the right side for RTL and left for LTR

inorichi:
Maybe configurable, because the resolution for the vertical is unclear.

Issue #67, opened 2016-01-25
Feature request: Fit to width 


higgster:
I read all my manga on my tablet and think this feature is a must for anyone reading in landscape on big screen devices

inorichi:
Duplicate of #40

Issue #40, opened 2016-01-20
Manga Scaling


MakinoTheHermit:
It would be nice to have some sort of scaling when reading manga, like scaling to width so that when i enter a chapter in landscape, i dont have to zoom in every page so that i can read the text.

j2ghz:
I would prefer that to be configurable. I would like it to automatically fit to the smaller of the two dimensions (now it fits the larger), so there are no black bars and I have to scroll only in one direction.

MakinoTheHermit:
Yea, thats how i would like it to be too. Was not specific/clear enough. Thank you for giving a more clear explanation.

inorichi:
Closing this, already implemented.

PR #103, opened 2016-02-02
Added different image fit styles

kollikolli:
As was requested by many people, a setting has been generated to fit an image to the width of the screen(very useful for landscape mode on tablets).
Available settings are: fit image to width, to height, to smaller dimension (smart_fit). 
Related to #40, #67, #85 and #92


inorichi:
I appreciate your effort, but I already did that (even the linked issues are referenced by those commits).

inorichi:
Well, #92 is still open, but I will use `setImage` method for the start position.

Issue #105, opened 2016-02-02
Transparently load the next chapter in the view while reading


icewind1991:
By automatically appending the pages of the next chapter to the list of pages in the reader the chapter boundary becomes transparent and they no longer break the reading flow.

The reader could automatically append the next chapter once you get near to the end of the current chapter and remove the previous chapter once you're a few pages into the next chapter to clean up.

I had a go at this myself and while appending the next chapter worth of pages was easy, getting the reader to properly handle a list of pages that belong to different chapters required to much changes for me to handle (updating the current chapter correctly while scrolling to pages, proper progress bar, etc)

inorichi:
Merged in 2566862

PR #112, opened 2016-02-03
Seamless chapter transition

icewind1991:
A mostly working version of #105, when you near the end of the chapter it appends the pages of the next chapter to the reader to make the chapter transition consistent.

- [x] going back a chapter currently isn't seamless
- [x] total page count isn't always correct (sometimes shows the pages for multiple
- [x] read status isn't updated during a chapter transition
- [x] I think there is some weird flickering sometimes but it's hard to confirm in the emulator
- [x] only normal left-to-right reader is tested atm
- [x] breaks when rotating the screen
- [x] webtoon reader shows the same chapter twice before moving to the next one
- [x] webtoon reader flickers sometimes when adding new pages
- [x] rtl reader is broken once the next chapter is added to the list of pages
- [x] seekbar always goes to the first chapter
- [x] next/previous chapter buttons don't work correctly


inorichi:
The flickering must be caused whenever `onPageListAppendReady` is called. `notifyDataSetChanged` recreates the fragments. It could be solved overriding `getItemId()` in the adapter and generating a random id when a `Page` object is created (the url can be null, so it's better not to use it).

PR #126, opened 2016-02-09
Download features

beschoenen:
I've added a couple of features to the chapters (download all / unread).
I also changed the download button to a pause button (made more sense to me) and added a clear queue button.


app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadFragment.java
```diff
@@ -77,8 +76,13 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
                 DownloadService.start(getActivity());
                 break;
-            case R.id.stop_queue:
+            case R.id.pause_queue:
+                DownloadService.stop(getActivity());
+                break;
+            case R.id.clear_queue: // Not sure if this is correct
                 DownloadService.stop(getActivity());
+                getPresenter().downloadManager.getQueue().clear();
+                onNextDownloads(new ArrayList<>());
```
inorichi: Create a method in the presenter `clearQueue` that calls `downloadQueue.clear();` and `start(GET_DOWNLOAD_QUEUE);` and let him signal the view to refresh. I'm not sure if the `start(x)` is required. Maybe it's automatically refreshed, because it's the same object.

xoryo:
:+1:

PR #163, opened 2016-02-20
[WIP] Added option te delete chapter after reading.

NoodleMage:
Go to Setting > Downloads to toggle: default = false.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.java
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.java
@@ -185,6 +185,14 @@ public class PreferencesHelper {
         return prefs.getBoolean(getKey(R.string.pref_download_only_over_wifi_key), true);
     }
 
+    /**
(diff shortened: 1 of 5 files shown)


LendLord:
Hi. Please add the option to remove penultimate chapter, because often you can accidentally go to the next chapter, but you can't go to previous chapter. This is particularly important for manga like noobless where pages are really long. 
Thanks.

NoodleMage:
@LendLord. What do you exactly mean? It now deletes the chapter after you switch to the next one or when you go back when you're on the last page of a chapter.

Am I correct in assuming that you want an option where the chapter is not deleted after switching to the next chapter but only when you are 2 chapters ahead? Or do you mean there is a different option where you delete all mangas but the last one?

LendLord:
Yes, I mean when you 2 charters ahead, it will be really great to have such option. 
Thanks!

NoodleMage:
Okay. Will see what I can do ;). Thanks for the suggestion ;).

NoodleMage:
Corrupted. Will rewrite + implement request from LendLord

PR #166, opened 2016-02-20
[WIP] Add russian parsers

frenzer:
Added parsers for russian sites: ReadManga, MintManga, Mangachan.
Also renamed catalogs to sort them by language:
(EN) Batoto
(EN) .....
(RU) Mangachan
.....

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/source/online/russian/Mangachan.java
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/source/online/russian/Mangachan.java
@@ -1,0 +1,250 @@
+package eu.kanade.tachiyomi.data.source.online.russian;
+
+import android.content.Context;
+import android.net.Uri;
+
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.nodes.Element;
+import org.jsoup.select.Elements;
+
+import java.text.ParseException;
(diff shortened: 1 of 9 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/source/online/russian/Mangachan.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        return 0;
+    }
+
+    // Return List of IMAGE URLS, NOT pages
+    // to get page url need to parse [div style="height:155px;background-color:#ccc;"] > a
+    // and on the page get the same urls but take only first one
+    @Override
+    protected List<String> parseHtmlToPageUrls(String unparsedHtml) {
```
inorichi: If all the image urls can be obtained from the first page, you can just return a list of empty string here, as long as the array has the same size than the number of images.

Then override `parseFirstPage` and assign each image url to each page with a loop. You can use Kissmanga as a reference for this. It will avoid unneeded requests and parsing.

inorichi:
A few notes:

- `SourceManager` class has been implemented in Kotlin and it's going to cause conflicts. It's easy to fix.
- The `getId` function will be removed, it's now assigned in its constructor from the manager.
- If the source list is going to grow, there should be a way to enable/disable specific sources. I will try to do this.

PR #171, opened 2016-02-21
Implement parsers for Readmanga, Mintmanga and Mangachan

Taumer:
Add parsers for 
http://readmanga.me/
http://mintmanga.com/
http://mangachan.ru/

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/source/SourceManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/source/SourceManager.kt
@@ -6,6 +6,9 @@ import eu.kanade.tachiyomi.data.source.online.english.Batoto
 import eu.kanade.tachiyomi.data.source.online.english.Kissmanga
 import eu.kanade.tachiyomi.data.source.online.english.Mangafox
 import eu.kanade.tachiyomi.data.source.online.english.Mangahere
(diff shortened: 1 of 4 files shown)


LendLord:
эхх, когда же сольют парсеры...
@Taumer, а такой еще вопрос планируешь ли в будущем добавлять еще парсеры?

inorichi:
I will add an overridable method in the source to add custom parsing of the chapter. Doing it like that is unnecessary. I also have to make the sources language aware, or allow to enable/disable them individually.

Taumer:
@LendLord, в ближайшее время - нет.

> I will add an overridable method in the source to add custom parsing of the chapter

@inorichi, thanks, I really don't like my approach to this. I'll rewrite this part when you add method

inorichi:
https://github.com/inorichi/tachiyomi/commit/d4bb092543e030ddf89677499a65a668c0f95f7d#diff-d64413e9693a4eff5305fc2472fb1b6e

Override that method to have custom parsing. Keep in mind the property `url` of the chapter doesn't have the domain name at that point, so it will be something like: `/my_manga/56.html` or similar.

inorichi:
Can you implement the new `getLang` method? It has to return `LanguageKt.getRU()`. Also remove ` (RU)` from `NAME`, it's not needed anymore.

When you're done, can you squash all the commits into a single one?

Taumer:
@inorichi, done.

PR #178, opened 2016-02-22
Category rewrite + FAB rewrite to Kotlin

NoodleMage:
Rewrote ScrollAwareFABBehavior.java to Kotlin. Can now implement FABAnimationBase to create different FAB animations


NoodleMage:
ezgif-429140755

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryActivity.kt
```diff
@@ -0,0 +1,273 @@
...
+        setContentView(R.layout.activity_edit_categories)
+
+        // Setup the toolbar.
+        setupToolbar(findViewById(R.id.toolbar) as Toolbar)
+
+        // Get new adapter.
+        adapter = CategoryAdapter(this)
+
+        // Create view and inject category items into view
+        val recycler = findViewById(R.id.categories_list) as RecyclerView?
```
inorichi: Either use `categories_list` instead of `recycler` or rename the id to `recycler`. No need to `findViewById`, Also, it won't be null.

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryActivity.kt
```diff
@@ -0,0 +1,273 @@
...
+                    .show()
+        })
+    }
+
+    /**
+     * Finishes action mode.
+     * Call this when action mode action is finished.
+     */
+    fun destroyActionModeIfNeeded() {
+        if (actionMode != null)
```
inorichi: No need for the `if` :P

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryActivity.kt
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /**
+     * Returns the selected categories
+     *
+     * @return list of selected categories
+     */
+    private fun getSelectedCategories(): MutableList<Category?>? {
+        // Create a blocking copy of the selected categories
+        return Observable.from<Int>(adapter.selectedItems).map({ adapter.getItem(it) }).toList().toBlocking().single()
```
inorichi: With Kotlin's collections methods we don't need RxJava for this. \o/
`adapter.selectedItems.map { adapter.getItem(it) }` should be enough.

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryActivity.kt
```diff
@@ -0,0 +1,273 @@
...
+
+        // Get selected item count
+        val count = adapter.selectedItemCount
+
+        // If no item is selected finish action mode
+        if (count == 0) {
+            actionMode?.finish()
+        } else {
+            // Set title equal to selected item
+            actionMode?.title = getString(R.string.label_selected, count)
```
inorichi: If you want (I let you choose), you can use `let` to not add always the `?` mark. Something like:

```
actionMode?.let {
    // This block will only run if actionMode is not null

    it.title = getString(R.string.label_selected, count)
    it.invalidate()

    val editItem = it.menu?.findItem(R.id.action_edit)
    editItem?.isVisible = count == 1
}
```

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryActivity.kt
```diff
@@ -0,0 +1,273 @@
...
+            actionMode?.invalidate()
+
+            // Show edit button only when one item is selected
+            val editItem = actionMode?.menu?.findItem(R.id.action_edit)
+            editItem?.isVisible = count == 1
+        }
+    }
+
+    /**
+     * Prevent action mode from being destroyed
```
inorichi: I'm not sure this is for preventing being destroyed. I think this is called after `onCreateActionMode`, but it's not used.

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryAdapter.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Called when ViewHolder is created
+     *
+     * @param parent parent View
+     * @param viewType int containing viewType
+     */
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): CategoryHolder {
+        // Inflate layout with item_edit_categories.xml
+        val inflater = activity.layoutInflater
```
inorichi: Extensions functions are awesome. https://github.com/inorichi/tachiyomi/blob/library-kotlin/app/src/main/java/eu/kanade/tachiyomi/util/ViewGroupExtensions.kt

That file is in my PR for the library, you can copy it if you want or I will change it later.

With that, you can just use `val view = parent.inflate(R.layout.item_edit_categories)`. Concise and clean :)

PR #183, opened 2016-02-25
Rewrote IOHandler to Kotlin

NoodleMage:


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
```diff
@@ -320,11 +320,11 @@ class LibraryFragment : BaseRxFragment<LibraryPresenter>(), ActionMode.Callback
                 val inputStream = context.contentResolver.openInputStream(data.data)
 
                 // Convert to absolute path to prevent FileNotFoundException
-                val result = IOHandler.downloadMediaAndReturnPath(inputStream as FileInputStream,
+                val result = IOHandler().downloadMediaAndReturnPath(inputStream as FileInputStream,
```
inorichi: If `getTempFileName` can throw an `IOException`, this method does too, so include this in the try block. And the inputstream above should probably be closed (and included in the try/catch).

inorichi: Also, make the methods static by declaring the functions at package level, if it's needed to call this function from Java it can be called with `IOHandlerKt` as the class name http://kotlinlang.org/docs/reference/java-to-kotlin-interop.html

NoodleMage: Why do you want this? Everything is rewritten to Kotlin so no java calls.

inorichi: I thought this was still used in the info fragment. Nevermind then (it can still be called from Java though :P).

PR #190, opened 2016-03-01
Rewrote Recent to Kotlin

NoodleMage:
Code review is welcomed :+1:


app/src/main/java/eu/kanade/tachiyomi/ui/recent/RecentChaptersAdapter.kt
```diff
@@ -0,0 +1,134 @@
...
+     * Called when ViewHolder is bind
+     *
+     * @param holder bind holder
+     * @param position position of holder
+     */
+    override fun onBindViewHolder(holder: RecyclerView.ViewHolder, position: Int) {
+        // Check which view type and set correct values.
+        when (holder.itemViewType) {
+            VIEW_TYPE_CHAPTER -> {
+                if (getItem(position) is MangaChapter) {
```
inorichi: The type of the view holder should be enough, I don't see a need for the `if`.

app/src/main/java/eu/kanade/tachiyomi/ui/recent/RecentChaptersAdapter.kt
```diff
@@ -0,0 +1,134 @@
...
+     * @param parent parent View
+     * @param viewType int containing viewType
+     */
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): RecyclerView.ViewHolder? {
+        val view: View
+
+        // Check which view type and set correct values.
+        when (viewType) {
+            VIEW_TYPE_CHAPTER -> {
+                view = parent.inflate(R.layout.item_recent_chapter, false)
```
inorichi: With default parameters there's no need for `false`, it's already false.

app/src/main/java/eu/kanade/tachiyomi/ui/recent/RecentChaptersHolder.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Color of unread chapter
+     */
+    private val unreadColor: Int
+
+    /**
+     * Adapter containing the recent chapters.
+     */
+    private val adapter = adapter
```
inorichi: You can delete this line and use 
`class RecentChaptersHolder(view: View, private val adapter: RecentChaptersAdapter, listener: FlexibleViewHolder.OnListItemClickListener) : FlexibleViewHolder(view, adapter, listener) {`

Do that also for the `view` in the constructor and then you can change every `itemView` with just `view`. A little more short

inorichi:
Reviewed. It looks good to me :+1:

app/src/main/java/eu/kanade/tachiyomi/ui/recent/RecentChaptersAdapter.kt
NoodleMage: Want to keep it nonetheless. I added this because I got a wrong class exception on date when I made a mistake in the observable. I know it probably won't happen for the MangaChapter object but better save then sorry in my opinion :)

inorichi: Ok. Then store a reference to `getItem(position)` before the `when`:
`val item = getItem(position)`

When `if (item is MangaChapter)` is called, `item` will be a `MangaChapter` object inside that block so that you don't have to cast it. Smart cast ftw.

Issue #27, opened 2016-01-15
Feature Request: view for downloaded chapters


Iliansic:
Could you please add view option, where we can see all series with currently downloaded chapters, like in Manga Rock, cause it's kinda getting hard to track them all after some time.

j2ghz:
Feature proposal: In the future this could be replaced by configurable and saveable (multiple presets) filters?

Filter: Next chapter downloaded, All remainig chapters downloaded, MAL status finished/publishing
Sort: Date updated, Date last read, Remaining chapters, % Completed

Issue #196, opened 2016-03-03
[Bug] Recent chapters is mixed up


BlackSynder:
Recent updates is mixed up, showing 5days ago at the top followed by 2 days ago and today is near the bottom.
---
Version: r528 (March 3rd)

PR #198, opened 2016-03-04
Added filter options. fix for Issue #27

NoodleMage:
- Filter Downloaded
- Filter Unread
- Remove filters

ezgif-429140755


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -91,6 +95,7 @@ class LibraryPresenter : BasePresenter<LibraryFragment>() {
 
     override fun onTakeView(libraryFragment: LibraryFragment) {
         super.onTakeView(libraryFragment)
+        this.libraryFragment = libraryFragment
```
inorichi: We can't have references to the view in the presenter.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
```diff
@@ -104,6 +115,8 @@ class LibraryFragment : BaseRxFragment<LibraryPresenter>(), ActionMode.Callback
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setHasOptionsMenu(true)
+        isFilterDownloaded = presenter.preferences.filterDownloaded().get() as Boolean
```
inorichi: I wouldn't turn this into a preference. I don't think it's needed, you can use the presenter to save them and store their value in a bundle if needed.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
NoodleMage: woops. Forgot to remove that one :). It isn't used btw.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
NoodleMage: I want to remember the filter on startup (convenient for unread filter). That doesn't work if it's stored via the presenter. Or does it?

inorichi: In that case it's needed to use the preferences.

inorichi:
PTAL https://github.com/inorichi/tachiyomi/commit/33386e2ff949e6920fe45408e18e9fad0d392ec6 ;)

sm7297:
will this be added to next version cool features

NoodleMage:
@inorichi will look at it tonight ;). And ty

@sm7297 when this pull is merged it will be added to the release and debug. So after some more tweaking, yes :)

sm7297:
awesome other fdroid app having it own problem while updating repos i have to download directly from github

sm7297:
can i be of any help if i learn the programming language in which this app is written for adding different catalogues?
if so which language it is
and which software to be used

NoodleMage:
@sm7297 Help is always welcomed :). Bit of a disclaimer though. I don't know how experienced you are with programming in general? I know you have to start somewhere but working on Tachiyomi is a bit advanced if you still have to start :).

If you are not afraid:) and want to help I recommend you learn Kotlin (we are rewriting everything to Kotlin so java will become obsolete for the most part) and a bit of java in combination with JavaRx .
Also you should get accustomed with Scala because otherwise the code will be very confusing :)

We use Android Studio 2 Beta 6 http://tools.android.com/download/studio/beta.

I hope this is the information you need :). If you have more question please pm me because I don't want this pull request to get to crowded with messages xD.

NoodleMage:
ezgif-4127530487

Now with super filter speed!

@inorichi  Should I remove the filter message now that it's this quick?

inorichi:
Sure. It shouldn't take too long even with a big library.

BTW, can you replace the drawable with a vector one? I replaced some already here 8c456a2. Android Studio has a "Vector Asset" tool to add them. It's a bit annoying it doesn't have a search box though (or I don't know where it's).

NoodleMage:
@inorichi  done :+1:

inorichi:
The workaround for tabs makes me sad. :(

Regarding the programming language in Tachiyomi, I would say you are fine with just Java. If you know Java, switching to Kotlin is very easy, because it's the same but with some additions and a better syntax.
Scala is not needed (I never used Scala).

The worst part is RxJava. I had to do a lot of trial and error and it's quite difficult to learn, but worth it in my opinion. Though you don't need to know RxJava to implement parsers.

inorichi:
I forgot to mention that if you remove a manga from the library, it won't appear in the downloaded filter.

A possible solution is asking to delete downloaded chapters when you remove it from the library.

NoodleMage:
@inorichi @sm7297  Scala should be lambda. Didn't have my evening caffeine yet I suppose :o.

@inorichi We should implement something like that yes.I got 200mb+ of read chapter files xD.  It will get better when "remove after read" is implemented :).

sm7297:
I have some knowledge of java topic like io, enum, exception handling, serialization, operators and in collections framework arraylist other than this which topic should i learn and can i directly implement it in android

NoodleMage:
@sm7297  I think if you read inorichi's and my comments you have a good idea on which knowledge you need to help out. Like inorichi told JavaRx is probably the most difficult things to understand (even I make some mistakes with it. Issue #196 uch... uch...). If you have any more questions please contact me on noodlemagedev@gmail.com.

PR #204, opened 2016-03-07
Converted all icon drawables to vector + Removed Android-Iconics library.

NoodleMage:
download


NoodleMage:
download

inorichi:
Such a repetitive job :(

From 4.52MB to 4.25MB, it's something.

NoodleMage:
@inorichi Yeah I know. 0.27 MB but still better then nothing ;). The funny thing is that the most time consuming action was getting the correct grey values. Thought that alpha was 0-255 but apparently in the vector case it's 0-1 ;(. But grey colors are know on par with the Google guidelines. So yeah xD

inorichi:
I see you are using "#FF000000" with alpha ".54". 

[Here](https://www.google.com/design/spec/style/color.html#color-themes) grey 600 is "#757575"

I don't know if there's any difference.

Have you tried setting the alpha in the first two "FF"? There you can use 0-255, but I don't know if you have to use the alpha property or not.

NoodleMage:
screenshot_20160308-011604

From same page :). Did my research this time xD. I think it's possible if you change the FF. But this was a bit easier and I don't know how accurate hex alpha is.

inorichi:
Ok. I was fine with the current implementation. Just pointing out my thoughts/doubts.

PR #208, opened 2016-03-08
Rewrote nav drawer to Kotlin + UI updates

NoodleMage:
## New navigation bar
screener_20160309 112148

## Information views
screener_20160308 21_25_13

screener_20160308 23-37-56

screener_20160308 23-38-34

## Launch Screen
screener_20160309 174029


NoodleMage:
# Updates [WIP]
## New navigation bar
screener_20160308 21_24_55

## Information view
screener_20160308 21_25_13

NoodleMage:
If anyone is interested in designing a material header for the navigation drawer please let me know ;-)

M2ck:
About navigation drawer, here is my suggestions: 
- Reduce header
- Use symbol from icon app in header
- Black icons

a00ebb16-e5e9-11e5-850d-6677f073cfcf

NoodleMage:
@m2ck thanks for the feedback.

I do think the header looks nice the way you designed it. But with your header it will be impossible to implement correct account view + ignores Android guidelines.

But as I said it does look nice. So if more people want it I can change the height. 

The logo with plain color looks nice. Will try to implement this and see how it looks with higher header. Still prefer an image so if anyone is interested ;-).

As for the Icons. I don't think black looks better than grey and it ignores the guidelines. So will not change it.

app/src/main/java/eu/kanade/tachiyomi/util/ImageViewExtensions.kt
```diff
@@ -0,0 +1,19 @@
...
+import android.widget.ImageView
+import org.jetbrains.annotations.Nullable
+
+/**
+ * Set information image
+ *
+ * @param drawable id of image
+ * @null makes image transparent
+ */
+fun ImageView.setInformationDrawable(@Nullable drawable: Int?) {
```
inorichi: how about `setDrawableCompat`? So that we can use this for everything instead of manually calling `ContextCompat` every time.

inorichi: And the `@Nullable` property shouldn't be needed, you are already using `?`, which means it's nullable.

NoodleMage:
Working on rewrite of navigation drawer + cleanup of theme values. Will open new pull request when done.

PR #218, opened 2016-03-14
Rewrote Nav Drawer to Kotlin + Dark Theme

NoodleMage:
## Light Theme
screener_20160317 200711
screener_20160317 092229
screener_20160314 220500

## Settings Option
screener_20160317 092238

## Dark Theme
screener_20160317 201154
screener_20160317 092334
screener_20160314 220452


NoodleMage:
Tested for SDK 16 / 21 / Android N.

inorichi:
"My library" is a bit unreadable in the picture with the black theme IMO.

NoodleMage:
I agree. But this is because the primary color isn't material.  Thus the only way around this is using a different primary color for the dark theme.

inorichi:
Okay. Any particular reason to remove `app/src/main/res/drawable-xhdpi/card_background.9.png`? It provides a grey shadow to the covers. It also looks fine in the dark theme.

NoodleMage:
I thought the background is white in dark theme when I use card_background. Can't test it at this time

inorichi:
Not very noticeable but it seems to look better:

NoodleMage:
Does the ripple effect on touch still work? If so will change it back then ;)

inorichi:
Yep, it works.

inorichi:
Let me do some review/changes. Then just pull my changes to your branch if they are fine.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerReaderFragment.kt
```diff
@@ -74,12 +74,12 @@ class PagerReaderFragment : BaseFragment() {
     /**
      * Text color for black theme.
      */
-    private val lightGreyColor by lazy { ContextCompat.getColor(context, R.color.light_grey) }
+    private val lightGreyColor by lazy { ContextCompat.getColor(context, android.R.attr.textColorHint) }
```
inorichi: This line makes the reader to crash when it's opened (only when using black background).

inorichi:
The only issue I see now is the drawer items colors and the text in the dark theme. I think we should fix this before merging. Also pull my changes from `theme_update2` when you can.

NoodleMage:
Pulled. Maybe we can use #607D8B as primary for the dark theme? What do you mean by text?

inorichi:
Same as my first comment: "My library" is a bit unreadable in the picture with the black theme IMO.

app/src/main/res/values/attrs.xml
```diff
@@ -20,4 +20,6 @@
 
     <attr name="selectable_list_drawable" format="reference|integer" />
     <attr name="divider_drawable" format="reference|integer" />
+    <attr name="status_bar_trans" format="reference|color"/>
+    colorPrimaryDarkTrans
```
inorichi: Uh... and no warning? :)

NoodleMage: Nope. Themes.xml just rolls with it xD

PR #225, opened 2016-03-19
Updated manga chapters UI

NoodleMage:
## UI
screener_20160318 001502

## Animation
ezgif-2349610590

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
@@ -1,7 +1,6 @@
 package eu.kanade.tachiyomi.ui.manga.chapter
 
 import android.os.Bundle
-import android.support.v4.content.ContextCompat
 import android.support.v7.view.ActionMode
 import android.support.v7.widget.LinearLayoutManager
 import android.view.*
@@ -71,7 +70,7 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
 
         swipe_refresh.setOnRefreshListener { fetchChapters() }
 
-        next_unread_btn.setOnClickListener { v ->
+        fab.setOnClickListener { v ->
             val chapter = presenter.getNextUnreadChapter()
             if (chapter != null) {
                 openChapter(chapter)
@@ -93,35 +92,37 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
 
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
         inflater.inflate(R.menu.chapters, menu)
+        menu.findItem(R.id.action_filter_unread).isChecked = presenter.onlyUnread()
+        menu.findItem(R.id.action_filter_downloaded).isChecked = presenter.onlyDownloaded()
     }
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         when (item.itemId) {
             R.id.action_display_mode -> showDisplayModeDialog()
             R.id.manga_download -> showDownloadDialog()
+            R.id.action_filter_unread -> {
+                item.isChecked = !item.isChecked
+                presenter.setReadFilter(item.isChecked)
+            }
+            R.id.action_filter_downloaded -> {
+                item.isChecked = !item.isChecked
+                presenter.setDownloadedFilter(item.isChecked)
+            }
+            R.id.action_filter_empty -> {
+                presenter.setReadFilter(false)
+                presenter.setDownloadedFilter(false)
+                activity.supportInvalidateOptionsMenu();
+            }
+            R.id.action_sort -> presenter.revertSortOrder()
             else -> return super.onOptionsItemSelected(item)
         }
         return true
     }
 
     fun onNextManga(manga: Manga) {
-        // Remove listeners before setting the values
-        show_unread.setOnCheckedChangeListener(null)
-        show_downloaded.setOnCheckedChangeListener(null)
-        sort_btn.setOnClickListener(null)
-
         // Set initial values
         setReadFilter()
         setDownloadedFilter()
-        setSortIcon()
-
-        // Init listeners
-        show_unread.setOnCheckedChangeListener { arg, isChecked -> presenter.setReadFilter(isChecked) }
-        show_downloaded.setOnCheckedChangeListener { v, isChecked -> presenter.setDownloadedFilter(isChecked) }
-        sort_btn.setOnClickListener {
-            presenter.revertSortOrder()
-            setSortIcon()
-        }
     }
 
     fun onNextChapters(chapters: List<Chapter>) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/eu/kanade/tachiyomi/util/ViewExtensions.kt
```diff
@@ -0,0 +1,86 @@
...
+}
+
+
+/**
+ * Returns coordinates of view.
+ * Used for animation
+ *
+ * @return coordinates of view
+ */
+fun View.getCoordinates() : IntArray
```
inorichi: This could return a `android.graphics.Point` instead:
`fun View.getCoordinates() = Point((left + right) / 2, (top + bottom) /2)`

(It's valid code btw, return type is inferred)

app/src/main/java/eu/kanade/tachiyomi/util/ViewExtensions.kt
```diff
@@ -0,0 +1,86 @@
...
+import android.view.ViewAnimationUtils
+
+/**
+ * Hides the animation view with a animation
+ *
+ * @param centerX x starting point
+ * @param centerY y starting point
+ * @param initialRadius size of radius of animation
+ */
+fun View.hideRevealEffect(centerX: Int, centerY: Int, initialRadius: Int) {
```
inorichi: I'm wondering if this should be a custom widget instead. What do you think? Would it be better to subclass `View` or have this extension method for every view? If it won't be widely used I think it's better a widget

NoodleMage: To be honest. Didn't like it like this either. Widget is a great idea. Will look into it when I have some free time ;)

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
```diff
@@ -71,10 +72,23 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
 
         swipe_refresh.setOnRefreshListener { fetchChapters() }
 
-        next_unread_btn.setOnClickListener { v ->
+        fab.setOnClickListener { v ->
             val chapter = presenter.getNextUnreadChapter()
             if (chapter != null) {
-                openChapter(chapter)
+                // Create animation listener
+                var revealAnimationListener: Animator.AnimatorListener = object : AnimatorListenerAdapter() {
```
inorichi: `val`! Is there any difference between `AnimatorListener` and `Animation.AnimationListener`? I created a simple implementation here https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/base/listener/SimpleAnimationListener.kt . I'm using this to hide views when the animation ends (somewhere in `ReaderActivity`, when toggling the menu), maybe it can be reused.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
```diff
@@ -71,10 +72,23 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
...
+                    override fun onAnimationEnd(animation: Animator) {
+                        // On done open chapter
+                        openChapter(chapter, true)
+                    }
+                }
+
+                // Get coordinates and start animation
+                var coordinates = fab.getCoordinates()
```
inorichi: `val` too. Also opening bracket on same line as `if`

inorichi:
Looks fine. :+1: 

I don't like the FAB though, it hides the downloaded text and the popup menu on a certain area. In my opinion that is killing usability. Sure, you can scroll down to hide it but:
1) It requires you to make an extra action to open a menu -> annoying
2) If the list isn't scrollable you have to resort to chapter selection with long click or even screen rotation -> annoying

NoodleMage:
I know you don't like fab. But I agree with Antiligh when he said "Design is one of the most important aspects of app development. A Fab is a great addition and improvement to that, making an app almost always look vastly better and thought out. "

I do think this way it looks way better. To counter you're complain the bottom chapter wasn't even visible with the previous bar so you still had to scroll ;).

Thank you for the code review. Will try to implement it tonight

inorichi:
Fair reply against the first point. For the second, imagine in your capture you have 8 chapters. You can't scroll and the FAB doesn't hide (or am I wrong? I haven't tried). This could happen 1 of 200 times, but it's still possible. I don't know if there's a way to hide the FAB even if the content is not scrollable.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
```diff
@@ -158,9 +184,13 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
     val isCatalogueManga: Boolean
         get() = (activity as MangaActivity).isCatalogueManga
 
-    protected fun openChapter(chapter: Chapter) {
+    protected fun openChapter(chapter: Chapter, hasAnimation: Boolean = false) {
         presenter.onOpenChapter(chapter)
         val intent = ReaderActivity.newIntent(activity)
+        if (hasAnimation)
```
inorichi: Brackets ;)

NoodleMage:
In this case you're completely right ;). 

Some work arounds are:
1) remove 3 dot pop-up and only use long press. (Dislike + no previous chapters as read option)
2) remove fab on long press (realy dislike)
3) live with the 1 200 (probably smaller) change

inorichi:
4) Hide it with vertical scrolling even if it can't scroll.

I think this one is better than those 3. But maybe not easy to implement.

LendLord:
I have problems with chapter length, in some mangas chapter number in the middle of name, so it isn't possible to understand what chapter is it, may be it is possible to make font a little smaller or when chapter name too long show it in two rows?
Also after sort and filter buttons moved to top, manga name have the same problem, but it isn't critical at all. 

screenshot_20160323-065743

NoodleMage:
@LendLord. Yeah. My mistake. Keep forgetting that not everyone has a 5.7" sceeen xD. Will test ui changes on my old moto g (720p 4.5") from now on. Thanks for the feedback will fix it when I have some free time ;)

LendLord:
@NoodleMage. Thanks, very much appreciated!

inorichi:
There's a setting in the chapters tab to display the chapter number instead of the title

LendLord:
@inorichi. Yeah, but almost all chapters have the same name.
I know its a problem from parser...
screenshot_20160323-132657

Taumer:
> I know its a problem from parser

@LendLord, это было сделано для того, чтобы сохранить правильный порядок сортировки глав.
This was done to correct sorting chapters

LendLord:
@Taumer да, я помню, ты писал об этом.

NoodleMage:
@LendLord  Made it the same as recent. It's still not optimal, but it's something with the parser so can't fix that ;).

screenshot_20160323-174002
Better?.

LendLord:
Better=) But sometimes still no chapter number,  I think the only way here is to make it in two rows, but it can be ugly... anyway thanks for this, I think I can live with that now=)

NoodleMage:
@LendLord Can't argue with good idea's :+1: 
screenshot_20160323-182439

It looks good to me. And it's only on longer titles. @inorichi  opinion?

LendLord:
Wow, I think it's great!!=))
Thanks!=))

inorichi:
:+1:

Issue #42, opened 2016-01-20
Auto download, delete


j2ghz:
Settings:
* [ ] How many next chapters to download upon reading a chapter
* [ ] Delete chapter when set as read

Menu in Library > Chapter list
* [ ] Download next 1/5/10/all chapters

BlackSynder:
So is this happening? Would be amazing if it is. If you turn off downloaded and turn on unread, the second you finish a chapter, it dissapears and its hard to keep track of whats downloaded.

jassibacha:
The "Download Next 1 / 5 / 10 / All" is an extremely useful thing to add, I didn't realize how much I used it until now.

PR #227, opened 2016-03-21
Fix Issue #42

NoodleMage:

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadsFragment.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadsFragment.kt
@@ -20,7 +20,6 @@ import java.io.File
 class SettingsDownloadsFragment : SettingsNestedFragment() {
 
     val downloadDirPref by lazy { findPreference(getString(R.string.pref_download_directory_key)) }
-
     companion object {
 
         val DOWNLOAD_DIR_CODE = 103
--- a/app/src/main/res/xml/pref_downloads.xml
+++ b/app/src/main/res/xml/pref_downloads.xml
@@ -3,26 +3,41 @@
     xmlns:android="http://schemas.android.com/apk/res/android">
 
     <Preference
-        android:title="@string/pref_download_directory"
-        android:key="@string/pref_download_directory_key"/>
+        android:key="@string/pref_download_directory_key"
+        android:title="@string/pref_download_directory"/>
 
     <SwitchPreferenceCompat
-        android:title="@string/pref_download_only_over_wifi"
+        android:defaultValue="true"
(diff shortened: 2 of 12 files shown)


inorichi:
I'm against a setting for every viewer. This setting should be either enabled or disabled (globally).

inorichi:
I see a lot of problems with automatically downloading a chapter from the reader. First of all it slows the active chapter (because it starts to download the next one almost at the same time). It also clashes with the preloader, so we're wasting resources (repeated image downloads). I don't see any advantage for this ATM, the preloader is always enabled for the next chapter, so the transition shouldn't be noticeable and it doesn't affect the loading of the active chapter.

NoodleMage:
@inorichi Agree with with automatically downloading. Didn't know it was already pre-loading. Shall remove it from commit then ;). Will also make previous chapter global. later today... ;)

j2ghz:
So no auto download? How about stopping download queue when preloader is downloading? Or making preloader save what it downloads and adding a setting for how much to preload (e.g. 10 chapters)

inorichi:
It adds too much complexity with the current implementation. For something like this, both preloader and downloader should work as one and it would mean a lot of changes, I believe.

And there would be almost no difference between autodownloading and the current preload implementation (unless you want to keep chapters downloaded forever). Next time you open the next chapter, it should continue where it was (unless the disk cache removes them because it needs space).

It's possible, but I think it's not worth the time. Manually using the download next 1/5/10 chapters is already enough IMO.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadsFragment.kt
```diff
@@ -33,6 +34,12 @@ class SettingsDownloadsFragment : SettingsNestedFragment() {
     }
 
     override fun onViewCreated(view: View, savedState: Bundle?) {
+
+        removeAfterReadPref.setOnPreferenceChangeListener { preference, any ->
+            // Enable previous when remove after read is enabled
+            removeAfterReadPreviousPref.isEnabled = !(preferences.removeAfterRead())
```
inorichi: Use `android:dependency` for this in the XML

app/src/main/res/xml/pref_downloads.xml
```diff
@@ -3,20 +3,41 @@
...
         android:entryValues="@array/download_slots"
-        android:defaultValue="1"
-        android:summary="%s"/>
+        android:key="@string/pref_download_slots_key"
+        android:summary="%s"
+        android:title="@string/pref_download_slots"/>
+
+    <PreferenceCategory
+        android:persistent="false"
+        android:title="Remove After Read"/>
```
inorichi: Make it a string resource, and don't use upper case on every word

PR #229, opened 2016-03-25
Added option to check if connected to power before updating. closes #192

### LINKED ISSUE #192: Library update enhancements
* Cancel update (maybe from notification)
* Update only on AC power

NoodleMage:
screenshot_20160325-144842


app/src/main/java/eu/kanade/tachiyomi/util/PowerUtil.kt
```diff
@@ -0,0 +1,17 @@
+package eu.kanade.tachiyomi.util
+
+import android.content.Context
+import android.content.Intent
+import android.content.IntentFilter
+import android.os.BatteryManager
+
+object PowerUtil {
```
inorichi: Can we merge these two to `DeviceUtil`? Having one class for just one method is awful :(

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -129,6 +126,18 @@ class LibraryUpdateService : Service() {
...
             Timber.i("Sync canceled, connection not available")
+            showWarningNotification(getString(R.string.notification_no_connection_title),
+                    getString(R.string.notification_no_connection_body))
+            AndroidComponentUtil.toggleComponent(this, SyncOnConnectionAvailable::class.java, true)
+            stopSelf(startId)
+            return Service.START_NOT_STICKY
+        }
+
+        // If user doesn't want to update while phone is not charging, cancel sync
+        if (!PowerUtil.isPowerConnected(this) && preferences.updateOnlyWhenCharging()) {
```
inorichi: Check the preference first, then check for connection etc

app/src/main/java/eu/kanade/tachiyomi/util/PowerUtil.kt
NoodleMage: Of course. Only created powerutil object because I thought networkutil was such a specific name that you wanted to keep it separate :)

inorichi: Nice. Also good job using `object`, I didn't know it works like this ;)

NoodleMage: Have to admit that I only discovered it because of the Kotlin convert of Network Util ;).

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -127,13 +127,27 @@ class LibraryUpdateService : Service() {
...
+            showWarningNotification(getString(R.string.notification_no_connection_title),
+                    getString(R.string.notification_no_connection_body))
             AndroidComponentUtil.toggleComponent(this, SyncOnConnectionAvailable::class.java, true)
             stopSelf(startId)
             return Service.START_NOT_STICKY
         }
 
+        // If user doesn't want to update while phone is not charging, cancel sync
+        if (preferences.updateOnlyWhenCharging()) {
+            if (!DeviceUtil.isPowerConnected(this)) {
```
inorichi: I meant  `if (preferences.updateOnlyWhenCharging() && !DeviceUtil.isPowerConnected(this))` The second one does not execute if the preference is disabled

NoodleMage: lol. didn't know that :+1: Will change

inorichi: It also happens with `||`, if the first condition is true, then it doesn't evaluate the others

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -64,20 +64,24 @@ class LibraryUpdateService : Service() {
...
     private lateinit var wakeLock: PowerManager.WakeLock
 
+
     // Subscription where the update is done.
     private var subscription: Subscription? = null
 
     companion object {
         val UPDATE_NOTIFICATION_ID = 1
 
+        var isForced: Boolean = false
```
inorichi: This should be passed as parameter to `getIntent` and added to the intent with `putExtra`, something like: 

```
fun getIntent(context: Context, isForced: Boolean = false): Intent {
    return Intent(context, LibraryUpdateService::class.java).apply { 
        putExtra("forced", isForced)
    }
}
```

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -142,7 +145,7 @@ class LibraryUpdateService : Service() {
         }
 
         // If user doesn't want to update while phone is not charging, cancel sync
-        else if (preferences.updateOnlyWhenCharging() && !isForced && !DeviceUtil.isPowerConnected(this)) {
+        else if (preferences.updateOnlyWhenCharging() && !((intent?.getBooleanExtra(UPDATE_IS_FORCED, false)) as Boolean) && !DeviceUtil.isPowerConnected(this)) {
```
inorichi: `!(intent?.getBooleanExtra(UPDATE_IS_FORCED, false) ?: false)` safe null pointer ;)

inorichi: On a side note, the intent can be null, (I've received crashes because the parameter wasn't nullable), but maybe it's because I'm using wrong return values (there's `STICKY`, `NOT_STICKY`, `REDELIVER`...).

I read a while ago the meaning of those and I didn't understand everything, so...

NoodleMage: It's a null safe call on the intent so it shouldn't be a problem?

inorichi: Yes, it won't be a problem, but it was a note to answer why the intent is nullable.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -145,13 +167,24 @@ class LibraryUpdateService : Service() {
                             showNotification(getString(R.string.notification_update_error), "")
                             stopSelf(startId)
                         }, {
-                            stopSelf(startId)
-                        })
+                    stopSelf(startId)
```
inorichi: Wrong indentation?

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -145,13 +167,24 @@ class LibraryUpdateService : Service() {
...
     }
 
     /**
+     * Creates a PendingIntent for LibraryUpdate broadcast class
+     * @param action id of action
+     */
+    fun getLibraryUpdateReceiverIntent(action: String): PendingIntent {
+        val intent = Intent(this, LibraryUpdateReceiver::class.java)
+        intent.action = action
+        return PendingIntent.getBroadcast(this, 0,
```
inorichi: Merge these lines

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
```diff
@@ -207,7 +207,8 @@ class LibraryFragment : BaseRxFragment<LibraryPresenter>(), ActionMode.Callback
                 // Apply filter
                 onFilterCheckboxChanged()
             }
-            R.id.action_refresh -> LibraryUpdateService.start(activity)
+        // Force refresh
```
inorichi: Indentation

inorichi:
These little changes and done :+1:

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
NoodleMage: Do you mean 
```
 fun getLibraryUpdateReceiverIntent(action: String): PendingIntent {
        return PendingIntent.getBroadcast(this, 0,
                Intent(this, LibraryUpdateReceiver::class.java).apply { this.action = action }, 0)
    }
```

Or did you have something else in mind?

inorichi: `return PendingIntent.getBroadcast(this, 0, intent, 0)`

instead of 

```
return PendingIntent.getBroadcast(this, 0,
                intent, 0)
```

Text isn't that long to split them :)

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
NoodleMage: It's the same as master. An this is how Android Studio indents it. So will keep it like this ;)

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
NoodleMage: I understand. Think apply looks a bit cleaner. and it removes one line. So will use that ;). 
And people say misunderstandings are bad xD

NoodleMage:
And squashed. As always thanks for the feedback ;)

PR #239, opened 2016-04-04
Rewrite source

inorichi:
Now the app can read custom sites from a YAML file:
[Example for mangafox](https://github.com/inorichi/tachiyomi/files/201654/mangafox.txt)
[Example for mangahere](https://github.com/inorichi/tachiyomi/files/201655/mangahere.txt)

They have to be placed under `/sdcard/Tachiyomi/parsers/[site].yml` (It's important to have the `.yml` extension). 

Most sites could use this method instead of hardcoding them in the app, but they are very limited at the moment (no support for login, custom chapter recognition, etc...)

Old sources have to be implemented with the new source, either as a YAML file or hardcoded, and if possible with the former (like mangafox and mangahere).

Note the YAML parser is probably not finished and it can change at any moment. I will improve/document it these days.


beepFTW:
Would this make it easier to implement latest chapter updates from sources too?

inorichi:
No difference, probably. It's almost the same.

Taumer:
Can you update this branch?

inorichi:
Yes, I was waiting to finish the cloudflare thing. I will start working on this today. It's going to be tough after all the changes

inorichi:
@na-ji Can you implement ReadMangaToday with the new sources? I would like to merge this ASAP and forget about rebasing every time.

You probably want to extend from `ParsedOnlineSource`, use the others implementations as a guide.

@Taumer Are russian sources ready for merge? Will you use the custom chapter recognition? If you don't, override the method and return a -2 so that MAL sync doesn't update wrong chapters. 

With sources ordering (af0cf9e), the chapter number won't be required for loading the next chapter from the reader, so it's okay to have every chapter recognized as -2, but MAL won't work

Taumer:
@inorichi I'm override chapter number for Readmanga/Mintmanga  to -2
For Mangachan chapters number parsed correctly

Yes, sources ready for merge.

na-ji:
@inorichi It's done. I finally dropped the JSON API because I noticed that the API doesn't work if the query is too short. Should have done that to begin with because a POST request is easier. :sweat_smile:

inorichi:
Finally rebased this monster. I'll test this PR for a bit (2 or 3 days maybe) and make sure everything is still fine. 

Given that chapter recognition isn't required anymore, I have to show an alert dialog when the user installs the new version telling him to update the library (in order to sync the chapters' order with the original source). I'll try to find a nice solution for this while I test the PR.

na-ji:
So from now on we have the choice to use yaml parser or hardcoded parser ? And we should use hardcoded parser when yaml parser is not capable of doing what we need ?
Also do we have a solution to easily share and install yaml file ? Like a host or something. It could also be useful to update yaml files automatically. 
Because I will make two or three other parser when this is merged.

inorichi:
Yes, you can choose which one to use, but the yaml parser is still pretty basic, and I haven't documented it yet (it's just two files though).

There isn't any UI nor a host to install the file. I haven't decided anything yet (open to suggestions), and it's a time consuming task. Currently the only advantage is that it doesn't have to be merged in the app for it to be available.

na-ji:
I don't think that parsers need to be updated very often, but it would be cool to have a place where we can find parsers easily. So how about a github repo to centralize the parsers where dev can push their parsers and user can download them manually ? Maybe this solution could be used for a downloader in the app, even if it's not required on the beginning.

Efreak:
When will there be documentation on the yaml format? Currently there's an issue with it--readcomicsonline.tv and two kinds both read backwards (next/previous chapters are reversed).

It would also be nice to have some way to split up chapters for webcomics like two kinds where each individual page is recognized as a chapter in tachiyomi. This comic reader does for me what the comic rocket app never did. (I'm assuming such function would be a function of the yml file, as it might change per provider). Whether this involves getting the chapter/volume list from the webcomic website, manually listing the page count of each chair in the yml file, or even just setting a fixed number of pages to download a time, anything would be welcome.

munrotm:
Does this work for you at all though? I've been waiting for an update as I would love to use this app with readcomics.tv but they don't even show up for me as options in my source lost, let alone fail on use. Installed latest version of dev branch of Tachiyomi available via FDroid, copy text from the two English comics yml files from git into .yml files in /sdcard/Tachiyomi/parsers/ and started up Tachiyomi...

Maybe I am doing something wrong. Would love to use this app alongside my ubooquity server for stuff I don't own or can't find.

Efreak:
These almost work for me. It would be perfect, but it downloads and opens
chapters in backwards order--they're numbered correctly, but it starts with
the last and ends with the first. I think I mentioned this somewhere, but
it doesn't personally bother me too much, as I use perfect viewer for
reading hi-res files anyways (PV has autocrop, gamma settings, etc, and
allows me to open files that I've moved to a network share).

Do the yml sources show up in settings>sources? I can't remember if I had
to enable them manually...

On Dec 19, 2016 2:31 AM, "munrotm" <notifications@github.com> wrote:

Does this work for you at all though? I've been waiting for an update as I
would love to use this app with readcomics.tv but they don't even show up
for me as options in my source lost, let alone fail on use. Installed
latest version of dev branch of Tachiyomi available via FDroid, copy text
from the two English comics yml files from git into .yml files in
/sdcard/Tachiyomi/parsers/ and started up Tachiyomi...

Maybe I am doing something wrong. Would love to use this app alongside my
ubooquity server for stuff I don't own or can't find.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/inorichi/tachiyomi/pull/239#issuecomment-267931756>,
or mute
the thread

gallifreyanarchitect:
Yeah, I followed the instructions for YAML sources (for readcomiconline.to) to the letter, in both versions 0.4.1 and 0.2.3, and the site just does not show up in the Catalogue/Source list. Any idea what's wrong? I'm at my wits end and am really just wanting to figure out what the issue is so I can get back to reading comics.

So, yeah, anyone have any idea what's up? Inorichi, anything??

inorichi:
If everything works well, soon catalogs will be installed separately as an apk and YAML will be deprecated. They will have to be installed/updated manually (at least you won't need to know the exact folder where you have to place them), until an UI for managing them is implemented.

joewandy:
Is this feature available in the 0.4.1 version from f-droid? I downloaded some yml files for parsing readcomicsonline from this link: http://my.mixtape.moe/wxhviq.zip then I placed them in /sdcard/Tachiyomi/parsers (https://imgur.com/a/8oDhY), but they still don't show up in my list of sources. What am I missing here?

inorichi:
It should work. The exact folder is the same as the first downloads directory shown in the preferences (replacing `/downloads` with `/parsers`).

If you are running Android 6.0 or higher, you also have to give the app disk permissions (a dialog should be shown the first time you open any manga).

joewandy:
Ah.. The permission is what was missing. Thanks!

PR #246, opened 2016-04-10
diagonally scrolling bug semi-fix #231

### LINKED ISSUE #231: Bug with scroll diagonally
Hi, there is really annoying bug with scroll. When you scroll page and move finger diagonally then nothing will happend, like there is a dead zone. 
To reproduce bug you need viewer one of: left to right, default, right to left and page that is longer than screen height, so if you move finger diagonally then nothing will happen. 

Here is the video of the bug. 
https://goo.gl/photos/AhuYnmHqHW2nqWr48
---
Version: r584 or v0.1.4
Nexus 5, Android 6.0.1, Root, Custom Rom

iBaer:
Fix for this "bug" by removing (just uncommented) a statement in the viewpager class

EDIT: Just pushed it to my real device, it works better, but it's far from perfect.
Gonna look into it some more.


iBaer:
will reopen it later, when the result is actually perfect

PR #255, opened 2016-04-17
Fixed backup/restore for 3rd party applications

iBaer:
Same MIME for restore/backup, so 3rd party applications like Google Drive find the proper files.
MIME changed to proper type for json files.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/backup/BackupFragment.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/backup/BackupFragment.kt
@@ -57,7 +57,7 @@ class BackupFragment : BaseRxFragment<BackupPresenter>() {
         restore_button.setOnClickListener {
             val intent = Intent(Intent.ACTION_GET_CONTENT)
             intent.addCategory(Intent.CATEGORY_OPENABLE)
-            intent.type = "application/json"
+            intent.type = "application/*"
             val chooser = Intent.createChooser(intent, getString(R.string.file_select_backup))
             startActivityForResult(chooser, REQUEST_BACKUP_OPEN)
         }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/backup/BackupFragment.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/backup/BackupFragment.kt
@@ -57,7 +57,7 @@ class BackupFragment : BaseRxFragment<BackupPresenter>() {
         restore_button.setOnClickListener {
             val intent = Intent(Intent.ACTION_GET_CONTENT)
             intent.addCategory(Intent.CATEGORY_OPENABLE)
-            intent.type = "application/octet-stream"
+            intent.type = "application/json"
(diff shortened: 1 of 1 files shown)


inorichi:
I couldn't select backups with the android chooser when the mime type was `application/json`. It only worked when it was `application/octet-stream`.

Using `text/plain` should allow to use more apps to save the file. I'll have to test this.

iBaer:
Here is the situation on (my device) prior to my "fix".
1. With backup you would push a text/plain file to Google Drive
2.1 With restore you would search for application/octet-stream files in Google Drive, thus you can't select the backup.
2.2 Same with the default file manager in Android. Since it will not search for files of the MIME which our backup has

You might be right with the amount of applications that might work with text/plain though. But technically all file browsers and cloud services should work with application/json.

inorichi:
This is what happens with the emulator (API 23).

With `application/json`:

qemu-system-i386_2016-04-17_19-23-33

With `application/octet-stream`:

qemu-system-i386_2016-04-17_19-25-03

I don't know if there's a way to have `application/json` and being able to select them...

iBaer:
Make a new backup file with the same MIME (`application/json`).

If your backup file has the `application/json` MIME you will also find it with the search.
It's funny though, that you can select your file with `application/octet-stream`. Since I could only do it if (same as above) the backup file and the search have the same MIME. Is your file of `text/plain` and you are searching for `application/octet-stream`?

inorichi:
I tried with `application/json` in both places. It still doesn't allow me to select the file.

And yes, the first one was created with `text/plain` and restored with `application/octet-stream`

inorichi:
Finally found it http://stackoverflow.com/questions/27092151/search-json-file-using-intent

iBaer:
Now this is funny. I'll try to get my emulator with API 23 running again. My real device I'm debugging on is using API 22.
I can't even backup the file with the default file browser (might be, because it's the one from Cyanogenmod) here using my real device.
I tried to backup the file with Drive and ES explorer. Restoring from Drive and the default explorer doesn't work with the wrong MIME types. Which results in what I tried to fix.

I will try a few things and will see what works for me and the emulator.

iBaer:
The emulator with API 23 doesn't give me a proper option (it wants to use only the messenger).
So what file browser are you using to backup the data?

inorichi:
You can just dismiss the dialog, the file is always saved in the external cache directory. `/sdcard/Android/.../cache`

iBaer:
Oh, I see what you did there.
The cached file doesn't have the MIME we intend to save it with yet. The type it gets the `application/octet-stream` MIME by default (ofc, since it's for unknown types).
I'll see what I can come up with to change the MIME of the File-object directly. If you already have a solution in mind, feel free to help me out on this one.

inorichi:
AFAIK you can't save MIME on a file. I think changing only the backup `Intent` from `text/plain` to `application/json` should be enough (and it should work with drive, can you test?)

iBaer:
We can go with this change temporaly.
It would still be better if either the cached file isn't used or a local file with a proper MIME type will be created,

Issue #260, opened 2016-04-18
No download failed/error notification 


pitman10:
I've added a bunch of titles to my library(from Batoto) and proceeded to download them all, 100+ chapters (going into each title -> select all -> download -> download all) and left it at that, came back later to find out that it only downloaded a fraction with no indication when/why it stopped.

Version: v2.0
Android: 6.0.1

PR #289, opened 2016-05-04
Chapter download notification

NoodleMage:
## Done
- [x] Code optimization + comments

## Screens
Note: Screenshots are WIP
### Simultaneous downloads = 1
screener_20160504 180736
screener_20160504 180758

### Simultaneous downloads > 1
screener_20160506 112343

### Error
screener_20160504 180806

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
@@ -5,6 +5,7 @@ import android.net.Uri
 import com.google.gson.Gson
 import com.google.gson.reflect.TypeToken
 import com.google.gson.stream.JsonReader
+import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.data.database.models.Chapter
 import eu.kanade.tachiyomi.data.database.models.Manga
 import eu.kanade.tachiyomi.data.download.model.Download
@@ -14,7 +15,10 @@ import eu.kanade.tachiyomi.data.preference.getOrDefault
 import eu.kanade.tachiyomi.data.source.SourceManager
 import eu.kanade.tachiyomi.data.source.base.Source
 import eu.kanade.tachiyomi.data.source.model.Page
-import eu.kanade.tachiyomi.util.*
+import eu.kanade.tachiyomi.util.DiskUtils
+import eu.kanade.tachiyomi.util.DynamicConcurrentMergeOperator
+import eu.kanade.tachiyomi.util.UrlUtil
+import eu.kanade.tachiyomi.util.saveImageTo
 import rx.Observable
 import rx.Subscription
 import rx.android.schedulers.AndroidSchedulers
@@ -35,6 +39,8 @@ class DownloadManager(private val context: Context, private val sourceManager: S
     val runningSubject = BehaviorSubject.create<Boolean>()
     private var downloadsSubscription: Subscription? = null
 
+    val downloadNotifier by lazy { DownloadNotifier(context) }
+
     private val threadsSubject = BehaviorSubject.create<Int>()
     private var threadsSubscription: Subscription? = null
 
@@ -48,10 +54,14 @@ class DownloadManager(private val context: Context, private val sourceManager: S
         private set
 
     private fun initializeSubscriptions() {
+
         downloadsSubscription?.unsubscribe()
 
         threadsSubscription = preferences.downloadThreads().asObservable()
-                .subscribe { threadsSubject.onNext(it) }
+                .subscribe {
+                    threadsSubject.onNext(it)
+                    downloadNotifier.multipleDownloadThreads = it > 1
+                }
 
         downloadsSubscription = downloadsQueueSubject.flatMap { Observable.from(it) }
                 .lift(DynamicConcurrentMergeOperator<Download, Download>({ downloadChapter(it) }, threadsSubject))
@@ -60,7 +70,9 @@ class DownloadManager(private val context: Context, private val sourceManager: S
                 .subscribe({
                     // Delete successful downloads from queue
                     if (it.status == Download.DOWNLOADED) {
+                        // remove downloaded chapter from queue
                         queue.del(it)
+                        downloadNotifier.onProgressChange(queue)
                     }
                     if (areAllDownloadsFinished()) {
                         DownloadService.stop(context)
@@ -68,7 +80,7 @@ class DownloadManager(private val context: Context, private val sourceManager: S
                 }, { e ->
                     DownloadService.stop(context)
                     Timber.e(e, e.message)
-                    context.toast(e.message)
+                    downloadNotifier.onError(e.message)
                 })
 
         if (!isRunning) {
@@ -114,6 +126,12 @@ class DownloadManager(private val context: Context, private val sourceManager: S
                 pending.add(download)
             }
         }
+
+        // Initialize queue size
+        downloadNotifier.initialQueueSize = queue.size
+        // Show notification
+        downloadNotifier.onProgressChange(queue)
+
         if (isRunning) downloadsQueueSubject.onNext(pending)
     }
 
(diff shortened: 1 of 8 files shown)


NoodleMage:
https://github.com/inorichi/tachiyomi/issues/260

inorichi:
I see a lot of problems showing the title of the "active" download. It works fine if you only have 1 simultaneous download, but it will go mad if you have more. I would just show an overall status without any specifics (at most a progress bar).

Also `setProgressObservable` is very troublesome. Generally you should have only one `.subscribe`, if you nest them, it won't end well.

NoodleMage:
Yeah I kinda forgot about simultaneous downloads. Never use it so didn't realize it was an option xD. It looks very funny but will fix it xD.

As for `.subscribe` this shouldn't be a problem would it. It does exactly what's needed. When a new download starts is subscribes to the progress and unsubscribes when the download is completed.

Constants.kt is not a bad idea. I assume you want it to work the same way as the `PreferenceKeys` class?

inorichi:
Those subscriptions never unsubscribe (you aren't calling unsubscribe, and they don't call `onCompleted` or `onError` so they stand there until the process is killed).

Anyways, you shouldn't need to use the queue's observables if you show a progress with remaining downloads.

About the Constants, yeah, it would be something like that, but without needing a context and using a singletion (`object Constants { }`). We could even use multiple object classes inside that file to separate them by something: `object NotificationConstants { ... }` `object xxxConstants { ... }` though I'm not sure it's a good idea to split them.

NoodleMage:
I want to move `onCompleted` and `onError `notifier calls to the `setProgressObservable `anyway because it gives better results (I hope). This means that the subscription will be closed if I'm correct or is it needed to call `unsubscribe` in the `oncomplete`?

As for showing remaining download progress. This is where I have to disagree. I don't care about the amount that I have downloaded. I can see that in the chapter overview. I want to know the progress of pages downloaded per chapter in my notification, showing the chapter download progress is in my opinion redundant.

inorichi:
`downloadQueue.getProgressObservable()` never calls `onCompleted` or `onError`, and you can't change that. It's behind a subject which is emitting a `Download` when a page is downloaded so the only way to unsubscribe from it is by calling `.unsubscribe` on the subscription.

Also, that observable already listens for every download in the queue, so you would have to subscribe only once.

NoodleMage:
Aah. That changes a lot. I thought the ovservable was chapter based. 

Will look at it when I have some time and see how I can implement it correctly. Thanks for help as always ;)

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
```diff
@@ -35,6 +38,15 @@ class DownloadManager(private val context: Context, private val sourceManager: S
...
     private var downloadsSubscription: Subscription? = null
 
+    /**
+     * Subscription of the download progress
+     */
+    private var progressSubscription: Subscription? = null
+
+    val downloadNotifier by lazy { DownloadNotifier(context) }
+
+    val multipleDownloadThreads by lazy { preferences.downloadThreads().getOrDefault() != 1 }
```
inorichi: Set this as a `var` and change its value in `threadsSubscription`:
```
threadsSubscription = preferences.downloadThreads().asObservable()
    .subscribe { 
        threadsSubject.onNext(it) 
        multipleDownloadThreads = it != 1
    }
```

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
```diff
@@ -161,34 +201,35 @@ class DownloadManager(private val context: Context, private val sourceManager: S
...
+        return Observable.defer {
+            pageListObservable
+                    .doOnNext { pages ->
+                        download.downloadedImages = 0
+                        download.status = Download.DOWNLOADING
+                    }
+                    // Get all the URLs to the source images, fetch pages if necessary
+                    .flatMap { download.source.getAllImageUrlsFromPageList(it) }
+                    // Start downloading images, consider we can have downloaded images already
+                    .concatMap { page -> getOrDownloadImage(page, download) }
```
inorichi: I would still try to avoid using `queue.getProgressObservable` and add below this `.concatMap` a `.doOnNext`. The notifier is too tied to the manager, so there shouldn't be a need to use that.

I like the idea of delegating notifications to a new class, but currently the manager has a lot of `if` statements because of the `multipleDownloadThreads`. I would make the manager as dumb as possible, and store `multipleDownloadThreads` in the notifier class. The manager will tell the notifier to update providing only the necessary elements (like a download, a page, the queue size...)

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
```diff
@@ -26,41 +28,40 @@ class DownloadNotifier(private val context: Context) {
...
-                isDownloading = true
-            }
+    var multipleDownloadThreads = false
 
-            if (download.chapter.name.length >= 33)
-                setContentTitle(download.chapter.name.slice(IntRange(0, 30)).plus("..."))
-            else
-                setContentTitle(download.chapter.name)
+    init {
+        multipleDownloadThreads = preferences.downloadThreads().getOrDefault() != 1
```
NoodleMage: This doesn't work like I wanted it to (Settings change is now only applied after app restart). So will probably change this to subscription like the DownloadManager

inorichi: Create a method in the notifier `fun setMultipleDownloads(active: Boolean) {}` and call it in the `threadsSubscription`

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
```diff
@@ -48,26 +55,41 @@ class DownloadManager(private val context: Context, private val sourceManager: S
...
                     DownloadService.stop(context)
                     Timber.e(e, e.message)
-                    context.toast(e.message)
+                    downloadNotifier.onError(e.message)
                 })
 
+        // Create subscription for clear library call
+        clearSubscription = queue.getClearObservable()
+                .observeOn(AndroidSchedulers.mainThread())
+                .doOnCompleted { clearSubscription?.unsubscribe() }
```
NoodleMage: I couldn't discover if this call is needed or that `isClearedSubject.onComplete()`  from `DownloadQueue ` is enough to unsubscribe?

btw. I think the code is now working as intended. So unless you see something you don't like I will start on the squash and merge.

inorichi: Sure. No need to complicate yourself with more observables/subjects. Just move the call to `downloadQueue.clear()` to 
```
downloadManager.clearQueue() {
    queue.clear()
    notifier.onClear()
```

Also, in the notifier remove the nullability on parameters that won't be null (like the queue)

NoodleMage: @inorichi  this is what I mean.... I don't even..... xD.

Made the queue non null-able but wan't to keep `onProgressChange `as one function call. So will keep download null-able.

Code should be working as intended now ;). Please let me know if I did anything completely unnecessary, nothing will surprise me at this point xD. but if not I will start on the merge.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
```diff
@@ -0,0 +1,179 @@
...
+     * Called when download progress changes
+     * Note: Only accepted when single download active
+     * @param download download object containing download information
+     */
+    internal fun onProgressChange(download: Download, queue: DownloadQueue) {
+        // If multi download mode return.
+        if (multipleDownloadThreads)
+            return
+        // Update progress.
+        onProgressChange(download, queue)
```
inorichi: Infinite recursion here :)

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
```diff
@@ -0,0 +1,179 @@
...
+     * Simultaneous download setting > 1.
+     */
+    internal var multipleDownloadThreads = false
+
+    /**
+     * Called when download progress changes.
+     * Note: Only accepted when multi download active.
+     * @param queue the queue containing downloads.
+     */
+    internal fun onProgressChange(queue: DownloadQueue) {
```
inorichi: Why not a single function?

```
internal fun onProgressChange(queue: DownloadQueue, download: Download? = null) {
    ....
}
```

Then you can call it with one parameter (queue) or with two (queue and download)

NoodleMage: Yeah I know. That's not why I used separate functions. I used it because the `multipleDownloadThreads` check is done in the notifier. 

```
if (!multipleDownloadThreads)
            return
```

If this is not the case every page download will call the function with a download which is not null and a queue which is not null. 

```
         // Create notification
        with (notificationBuilder)
        {
            // Check if icon needs refresh
            if (!isDownloading) {
                setSmallIcon(android.R.drawable.stat_sys_download)
                isDownloading = true
            }

            if (multipleDownloadThreads) {
                    setContentTitle(context.getString(R.string.app_name))

                    setContentText(context.getString(R.string.chapter_downloading_progress)
                            .format(initialQueueSize - queue.size, initialQueueSize))
                setProgress(initialQueueSize, initialQueueSize - queue.size, false)
            } else {
                download?.let {
                    if (it.chapter.name.length >= 33)
                        setContentTitle(it.chapter.name.slice(IntRange(0, 30)).plus("..."))
                    else
                        setContentTitle(it.chapter.name)

                    setContentText(context.getString(R.string.chapter_downloading_progress)
                            .format(it.downloadedImages, it.pages.size))
                    setProgress(it.pages.size, it.downloadedImages, false)

                }
            }
        }
        // Displays the progress bar on notification
        context.notificationManager.notify(notificationId, notificationBuilder.build())
```
thus the above code will be called ~21 times per chapter instead of one time in the case of simultaneous downloads enabled.

inorichi: I see, but you can still reduce it to two methods. The public one would send a null download to the private one.

NoodleMage: Then you still have useless notification updates do you not?

inorichi: Ahhh, you're right. It's a bit confusing :(

Issue #233, opened 2016-03-27
Chapter skipping


j2ghz:
When reading http://www.readmanga.today/ubel-blatt chapter 4 got skipped. Both `4` and `004.extra` got recognized as chapter 4.

<img src="https://cloud.githubusercontent.com/assets/1708197/14065628/b98221fe-f431-11e5-9041-70cdf3a9134a.png" width="49%"> <img src="https://cloud.githubusercontent.com/assets/1708197/14065637/f621fed6-f431-11e5-9d8e-007c5ee70e0c.png" width="49%">
---
Version: r590

Issue #9, opened 2016-01-01
Wrong chapter ordering


j2ghz:
* [x] As you can see there is one extra chapter `Onepunch-Man Punch Ver002 028`

https://cloud.githubusercontent.com/assets/1708197/12069149/7922750e-b021-11e5-8005-a8436aa57791.png
https://cloud.githubusercontent.com/assets/1708197/12069150/98cd49ba-b021-11e5-801d-7d1fa1401f47.png

Version `r285` `31 Dec 2015 7:21pm`

j2ghz:
It seems the problem is you get the chapters from html and then parse the number of each chapter and then sort them. OnePunch Man on kissmanga uses different type of naming:
`OnePunch Man 001 - 056`
`OnePunch Man Ver002 001 - 0086`
Which is like two mangas in one? (I think) (`OnePunch Man 001` starts the same as `OnePunch Man Ver002 001`)

Chapter `Onepunch-Man Punch Ver002 028` is parsed as chapter number 2028 (confirmed  when  submitted to MAL) and thus sorted as the last one.
The others are sorted like this:
1. `OnePunch Man 001`
2. `OnePunch Man Ver002 001`
3. `OnePunch Man 002`
4. `OnePunch Man Ver002 002`

inorichi:
I fixed the chapter being recognized as 2028 here https://github.com/inorichi/mangafeed/blob/master/app/src/test/java/eu/kanade/mangafeed/ChapterRecognitionTest.java#L117

The chapters have to be deleted in order to recognize the number again. If the "Ver002" is a common keyword in Kissmanga (I don't usually read here) we could add another pattern for it.

I want to add an additional sorting method based on the upload date, but I have to save sorting options first.

inorichi:
I improved the recognition again. I will keep this issue open for future improvements.

If anyone finds a wrong chapter order, post an example here to try to fix it.

myuc:
Take a look at capeta both mangafox and batoto.

For mangafox: http://mangafox.me/manga/capeta/
Chapters for each volume are divided into chpt 1-4 and in the app, its sorted in alphabetical order ignoring volume.

For batoto: http://bato.to/comic/_/comics/capeta-r628
Although you can see Vol:xx Ch.XX, it is still sorted in ch no only and ignore the Vol:xx in the title.

PS: not sure if sorting by date is a good idea as sometimes earlier chapters could be updated/added later on.

inorichi:
[...]

PR #293, opened 2016-05-07
Chapter recognition rewrite

NoodleMage:
Rewrote ChapterRecognition to Kotlin


app/src/main/java/eu/kanade/tachiyomi/util/ChapterRecognition.kt
```diff
@@ -0,0 +1,175 @@
+package eu.kanade.tachiyomi.util
+
+import eu.kanade.tachiyomi.data.database.models.Chapter
+import eu.kanade.tachiyomi.data.database.models.Manga
+import java.util.*
+
+object ChapterRecognition {
+    private val cleanWithToken = Regex("ch[^0-9]?\\s*(\\d+[\\.,]?\\d+)($|\\b)")
```
inorichi: Can you test if it works with triple quotes and not using escaping backslash?

```private val cleanWithToken = Regex("""ch[^0-9]?\s*(\d+[\.,]?\d+)($|\b)""")```

Comment (NoodleMage) [inline app/src/main/java/eu/kanade/tachiyomi/util/ChapterRecognition.kt]: Yeah it works \o/.

**Comment by NoodleMage** [comment]
#9

File: app/src/main/java/eu/kanade/tachiyomi/util/ChapterRecognition.kt [hunk 2]
```diff
@@ -0,0 +1,236 @@
...
+        //If chapter number is known return.
+        if (chapter.chapter_number != -1f)
+            return
+
+
+        // Safest option, the chapter has a token prepended and nothing at the end of the number
+        var name = chapter.name.toLowerCase()
+
+        basic.find(name)?.let {
+            chapter.chapter_number = it.groups[2]?.value?.toFloat() as Float
```
inorichi: Isn't `toFloat() as Float` redundant?

NoodleMage: The IDE doesn't accept it without `as Float`. There's probably a better way to convert floats. But I fist want to complete the regex before spending time on that ;).

inorichi: Then use `toFloat()!!` it will throw an exception with both approaches if anything goes wrong (it shouldn't though)

app/src/test/java/eu/kanade/tachiyomi/ChapterRecognitionTest.kt
```diff
@@ -0,0 +1,229 @@
...
+        return chapter
+    }
+
+    @Before
+    fun setUp() {
+        randomManga.title = "Something"
+    }
+
+    @Test
+    fun testWithOneDigit() {
```
inorichi: I haven't tried it, but with Kotlin you can do: 
```
fun `test with one digit`() {
    ...
}
```

Of course this should be used only for testing :)

NoodleMage: Doesn't work in Android ;(

j2ghz:
closes #233

NoodleMage:
Closes #9

inorichi:
It seems it only closes them if the comment is in the first post :)

PR #303, opened 2016-05-16
UI tweaks

NoodleMage:
- Fixed some incorrect colors
- MyAnimeList card now looks correct.
- Empty view is now a custom view.

screener_20160516 153245
screener_20160516 153301

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/widget/EmptyView.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/widget/EmptyView.kt
@@ -1,0 +1,36 @@
+package eu.kanade.tachiyomi.widget
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.widget.RelativeLayout
+import eu.kanade.tachiyomi.R
+import eu.kanade.tachiyomi.util.getResourceColor
+import eu.kanade.tachiyomi.util.setVectorCompat
+import kotlinx.android.synthetic.main.view_empty.view.*
+
+class EmptyView @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) :
+        RelativeLayout (context, attrs) {
+
+    init {
+        inflate(context, R.layout.view_empty, this)
+    }
+
+    /**
+     * Hide the information view
+     */
+    fun hide() {
+        this.visibility = View.GONE
+    }
+
+    /**
+     * Show the information view
+     * @param drawable icon of information view
+     * @param textResource text of information view
+     */
+    fun show(drawable: Int, textResource: Int) {
+        image_view.setVectorCompat(drawable, context.theme.getResourceColor(android.R.attr.textColorHint))
+        text_label.text = context.getString(textResource)
+        this.visibility = View.VISIBLE
+    }
+}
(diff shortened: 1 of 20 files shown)


app/src/main/res/values/themes.xml
```diff
@@ -25,13 +25,14 @@
...
         <item name="windowActionModeOverlay">true</item>
-        <item name="actionModeBackground">?colorPrimaryDark</item>
+        <item name="actionModeBackground">?colorPrimary</item>
         <item name="actionBarTheme">@style/Theme.ActionBar</item>
         <item name="actionBarPopupTheme">@style/ThemeOverlay.AppCompat.Light</item>
         <item name="preferenceTheme">@style/PreferenceThemeOverlay.v14.Material</item>
 
         <!-- Custom Attributes-->
         <item name="selectable_list_drawable">@drawable/list_item_selector_light</item>
+        <item name="selectable_library_drawable">@drawable/library_item_selector_light</item>
```
inorichi: Isn't `<ripple>` for API >= 21?

NoodleMage: yes. but the drawable has an API v21+ version and a everything else version

inorichi: Right. Didn't look at the full path :)

app/src/main/java/eu/kanade/tachiyomi/widget/InformationView.kt
```diff
@@ -0,0 +1,36 @@
...
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.widget.RelativeLayout
+import eu.kanade.tachiyomi.R
+import eu.kanade.tachiyomi.util.getResourceColor
+import eu.kanade.tachiyomi.util.setVectorCompat
+import kotlinx.android.synthetic.main.view_information.view.*
+
+class InformationView @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) :
```
inorichi: How about calling it EmptyView? It makes some more sense IMO.

app/src/main/java/eu/kanade/tachiyomi/widget/EmptyView.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    init {
+        inflate(context, R.layout.view_empty, this)
+    }
+
+    /**
+     * Hide the information view
+     */
+    fun hide() {
+        this.visibility = View.INVISIBLE
```
inorichi: Have you tried `View.GONE`? I think it's better than this one

app/src/main/res/values/themes.xml
```diff
@@ -25,13 +25,14 @@
 
         <!-- Themes -->
         <item name="windowActionModeOverlay">true</item>
-        <item name="actionModeBackground">?colorPrimaryDark</item>
+        <item name="actionModeBackground">?colorPrimary</item>
```
inorichi: Not sure about this one. IMO it should be different from the default toolbar color. For example, the gmail app has a totally different color when selection mode is enabled. But I agree `colorPrimaryDark` isn't any better.

NoodleMage: I liked the way it looked in Solid Explorer (also same color as toolbar). 
Will try to make it look like Google inbox one (white for normal theme, black for dark theme), still prefer the main color especially when you select chapter, the contrast isn't as big.

PR #307, opened 2016-05-21
Recently read tab

### LINKED ISSUE #111: Recently read tab
Now that I've used Tachiyomi for awhile I often find myself in need of a Recently read tab, that lets me see the latest chapters I've read for each series.

Check out Manga Rock's "Recents" tab or Manga DLR's "Recently read" tab for similiar functionality.

NoodleMage:
Closes #111 

## Screens
screener_20160522 15_32_55
screener_20160522 21_45_10
screener_20160522 21_45_21
screener_20160522 21_45_29


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -285,6 +288,11 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
         }
     }
 
+    private fun setMangaLastRead() {
+        manga.last_read = Date().time
+        db.insertManga(manga).executeAsBlocking()
+    }
```
NoodleMage: This is the best location to place this as far as I could think of (and it works). @inorichi  if you have a better idea please let me know ;).

app/src/main/java/eu/kanade/tachiyomi/data/database/DbOpenHelper.kt
```diff
@@ -33,7 +33,11 @@ class DbOpenHelper(context: Context) : SQLiteOpenHelper(context, DbOpenHelper.DA
     }
 
     override fun onUpgrade(db: SQLiteDatabase, oldVersion: Int, newVersion: Int) {
-
+        if (newVersion == 2) {
```
NoodleMage: Don't know if this works or that Android fixed the "last_read column not found" by coincidence. Will try to debug this.

inorichi: Yes, it's done like that, but you should use `if (oldVersion < 2)`. Though in this particular case it doesn't matter because it can only go from 1 to 2, but if it where 3, old version is 1 and new version is 3.

inorichi:
History would have it's own table referencing the chapter id, as it can grow if we decide to show time spent reading the chapter or have statistics.

NoodleMage:
Cool idea. Will change it and implement it with that in mind ;).

NoodleMage:
Working on rewrite. Will open new pull when done

PR #310, opened 2016-05-23
Fixed spinner + improved ripple

NoodleMage:


inorichi:
LGTM, though I would rename `textColorPrimary` to `text_color_primary`, basically because it's easier to differentiate android's attributes from custom attributes.

BTW, yesterday I saw a video from the IO about the support library and it seems finally we will be able to use attributes' references in selectors.

NoodleMage:
Done :+1:

PR #318, opened 2016-05-29
Rename "Only update incomplete manga" string to "ongoing"

wopian:
All English manga sources use either Ongoing or Completed, so Tachiyomi using a different term for ongoing manga feels off.


inorichi:
Technically it wouldn't be true, there's unkown, ongoing, completed and licensed.

Issue #268, opened 2016-04-19
Resuming chapters downloads read pages


wopian:
When resuming a chapter you've partially read (e.g read up to page 26 of 46) after the chapter cache has been overwritten by other chapters, the app proceeds to download all the pages again from page 1.

For the user, this results in them just starring at the spinner for ages until it finally starts downloading the page you were on...after wasting several dozen MBs downloading pages you'll never see.

Version: r661

Bloodwork:
This one is particulary bad on Bato.to which is slower than other sources by default. Sometimes it takes up to two minutes to resume at the end of a chapter for manga with particulary large scan sizes.

PR #350, opened 2016-06-14
Rewrite DB models, tests and add a chapter loader.

inorichi:
This PR have the following changes:

* Replaced Dagger2 with Injekt, it's easier to inject dependencies and no code generation (apt) is required.
* Converted all database models to Kotlin, making some of them interfaces, allowing UI models to store additional information (like the `isDownloaded` flag, which is not stored in the database). Non database fields should extend DB models.
* Converted all tests to Kotlin.
* Added a new chapter loader, which implements #268. It allows to load pages from the starting one and on demand (starting on page 1 and jumping to page 28 with the slider will prioritize loading page 28 over the previous ones.

The rest of changes are minor ones and shouldn't break anything, but any testing is welcome.


inorichi:
Commits slightly reorganized, but partial commits won't compile.

PR #355, opened 2016-06-18
Mangasee as image source

Gilfar:
As batoto is sometimes removing scanlations there is a need for comparable quality scans.


inorichi:
I don't like the method `getAbsoluteUrl`. I'm going to try removing it, because it will always need improvements and dirty fixes.

inorichi:
Done. Replace `Jsoup.parse(response.body().string())` with `response.asJsoup()` and `.attr("href")` with `.absUrl("href")` where needed.

Issue #288, opened 2016-05-02
Add readcomiconline.me to the app


Rumple007:
The site seems to be connected to kissanime so I was wondering if it was possible

Soumatan:
Yeah this would be great. That way I can actually start reading comics. Tho it's readcomiconline.to right now. I think they changed it like kissanime.

Neverest:
Agree, would be nice to have comics too.

j2ghz:
I tried: https://gist.github.com/j2ghz/fe656fc9efe86de6fee6457bc9454170

Couldn't finish (every page gets the image of first page), because I'm getting CAPTCHAs after reading a few chapters while testing. The error `page list is empty` means CAPTCHA in that case.

So even if I finish it, it's possible it will be unusable.

Neverest:
Hmm, since the website is basically a comic version of kissanime.to, would copying the same settings as kissmanga, but changing the relevant info to readcomiconline make it possible to fix it?

j2ghz:
I did, getting the same CAPTCHA error. You can try for yourself, maybe cloudflare marked my IP as suspicious. [Download APK](https://github.com/inorichi/tachiyomi/files/322226/tachiyomi.zip) (it installs separately from release and dev)

Rumple007:
So I just tried it, I tried reading the first chapter of Injustice Gods among us and I managed to read all the pages

Rumple007:
And tried downloading the chapter for online reading and it works

Neverest:
Hehe, you got me  just when I waked up. :)

Gave it a try, and same as Rumple007, it worked fine. Browse some chapters back to back, changed comics, tried webtoon and no problems.

j2ghz:
As @inorichi said, because it can be added with YAML source, he won't be adding it directly. Head over to [wiki](https://github.com/inorichi/tachiyomi/wiki/YAML-Sources) for instructions.

PR #357, opened 2016-06-19
Add ReadComicOnline

j2ghz:
I get CAPTCHAs on http://readcomiconline.to/, so I can't test it. But people in #288 report it's working, so maybe it's just me. (I get `Page listing is empty`, debugger shows the response is Are you human page from cloudflare)

[Download APK](https://github.com/inorichi/tachiyomi/files/322226/tachiyomi.zip) (it installs separately from release and dev)

There's also [YAML version](https://gist.github.com/j2ghz/fe656fc9efe86de6fee6457bc9454170) available, but every page gets the same image as first page and again I can't finish it because CAPTCHAs.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/source/SourceManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/source/SourceManager.kt
@@ -25,8 +25,9 @@ open class SourceManager(private val context: Context) {
     val MINTMANGA = 6
     val MANGACHAN = 7
     val READMANGATODAY = 8
+    val READCOMICONLINE = 9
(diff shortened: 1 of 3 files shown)


inorichi:
Since this is for comics and the YAML parser could support it, I prefer this one to be optional.

I will to improve the parser to fix that issue.

BTW, I allowed an integer as id to have compatibility if I decided to remove mangafox/mangahere, but you should use a string, like "readcomiconline".

inorichi:
Closing for now, unless it's not possible to parse with YAML I won't add it to the app. I'll work on the parser these days.

Issue #340, opened 2016-06-07
Add share intent for manga


jsilvermist:
Please add a share button below the `Open in browser` button in the overflow menu on the manga `Info` page.

Currently this requires opening in browser, pressing again to open in chrome, and then once more to share. Also, chrome's share menu doesn't support direct share, adding yet more steps to sharing a manga from the app.

Issue #312, opened 2016-05-25
AniList Sync Support


NotLemon:
Hi,

Would it be possible to look into supporting sync with Anilist?
Our JSON API is fully documented [here](http://anilist-api.readthedocs.io/en/latest/) and not difficult to work with. Other users have successfully managed to create syncing apps.

Any issues/questions let me know.
The developer section can be accessed at [http://anilist.co/settings/developer
](http://anilist.co/settings/developer)

Thanks

inorichi:
OAuth makes it quite hard to work with. Since I won't provide a client url, it would have to use the authorization pin. I played with it for a while with a WebView and tried to copy the pin with [a solution like this](http://stackoverflow.com/questions/3479833/is-it-possible-to-get-the-html-code-from-webview), but I don't like how it grabs the pin (looking a <h2> tag in the HTML). It would also be possible to let the user copy it though, but that's even more steps.

joshstar:
You could use the authorization code grant with an android URI scheme, this will redirect direct to your application with the auth code so you wouldn't need to mess around with scraping web views. 

As for keeping the client secret private it should be stored locally on each users device once they've initially authenticated via oauth. It wouldn't be something you store for everyone in a central or public place.

inorichi:
I mean the `client id` and `client secret` obtained from the developer section of AniList.

According to [this](http://anilist-api.readthedocs.io/en/latest/authentication.html#which-grant-type-to-use): 

> Both the current grant types have a client secret key, you are required to keep this private.

But then, they have to be known (compile time) before oauth, or am I wrong? Of course the refresh token would be stored locally for each device.

joshstar:
Ah right the client secret, you can either put it in your source code, which would be the easiest way and plenty of other AniList API devs do it. Theres a risk someone else could use your details, but even then they won't be able to authenticate a user without the user secret which is stored on device, so its not a huge concern.

If we spot huge crazy spikes in the API logs we usually contact the app dev and issue them new details if someone else is abusing their API keys.

The better way to do it is to store your client keys in a separate (.gitignored) file and read from it on the device, if thats possible with your build setup.

inorichi:
I could exclude the file from the source code, but then the F-Droid build won't be able to use the API, so my only option if I don't want to exclude them is to hardcode the values. I'll do that as you said it shouldn't be a problem.

Thanks for the clarification.

inorichi:
[...]

Issue #200, opened 2016-03-05
Tracking default status Plan to read


j2ghz:
- When binding a manga without any read chapters, make status Plan to read
- When changing chapter count from 0 and status is Plan to read, make status Reading

IceKreem:
I second this +1

inorichi:
I could make the default status Plan to read. For the second point, I think it's better to let the user decide, some may want to keep them in on-hold, for example.

j2ghz:
Without the second I wouldn't like the first, because the I would forget to set the right status all the time.

The change is supposed to happen just when the number of read is 0 and it's changing to something other than 0 and the status is plan to read. I think there are hardly any cases when all of this is true and you still want to keep it plan to read.

hyoretsu:
Does bumping work? I wrongly populate my reading list by pre-downloading new manga.

PR #384, opened 2016-07-09
Rewrote Manga Info and Manga Sync => new Manga Info (with Anilist support)

NoodleMage:
#312, #340

## TODO

- [x] Comments TODO

- [x] Default Sync Service (in settings)

- [x] Remove old classes

- [x] Test API 16

- [x] Bug fixes

- [x] #200 

## Screens

Feedback is appreciated!


app/src/main/java/eu/kanade/tachiyomi/data/database/models/MangaSyncImpl.kt
```diff
@@ -10,6 +10,8 @@ class MangaSyncImpl : MangaSync {
 
     override var remote_id: Int = 0
 
+    override var remote_score: Float = 0f
```
inorichi: Why do we need two scores for every sync?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/sync/MangaSyncFragment.kt
```diff
@@ -0,0 +1,509 @@
...
+        }
+    }
+
+    /**
+     * Called to open the [Manga] in browser.
+     */
+    fun openInBrowser() {
+        val source = presenter.source as? OnlineSource ?: return
+        try {
+            val url = Uri.parse(source.baseUrl + presenter.manga.url)
```
inorichi: I know I wrote that, but this line should do `source.mangaDetailsRequest(manga).url().toString()` or something like that for better consistency.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/sync/MangaSyncFragment.kt
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    /**
+     * Called when [MangaSyncPresenter] has fetched [MangaSync] items
+     * When fetched update view with Sync information
+     * @param mangaSync [MangaSync] object containing information about [MangaSync].
+     */
+    fun onNextMangaSync(mangaSync: MangaSync) {
+        when (mangaSync.sync_id) {
+            MangaSyncManager.MYANIMELIST -> {
```
inorichi: In my opinion the view shouldn't know about the implementations of services. The view should work with the service manager and the interface. Every custom action should be added to the interface and implemented in the services

NoodleMage: I agree. Will move some functions to presenter 👍

app/src/main/java/eu/kanade/tachiyomi/data/database/models/MangaSyncImpl.kt
NoodleMage: The remote score is the average score from the service. Score is the score given by user.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/sync/MangaSyncPresenter.kt
```diff
@@ -315,6 +315,18 @@ class MangaSyncPresenter : BasePresenter<MangaSyncFragment>() {
     /**
      * TODO
      */
+    fun getStatusPosition(mangaSync: MangaSync): Int = with(context)
+    {
+        when (mangaSync.sync_id) {
```
inorichi: These can be abstracted even more, something like `manager.getService(mangaSync.id)?.getStatusPosition(...) ?: throw ...`

app/src/main/java/eu/kanade/tachiyomi/data/mangasync/MangaSyncService.kt
```diff
@@ -35,6 +35,48 @@ abstract class MangaSyncService(private val context: Context, val id: Int) {
...
+    abstract fun getList(): Observable<List<MangaSync>>
+
+    abstract fun search(query: String): Observable<List<MangaSync>>
+
+
+    /**
+     * Returns the spinner position of the status
+     * @return spinner position
+     */
+    abstract fun getPositionFromStatus(status: Int): Int
```
inorichi: There should be a better approach than spinner positions. I think this class should have:

`getStatusList()` which returns a list of the available status codes as an int (the value stored in the database)

`getStatus(status: Int): String` which returns the string from the code

```
// Maps a code to a string
val statusMap by lazy {
    getStatusList().associate { it to getStatus(it) }
}
```

```
// Returns the code from the string
fun getStatus(name: String): Int {
    statusMap.forEach { if (it.value == name) return it.key }
    throw Exception("Unknown status")
}
```

And for example, to display a list of status to the user, we can do: `statusMap.values`

I don't know if you are using it for something else, but IMO the services shouldn't know anything about positions.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/MangaSyncImpl.kt
inorichi: But why do we need two scores for every manga? A local score is pointless if you have the other IMO (and local score would belong to `Manga`).

I'll try this weekend to do a full review, don't merge it yet.

NoodleMage: untitled
Screenshot data is from Anilist and to lazy to make new one xD. I hope you get the gist.


The database has to be altered and the code gets unnecessary complicated if we move local score to the manga object (as far as I can think of). That's why I implemented it like this

inorichi: Ah, now I get what the remote score is. I don't think we should support that. At most I would provide a link to open in the browser that manga in the sync service.

The more fields we support the harder it will be to maintain (that score will probably be outdated after a few days and we have another field to refresh).

NoodleMage: I added it because it improves the layout in my opinion. But I understand where you're coming from. 

If I remove the score I will have to move it to the sync options. But I will do that some other time because I'm kinda done with this pull for a few days ;).

inorichi: Yeah I get it, this PR is big, I can help you this weekend.

I would prefer an approach without that score, otherwise we would have more things to consider, like if a manga is synced with 2 services, would that field show the average of the two average (remote) scores?

In the end it will just give more problems, I think opening a browser tab is easier and the user can see a lot more data, but that could be in a separate PR.

Hylak:
If feedback is welcomed, I would like to say something.

When looking for something to read the summary is the most important thing, and the new design is gonna force the scroll putting the new "buttons" and the "info" bar. So I think the bar should be removed, you already are on the info tab. 

Regarding the option buttons, in my opinion, would be better in the overflow menu because most of them are used never/once or in very few occasions (like share, sync options or open in browser). The FAB to add/remove from library is very useful as it is now, it's the most noticeable button as I think it should still be.

tl;dr: keep it simple and keep the summary and the useful data the most visible as possible.

Thanks for your work.

inorichi:
I agree share and open in browser won't be used a lot and it would make sense to place them in the overflow menu and have more room for the relevant data (the manga's information). But IMO sync options should be more accessible (maybe an always visible button in the toolbar if the user uses sync services?). The "MAL" tab was very simple/ugly and we thought about merging it in the info tab, but maybe it wasn't the best idea...

Hylak:
>maybe an always visible button in the toolbar if the user uses sync services?

I think that's a good solution, just 2 tabs seems fine and share/open in browser in the overflow menu.

NoodleMage:
Will do a mock-up with your feedback when I have some energy for it xD. 

If someone with more UX experience and a feel for material design could make a mock-up would be fantastic.

NoodleMage:
Closed because only collecting dust at the moment ;). Will reopen when there is more information about the design and the way sync gets implemented.

NotLemon:
Any updates on this one?

NoodleMage:
Because of the size of this pull I want to have enough free time to update the code in one go (including code review etc). Because this is only the case during the holidays I probably won't be able to update it before Christmas.

So unless someone else picks it up it will not be implemented before that time, sorry.

inorichi:
I'll work on this one next, however I'm going to keep the third tab (I guess I'll rename it to "Tracking" unless something better is suggested).

I had the idea of showing an icon in the tab if the manga has a binding so that you can easily know its status.

Flat:
Anything that needs to be done to move forward with this?

inorichi:
It's almost done, it's only missing a custom layout for every score type (currently it shows a number picker from 0 to 100).

Although there's a problem. I wrote the layout with the new ConstraintLayout, and it doesn't compile in FDroid (at least when I checked it last month with other apps using it).

PR #393, opened 2016-07-15
Changed minSdkVersion to 15 instead of 16

psychowood:
I’ve got an Onyx Book Android ebook running 4.0.4 . minSdkVErsion 16 prevented the app from being installed, while changing to 15 allowed it to build and install correctly (and, first tests confirmed that is compatible).

Are you using any >15 specific functionality I need to check for compatibility (and in case try to implement a workaround)?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,7 +35,7 @@ android {
 
     defaultConfig {
         applicationId "eu.kanade.tachiyomi"
-        minSdkVersion 16
+        minSdkVersion 15
         targetSdkVersion 23
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
         versionCode 9


inorichi:
API 16 added a huge lot of new methods and are used all over the app. I remember usages of these: 

[SQLiteDatabase::onConfigure](https://developer.android.com/reference/android/database/sqlite/SQLiteOpenHelper.html#onConfigure(android.database.sqlite.SQLiteDatabase))
[TaskStackBuilder](https://developer.android.com/reference/android/app/TaskStackBuilder.html)

But I would say there are a lot more. Some third-party library could also require API 16 (probably only the UI related ones).

psychowood:
Ok, so probably I was a little bit too optimistic :) I understand why you won't merge this pull request.

Anyway, since after using that build for a few hours I didn't notice any crash at all, maybe I was just lucky, even if I tested at least the main features (backup/restore, sync, browsing, download, reading), if it is ok for you I'll try to mantain an unofficial build just for SDK 15 (there are quite a few eink ereaders with Android 4.0, I'd dare to say most of them are below 4.1)

j2ghz:
I was also using API 15 device without any problems. I don;t have it anymore, but I still have the build configuration for it in https://github.com/inorichi/tachiyomi/wiki/FDroid-for-dev-versions , currently inactive. If there's interest in it, I can enable it.

inorichi:
I'm fine if you want to support a build below API 16, but I already have to spend a lot of time trying to support 16.

That said, I always try to use support libraries so almost everything should work (I think changing the app's theme will crash though, because of the TaskStackBuilder).

PR #394, opened 2016-07-18
Added the ability to view the library as a list

### LINKED ISSUE #224: Allow to display manga from library as a simple list
#35 for library

QuestofIranon:
Hi, I've been using the app for a while and wanted to help out so I've been working on some of the feature requests. In this one I implemented the feature requested in #224, the ability to display the library as a list.

Style comparisons:
With the frameLayout:
with-frame-layout

without the frameLayout:
without-frame-layout

Closes #224


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryAdapter.kt
```diff
@@ -5,6 +5,7 @@ import android.support.v4.app.FragmentManager
 
 import eu.kanade.tachiyomi.data.database.models.Category
 import eu.kanade.tachiyomi.ui.base.adapter.SmartFragmentStatePagerAdapter
```
inorichi: I would appreciate reverting the changes on this file, since they do nothing at all

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryCategoryAdapter.kt
```diff
@@ -84,11 +84,17 @@ class LibraryCategoryAdapter(val fragment: LibraryCategoryFragment) :
...
+        //depending on preferences, display a list or display a grid
+        val view = if(parent.id == R.id.library_list)
+            parent.inflate(R.layout.item_library_list)
+        else
+            parent.inflate(R.layout.item_catalogue_grid).apply {
+                card.layoutParams = FrameLayout.LayoutParams(MATCH_PARENT, coverHeight)
+                gradient.layoutParams = FrameLayout.LayoutParams(MATCH_PARENT, coverHeight / 2, Gravity.BOTTOM)
+            }
+
+        return LibraryListHolder(view, this, fragment)
```
inorichi: You are inflating the correct layout, but always returning the same holder. Change it to the usual if.

```
if (...) {
    val view = parent.inflate(...)
    return Grid(view, this, fragment)
} else {
    val view = parent.infalte(...)
    return List(view, this, fragment)
}
```

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryCategoryFragment.kt
```diff
@@ -94,6 +117,18 @@ class LibraryCategoryFragment : BaseFragment(), FlexibleViewHolder.OnListItemCli
             adapter.updateDataSet()
         }
 
+        toggleViewSubscription = libraryPresenter.viewToggleNotifier.subscribe() {onViewModeChange(it)}
```
inorichi: You could directly subscribe to the preference, 

```
toggleViewSubscription = preferences.libraryAsList()
        .subscribe { onViewModeChange(it) }
```

The event class created shouldn't be needed, nor the `BehaviorSubject` in the presenter. In the catalog I needed it to do something else, but here I don't see the need.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryCategoryFragment.kt
```diff
@@ -94,6 +117,18 @@ class LibraryCategoryFragment : BaseFragment(), FlexibleViewHolder.OnListItemCli
...
+
+        if(libraryPresenter.displayAsList != displayAsList)
```
inorichi: When subscribing to the preference above, you get an initial call to `onViewModeChange`, this shouldn't be neede.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
```diff
@@ -20,7 +20,9 @@ import eu.kanade.tachiyomi.ui.category.CategoryActivity
 import eu.kanade.tachiyomi.ui.main.MainActivity
 import eu.kanade.tachiyomi.util.toast
 import kotlinx.android.synthetic.main.activity_main.*
+import kotlinx.android.synthetic.main.fragment_catalogue.*
```
inorichi: Unwanted imports?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
```diff
@@ -178,6 +185,18 @@ class LibraryFragment : BaseRxFragment<LibraryPresenter>(), ActionMode.Callback
                 return true
             }
         })
+
+        //set the icon for the display mode button
+        displayMode = menu.findItem(R.id.action_library_display_mode).apply {
+            val icon = if (presenter.displayAsList)
```
inorichi: `if (preferences.libraryAsList().getOrDefault())`

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -47,6 +47,11 @@ class LibraryPresenter : BasePresenter<LibraryFragment>() {
     val libraryMangaSubject = BehaviorSubject.create<LibraryMangaEvent>()
 
     /**
+     * notifier of view switch
+     */
+    val viewToggleNotifier = BehaviorSubject.create<libraryToggleViewEvent>()
```
inorichi: Shouldn't be needed

inorichi:
Overall it looks good, thanks for helping!

If you need it, you can store the current mode in the presenter, but there's no need to create additional subscriptions and the Subject. If you want I can take care of this, but it will probably have to wait until the weekend.

I want to release a new version too before merging any commit (only bugfixes).

QuestofIranon:
I created the subscription/subject because if there is more than one category it wouldn't always update the view mode (even on the current category). I'll look it over and see if I can change it elsewhere to resolve the problem.

inorichi:
If every `LibraryCategoryFragment` (aka each category) subscribes to the preference, it will be updated, and new ones will be able to set the initial value with the subscription too.

inorichi:
And BTW, edit your first message to [automatically close the issue when merged.](https://help.github.com/articles/closing-issues-via-commit-messages/)

I edited the message in the other PR.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryCategoryFragment.kt
```diff
@@ -117,10 +124,9 @@ class LibraryCategoryFragment : BaseFragment(), FlexibleViewHolder.OnListItemCli
             adapter.updateDataSet()
         }
 
-        toggleViewSubscription = libraryPresenter.viewToggleNotifier.subscribe() {onViewModeChange(it)}
```
inorichi: Change this line to `subscribe { onViewModeChange(it) }` (note the absence of `()`).

QuestofIranon:
This is more of a style decision so I thought I'd ask. Right now I have the listview set up so the cover of the manga covers the entire list object. However, this results in the highlighting not showing up when the item is selected. I added a framelayout like the grid objects have to test it out and the results seem decent.

Should I leave it as it is or commit the framelayout? Would it be better to not have the cover at all?

I added comparison pictures to the pull request message.

inorichi:
I'd raher use a circle image view library like https://github.com/hdodenhof/CircleImageView to have a material design list: https://material-design.storage.googleapis.com/publish/material_v_8/material_ext_publish/0B6Okdz75tqQsNlAwTVZUcXB1M0U/components_lists_keylines_single9.png

QuestofIranon:
screenshot_20160719-190603

wopian:
^ imo it'd look even nicer if the unread chapters was grey text with no background.

QuestofIranon:
I was thinking about changing the unread chapters background to a box with rounded corners that uses the same blue as the title bar. 

I think you're right that it'd look better with just grey text though.

QuestofIranon:
Yeah, that definitely looks nicer.

screenshot_20160720-115522
screenshot_20160720-115501

inorichi:
From looking at the pictures I'd say the margins don't match with the spec. [They should be like this](https://github.com/DmitryMalkovich/material-design-dimens#metrics--keylines) (that library is already included).

NoodleMage:
@inorichi Sorry apparently merged to soon. Though is was done ;).

@QuestofIranon I want to update all the views with the values inorichi gave so I will also fix the overflow menu issue.

QuestofIranon:
@NoodleMage Ok

PR #395, opened 2016-07-18
Added a startup screen preference option

### LINKED ISSUE #348: Select starting screen
Choose whether to open Library, recently read or recent updates on startup.

Other idea is to allow creating shortcuts to open Library, recently read or recent updates directly.

QuestofIranon:
Added in the startup screen option requested in #348

Closes #348

### DIFF
--- a/app/src/main/res/values/keys.xml
+++ b/app/src/main/res/values/keys.xml
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
-    <string name="pref_start_screen_key">pref_start_screen_key</string>
-
     <string name="pref_category_general_key">pref_category_general_key</string>
     <string name="pref_category_reader_key">pref_category_reader_key</string>
     <string name="pref_category_sync_key">pref_category_sync_key</string>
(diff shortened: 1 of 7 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -93,6 +93,10 @@
     <string name="pref_theme">Application theme</string>
     <string name="light_theme">Main theme</string>
     <string name="dark_theme">Dark theme</string>
+    <string name="pref_start_screen">Start Up Screen</string>
+    <string name="start_screen_my_library">My Library</string>
+    <string name="start_screen_recent_read">Recently Read</string>
+    <string name="start_screen_recent_update">Recent Updates</string>
```
NoodleMage: No need for capitals. e.g. Start up screen -> Start up screen and My Library -> My library

inorichi: Better yet, don't create those and use the ones at the top of the file, `label_xxx`

app/src/main/res/values/keys.xml
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
+
+    <string name="pref_start_screen_key">pref_start_screen_key</string>
```
inorichi: Place this one where the other related preferences (keys from general tab) and remove `pref` `key` from the value (but keep that nomenclature in the key), I'm doing that for new settings, it was a bad decision I made months ago. :)

QuestofIranon: ah, so just start_screen then?

inorichi: Like this one, for example: `<string name="pref_last_used_category_key">last_used_category</string>`

QuestofIranon:
this should cover those changes

PR #411, opened 2016-08-01
Use RxPresenter from upstream

Gilfar:
The RxPresenter should be used from dependencies instead of having a copy in repository.


inorichi:
I added all the `startable*` methods there, but lately I'm just using `add()`. Did I remove all the usages on `startable*` methods?

Gilfar:
They weren't used when I tested this change, looks like they are now.
Still it's bad practice to do it this way, but I see why you did it this way to access private fields.
I'll just abandon this change for now.

PR #428, opened 2016-08-23
Add genre filter for catalogue

icewind1991:
Adds filtering by genre while browsing the catalogs (only implemented for kissmanga and batoto atm)

List of available filters is retrieved from the source, sources that don't implement filtering (or return 0 possible filters) have the filter menu item grayed out

cc @inorichi


inorichi:
I think the list of filters should be created manually in the source implementation. Maybe a `Map<String, String>` with key the label to show and value its internal value (for creating the request). It will avoid unneeded requests and will probably need less changes for the View/Presenter.

I think the `Filter` class should be in `OnlineSource` instead of `Source`. All sources in the catalogue will (at least for now) inherit from `OnlineSource` and I don't think the interface will ever need a `Filter` (I may be wrong though).

icewind1991:
I didn't want to hardcode them since that means the app needs to be updated when one of the sources decides to change the list of genres, this was a problem with the previous manga app I used once that stopped being updated.

If we just want to avoid making requests we could cache the filters and only update them infrequently but that would add even more code changes

inorichi:
IMO they don't change too often for it to be a problem. The same problem applies if the web structure changes (and that's the reason I don't like hardcoded sources at all, but since they already are it's not a big problem)

wopian:
Batoto genres change quite a bit. Also what would happen with new manga that get their genres added after you've browsed the category listings?

icewind1991:
Moved `Filter` stuff to `OnlineSource` and hard coded the list of filters, providing a small js snippet to get the list of filters for easy updating

inorichi:
@wopian They will show. It still uses an online search, but the genre list shown in the info tab won't be updated (unless you do it manually).

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/CataloguePresenter.kt
```diff
@@ -110,18 +113,23 @@ class CataloguePresenter : BasePresenter<CatalogueFragment>() {
         super.onCreate(savedState)
 
         source = getLastUsedSource()
+        view().subscribe {
```
inorichi: This line probably leaks the view.

I'll try this week to review and probably rewrite how the catalogue works. It's been a long time since I wrote this and there are some things I don't like. So don't mind this, I will take care of this part (self reminder).

app/src/main/java/eu/kanade/tachiyomi/data/source/online/OnlineSource.kt
```diff
@@ -428,4 +429,7 @@ abstract class OnlineSource(context: Context) : Source {
 
     }
 
+    data class Filter(val id: Int, val name: String)
```
inorichi: Id will probably have to be a string, mangafox uses an string in the query for example.

icewind1991:
Switched filter id to string and added mangafox support

inorichi:
https://github.com/inorichi/tachiyomi/commit/079d2de5ac993fe61e56af2c3677b92ea9214924 should survive rotation changes (moving the selected filters to the presenter) and it shows the progress bar when the filters are changed.

Though I haven't tested too much if everything is okay.

icewind1991:
@inorichi added the commit to this pr

inorichi:
https://github.com/inorichi/tachiyomi/commit/698b81592e1976c2af7d672090fecfea710d992a Slightly improved the implementation

inorichi:
I think this PR is ready for merge.

icewind1991:
added the commit

PR #431, opened 2016-08-28
Added read filter to chapter select.

### LINKED ISSUE #390: Delete ALL Downloaded and Read chapters
Version: r2. 2-1

Please add an option to delete all read and downloaded chapter in a selected Manga. The delete after reading does Not work for some reason. 

If possible add an option to  auto delete the penultimate chapter. 

Also please add "read" in the sorting. Currently "unread" and "downloaded" is not enough. 

Thanks and more great things may come your WY.

NoodleMage:
- Added read filter to chapter select.
- Can now select how far back the chapter should be deleted after read.

Closes #390 

screenshot_20160828-231601_framed
screenshot_20160905-102724_framed
screenshot_20160901-162002_framed

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
@@ -116,8 +116,18 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
 
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
         inflater.inflate(R.menu.chapters, menu)
+        menu.findItem(R.id.action_filter_read).isChecked = presenter.onlyRead()
         menu.findItem(R.id.action_filter_unread).isChecked = presenter.onlyUnread()
         menu.findItem(R.id.action_filter_downloaded).isChecked = presenter.onlyDownloaded()
+    }
+
+    override fun onPrepareOptionsMenu(menu: Menu) {
+        if (presenter.onlyRead())
+            //Disable unread filter option if read filter is enabled.
+            menu.findItem(R.id.action_filter_unread).isEnabled = false
+        if (presenter.onlyUnread())
+            //Disable read filter option if unread filter is enabled.
+            menu.findItem(R.id.action_filter_read).isEnabled = false
     }
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
@@ -127,7 +137,13 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
             R.id.action_sorting_mode -> showSortingDialog()
             R.id.action_filter_unread -> {
                 item.isChecked = !item.isChecked
+                presenter.setUnreadFilter(item.isChecked)
+                activity.supportInvalidateOptionsMenu()
+            }
+            R.id.action_filter_read -> {
+                item.isChecked = !item.isChecked
                 presenter.setReadFilter(item.isChecked)
+                activity.supportInvalidateOptionsMenu()
             }
             R.id.action_filter_downloaded -> {
                 item.isChecked = !item.isChecked
@@ -145,8 +161,7 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
 
     fun onNextManga(manga: Manga) {
         // Set initial values
-        setReadFilter()
-        setDownloadedFilter()
+        activity.supportInvalidateOptionsMenu()
     }
 
     fun onNextChapters(chapters: List<ChapterModel>) {
@@ -393,13 +408,5 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
 
     private fun setContextTitle(count: Int) {
         actionMode?.title = getString(R.string.label_selected, count)
-    }
-
-    fun setReadFilter() {
-        activity.supportInvalidateOptionsMenu()
-    }
-
-    fun setDownloadedFilter() {
-        activity.supportInvalidateOptionsMenu()
     }
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
@@ -116,18 +116,25 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
 
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
         inflater.inflate(R.menu.chapters, menu)
-        menu.findItem(R.id.action_filter_read).isChecked = presenter.onlyRead()
-        menu.findItem(R.id.action_filter_unread).isChecked = presenter.onlyUnread()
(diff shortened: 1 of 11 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
```diff
@@ -105,9 +109,10 @@ interface Manga : Serializable {
         const val SHOW_UNREAD = 0x00000002
         const val SHOW_READ = 0x00000004
         const val READ_MASK = 0x00000006
+        const val UNREAD_MASK = 0x00000008
```
inorichi: You don't need the `UNREAD_MASK`. Use `READ_MASK`.

The mask is used to allow only one value from the values it covers, and `0x6` covers `SHOW_ALL`, `SHOW_READ`, and `SHOW_UNREAD`.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
```diff
@@ -105,9 +109,10 @@ interface Manga : Serializable {
...
 
-        const val SHOW_DOWNLOADED = 0x00000008
-        const val SHOW_NOT_DOWNLOADED = 0x00000010
+        const val SHOW_DOWNLOADED = 0x00000010
+        const val SHOW_NOT_DOWNLOADED = 0x00000012
```
inorichi: These two should be reverted. It breaks compatibility and there's no need to change them.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
```diff
@@ -80,6 +80,10 @@ interface Manga : Serializable {
         get() = chapter_flags and READ_MASK
         set(filter) = setFlags(filter, READ_MASK)
 
+    var unreadFilter: Int
```
inorichi: These can be removed too. You can achieve the same with `readFilter`

app/src/main/res/values/arrays.xml
```diff
@@ -48,6 +48,22 @@
         <item>3</item>
     </string-array>
 
+    <string-array name="remove_after_read_slots">
+        <item>Last chapter</item>
```
inorichi: This should be pointing to a string resource to allow translations.

Also I'm not sure if "penultimate" is used in english. It would be nice if a native could take a look here.

app/src/main/res/values/keys.xml
```diff
@@ -41,6 +41,7 @@
 
     <string name="pref_download_directory_key">pref_download_directory_key</string>
     <string name="pref_download_slots_key">pref_download_slots_key</string>
+    <string name="pref_remove_after_read_slots_key"> pref_remove_after_read_slots_key</string>
```
inorichi: Remove the space and the prefix "pref_" and suffix "_key" from the value.

Old keys haven't changed because compatibility, but new preferences shouldn't use them.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -355,20 +356,20 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
...
-                                if (prevChapter != null) {
-                                    deleteChapter(prevChapter, manga)
-                                }
-                            } else {
+                            if (prefs.removeAfterReadSlots() == 0) {
+                                // Remove current read chapter
                                 deleteChapter(chapter, manga)
+                            } else {
+                                // Remove previous chapter specified by user in settings.
+                                getAdjacentChaptersStrategy(chapter,prefs.removeAfterReadSlots()).first?.let {  deleteChapter(it, manga) }
```
inorichi: `prefs.removeAfterReadSlots()` is called twice in this method. Introduce a local variable for it.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersFragment.kt
```diff
@@ -116,18 +116,34 @@ class ChaptersFragment : BaseRxFragment<ChaptersPresenter>(), ActionMode.Callbac
 
     override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
         inflater.inflate(R.menu.chapters, menu)
+        menu.findItem(R.id.action_filter_read).isChecked = presenter.onlyRead()
```
inorichi: It's probably better to move these 3 to "onPrepareOptionsMenu" and use less calls to `menu.findItem()`

app/src/main/res/values/arrays.xml
null-dev: http://www.merriam-webster.com/dictionary/penultimate

NoodleMage: I used penultimate from issue. (Fun fact: four to last is apparently Preantepenultimate ;)).

But to avoid confusion and make it more coherent I suggest:
- Last read chapter
- Second to last chapter
- Third to last chapter
- Etc.

As for using hard coded strings. Shame on me ;(.

Flat:
>Also I'm not sure if "penultimate" is used in english. It would be nice if a native could take a look here.

As a native English speaker: Penultimate is very rarely used and I wouldn't be surprised if many don't know what it means. Most people could probably figure it out from context though.

wopian:
As well as penultimate being rarely used, it feels out of place with the rest of the options being "X to last chapter".

Using "Second to last chapter" would flow nicer and is more commonly used.

app/src/main/res/xml/pref_downloads.xml
```diff
@@ -35,15 +35,16 @@
         <SwitchPreference
             android:defaultValue="false"
             android:key="@string/pref_remove_after_read_key"
-            android:summary="@string/current_chapter"
+            android:summary="@string/sum_remove_after_read"
```
inorichi: I think this setting should be merged with the ListPreference. The first value wouldn't delete anything.

NoodleMage: I don't exactly know what you mean by this?
Without this SwitchPreference it won't be possible to either not delete read chapter or delete the last read chapter.

Also with a offset of -1 you will delete the next chapter if downloaded?!

inorichi: What I mean is, the preference would be:

- Disabled (-1, default value).
- Last read chapter (0).
- Second to last (1).
- Third to last (2).
- Fourth to last (3).
- Fifth to last (4).

Then in the `ReaderPresenter` you would do:

```
if (chapter.read) {
	// Check if remove after read is selected by user.
	val removeAfterReadSlots = prefs.removeAfterReadSlots()
	when {
		// Remove current read chapter
		removeAfterReadSlots == 0 -> deleteChapter(chapter, manga)
		// Remove previous chapter specified by user in settings.
		removeAfterReadSlots > 0 -> getAdjacentChaptersStrategy(chapter, removeAfterReadSlots)
				.first?.let { deleteChapter(it, manga) }
	}
}
```

You can use an `if, else if` if you don't like the `when`.

inorichi: Even better:

```
when (removerAfterReadSlots) {
    // Setting disabled
    -1 -> {}
    // Remove current read chapter
    0 -> deleteChapter(chapter, manga)
    // Remove previous chapter specified by user in settings.
    else -> getAdjacentChaptersStrategy(chapter, removeAfterReadSlots)
                .first?.let { deleteChapter(it, manga) }
}
```

PR #433, opened 2016-08-29
Add sync integration with TachiWeb.

null-dev:
This PR attempts to add sync integration with TachiWeb.

Known issues:
- Syncing the library breaks some thumbnails
- Not sure what to use as the icon for the sync button in the Library screen
- Sync is slower than I would like it to be

Don't merge this yet as I have yet to do any extensive testing.

The live demo is already updated with the sync framework so you can test with `http://64.137.238.70:4567/api` as the "sync endpoint" if you want.

If you would like to test with your own server the format for the "sync endpoint" is simply: `http://server-ip:server-port/api`

screener_20160829 12-46-31
screener_20160829 12-47-02
screener_20160829 12-47-34
screener_20160829 12-47-20


inorichi:
* So... covers that were working in Android are broken in both applications?
* That icon is fine IMO.
* I would remove the "Sync favorites only" preference and make it always true. It's not viable to synchronize everything.

Maybe it's time to rename "Sync" to "Tracking" or something like that? Having "Sync" and "Library sync" looks weird to me.

I haven't started reviewing the PR, and I probably won't have much time until the weekend to fully review it.

null-dev:
The covers only break in Tachiyomi (and it I THINK it only affects Batoto).

I can't reproduce it anymore :(...

I'll mess around with it and get back to you if I can reproduce it still...

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
```diff
@@ -209,6 +216,11 @@ class LibraryFragment : BaseRxFragment<LibraryPresenter>(), ActionMode.Callback
             }
         })
 
+        //Initialize sync button
+        val syncItem = menu.findItem(R.id.action_sync_library)
+        preferences.enableLibrarySync().asObservable().subscribe {
```
inorichi: This leaks the fragment. You should unsubscribe in `onDestroyOptionsMenu`. 

If its lifecycle is not consistent (I don't think I tried subscriptions inside the options menu), use a non observable approach (`.getOrDefault()` instead of `.asObservable`) and subscribe during `onViewCreated` with a `.skip(1)` to ignore the first event, and inside the subscription invalidate the options menu. Always unsubscribe in the opposite lifecycle, which in this last case is `onDestroyView`.

null-dev: Oh wow, I didn't know I have to unsubscribe from observables.
Uh oh, now I have quite the mess to clean up...
I'll work on it later today.

inorichi: Observables are unsubscribed when either `onCompleted` or `onError` is called. But some observables are always active (they don't emit `onCompleted` or `onError`).

In this case the preference is always waiting for updates, that's why you have to manually unsubscribe.

`Observable.just(1, 2, 3).subscribe { println(it) }` doesn't really need an unsubscription, because the source observable emits `onCompleted()` after `onNext(3)`.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
```diff
@@ -48,6 +50,11 @@ class LibraryFragment : BaseRxFragment<LibraryPresenter>(), ActionMode.Callback
     val preferences: PreferencesHelper by injectLazy()
 
     /**
+     * Sync
+     */
+    val sync: LibrarySyncManager by injectLazy()
```
inorichi: I would inject this in the presenter, the only injection I use in views are the preferences because they are neede everywhere.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
null-dev: Yeah, I totally forgot to consider that.

I had a feeling that RxJava was too good to be true the whole time, I was like: "Oh, I can just leave these here and the whole thing will just dynamically update, magical!".

app/src/main/java/eu/kanade/tachiyomi/ui/library/sync/LibrarySyncDialogFragment.kt
```diff
@@ -0,0 +1,43 @@
...
+        //Create new dialog, call this before runSync to ensure runSync has a dialog to work with
+        val dialog = librarySyncDialog.createDialog()
+        //Run sync if it has not yet
+        if (savedState == null) {
+            librarySyncDialog.runSync()
+        }
+        return dialog
+    }
+
+    //http://stackoverflow.com/questions/14657490/how-to-properly-retain-a-dialogfragment-through-rotation
```
inorichi: I always try to avoid fragment retaining. You should consider creating a presenter for this dialog and doing your background tasks there, then sending the updates to the ui with RxJava's subscriptions. (Everything is done that way in the rest of the app).

null-dev:
I think I have applied most of your suggestions (let me know if the way I implemented the presenter was bad).

Travis CI has decided not to build my PR because it can't connect to Google :cry: ...

null-dev:
Oh, and should I rename the old "Sync" to "Tracking" now?

inorichi:
Let's rename it in a different PR, this one is big enough already :)

Rebase without squashing for now.

app/src/main/java/xyz/nulldev/ts/sync/RxSyncClient.kt
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Sync libraries asynchronously
+     * @param oldLibrary             The old library
+     * @param newLibrary             The new library
+     */
+    fun syncLibrariesWithProgress(
+            oldLibrary: String,
+            newLibrary: String): Observable<SyncResult> {
+        return Observable.create {
```
inorichi: You can name the parameter inside the lambda. `Observable.create { sub ->`

A few tips:
* You can use `subscribe` only once at the end of the sequence, everything else has to be done chaining operators, probably with just `flatMap` you can achieve the same.
* Avoid `Observable.create` that's the last resort. Start the sequence with your `trySync`. 

A quick example:

```
trySync(params)
        .flatMap { taskId ->
            // do something with your task and return an observable
        }
```

null-dev:
OK, I will close this for now as I feel TachiWeb is not stable enough. I don't want to have to keep making PRs when the API changes so I will wait until TachiWeb stabilizes before making another PR.

I will keep a fork of Tachiyomi in the TachiWeb repo however, up to date with the latest release so if you do want a version that has syncing implemented, you can just download it from there.

PR #434, opened 2016-08-29
Custom color filter for reader

### LINKED ISSUE #432: "Night Mode" colour temperature option
I'd like to see an optional filter that colours the screen. You currently have something similar implemented in that adjusting the brightness below 0% introduces a black filter. What I'd like is a separate option to choose a "permanent"/"always on" colour filter regardless of brightness level.

I specifically put "night mode" in the title because I was thinking of a toggle that made reader mode turn to a warmer colour filter (red or orange) to aid in reading in dark/dimly lit areas, but i recall using a reader a long time ago that gave the user a choice from preset colours if this becomes something other people like and want a different colour other than red/orange.

NoodleMage:
screenshot_20160916-222408_framed
screenshot_20160916-222429_framed
If someone has a royalty free manga image that is better than the current stock image please let me know ;).

Closes #432


inorichi:
It may be a good idea to also move the brightness setting to this dialog.

I would store everything in a single integer (if it's not too complicated), in ARGB format for example.

`0xff000000` mask for alpha,
`0x00ff0000` mask for red.
`0x0000ff00` mask for green.
`0x000000ff` mask for blue.

inorichi:
Testing this, I've noticed the UI lags when moving the seekbar.

Adding `.throttleLast(100, TimeUnit.MILLISECONDS, AndroidSchedulers.mainThread())` to the preference subscriptions improves it.

app/src/main/java/eu/kanade/tachiyomi/widget/NegativeSeekBar.kt
```diff
@@ -0,0 +1,61 @@
...
+        })
+    }
+
+    override fun setProgress(progress: Int) {
+        super.setProgress(Math.abs(minValue) + progress)
+    }
+
+    fun setMinSeek(minValue: Int) {
+        this@NegativeSeekBar.minValue = minValue
+        super.setMax(this@NegativeSeekBar.maxValue - this@NegativeSeekBar.minValue)
```
NoodleMage: Kinda annoying that it doesn't disappear after code change ;-(.

PR #435, opened 2016-08-30
Open from homescreen/add shortcut to launcher

null-dev:
_I know I'm supposed to focus on one PR at a time but..._

**This PR enables opening manga from the homescreen via launcher shortcuts!**

Known issues:
- None currently? If you find one, please tell me.

screener_20160830 14-25-50


Rumple007:
Just wondering, would it be like difficult to use a widget instead, like a scrollable home screen widget that shows like a catogery of the library

null-dev:
@Rumple007 Actually, I really wanted a widget but I have zero experience writing widgets and the process of creating a widget is probably far more complex than just adding a couple of shortcuts.

Still, I would recommend you open an issue if you want this feature and maybe me or another developer can take a shot at it later.

inorichi:
Sorry I've delayed this PR for too long.

I've tried it and I would only support the manga image as icon. Shapes are still fine.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoFragment.kt
```diff
@@ -159,6 +186,134 @@ class MangaInfoFragment : BaseRxFragment<MangaInfoPresenter>() {
...
+        fun BitmapRequestBuilder<out Any, Bitmap>.toIcon(): Bitmap {
+            return this.into(96, 96).get()
+        }
+
+        MaterialDialog.Builder(activity)
+                .title(R.string.icon_shape)
+                .negativeText(android.R.string.cancel)
+                .items(modes.map { getString(it) })
+                .itemsCallback { dialog, view, i, charSequence ->
+                    thread {
```
inorichi: Is this that slow to require a thread? I prefer to block the UI just a little.

null-dev: I don't think it's slow but Glide forces you to do it in another thread (or it throws an Exception telling me to use a background thread).

Should I convert it into a runnable?

inorichi: Oh, I would use the observable approach.

```
Observable.fromCallable {
    //bitmap conversion
}
.subscribeOn(Schedulers.io())
.observeOn(AndroidSchedulers.mainThread())
.subscribe({ bitmap -> createShortcut(addIntent, bitmap)}
```

If it can throw exceptions, add the error lambda to the subscription.

PR #444, opened 2016-09-01
Also check if we need to load the next chapter when entering the pager

icewind1991:
This is needed for single page chapters or when you're resuming the chapter on the last page


inorichi:
> or when you're resuming the chapter on the last page

Are you sure? `setActivePage(currentPage)` (line 222) should call `onPageChanged()`

inorichi:
Also I'm not sure that's the best place for that. In the case the view isn't initialized yet, it won't be called (I don't remember if it can happen though), but `setPagesOnAdapter` should be better.

It would be nice that `onPageChanged` is called only once when you open the reader, `setActivePage(currentPage)` only calls the method if the index is different IIRC. In that case you should only call `onPageChanged` if `currentPage == pager.currentItem`

icewind1991:
Ok, I wasn't sure about the resuming use case, I remembered having issues with it sometimes while submitting the PR

inorichi:
7c3cd10

PR #449, opened 2016-09-03
Complete auto updates checker

inorichi:
This one has been around here for months without being completed. Today I remembered about this feature and decided to finish this.

* ~~The updates checker and downloader now run inside a `IntentService` (instead of an `AsyncTask`), and the alarm only starts the service instead of handling background work.~~ Rewritten with a better approach, `GcmNetworkManager`.
* The logic to find out if there's a new version has been removed from the UI. To achieve this the method from the API now returns a sealed class `GithubUpdateResult`, which can be `NewUpdate(val release: GithubRelease)` or `NoNewUpdate()`. The UI now works with these two.

cc @NoodleMage (original author of this feature) code review? :)


app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderAlarm.kt
```diff
@@ -70,40 +63,10 @@ class UpdateDownloaderAlarm : BroadcastReceiver() {
 
     override fun onReceive(context: Context, intent: Intent) {
         when (intent.action) {
-        // Start the alarm when the system is booted.
+            // Start the alarm when the system is booted.
             Intent.ACTION_BOOT_COMPLETED -> startAlarm(context)
```
NoodleMage: Alarm is only started on boot or by [toggle switch](https://github.com/inorichi/tachiyomi/blob/d3998f892c25f8625ff5eaec4121101d5666e4bb/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAboutFragment.kt#L66).

Because the switch is enabled by default the automatic check for updates won't start until reboot. This can be problematic for tablets that don't get rebooted for example. Unless I'm missing something of course ;).

app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderService.kt
```diff
@@ -0,0 +1,262 @@
...
+         */
+        const val ACTION_CHECK_VERSION = "eu.kanade.CHECK_UPDATE"
+
+        /**
+         * Action to download a new update.
+         */
+        const val ACTION_DOWNLOAD_UPDATE = "eu.kanade.DOWNLOAD_UPDATE"
+
+        /**
+         * Look for updates and notify the user if there's a new version with a notification.
```
NoodleMage: I see you formatted the comments as:

```
/**
* this is a comment.
* @param value this is a value
*/
```
In other commits you used:
```
/**
* this is a comment.
*
* @param value this is a value
*/
```

Which one do you want to use in the future?

NoodleMage:
Rest looks good to me ;) 👍

app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderService.kt
inorichi: According to the [official doc](https://kotlinlang.org/docs/reference/kotlin-doc.html) only classes documentation have a line break.

I've been using a lot the line break, but I guess it should be removed if we want to follow the doc.

app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderAlarm.kt
inorichi: Valid point. I'll check if an application restart triggers the alarm.

inorichi: I think I'll rewrite this with [GcmNetworkManager](https://developers.google.com/android/reference/com/google/android/gms/gcm/GcmNetworkManager).

inorichi:
Ended up rewriting almost everything. Now there's the updates checker and the update downloader, both as separate services. The first one is used by the Gcm periodic task, and the second one as the result of clicking download.

Still have to test if the periodic task is automatically restored after an actual app update.

PR #456, opened 2016-09-07
HotFix for 1-page mangas not properly updating with MAL

iBaer:
The updating routine starts with `onBackPressed()` in `ReaderActivity.kt`
It will ask for the chapter number which should be used for the update with `getMangaSyncChapterToUpdate()` in `ReaderPresenter.kt`

The problem here is that 1-page chapters are not marked as read when this happens.
This will happen with `onChapterLeft()` in `ReaderPresenter.kt` but only after the previous function was called, thus I made a simple fix which checks if the read chapter was only 1 page in size and uses this for the chapter number.

There might be a better solution if `onChapterLeft()` would be called before `onBackPressed()` (or `getMangaSyncChapterToUpdate()` in particular). Maybe `onChapterLeft()` could be called from within `onBackPressed()` instead of in `onDestroy()` (`ReaderPresenter.kt`) where I think it's currently called from.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
@@ -421,8 +421,11 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
 
         val prevChapter = prevChapter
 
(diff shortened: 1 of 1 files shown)


inorichi:
#444 Should fix this issue

iBaer:
You are right. #444 might also fix a bunch of other issues at once.

inorichi:
7c3cd10 should fix it

PR #466, opened 2016-09-10
Hide catalogues

### LINKED ISSUE #447: Hide Individual Sources
I'm starting this Issue because I don't think a pull request is the place to discuss a new feature

I imagine the simplest way of doing this would be something along the lines of adding a new entry to pref_sources, a list of some sort.

What I want to do is much more work and I probably can't even handle it (I detest UI) but I'm going to try.
I want to make the Spinner swipe-able, in that a user can remove a catalog from the spinner directly.

inorichi:
Closes #447 

This PR allows to select a list of languages and individual sources to display, allowing to hide the ones unused by the user.

I think this selection should be available in the "Sources" screen (see screenshots below), merging the functionality of login with enabling/disabling individual sources.

Known issues:

* Deselecting all the sources will cause a crash in the catalogue (it doesn't expect an empty source list).

Screens:

qemu-system-i386_2016-09-10_18-53-34
qemu-system-i386_2016-09-10_18-54-06


inorichi:
Made it top level with login support.

qemu-system-i386_2016-09-11_12-09-28

ThePreviousOne:
Looks good to me, but what about just "Don't have a checkbox for Batoto" put the login icon there instead?

inorichi:
I think I'll leave it like this for now. Maybe one day sources with login will have additional functionality.

PR #481, opened 2016-09-29
Added option to download page or set page as cover

### LINKED ISSUE #457: Download single page while reading
A lot of times I'm reading something and found just one page amusing and try to download it but it looks like you can't do it on tachiyomi while reading. Is there any way or is it totally not possible?

NoodleMage:
Closes #457 

TODO
- [x] code/ui cleanup
## Screens [WIP]

screenshot_1475852335_framed
- Set as cover
- Share
- Save

screenshot_1475852464_framed


inorichi:
Why not just use the `imagePath` from the `Page` object and copy the file?

NoodleMage:
https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/util/OkioExtensions.kt#L55

inorichi:
I'm going to remove that function some day. If it needs to be a jpg it's fine, but you can use the downloaded image as a `BufferedSource` , no need to go through the network.

NoodleMage:
I think I don't fully understand you. What has to be a jpg?

Also the downloaded image is already a `BufferedSource `->` response.body().source()`

If you mean the download image of the page object, then I still have the problem that the quality is lower than the original image if someone enabled reencode to compress the image.

inorichi:
I plan to remove the reencode setting, it's not meant to be always enabled.

> I think I don't fully understand you. What has to be a jpg?

I thought you wanted to use only a jpg when you linked that line because the image format may not be known (the cache doesn't store the extension).  If the image comes from the internal cache you don't know the file format, but you can retrieve it with the `imageUrl` from `Page` (I did that in the download manager).

So, it's better to just use the already downloaded image.

NoodleMage:
> I thought you wanted to use only a jpg when you linked that line 

Sorry, should've been a bit more clear on what I meant ;). 

The image is now copied from the image location given by the `Page `object,

inorichi:
Now that I think about it, there's not even a need to copy the file, you can use the original for everything.

And... I wouldn't add more menu items, IMO this feature fits better with a long tap on the image (showing a dialog). With the webtoon reader it's not too clear which image you are trying to use/save, and you may not be able to use the first one with the menu (if the second image is partially shown).

NoodleMage:
> Now that I think about it, there's not even a need to copy the file, you can use the original for everything.

The image need to copied to a location the user can access. (in this case /sdcard/pictures). Otherwise you have to download > get notification > share > copy with explorer (if you have one) and place it in the desired location. This will make the feature kind of useless.

> And... I wouldn't add more menu items, IMO this feature fits better with a long tap on the image (showing a dialog). 

I like this idea. Will try to code it and check if it doesn't interfere (unnecessary pop-up) while reading.

inorichi:
> Otherwise you have to download > get notification > share > copy with explorer (if you have one) and place it in the desired location. This will make the feature kind of useless.

Can't it start from the "share" step? Download is not needed using the source file, and the notification is used to click on share or delete. Delete is also not needed because it uses the source image.

Basically, I don't see any need for a notification.

NoodleMage:
> Can't it start from the "share" step? Download is not needed using the source file, and the notification is used to click on share or delete. Delete is also not needed because it uses the source image.

IMO no. The purpose of this feature for me is that I can quickly download a file and it's placed in a folder. If I have to save the file via an explorer I can just download the manga which is just as quick. Only using share would more annoy me then appreciate the download. 

Let me give a scenario why. Let's say I want to download a picture for later but also share it with someone. 

With notification (note that I can also look at the picture when I press it) -> moved to pictures (thus in gallery) + share option for friend. 

Without I have to download. I get options to share but I want to save. I don't know how because I'm a basic user who has never heard of a file explorer ;). But let's say I do have one. I have to save the file. Then open my explore. Go to the file. Select that file and share it. Which is really inconvenient.

inorichi:
If the user wants to share an image through a messaging app to a friend, he has to copy the image to the pictures folder, and then click the share button. IMO this is not ideal at all.

The best solution I can think of is to have two options with a long tap in the image: Share and save.

The first one opens the share intent, the second one copies the image to a folder (without any notification). I'm not sure if it would be better to copy the image to the downloads directory instead.

NoodleMage:
Adding a share option is possible (though redundant IMO).
Removing the notification... Not an option in my opinion for multiple reasons. You complicate a lot of use cases. E.g. setting an image as background needs to be done with a different app. With the notification you're done with one click wherein just a download forces the user to search for the image. 
Notification is also convenient with webtoons to see a preview of the downloaded image. 

Maybe someone else can give a different view because I don't feel like we can agree on this one ;)

inorichi:
To set an image as background, you can use the share, open with the gallery app, and use the set as background from there (or you can add another option in the dialog to open the image with the gallery app).

Also it's possible that the image preview crashes the app because of an OutOfMemory error. The proper solution is to read the image dimensiones before trying to decode, and then load the image with a lower resolution if needed.

The only advantage I see in the notification is that you have a preview, but you are already seeing the image in the reader.

inorichi:
screenshot_2016-10-07-18-24-55-852_eu kanade tachiyomi debug

Another reason notifications aren't reliable. Every SO does what he wants. The screenshot is from MIUI 8, the only thing I can do is click in the notification and it seems to execute the first action (share).

But ignoring that issue, here's what I would like to see:

*  I can accept a notification only when the user clicks in "save image", but adding new settings for notifications is just overkill.

* Share image and Save image should use `action_share` and `action_save` as strings respectively.

* A long press should provide a `Page?` as parameter (the issue with the webtoon reader is still there). I guess this involves calling methods from the layout manager to retrieve the item at a certain tap.

app/src/main/java/eu/kanade/tachiyomi/data/download/ImageNotifier.kt
```diff
@@ -0,0 +1,138 @@
...
+import com.bumptech.glide.request.target.SimpleTarget
+import eu.kanade.tachiyomi.Constants
+import eu.kanade.tachiyomi.R
+import eu.kanade.tachiyomi.util.notificationManager
+import java.io.File
+
+/**
+ * Class used to show BigPictureStyle notifications
+ */
+class ImageNotifier(private val context: Context) {
```
inorichi: Move this class and the broadcast receiver to `ui/reader`, it's too tied to the reader.

I would also implement the broadcast receiver inside this one, if it doesn't make the entire class too big.

app/src/main/java/eu/kanade/tachiyomi/data/download/ImageNotifier.kt
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Status of download. Used for correct notification icon.
+     */
+    private var isDownloading = false
+
+    /**
+     * Called when download progress changes.
+     * @param progress progress value in range [0,100]
+     */
+    fun onProgressChange(progress: Int) {
```
inorichi: There shouldn't be any progress to show afaik.

app/src/main/java/eu/kanade/tachiyomi/data/download/ImageNotifier.kt
```diff
@@ -0,0 +1,138 @@
...
+                /**
+                 * The method that will be called when the resource load has finished.
+                 * @param resource the loaded resource.
+                 */
+                override fun onResourceReady(resource: Bitmap?, glideAnimation: GlideAnimation<in Bitmap>?) {
+                    setLargeIcon(resource)
+                    onNotify()
+                }
+            })
+            Glide.with(context).load(file).asBitmap().into(object : SimpleTarget<Bitmap>(720, 1280) {
```
inorichi: Can you try using this resource for `setLargeIcon` too? Does it look worse? This way we would only decode 1 time.

app/src/main/java/eu/kanade/tachiyomi/data/download/ImageNotifier.kt
```diff
@@ -0,0 +1,138 @@
...
+            // Share action
+            addAction(R.drawable.ic_share_white_24dp,
+                    context.getString(R.string.action_share),
+                    ImageNotificationReceiver.shareImageIntent(context, file.absolutePath, notificationId))
+            // Delete action
+            addAction(R.drawable.ic_delete_white_24dp,
+                    context.getString(R.string.action_delete),
+                    ImageNotificationReceiver.deleteImageIntent(context, file.absolutePath, notificationId))
+        }
+        onNotify()
```
inorichi: I think it's better to wait for the image to be decoded instead of notifying multiple times.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -228,6 +228,21 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
         // Ignore
     }
 
+    fun onLongPress(page: Page) {
+        MaterialDialog.Builder(this).apply {
+            title = "Choose"
```
inorichi: String resource

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -393,16 +408,16 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
 
     private fun setRotation(rotation: Int) {
         when (rotation) {
-            // Rotation free
+        // Rotation free
```
inorichi: This change looks worse IMO.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -32,6 +42,11 @@ import java.util.*
 class ReaderPresenter : BasePresenter<ReaderActivity>() {
 
     /**
+     * Network helper
+     */
+    private val network: NetworkHelper by injectLazy()
```
inorichi: Not used?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerReader.kt
```diff
@@ -183,6 +183,11 @@ abstract class PagerReader : BaseReader() {
                 }
                 return true
             }
+
+            override fun onLongPress(e: MotionEvent?) {
+                super.onLongPress(e)
+                readerActivity.onLongPress(adapter.pages!![pager.currentItem])
```
inorichi: Just in case use `.getOrNull(pager.currentItem)` and call the activity if it's not null (or handle the nullability in the activity).

Also wrap the call inside `if (isAdded) { ... }`. I changed it for the single press a few days ago (it crashed the app sometimes).

Btw, is the super actually executing something?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonReader.kt
```diff
@@ -138,6 +138,13 @@ class WebtoonReader : BaseReader() {
                 }
                 return true
             }
+
+            override fun onLongPress(e: MotionEvent) {
+                super.onLongPress(e)
+                val child = recycler.findChildViewUnder(e.rawX, e.rawY)
+                val position = recycler.getChildAdapterPosition(child)
+                readerActivity.onLongPress(adapter.getItem(position))
```
inorichi: Also use `getOrNull` here

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -89,6 +103,20 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
     private val source by lazy { sourceManager.get(manga.source)!! }
 
     /**
+     * Used to show image notification
+     */
+    val imageNotifier by lazy { ImageNotifier(context) }
```
inorichi: If its usage is only inside a method, create it inside the method.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerReader.kt
```diff
@@ -183,6 +184,16 @@ abstract class PagerReader : BaseReader() {
...
                 return true
             }
+
+            override fun onLongPress(e: MotionEvent?) {
+                if (isAdded) {
+                    val page = adapter.pages?.getOrNull(pager.currentItem)
+                    if (page != null)
+                        readerActivity.onLongPress(page)
+                    else
+                        context.toast(getString(R.string.saving_picture_error))
```
inorichi: `R.string.unknown_error`? Just to save further translations

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -228,6 +228,21 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
         // Ignore
     }
 
+    fun onLongPress(page: Page) {
+        MaterialDialog.Builder(this).apply {
```
inorichi: I wouldn't use `apply` here. It will feel better without it.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -508,4 +531,84 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
         db.insertManga(manga).executeAsBlocking()
     }
 
+    /**
+     * Update cover with page file.
+     */
+    internal fun setCover(page: Page) {
+            // Update cover to selected file, show error if something went wrong
```
inorichi: 8 spaces here

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -508,4 +531,84 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
...
+            try {
+                if (editCoverWithStream(File(page.imagePath).inputStream(), manga)) {
+                    context.toast(R.string.cover_updated)
+                } else {
+                    throw Exception("Stream copy failed")
+                }
+            } catch(e: Exception) {
```
inorichi: Call it `error`, and log `error`, not just the message

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -508,4 +531,84 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
...
+
+    /**
+     * Called to copy image to cache
+     * @param inputStream the new cover.
+     * @param manga the manga edited.
+     * @return true if the cover is updated, false otherwise
+     */
+    @Throws(IOException::class)
+    private fun editCoverWithStream(inputStream: InputStream, manga: Manga): Boolean {
+        if (manga.thumbnail_url != null && manga.favorite) {
```
inorichi: When this method returns false, it's just going to throw an exception (in the method above). I would get rid of this method and implement the other like:

```
try {
    if (manga.thumbnail_url != null && manga.favorite) {
        coverCache.copyToCache(manga.thumbnail_url!!, inputStream)
        context.toast(R.string.cover_updated)
    } else {
        throw Exception("Thumbnail url null or manga not favorite")
    }
} catch (error: Exception) {
    context.toast(R.string.notification_manga_update_failed)
    Timber.e(error)
}
```

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -508,4 +531,84 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
...
+    }
+
+    fun shareImage(page: Page) {
+            val shareIntent = Intent().apply {
+                action = Intent.ACTION_SEND
+                putExtra(Intent.EXTRA_STREAM, Uri.parse(page.imagePath))
+                flags = Intent.FLAG_ACTIVITY_NEW_TASK
+                type = "image/jpeg"
+            }
+            context.startActivity(Intent.createChooser(shareIntent, context.resources.getText(R.string.action_share))
```
inorichi: Definitely not!

The presenter mustn't control navigation. (In fact I wanted to remove the context from the presenter but it's handy sometimes). At most I allow showing toasts or data related things, like obtaining a directory.

This method should be implemented in the view.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -508,4 +531,84 @@ class ReaderPresenter : BasePresenter<ReaderActivity>() {
...
+
+        //Check if file doesn't already exist
+        if (destFile.exists()) {
+            imageNotifier.onComplete(destFile)
+        } else {
+            if (inputFile.exists()) {
+                // Copy file
+                Observable.fromCallable { inputFile.copyTo(destFile) }
+                        .observeOn(AndroidSchedulers.mainThread())
+                        .subscribeOn(Schedulers.io())
```
inorichi: place `subscribeOn` above. You also have to close both `inputFile` and `destFile` or very bad things will happen.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/notification/ImageNotificationReceiver.kt
```diff
@@ -0,0 +1,108 @@
...
+     */
+    private fun shareImage(context: Context, path: String) {
+        val shareIntent = Intent().apply {
+            action = Intent.ACTION_SEND
+            putExtra(Intent.EXTRA_STREAM, Uri.parse(path))
+            flags = Intent.FLAG_ACTIVITY_NEW_TASK
+            type = "image/jpeg"
+        }
+        context.startActivity(Intent.createChooser(shareIntent, context.resources.getText(R.string.action_share))
+                .apply { flags = Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_MULTIPLE_TASK })
```
inorichi: `flags` being updated twice?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/notification/ImageNotifier.kt
```diff
@@ -0,0 +1,92 @@
...
+     * Id of the notification.
+     */
+    private val notificationId: Int
+        get() = Constants.NOTIFICATION_DOWNLOAD_IMAGE_ID
+
+    /**
+     * Called when image download/copy is complete
+     * @param file image file containing downloaded page image
+     */
+    fun onComplete(file: File) {
```
inorichi: I would create an additional method `onBitmapReady(file, bitmap)` and call it inside the anonymous class.

Also disable disk cache for this bitmap (it shouldn't be used more than once per image).

app/src/main/java/eu/kanade/tachiyomi/ui/reader/notification/ImageNotifier.kt
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Clears the notification message
+     */
+    fun onClear() {
+        context.notificationManager.cancel(notificationId)
+    }
+
+    private fun onNotify() {
```
inorichi: This shouldn't be called `on`. It's a private method and you aren't calling it from outside, so it makes less sense. Something like `notify` or `updateNotification` is a lot better.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonReader.kt
```diff
@@ -121,7 +123,6 @@ class WebtoonReader : BaseReader() {
 
     /**
      * Creates the gesture detector for the reader.
-     *
```
inorichi: Don't hate me, but I'm seeing more projects with this line break than without. :(

I think most of our code has them, and they look better if they are multiline IMO. So I think I'm going to ignore the KDoc in their examples.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/notification/ImageNotifier.kt
NoodleMage: I updated all other request (thanks for the thorough feedback). This one I don't get. Why do you want a onBitmapReady when we have OnResourceReady?

I assume you meant `.diskCacheStrategy(DiskCacheStrategy.NONE) `with disable disk cache?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonReader.kt
NoodleMage: I won't hate you for it, but please make up your mind. You have changed your opinion about this 3 times now ;)

app/src/main/java/eu/kanade/tachiyomi/ui/reader/notification/ImageNotifier.kt
inorichi: Call that method inside `onBitmapReady` to avoid too much indentions. It makes it a little more readable (and understandable).

Yes, I meant that.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonReader.kt
inorichi: Sorry about that, yeah, this is the last one, I promise 😅

app/src/main/java/eu/kanade/tachiyomi/ui/reader/notification/ImageNotificationReceiver.kt
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Called to show image in gallery application
+     * @param context context of application
+     * @param path path of file
+     */
+    private fun showImage(context: Context, path: String) {
```
NoodleMage: This crashes on API 24. Working on fix

inorichi:
Ugh... I didn't notice it wasn't a squash 😰

PR #482, opened 2016-09-30
Cleaned up navigation drawer

NoodleMage:
Updated the navigation bar + moved backup to settings
## Screens

screenshot_1475845751_framed
screenshot_1475845754_framed
screenshot_1475845925_framed


inorichi:
I think sync shouldn't be merged with backup. In fact, that one would be renamed to "Tracking" (or a similar one) if cloud sync is implemented.

About the drawer, I would like the word "library" not to be repeated, but I can't think of something better.

NoodleMage:
The reason I suggested the merge is because as you can see backup doesn't have many options (this gets better if auto backup is added).
screenshot_1475763567_framed


But the settings menu itself seems to be getting a bit crowded ;).
screenshot_1475763738_framed

Will probably finish the pull tomorrow

inorichi:
A rotation will destroy the backup/restore process. I'll need to create a presenter for that fragment or use the ugly retain fragment instance.

j2ghz:
Library -> Local, Device
Catalogue -> Remote, Online

NoodleMage:
@inorichi things can never be easy can they ;(. But I will leave this one open then so you can commit your changes (will not update anything myself).

> PS: I think "My library" shouldn't be changed, but the header. Some feedback on the matter will be appreciated.

As for the "my library" part, this is my solution (for some reason I really dig "Discover" in catalogue xD). If you can think of something better you may change it ;).

@j2ghz Local and device are a bit confusing. The manga is not on your device/local until you download it.

NoodleMage:
Navigation will have less items when backup is done, so not needed anymore.

PR #496, opened 2016-10-18
Added option to bookmark single chapter

### LINKED ISSUE #89: Bookmarking
As the title suggests, bookmarking option would be great.

NoodleMage:
Closes #89 
screenshot_1476814528_framed

Multi-select isn't enabled for bookmark because I think it gets too crowded, and it's not really necessary for the bookmark use case.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/tables/ChapterTable.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/tables/ChapterTable.kt
@@ -14,6 +14,8 @@ object ChapterTable {
 
     const val COL_READ = "read"
 
+    const val COL_BOOKMARKED = "col_bookmarked"
+
     const val COL_DATE_FETCH = "date_fetch"
 
     const val COL_DATE_UPLOAD = "date_upload"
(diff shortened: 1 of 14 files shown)


Syborg49:
Who's the man? [You are!] (https://media.giphy.com/media/l0MYw5cmuDoC4ocPS/giphy.gif). Thanks a lot for this @Noodlemage

app/src/main/java/eu/kanade/tachiyomi/data/database/tables/ChapterTable.kt
```diff
@@ -14,6 +14,8 @@ object ChapterTable {
 
     const val COL_READ = "read"
 
+    const val COL_BOOKMARKED = "col_bookmarked"
```
inorichi: `COL_BOOKMARK`. The value shouldn't be prefixed with "col", maybe "ch"? I would also remove the trailing "ed"

inorichi:
I think we should create a custom JSON serializer for chapter, and remove the default values (0 and false) from the output. This will likely increase the size of backups.

PR #520, opened 2016-11-11
Force categories names to be unique.

null-dev:
This commit forces category names to be unique by adding a check during category creation as well as a tiny bit of code cleanup.

~~If you want, I can also add a database constraint.~~
Actually, I will add a database constraint because I forgot that we have to merge any old categories that have the same name so don't merge this yet.


inorichi:
Partially implemented in a0064a1. I will add the database constraint in 0.4.1 with a full database migration (I want to prefix every column to avoid needing aliases in join queries).

PR #521, opened 2016-11-11
Update Mangasee due to webpage changes

Gilfar:
Update for Mangasee source due to webpage changes.
From my understanding there is no way to update existing mangas added to library and you have to remove and readd them one by one.


inorichi:
Have the urls to chapters also changed? If so, I think it would be better to change the source id.

ThePreviousOne:
Yes Mangasee is basically a new source

Gilfar:
Yes the urls have changed, but at the moment all the old ones redirect to the new links so it might work if client follows redirects.
I unfortunately do not have any test cases on my developer machine to test this.

PR #523, opened 2016-11-12
Added Volume and Title to chapters from MangaHere

### LINKED ISSUE #522: Problems with chapters with the same name
There are some sources that use the same title for different chapters (e.g. [Project ARMS on MangaHere](http://www.mangahere.co/manga/project_arms/?no_warning=1) where numbering restarts from 1 for each volume, and the volume number is not part of the chapter title).

- Reading without downloading the chapter works and follows the correct order (keeping source sorting)
- Downloading a chapter with the same title of another one overwrites the files of another chapter (since the directory name is the same) and marks every chapter with the same title as downloaded
- "Download next XXX chapters" seems to always sort by name instead of using the selected criteria (e.g. source, in this case): if I select the 3rd "Chapter 5" in the list (actually the 25th, with 10 chapters per volume), and I want to download the next 3 chapters, the app downloads the first three "Chapter 6" instead of the following "Chapter 6,7,8".
- Mark previous as read is not working, either
- Display by Chapter Number shows "Chapter -1" for every chapter

Version: r934

psychowood:
Fixes #522 
A cleaner implementation would remove the Manga name from the beginning of the chapter title, but an instance of Manga is not available in `chapterFromElement` it would require a few more changes to be feasible.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/source/online/english/Mangahere.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/source/online/english/Mangahere.kt
@@ -80,10 +80,22 @@ class Mangahere(override val id: Int) : ParsedOnlineSource() {
     override fun chapterListSelector() = ".detail_list > ul:not([class]) > li"
 
     override fun chapterFromElement(element: Element, chapter: Chapter) {
-        val urlElement = element.select("a").first()
+        val parentEl = element.select("span.left").first()
+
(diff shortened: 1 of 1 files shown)


inorichi:
Hm... I think I can provide a method to do some changes before inserting to database. Will let you know later.

inorichi:
You should be able to do that now with `prepareNewChapter(chapter, manga)`

Issue #519, opened 2016-11-08
Problems Continuing With MangaSee Chapters


basilg3:
Although the issue seemed to have been solved with the latest commit (I appreciate the speed and efficiency with which the work had been done) the problems with MangaSee chapters not loading are continuing into version r933.

At first, after downloading version r933, I tried reading a chapter as if nothing was wrong, but the same error message of "Attempt to invoke virtual method 'java.lang.String.org.jsoup.nodes.Element.attr(Java.lang.String)' on a null object reference" appeared. 

I then attempted to delete the MangaSee Chapters from my library, clear the database, and readd the manga I'd like to have, but no chapters could be loaded. When I tried to load chapters from any series from MangaSee, a dialogue box appeared saying "No chapters found." However, just like in version r931, when opening the manga from your browser, you can clearly see all chapters are accounted for, and working. Perhaps something else is the matter than the URL? Thank you so much for your help.

ThePreviousOne:
r934 is the "possible" fix

basilg3:
Alright, that makes sense. So, is version r934 downloadable as of now, or is it yet to be made? And if more was changed, is it possible to fix? Thank you.

ThePreviousOne:
Yes It appears to be.
And yes It can its just more work, as to how much "without checking I can't be sure".

basilg3:
Alright, after downloading version r934, the chapters still dont work. Also, when trying to open the manga in my browser, it simply displays a screen saying "Page Not Found," which is a bit of a step back from before because the chapters could be opened in the browser back then. In adittion, the whole catalog displays no manga. It seems there's more to the issue.

ThePreviousOne:
alright I'll look into this

basilg3:
My knowledge when it comes to computers and how they work is not too extensive, but allow me to point out something that I noticed which may or may not be helpful in the solving of this problem. In the app, when trying to open a manga series from the browser (One Piece for example) it tries to use the url http://mangaseeonline.net/manga/?series=OnePiece which does not exist. When opening the same series through MangaSee's website instead of Tachiyomi, the URL for the series of One Piece is http://mangaseeonline.net/manga/One-Piece . Slightly different from the one the app uses. Perhaps this could help you in your programming of the app? Thanks.

GitHubAdi:
[...]

PR #526, opened 2016-11-14
Update for seasonal manga from Mangasee

Gilfar:
With the new layout mangas that Mangasee considers having seasons are broken.


inorichi:
Does this fix the issue at #519 ?

Gilfar:
Yes and no.
The current state after testing is:
- Old mangas keep the old relative url and return no chapters found / 404 (would need to transform old links to new links)
- You can add new mangas and they are alongside the old ones ( I guess they are kept unique via source + path?)
- The only 1st page bug on some mangas is fixed via this commit

GitHubAdi:
Re-adding the manga is absolutly no problem for me if it allows us to keep on reading the higher quality scans from mangasee. Just installed r955 and i am really glad to be able download from mangasee again. Thanks a lot Gilfar.

PR #535, opened 2016-11-17
Download manager rewrite

### LINKED ISSUE #25: Download queue items are not remembered when app reopened
When closing the app via the Recent Apps view, reopening the app greets you with a blank download queue, even if you had several chapters downloading or queued for download.

https://gfycat.com/CrispBaggyBeetle

### LINKED ISSUE #511: Error creating folder when downloading
When a folder or file ends in " ." (space+one ore more points) result in a forbidden name as android ignores the trailing points and no folder or file can end in space, then in the case of a chapter ending in " ." the program tries to create the folder and fails as it don't remove the space before the trailing point/s. Instead of removing the space there is a alternative and simpler solution as is to append something to the name like the download website, time, release number...
 
I can point to a chapter example if toy want.

this affects all tested versions.

### LINKED ISSUE #527: Download notification stays forever stuck download
When downloading I've noticed that the download notification kept showing (http://i.imgur.com/CXQ5khQ.png)  despite finishing all the downloads and it did show that it failed to download 2 chapters. 
At least the notification can be cleared easily.

inorichi:
**Compatibility with previously downloaded chapter isn't guaranteed.**

This rewrite fixes a few bugs and improves compatibility with newer Android versions, where `File` can't be used due to the security mechanisms introduced.

Download discovery performance was also improved due to the way the downloader works now. It downloads every file (which is stored with a `.tmp` extension) to a temporary directory `xxxx_tmp`. When the download is properly completed, the directory is renamed (the `_tmp is removed from the path`).
Thanks to that, a chapter can be marked as downloaded only checking if the directory exists, previously it had to parse the `index.json` file (which is not needed anymore) and analyze if all the pages were downloaded (this added a lot of overhead when a manga had a lot of downloaded chapters).

- [x] Allow to write to any folder in an external SD card (except in KitKat)
- [x] Replace `File` with `Uri` everywhere to use the Storage Access Framework.
- [x] Retain downloads after restarts. Closes #25
- [x] Titles ending with dots or spaces can't be downloaded. Fixes #511 
- [x] Downloader notification not being dismissed on errors. Fixes #527 (Not very tested though)
- [x] Removed the need of the `index.json` file
- [x] Better image extension discovery
- [x] Split `DownloadManager` in multiple classes.
- [x] Refactor the new `Downloader` class
- [x] Documentation
- [x] Test every API

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -116,7 +116,7 @@ dependencies {
     compile 'com.evernote:android-job:1.1.3'
     compile 'com.google.android.gms:play-services-gcm:9.8.0'
 
-    compile 'com.github.seven332:unifile:0.1.4'
+    compile 'com.github.seven332:unifile:0.2.0'
 
     // ReactiveX
     compile 'io.reactivex:rxandroid:1.2.1'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,7 +38,7 @@ android {
         minSdkVersion 16
         targetSdkVersion 25
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
-        versionCode 13
+        versionCode 14
         versionName "0.3.2"
 
         buildConfigField "String", "COMMIT_COUNT", "\"${getCommitCount()}\""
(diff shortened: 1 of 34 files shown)


shogunreaper:
as far as i can see the only reason previous chapters don't work is issues with characters in title

like apostrophes and commas are now included where as before they weren't, changing them fixes it

PR #536, opened 2016-11-17
Added option to sort library

### LINKED ISSUE #374: Library sorting
can you add sort my library by : for example "lastest update, last added, alpha, etc"

NoodleMage:
screenshot_1479852338

closes #374

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -254,7 +254,11 @@ class LibraryUpdateService : Service() {
                             .map { manga }
                 }
                 // Add manga with new chapters to the list.
-                .doOnNext { newUpdates.add(it) }
+                .doOnNext {
+                    it.last_update = Date().time
+                    db.insertManga(it).executeAsBlocking()
+                    newUpdates.add(it)
+                }
                 // Notify result of the overall update.
                 .doOnCompleted {
                     if (newUpdates.isEmpty()) {
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -255,8 +255,10 @@ class LibraryUpdateService : Service() {
                 }
                 // Add manga with new chapters to the list.
                 .doOnNext {
+                    // Set last updated time
                     it.last_update = Date().time
                     db.insertManga(it).executeAsBlocking()
+                    // Add to the list
                     newUpdates.add(it)
                 }
                 // Notify result of the overall update.
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -254,12 +254,12 @@ class LibraryUpdateService : Service() {
                             .map { manga }
                 }
                 // Add manga with new chapters to the list.
-                .doOnNext {
+                .doOnNext { manga ->
                     // Set last updated time
-                    it.last_update = Date().time
-                    db.insertManga(it).executeAsBlocking()
+                    manga.last_update = Date().time
+                    db.updateLastUpdated(manga).executeAsBlocking()
                     // Add to the list
-                    newUpdates.add(it)
+                    newUpdates.add(manga)
                 }
                 // Notify result of the overall update.
                 .doOnCompleted {
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
@@ -201,7 +201,6 @@ class LibraryFragment : BaseRxFragment<LibraryPresenter>(), ActionMode.Callback 
         val sortModeAlpha = menu.findItem(R.id.action_sort_alpha)
         val sortModeLastRead = menu.findItem(R.id.action_sort_last_read)
         val sortModeLastUpdated = menu.findItem(R.id.action_sort_last_updated)
-        val sortModeDateAdded = menu.findItem(R.id.action_sort_date_added)
 
         // Set correct checkbox filter
         filterDownloadedItem.isChecked = isFilterDownloaded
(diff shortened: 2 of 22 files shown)


inorichi:
Would the date added sorting really be used? I mean, it's another row in db and I don't it so useful (I may be wrong).

Also, would it make sense to use the filter icon for both filters and sorting to avoid too many items?

A menu with something like:

```
+------------------+----------------+
|       Sort       |     Filter     |
+------------------+----------------+
| [*] Alphabetic   | [x] Unread     |
| [ ] Last read    | [ ] Read       |
| [ ] Last updated | [x] Downloaded |
| ...              | ...            |
+------------------+----------------+
```

NoodleMage:
I looked into multi menu during the pull. But I personaly didnt like the look of it. As far as I know its not possible to create a subheader (smaller text) in a menu. So I had to use a non checkable item which looks weird. 

Gonna do a rename though. No idea why I added sort by for every option. 

Date added was in the pull request. That's why I added it.

inorichi:
There's [this StackOverflow answer](http://stackoverflow.com/a/36215592/3263582), but it also doesn't use headers. And seeing the picture we could also include the show as grid/list there, but maybe it won't look that good on smaller screens (needing to scroll).

Regarding date added, I vote for not including it (at least yet).

NoodleMage:
The stack overflow answer is for a single popup menu. Maybe I can rewrite it to work for a sub menu but it will get even more messy, also it's not tested on a lot of devices (only API's).

Removed date added sort as option, kept the database in case we might need it in the future, or do you want to exclude that as well?

inorichi:
Well, I said it to avoid updating the database unless the field is really useful.

I agree the popup menu won't be easy to implement and too messy. Maybe we could show a dialog with all the options (here everything should fit the screen). It would avoid three levels (overflow icon + sort icon + selection) and also keep the dialog if the user wants to select two filters (at the moment he has to click the filter icon, select a filter, and open it again because the menu is closed). What do you think?

NoodleMage:
Woops my brain can't read back that far :O. Will remove the field then ;).

I kinda like the idea. But wouldn't it reduce the effectiveness of the sort and filter. In my experience I only use filter to quickly check something. If you have to open a dialog >  select the filter > OK. everytime you want to see if you have any downloaded manga could be frustrating.  

Also if in the future if new filter/sort/view gets added the dialog will be to large for some devices?

inorichi:
Fair enough, let's keep it like this for now.

One last thing, I think the "Sort" option should be removed (the one in the menu of the screenshot) since the other menus don't show a title. Or was there a reason (material design spec, etc..) for that?

NoodleMage:
It's because sort is a sub menu. If It's not shown as action android automatically adds the title. So I think it's supposed to happen, e.g.

filter_sub
filter_sub2

inorichi:
Oh ok. Ignore my comment then.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -254,7 +254,13 @@ class LibraryUpdateService : Service() {
                             .map { manga }
                 }
                 // Add manga with new chapters to the list.
-                .doOnNext { newUpdates.add(it) }
+                .doOnNext {
+                    // Set last updated time
+                    it.last_update = Date().time
+                    db.insertManga(it).executeAsBlocking()
```
inorichi: This should be done with a custom PutResolver, similar to `MangaFlagsPutResolver` but for `last_update` field.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryFragment.kt
```diff
@@ -179,12 +186,35 @@ class LibraryFragment : BaseRxFragment<LibraryPresenter>(), ActionMode.Callback
...
+        val sortModeAlpha = menu.findItem(R.id.action_sort_alpha)
+        val sortModeLastRead = menu.findItem(R.id.action_sort_last_read)
+        val sortModeLastUpdated = menu.findItem(R.id.action_sort_last_updated)
+
+        // Set correct checkbox filter
+        filterDownloadedItem.isChecked = isFilterDownloaded
+        filterUnreadItem.isChecked = isFilterUnread
+
+        // Set correct radio button sort
+        if (sortingMode == Constants.SORT_LIBRARY_ALPHA) sortModeAlpha.isChecked = true
```
inorichi: This should look better with `when (sortingMode)`

app/src/main/res/menu/library.xml
```diff
@@ -1,48 +1,72 @@
...
+      xmlns:tools="http://schemas.android.com/tools"
+      tools:context=".MainActivity">
 
     <item
-        android:id="@+id/action_search"
-        android:icon="@drawable/ic_search_white_24dp"
-        android:title="@string/action_search"
-        app:actionViewClass="android.support.v7.widget.SearchView"
-        app:showAsAction="collapseActionView|ifRoom"/>
+            android:id="@+id/action_search"
```
inorichi: Why 8 spaces?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -133,10 +134,12 @@ class LibraryPresenter : BasePresenter<LibraryFragment>() {
      */
     fun getLibraryMangasObservable(): Observable<Map<Int, List<Manga>>> {
         return db.getLibraryMangas().asRxObservable()
-                .flatMap { mangas ->
-                    Observable.from(mangas)
+                .flatMap {
+                    Observable.from(it)
                             // Filter library by options
                             .filter { filterManga(it) }
+                            .toSortedList { manga1, manga2 -> sortManga(manga1, manga2) }
```
inorichi: I'll merge these changes but just FYI I'm going to rewrite some of this. Every filter/sort update is triggering a database update and this is not efficient. A better approach will store a local, unfiltered list, and filter updates will query that list. That's how it's done for the chapters.

inorichi:
I just had an idea of a different approach (but in another PR, in the case we decide to implement it).

It would be another drawer in the right side of the screen with a custom view. We could open it by dragging by the edge (like the other drawer), or optionally we can change the current filter button to open that (meh).

In that drawer, we would show everything related to how the library is displayed:

* Filters (unread, downloaded, ...)
* Sorting (alphabetically, last updated, ...). We should also consider ascending and descending sorting, either by using a tri-state checkbox (disabled, ascending, descending) or with another section with a radio button. I prefer the first one.
* Display (grid, list).

This is basically a different implementation of the dialog I mentioned last week, which could be too intrusive. With the drawer, you can still see the library screen, update every filter at once (and this will be better if filters are saved by category in the future) and remove 3 menu items, which is starting to look bloated.


As said before, these are just my thoughts and they're not going to be implemented in this PR.

NoodleMage:
Nice Idea. 
Plex implemented it like that: http://i.imgur.com/JhFVEg8.png

I would personally only trigger it when the filter button is pressed though because this comes more natural than swiping from the right.

inorichi:
Awesome, it looks better than I thought.

Ok, we will trigger it with the icon.

PR #538, opened 2016-11-19
Add optional to automatically download new chapers

### LINKED ISSUE #116: Download new chapters automatically
This is a feature of Mango that I used to have turned on.
Whenever the app checks for new chapters, it will automatically download any new updates.

icewind1991:
Fixes #116

The changed `readded` logic was needed to prevent all non-read chapters from sources such as MangaFox being detected as new

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -13,7 +13,9 @@ import eu.kanade.tachiyomi.Constants
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.data.database.DatabaseHelper
 import eu.kanade.tachiyomi.data.database.models.Category
+import eu.kanade.tachiyomi.data.database.models.Chapter
 import eu.kanade.tachiyomi.data.database.models.Manga
+import eu.kanade.tachiyomi.data.download.DownloadManager
 import eu.kanade.tachiyomi.data.library.LibraryUpdateService.Companion.start
 import eu.kanade.tachiyomi.data.preference.PreferencesHelper
 import eu.kanade.tachiyomi.data.preference.getOrDefault
@@ -52,6 +54,8 @@ class LibraryUpdateService : Service() {
      * Preferences.
      */
     val preferences: PreferencesHelper by injectLazy()
+
+    val downloadManager: DownloadManager by injectLazy()
 
     /**
      * Wake lock that will be held until the service is destroyed.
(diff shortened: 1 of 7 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -262,13 +273,23 @@ class LibraryUpdateService : Service() {
...
     }
 
+    fun downloadChapters(manga: Manga, chapters: List<Chapter>) {
+        // we need to get the chapters from the db so we have chapter ids
+        val mangaChapters = db.getChapters(manga).executeAsBlocking()
+        val dbChapters = chapters.map {
+            mangaChapters.find { mangaChapter -> mangaChapter.url == it.url }!!
+        }
+        downloadManager.downloadChapters(manga, dbChapters)
+        downloadManager.startDownloads()
```
inorichi: This method is for internal use only. Use `DownloadService.start(this)` instead.

Also don't start the downloads until the entire update finishes. Less simultaneous calls to the website will probably be better.

j2ghz:
Suggestion: Add filter by category

inorichi:
It could be possible using the categories' flags, but there are a lot of things to consider first. The default category isn't in db, so we have to either add a default category to db or use the preferences to store them (I don't like the latter, but the first is probably not going to be easy).

icewind1991:
I was planning on adding (in a separate) pr some options to filter which mangas do download chapters for.

Things like only download chapters for mangas which are already downloaded or mangas that are actively being read

icewind1991:
@inorichi *poke*

app/src/main/java/eu/kanade/tachiyomi/util/ChapterSourceSync.kt
```diff
@@ -73,14 +70,16 @@ fun syncChaptersWithSource(db: DatabaseHelper,
                 // Try to mark already read chapters as read when the source deletes them
                 if (c.isRecognizedNumber && c.chapter_number in deletedReadChapterNumbers) {
                     c.read = true
-                    readded++
+                }
+                if (c.isRecognizedNumber && c.chapter_number in deletedChapterNumbers) {
```
inorichi: Hm... can't this logic be merged with the other somehow? I'm quite tired atm and I didn't dig too much into it, but at first glance it seems both sets are trying to do the same.

I'll review it in more detail tomorrow.

icewind1991: One is for all chapter numbers that were deleted, one is only for chapters that are marked as read.

They are needed separately since with one of them we need to mark them as read again

PR #551, opened 2016-11-25
Italian language

paronos:
I have translated the strings.xml in Italian.
I had doubts about some decisions (such as whether/how to translate Download), I would appreciate if someone else would give a look.


inorichi:
Should I merge this or wait for feedback? I can't help in the decision of whether to translate download or not.

paronos:
I think you can merge this, I checked the translation, and I don't think that there are errors (I don't think it's worth to wait even longer).

Issue #558, opened 2016-12-02
[request] Language selection


Snake87r:
Add a language selection for the app,
Usually I have my tablet in my native language, but i prefer some app (like tachyiomi) in english, so if it's possible add an option called "Select language", with Automatic option setted by default, and list all the available translation so every user choose the preferred language for the app..

Version: r982

inorichi:
Changing the locale in runtime is a real mess from what I've seen. I don't think this will be implemented.

You could try to remove the translation for your language, [here's a tutorial](http://forum.xda-developers.com/showthread.php?t=1890521) I found, but I haven't tried it, so it may not work.

I also wanted the english translation in my tablet, so I ended up changing the system's language to english.

PR #563, opened 2016-12-04
Implement language switcher

Flat:
Implements request #558 

In my opinion I don't think it's that much of a mess. There are two deprecated java functions used, these cannot be changed unless the min android API is set to 17 instead of 16.

Feel free to point out any changes needed or just close if this isn't something you want.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/App.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/App.kt
@@ -6,7 +6,6 @@ import android.content.res.Configuration
 import android.support.multidex.MultiDex
 import com.evernote.android.job.JobManager
 import eu.kanade.tachiyomi.data.library.LibraryUpdateJob
-import eu.kanade.tachiyomi.data.preference.PreferencesHelper
 import eu.kanade.tachiyomi.data.updater.UpdateCheckerJob
 import eu.kanade.tachiyomi.util.LocaleHelper
 import org.acra.ACRA
@@ -14,9 +13,7 @@ import org.acra.annotation.ReportsCrashes
 import timber.log.Timber
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.InjektScope
-import uy.kohesive.injekt.injectLazy
 import uy.kohesive.injekt.registry.default.DefaultRegistrar
-import java.util.*
 
 @ReportsCrashes(
         formUri = "http://tachiyomi.kanade.eu/crash_report",
--- a/app/src/main/java/eu/kanade/tachiyomi/App.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/App.kt
@@ -27,13 +27,6 @@ import java.util.*
 )
 open class App : Application() {
 
-    val preferences: PreferencesHelper by injectLazy()
-
-    //get lang code by pref
-    private val langCode by lazy {
-        LocaleHelper.intToLangCode(preferences.lang())
-    }
-
     override fun onCreate() {
         super.onCreate()
         Injekt = InjektScope(DefaultRegistrar())
@@ -44,8 +37,6 @@ open class App : Application() {
         setupAcra()
         setupJobManager()
 
-        //set language from preferences
-        LocaleHelper.setLocale(Locale(langCode))
         LocaleHelper.updateCfg(this, baseContext.resources.configuration)
     }
 
@@ -56,9 +47,9 @@ open class App : Application() {
         }
     }
 
-    override fun onConfigurationChanged(newConfig: Configuration?) {
+    override fun onConfigurationChanged(newConfig: Configuration) {
         super.onConfigurationChanged(newConfig)
-        LocaleHelper.updateCfg(this, newConfig!!)
+        LocaleHelper.updateCfg(this, newConfig)
     }
 
     protected open fun setupAcra() {
--- a/app/src/main/java/eu/kanade/tachiyomi/App.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/App.kt
@@ -2,16 +2,21 @@ package eu.kanade.tachiyomi
 
 import android.app.Application
 import android.content.Context
+import android.content.res.Configuration
 import android.support.multidex.MultiDex
 import com.evernote.android.job.JobManager
 import eu.kanade.tachiyomi.data.library.LibraryUpdateJob
+import eu.kanade.tachiyomi.data.preference.PreferencesHelper
 import eu.kanade.tachiyomi.data.updater.UpdateCheckerJob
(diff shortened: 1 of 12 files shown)


inorichi:
The language is lost after a rotation in the settings activity. That logic should be placed in the BaseActivity. 

[This](http://stackoverflow.com/a/32037685/3263582) (which I think is the best approach) says the solution you used won't work in Android M, though it seems to be working fine.

Also I'm not sure your PR will properly handle configuration changes, from the SO answers, the locale config change should be added to every activity in the manifest.

Finally, by reading the comments, this causes activity flashes if the `Configuration` isn't copied (and one said even copying it didn't help, so this must be one of those bugs specific to some devices).

That's why I said locale handling is a mess, because there isn't a clear solution, but by no means I'm against it (that's why I kept the issue open).

Flat:
Yeah, unfortunately there doesn't seem to be a "proper" way to switch languages on the fly. I'll work on this some more this week.

Testing on N the only issue I've seen myself is the language not sticking on orientation change (Which should be fixed by using the solution you linked). Not to say that there aren't other issues with it; but none that I've experienced on my device.

Flat:
New commit shouldn't lose locale due to orientation change, and the locale should persist through other configuration changes.

app/src/main/java/eu/kanade/tachiyomi/App.kt
```diff
@@ -40,6 +56,11 @@ open class App : Application() {
         }
     }
 
+    override fun onConfigurationChanged(newConfig: Configuration?) {
```
inorichi: Make `Configuration` non nullable instead of using `!!`.

app/src/main/java/eu/kanade/tachiyomi/App.kt
```diff
@@ -22,6 +27,13 @@ import uy.kohesive.injekt.registry.default.DefaultRegistrar
 )
 open class App : Application() {
 
+    val preferences: PreferencesHelper by injectLazy()
```
inorichi: I don't like injecting in the same class where they are created. It doesn't crash because `langCode` is lazy too.

Instead of calling `LocaleHelper.setLocale(Locale(langCode))`, create and call another method `LocaleHelper.init()`. Then you inject the preferences inside that method and set the locale.

Edit: You could also not create the `init()` method and avoid a `lateinit` var in `LocaleHelper`. Since `object` classes are initialized lazy, you can set the default value in `LocaleHelper` (with `private var locale = Locale(/* ... */)`) and it should be initialized when `LocaleHelper.updateCfg(this, baseContext.resources.configuration)` is called in the `onCreate` of the `App` class.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsActivity.kt
```diff
@@ -78,6 +78,7 @@ class SettingsActivity : BaseActivity(),
     companion object {
         const val FLAG_THEME_CHANGED = 0x1
         const val FLAG_DATABASE_CLEARED = 0x2
+        const val FLAG_LANG_CHANGED = 0x3
```
inorichi: This is `0x4`. With `0x3` you are setting `0x1` and `0x2` (and in the `MainActivity` it's calling the database changed method).

app/src/main/java/eu/kanade/tachiyomi/util/LocaleHelper.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun updateCfg(app: Application, config: Configuration){
+        if(Build.VERSION.SDK_INT < Build.VERSION_CODES.JELLY_BEAN_MR1){
+            config.locale = pLocale
+            app.baseContext.resources.updateConfiguration(config, app.baseContext.resources.displayMetrics)
+        }
+    }
+
+    fun intToLangCode(i: Int): String {
+        return when(i){
+            0 -> ""
```
inorichi: I would remove this case and add a comment in the else `// System language`.

app/src/main/res/values/keys.xml
```diff
@@ -21,6 +21,7 @@
     <string name="pref_theme_key">pref_theme_key</string>
     <string name="pref_library_update_restriction_key">library_update_restriction</string>
     <string name="pref_start_screen_key">start_screen</string>
+    <string name="pref_language">pref_language</string>
```
inorichi: Remove `pref_` from the value (not the key). The ones still prefixed are for compatibility, but new preferences shouldn't prefix.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsActivity.kt
Flat: Ha, oops...

Flat:
Think I've covered everything that you reviewed, let me know if there's anything else!

app/src/main/res/xml/pref_general.xml
```diff
@@ -60,7 +60,7 @@
             android:defaultValue="0"
             android:entries="@array/languages"
             android:entryValues="@array/languages_values"
-            android:key="pref_language"
+            android:key="language"
```
inorichi: This is `@string/pref_language_key`. You will also need to add `_key` in `keys.xml`.

Everything else looking fine.

inorichi:
Oh, and please cleanup now unused imports in `App` (Ctrl + Alt + O in Android Studio)

app/src/main/res/xml/pref_general.xml
Flat: should the value of pref_language_key contain pref_?

inorichi: No, just the key

Flat:
Okay, think that should do it.

inorichi:
This still has translation issues.

On API >= 17, strings resolved through the application context or a service/broadcast context (all the notifications for example) ignores the selected language and always use the system's. Wrapping BaseActivity isn't good after all.

I don't know yet if changing the application's locale up to Marshmallow (23) is feasible, but Nougat still needs a different approach after the changes they made.

Flat:
I suspect it has something to do with the way services get the application context, I'm working on seeing why the updated configuration is not passed onto the services (and therefore notifications) but it seems there is no support for kotlin line breaks in android studio yet. (that I can find)

inorichi:
The locale change is only applied to activities, so it makes sense it doesn't work in a non activity context.

Breakpoints have always worked fine for me.

I can take a look at this after I finish most of the work with the tracking services, probably next week.

inorichi:
It should work fine now.

Flat:
Sorry I wasn't able to help more on that. I feel like I've created more work for you than help with this feature.

inorichi:
No worries. 👍

PR #586, opened 2016-12-18
Experimental Anilist and Kitsu support

### LINKED ISSUE #312: AniList Sync Support

### LINKED ISSUE #574: Kitsu Sync Support
[Hummingbird](https://forums.hummingbird.me/t/hummingbird-will-be-read-only-as-we-migrate-data-over-to-kitsu/33359) is now [Kitsu](https://kitsu.io). 

It also brings along a new API following the JSON API spec with full manga support.

[API Documentation](http://docs.kitsu17.apiary.io)

inorichi:
Closes #312
Closes #574 

Some things may not work as expected or at all, so any feedback will be appreciated.

Score needs to be improved. Only integers can be selected, but it's translated properly.

If you change the score system in Anilist, you will have to login again, even if you already are.


na-ji:
For now I have one feedback for Anilist : when I search a name and the name is not found, the search animation continue indefinitely. I do not have this problem with My Anime List.

wopian:
You'll need to show what media type it is (manga, one-shot, light novel etc) since you end up with this problem:

wopian:
Kitsu rating is shown in the proper 0.5-5 scale on the tab page, but in the dropdown it becomes 1-10

wopian:
Linking some manga to my Kitsu library fails with `chapterCount is not found`.

Such as 'Aho Girl' and '20th Century Boys'

inorichi:
Yeah I said the score isn't the same but it's properly translated.

I can't reproduce the Aho Girl issue.

wopian:
Trying to sync it always gives the error, after manually changing values and deleting it from library.

Source: Batoto
Manga: Aho Girl
All 73 available chapters marked as read

Deleting, clearing cache and re-adding also doesn't work

https://gfycat.com/AbandonedWideeyedHairstreak

inorichi:
I had some local changes not pushed, however I don't remember changing anything related to `chapterCount`, and I think its nullability was handled. Try again with these changes.

The correct score is already shown for all the services.

wopian:
Still happens on 1017 😂

inorichi:
And I'm still not able to reproduce the issue. I've tried using that same manga, with the same status, rating and score. It works fine.

I'm 99% sure that is a response error from the API.

inorichi:
Fixed the infinite loading progress in Anilist when no results are found.

The chapterCount issue should be fixed too (I still don't know why it doesn't crash for me).

wopian:
`chapterCount` issue is now fixed in r1028 👍

Now have the same issue with `mangaType`

inorichi:
Fixed too.

PR #594, opened 2016-12-20
Notification Improvements

NoodleMage:
## TODO
- [x] Clean up

- [x] Testing

Improvements to Notifications + download notification rework

- Added a Notification Service.
- Added a Notification Activity Handler.
- Removed unneeded global broadcasts.
- Added actions to download notification.

NotificationReceiver -> Runs on main/UI thread should handle UI + notification (Only used for pending broadcasts)
NotificationService-> Runs default on different thread should run background tasks (uses own local broadcast for UI calls)
NotificationHandler -> Returns pending activities.

As always feedback is appreciated.


inorichi:
With a quick look, there are two things that concerns me.

* Why an activity?
* Downloads having a "paused" state is going to cause some issues with the current implementation. Do you really need it? I think you could use the queued state.

NoodleMage:
> Why an activity?
I can't call a fragment from notification. A possible solution for this is calling the MainActivity and setting the fragment to download manager. Problem is that you can call the notification from every activity. Thus also in the reader. So I have to lock the navigation bar (otherwise you can still swipe the nav bar and create a activity loop). Set the menu to arrow and change it's behavior which gets messy real fast. 

That's why I implemented it like this. It gives the desired behavior without creating all kinds of work arounds.

> Downloads having a "paused" state is going to cause some issues with the current implementation. Do you really need it? I think you could use the queued state.
I haven't encountered any issues (could you give an example where it would go wrong?). Queued is possible but I liked the way this looks (especially in multidownload).

inorichi:
The solution for that is to use the intent with clear top flag, but it depends in the expected behavior. If you want to place the downloads above the reader, an activity. If you want to exit the reader, clear top and fragments.

I don't know which one is better.

About the pause state, IIRC there are some places where I use integer range comparison in the status, like `download.status >= x`. I think the method `areAllDownloadsFinished` (don't recall the exact name) uses it.

NoodleMage:
The behavior is:
Main -> Reader -> Download Queue

When back pressed Download Queue is dropped and activity becomes.
Main -> Reader

Clear top gave weird behavior when switching between apps. That's why I used `android:launchMode="singleTask"` which is only possible with an activity.

> I think the method areAllDownloadsFinished (don't recall the exact name) uses it.

Will remove pause and change it to queue then ;).

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
```diff
@@ -88,24 +106,33 @@ internal class DownloadNotifier(private val context: Context) {
...
-            // Check if icon needs refresh
+            // Check if first call.
             if (!isDownloading) {
                 setSmallIcon(android.R.drawable.stat_sys_download)
+                setAutoCancel(false)
+                clearActions()
+                // Open download manager when clicked
+                setContentIntent(DownloadNotificationReceiver.openDownloadManagerIntent(context))
+                // Pause action
+                addAction(R.drawable.ic_av_pause_grey_24dp_img,
```
NoodleMage: Think I'm gonna remove this action and only let the user pause a download via the manager. I don't like how
```
 // This code prevents that pause action is unclickable  TODO improve (if possible)
            val currentTimeMillis = System.currentTimeMillis()
            if (currentTimeMillis - lastOnProgressCall < 750) {
                  return
              }
```
looks in the notification ;).

app/build.gradle
```diff
@@ -101,7 +101,7 @@ dependencies {
     compile 'com.github.inorichi:subsampling-scale-image-view:c4db85c'
 
     // Android support library
-    final support_library_version = '25.0.1'
+    final support_library_version = '25.1.0'
```
inorichi: Please don't include this. From my experience something weird always happen and I need to test this thoroughly.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
```diff
@@ -134,16 +149,56 @@ internal class DownloadNotifier(private val context: Context) {
...
         with(notification) {
-            setContentTitle(download?.chapter?.name ?: context.getString(R.string.app_name))
+            val title = download.manga.title.chop(15)
+            val chapter = download.chapter.name.replaceFirst(download.manga.title, "", true)
+            setContentTitle("$title - $chapter".chop(30))
             setContentText(context.getString(R.string.update_check_notification_download_complete))
             setSmallIcon(android.R.drawable.stat_sys_download_done)
+            setAutoCancel(true)
+            clearActions()
+            setContentIntent(NotificationHandler.openMangaPendingActivity(context, download.manga, download.chapter))
```
inorichi: I'm not sure if this is really worth. If you download chapters 1,2,3,4,5 will it open chapter 5?

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -121,6 +121,17 @@ class Downloader(private val context: Context, private val provider: DownloadPro
     }
 
     /**
+     * Pauses the downloader.
+     */
+    fun pause() {
+        DownloadService.stop(context)
```
inorichi: call `destroySubscriptions` before modifying the queue. I'm still worried about threading issues, but I don't know a better way.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -235,13 +235,10 @@ class LibraryUpdateService : Service() {
...
         val failedUpdates = ArrayList<Manga>()
 
-        val cancelIntent = PendingIntent.getBroadcast(this, 0,
-                Intent(this, CancelUpdateReceiver::class.java), 0)
-
         // Emit each manga and update it sequentially.
         return Observable.from(mangaToUpdate)
                 // Notify manga that will update.
-                .doOnNext { showProgressNotification(it, count.andIncrement, mangaToUpdate.size, cancelIntent) }
+                .doOnNext { showProgressNotification(it, count.andIncrement, mangaToUpdate.size, NotificationService.cancelLibraryUpdatePendingService(this, notificationId)) }
```
inorichi: Why not cache this object? It looks like unnecessary allocation

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationService.kt
```diff
@@ -0,0 +1,255 @@
...
+     * Download manager.
+     */
+    private val downloadManager: DownloadManager by injectLazy()
+
+    override fun onHandleIntent(intent: Intent) {
+        when (intent.action) {
+            // Resume the download service
+            ACTION_RESUME_DOWNLOADS -> DownloadService.start(this)
+            // Clear the download queue
+            ACTION_CLEAR_DOWNLOADS -> downloadManager.clearQueue(true)
```
inorichi: This method should be called in the main thread. I don't know if it can lead to errors, but just in case.

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationService.kt
```diff
@@ -0,0 +1,255 @@
...
+            }
+            context.sendLocalBroadcast(intent)
+        }
+    }
+
+    /**
+     * Local [BroadcastReceiver] that runs on UI thread
+     * Notification calls from [NotificationService] should be made from here.
+     */
+    private val notificationReceiver = object : BroadcastReceiver() {
```
inorichi: Nitpick: Fields should be at the top of the class.

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationService.kt
```diff
@@ -0,0 +1,255 @@
...
+            // Dismiss notification
+            ACTION_DISMISS_NOTIFICATION -> dismissNotificationBroadcast(this,
+                    intent.getIntExtra(EXTRA_NOTIFICATION_ID, -1))
+        }
+    }
+
+    override fun onCreate() {
+        // Register the local receiver
+        registerLocalReceiver(notificationReceiver, IntentFilter(INTENT_FILTER_NAME))
+        return super.onCreate()
```
inorichi: ~~return~~

inorichi: And if possible register after calling super

app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderService.kt
```diff
@@ -67,20 +91,24 @@ class UpdateDownloaderService : IntentService(UpdateDownloaderService::class.jav
                 val progress = (100 * bytesRead / contentLength).toInt()
                 if (progress > savedProgress) {
                     savedProgress = progress
-
-                    progressNotification.update { setProgress(100, progress, false) }
+                    intent = Intent(INTENT_FILTER_NAME).apply {
+                        putExtra(EXTRA_ACTION, NOTIFICATION_UPDATER_PROGRESS)
+                        putExtra(EXTRA_PROGRESS, progress)
+                    }
+                    // Prevents not showing of install notification TODO find out why notification isn't showing.
```
inorichi: It's probably being overwritten. I guess `sendLocalBroadcastSync` sends the intent to the event looper, while the install notification is executed synchronously, thus a race condition.

app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderService.kt
```diff
@@ -89,48 +117,158 @@ class UpdateDownloaderService : IntentService(UpdateDownloaderService::class.jav
...
+    }
+
+    /**
+     * Returns [PendingIntent] that starts a service which downloads the apk specified in url.
+     *
+     * @param url the url to the new update.
+     * @return [PendingIntent]
+     */
+    private fun downloadApkPendingService(url: String): PendingIntent {
+        val intent = Intent(this@UpdateDownloaderService, UpdateDownloaderService::class.java).apply {
```
inorichi: Isn't `@UpdateDownloaderService` redundant here?

app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderService.kt
```diff
@@ -89,48 +117,158 @@ class UpdateDownloaderService : IntentService(UpdateDownloaderService::class.jav
...
+    private val EXTRA_APK_PATH = "APK_PATH"
+
+    // Value containing apk url
+    private val EXTRA_APK_URL = "APK_URL"
+
+    /**
+     * Local [BroadcastReceiver] that runs on UI thread
+     * Notification calls from [UpdateDownloaderService] should be made from here.
+     */
+    private val updaterNotificationReceiver = object : BroadcastReceiver() {
```
inorichi: Fields at the top.

If you believe this object is less relevant and should be at the bottom, convert it to a class:

`class UpdaterBroadcastReceiver : BroadcastReceiver() { ... }`

And the field at the top:

`private val updaterNotificationReceiver = UpdaterBroadcastReceiver()`

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadActivity.kt
```diff
@@ -0,0 +1,41 @@
...
+import eu.kanade.tachiyomi.R
+import eu.kanade.tachiyomi.ui.base.activity.BaseActivity
+import kotlinx.android.synthetic.main.activity_main.*
+import kotlinx.android.synthetic.main.toolbar.*
+
+/**
+ * Wraps [DownloadFragment] into an activity.
+ * Note: Call the fragment if possible
+ */
+class DownloadActivity : BaseActivity() {
```
inorichi: We should only be allowed to open one instance of the fragment. You can open the download queue from the drawer and a this activity on top. I've already seen weird behavior (only one of the screens is notified of page updates).

I would fix this by making the drawer open this activity (and the fragment is not needed anymore, so the logic from `DownloadFragment` can be placed here.

inorichi:
I haven't worked enough with the notification system to know its best practices, but it seems you're running into some race conditions, and that's worrying.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
NoodleMage: I added it because it can be convenient if you download only one chapter you can instantly open it. I can add a check to test when queue > 1 -> dismiss()?

NoodleMage:
The documentation isn't super clear on notifications. I concluded two things. Local broadcasts are better (see title) and the following classes should be used:
Broadcast -> Runs on main/UI thread should handle UI + notification
Service -> Runs default on different thread should run background tasks
ActivityHandler -> Returns pending activities. 

Downside of ActivityHandler is that you can't cancel the notification after an action that's why I moved some of them to the service. Downside of Local Broadcast is that it should be registered before using which means the pendingintent doesn't work unless you use the service to register first and then send broadcast. 

As for the race condition. Here's where things get weird. It's only in the updater for some reason. But maybe is because this is the only one where broadcast is called real time instead of ONE_SHOT_FLAG pendingintents. The weird thing for me was that when I added a `Timber.e("you called function")` to the code it was executed in the correct order, which should be the case with the sendLocalBroadcastSync call. Will try to call the notification directly from the service (which I didn't like because it's on a background thread) to see if I get better results, I thought I already did this check but not sure.

### Local Broadcast Reciever
Helper to register for and send broadcasts of Intents to local objects within your process. This has a number of advantages over sending global broadcasts with sendBroadcast(Intent):

You know that the data you are broadcasting won't leave your app, so don't need to worry about leaking private data.
It is not possible for other applications to send these broadcasts to your app, so you don't need to worry about having security holes they can exploit.
It is more efficient than sending a global broadcast through the system.

NoodleMage:
Some more information on race condition on updater. Tested it on API 22 and 116 (without the  `if (progress <= 95) {`) works perfectly, Android N hangs at around 95%. I'm starting to think that #595 was just good (or bad) luck

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
NoodleMage: Download notifier is now only shown when 1 chapter is downloaded.

NoodleMage:
I kept the 95% only for Android N. It works without on first launch but when you restart the application it hangs on ~97%.

The notifications works on every emulator and device I tested it on (API 16 emu / API 22 emu / Android 7.1.1 Nexus 6P). 

But for some reason the pause notification is not shown (at random!) on API 25 emu. 
As there is no error nor warning, it works when I place a breakpoint to check if it reaches the code and it works on my device which also runs API 25 I think it's a bug in the emulator..

inorichi:
Probably https://code.google.com/p/android/issues/detail?id=231117

NoodleMage:
Big change that's the one ;). In that case I think it's best to hold this pull until it's resolved?

inorichi:
I would like to see a workaround, but it may take some time. I don't know how they overlooked that, it's a very problematic issue... The 95% limit on N may not be enough if the download is very fast, but at least it should work most of the times. Another possibility is to delay a bit the last notification, but let's wait a bit for now.

In the meantime I would look for a better alternative to `NotificationService`. For example, using a `Service` instead of `IntentService` will avoid the thread creation. I still don't know how the notification system should be used, so I'll try this week to look for articles and implementations and provide a better feedback.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
inorichi: Can this open one reader above the other? I want the reader instance to always be unique.

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationService.kt
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    override fun onDestroy() {
+        // Unregister the local receiver
+        unregisterLocalReceiver(notificationReceiver)
+        super.onDestroy()
+    }
+
+    override fun onHandleIntent(intent: Intent) {
+        when (intent.action) {
```
inorichi: Can't all these 3 methods run inside a `BroadcastReceiver` instead? The only blocking operation is the delete file, but it should be fast. I've read a bit about notifications but everything is just simple usage, and somehow I still can't seem to like this service.

NoodleMage: While working on the backup I noticed that the local broadcast receiver is best suited for communicating between a activity/fragment and an IntentService. 

If we change the notificationservice to Service like you suggested every pending intent can be managed from the service so we can keep everything internal and the broadcast isn't needed anymore. 

The updater will keep its local broadcast for the notification because these shouldn't be created from a background thread afaik.

NoodleMage: Edit. I moved everything to a single broadcast as this seems to be the best solution. In my opinion it's a bit weird to have something that is send to all receivers + have a constant listening receiver just for internal calls but apparently it's the best implementation.

I will read through the code once more and then everything should be in order. At least we don't have 4 receivers listening constantly anymore ;). And apologies for the sh*tshow this notifications have become.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
NoodleMage: Changed to flag to FLAG_ACTIVITY_MULTIPLE_TASK so the activity is always created as new

inorichi: It still opens one reader above the other, and it loads the already opened chapter again. I would ignore the new instance in the unlikely case (but possible) the reader is already open when the notification is clicked.

NoodleMage: sorry, I don't fully understand you. Do you want a maximum of one instance of the reader activity or should a new instance be created when this notification is clicked?

inorichi: I want always at most only one instance of `ReaderActivity` in the entire application. If there's one already open, it should close (gracefully) the other activity and start a new one.

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
```diff
@@ -0,0 +1,278 @@
...
+import eu.kanade.tachiyomi.BuildConfig.APPLICATION_ID as ID
+
+/**
+ * Global [BroadcastReceiver] that runs on UI thread
+ * Pending Broadcasts should be made from here.
+ * NOTE: Use local broadcasts if possible.
+ */
+class NotificationReceiver : BroadcastReceiver() {
+
+    companion object {
```
inorichi: Every `val` should be `const` here. I would also place the companion object at the end of the class (too big and less relevant).

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
```diff
@@ -0,0 +1,278 @@
...
+        // Value containing chapter id.
+        private val EXTRA_CHAPTER_ID = "$ID.$NAME.EXTRA_CHAPTER_ID"
+
+        /**
+         * Returns a [PendingIntent] that resumes the download of a chapter
+         *
+         * @param context context of application
+         * @return [PendingIntent]
+         */
+        internal fun resumeDownloadsPendingBroadcast(context: Context): PendingIntent? {
```
inorichi: This method and the one below can't return null.

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
```diff
@@ -0,0 +1,278 @@
...
+
+    /**
+     * Method called when user wants to stop a library update
+     *
+     * @param context context of application
+     * @param notificationId id of notification
+     */
+    private fun cancelLibraryUpdate(context: Context, notificationId: Int) {
+        LibraryUpdateService.stop(context)
+        dismissNotification(context, notificationId)
```
inorichi: There's a very small chance the notification stays alive here.

The service is stopped after the `dismissNotification` call so it could trigger another update after it was dismissed. You could use `Handler().post { dismissNotification(context, notificationId) }` to avoid this.

app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderReceiver.kt
```diff
@@ -0,0 +1,144 @@
...
+import eu.kanade.tachiyomi.BuildConfig.APPLICATION_ID as ID
+
+/**
+ * Local [BroadcastReceiver] that runs on UI thread
+ * Notification calls from [UpdateDownloaderService] should be made from here.
+ */
+internal class UpdateDownloaderReceiver(val context: Context) : BroadcastReceiver() {
+
+    companion object {
+        private val NAME = UpdateDownloaderReceiver::class.java.name
```
inorichi: `const` everything!

app/src/main/java/eu/kanade/tachiyomi/data/updater/UpdateDownloaderService.kt
```diff
@@ -35,30 +41,64 @@ class UpdateDownloaderService : IntentService(UpdateDownloaderService::class.jav
...
+                putExtra(EXTRA_DOWNLOAD_URL, url)
+            }
+            return PendingIntent.getService(context, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT)
+        }
     }
 
     /**
+     * Name of Local BroadCastReceiver.
+     */
+    private val INTENT_FILTER_NAME = UpdateDownloaderService::class.java.name
```
inorichi: I think this should be inside the companion.

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadActivity.kt
```diff
@@ -46,38 +47,22 @@ class DownloadFragment : BaseRxFragment<DownloadPresenter>() {
...
-        setHasOptionsMenu(true)
-    }
-
-    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedState: Bundle?): View {
-        return inflater.inflate(R.layout.fragment_download_queue, container, false)
-    }
-
-    override fun onViewCreated(view: View, savedState: Bundle?) {
+        setContentView(R.layout.activity_download_manager)
+        setupToolbar(toolbar, backNavigation = false)
```
inorichi: I think it's better to have `backNavigation` here and remove `android.R.id.home` from `onOptionsItemSelected`

inorichi:
And I think that's it. FYI after this PR I'm going to move the `source` and `network` packages outside `data`, so any imports will have to change. I want to release the first extensions lib soon, hopefully this weekend.

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationService.kt
inorichi: In the end it's improving so I'm happy with that. :)

NoodleMage:
Implemented the changes. Ty for the help as always ;).

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -144,6 +149,7 @@ class LibraryUpdateService : Service() {
      * the alarm and release the wake lock.
      */
     override fun onDestroy() {
+        notificationManager.cancel(notificationId)
```
inorichi: I just remembered this will also remove the new updates notification, so we should use the `Handler` in the other class. :sweat:

PR #597, opened 2016-12-21
Chapters download notifications now opens download queue on tap

blackgin:
When tapping on the chapters download notification, the app is opened displaying the download queue.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
@@ -1,12 +1,16 @@
 package eu.kanade.tachiyomi.data.download
 
+import android.app.PendingIntent
 import android.content.Context
+import android.content.Intent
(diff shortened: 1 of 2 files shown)


NoodleMage:
I'm also working on this in #594 :).

The problem I have with this implementation is that you don't always open the download manager from your library. Let's say I have opened a chapter and want to see the progress of my downloads.

Then the activity is:
Main -> Manga -> Reader
On click
Main -> Download Fragment 
After I have checked the download notification I want to go back to my manga by pressing back. The problem is that this activity is removed from the stack, so I return to the library overview.

This is why I did it with a separate [activity](https://cloud.githubusercontent.com/assets/9655378/21367275/cd81d15a-c6fd-11e6-8e87-ded034fae45b.png):

@inorichi and @blackgin  which one do you prefer?

inorichi:
An activity seems better, as long as there isn't any problem by placing it above the reader (not that it will happen everyday though).

blackgin:
Ok! So I'm closing this one, and I'll follow #594 evolution

PR #611, opened 2016-12-25
Add support for user added sources in jar format (written in Kotlin instead of yaml)

paronos:
Since the sources in yaml format have limitations, I had the idea to enable binary sources loading.
If this feature is introduced a security mechanism could also be added, such as enabling only the loading of signed .jar (perhaps asking the user for unsigned .jar).
In this case it could also be possible to completely remove the embedded sources and download them from an online repository.

Anyway I created a sample source (similar to the existing Mangafox) in this manner:
- Add the new source (Mangafox**2**.kt) to the tachiomy project an build it;
- run `jar cvf Mangafox2.jar eu/kanade/tachiyomi/data/source/online/english/Mangafox2.class` (e.g. from `tachiyomi/app/build/intermediates/classes/fdroid/debug`);
- run `dx --dex --output=eu.kanade.tachiyomi.data.source.online.english.Mangafox2.jar Mangafox2.jar`

The file `eu.kanade.tachiyomi.data.source.online.english.Mangafox2.jar` should be placed in `/sdcard/Tachiyomi/parsers/`

([Mangafox2.zip](https://github.com/inorichi/tachiyomi/files/672309/Mangafox2.zip) contains both the .tk and .jar files)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/source/SourceManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/source/SourceManager.kt
@@ -15,6 +15,7 @@ import eu.kanade.tachiyomi.util.hasPermission
 import org.yaml.snakeyaml.Yaml
 import timber.log.Timber
 import java.io.File
+import dalvik.system.DexClassLoader
 
 open class SourceManager(private val context: Context) {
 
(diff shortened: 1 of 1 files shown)


inorichi:
I think this approach increases the startup time of the app quite a lot, @null-dev tried this before and said that.

I started this weekend a different approach by installing sources as an application (apk). It's already working, but not polished yet. Installing it as an apk is easier for the user, and I haven't noticed any performance issue.

paronos:
On my device (an old Android 4.1) I have not noticed a significant increase in startup times. Even by adding 10 external sources I have not noticed significant differences (here the files with which I tried: [Mangafox2-11.zip](https://github.com/inorichi/tachiyomi/files/672326/Mangafox2-11.zip)).
I do not know if the startup time increases detected by @null-dev are due to a different Android version/device or to a different loading code.

I had also thought of sources as application (.apk) but to me seemed something a bit too "heavy" (also because I imagined a download/update functionality inside Tachiyomi, which would have involved each time the Android install/update confirm dialog).

For now, I leave this PR open in case you want to test the startup time on another device (otherwise you can close it).

inorichi:
I think it should be better by wrapping the filter in a new class (state could be an enum or an Int):
 `FilterState(filter: Filter, state: State)`

Then we provide a `List<FilterState>` to the methods.

And now is the best time for any change to the source classes.

inorichi:
We'll use apks. They have a few more advantages over jars, like adding URI handlers to the manifest of every catalog (to allow to open a manga from the browser in tachiyomi).

Regarding the catalogues interfaces, now that you have worked with them, if there's any change or improvement you can think of, please tell me. Also, we have a slack channel for devs, if you want to discuss something (like the filters thing), send me a mail.

Closing.

PR #615, opened 2016-12-27
Improve catalog search filters

paronos:
This initial version implements the three state (include/exclude/ignore) search filters. It only works for MangaFox and MangaHere (I have not yet fixed the other catalogs).

Each filter has now a type (two states / three states). More types (e.g for the "year" search field) could complicate the visual interface, so I do not think that I will implement them.

I added icons for the three checkbox states ic_check_box_outline_blank_24pxic_check_box_24pxic_check_box_x_24px from https://material.io/icons/ (although ic_check_box_x_24px is a my edit), but perhaps it would be appropriate to replace them with colored versions (✓green, ✗red).


inorichi:
I would use material design icons as svg so that we can tint the icon. Maybe we can find the one missing (a cross in a box) somewhere else.

paronos:
My version is already a .svg (converted to pngs using https://github.com/sterlp/svg2png)

inorichi:
I don't know why I thought they weren't from material design, probably the lack of sleep. In that case you should use them in svg format (in a xml like the others).

paronos:
I completed editing of the remaining catalogs.
[Mangasee](http://mangaseeonline.net/search/) seems to be one of the sources with more search options. It has the following filters with more than three states:
- Sort By (Alphabetical A-Z, Alphabetical Z-A, Newest, Oldest, Most Popular, Least Popular)
- Scan Status (Any, Complete, Discontinued, Hiatus, Incomplete, Ongoing)
- Publish Status (Any, Cancelled, Complete, Discontinued, Hiatus, Incomplete, Ongoing, Unfinished)
- Type (Any, Doujinshi, Manga, Manhua, Manhwa, OEL, One-shot)

inorichi:
https://materialdesignicons.com/

Search "close", the second result is the svg we are looking for. Copy the svg path to an xml like the others.

paronos:
I replaced the current path with the one taken from the site (even if they are almost identical).

Now there are 4 filter types:
- TYPE_IGNORE_INCLUDE (simple checkbox)
- TYPE_IGNORE_INCLUDE_EXCLUDE (three state checkbox)
- TYPE_LIST (list in a Spinner)
- TYPE_TEXT (user entered text in a EditText)

mangasee_filter
This is their appearance (from Mangasee).

Apart from the possible addition of more filters to the remaining catalogs I think I've finished, let me know what do you think about it.

inorichi:
There's still a png drawable (that's why I posted the link).

For the filters, I think a sealed class is better:

```kt

sealed class Filter(/** common attrs **/) {

  class Checkbox(/** common + specific attrs **/): Filter(/** common attrs **/)
  class TriState(...): Filter(...)
  class List(...): Filter(...)
  class Text(...): Filter(...)
}
```


and used like this:

```
when (filter) {
  is Filter.Checkbox -> { ... }
  is ...
}
```

I'm not sure if genres should be another filter containing a list of checkbox filters, because in the picture everything has a header except the genres (not a big deal tho).

wopian:
I think it works better as a sidebar, like the search filters for the library view.

paronos:
Now i have deleted the pngs drawables.

Regarding the sealed class, since I have used the `states` attribute also for the checkboxes (I have used it in catalogs to represent info on two/three states, but it is not strictly necessary), the four class are almost equal (only `Text` miss `states`).
The sealed class would be:
```kotlin
    sealed class Filter(val aid: String, val aname: String, val adefaultState: Any) {
        class CheckBox(val id: String, val name: String, val states: Array<Any> = arrayOf(STATE_IGNORE, STATE_INCLUDE), val defaultState: Any = states.first()) : Filter(id, name, defaultState)
        class TriState(val id: String, val name: String, val states: Array<Any> = arrayOf(STATE_IGNORE, STATE_INCLUDE, STATE_EXCLUDE), val defaultState: Any = states.first()) : Filter(id, name, defaultState)
        class List(val id: String, val name: String, val states: Array<Any>, val defaultState: Any = states.first()) : Filter(id, name, defaultState)
        class Text(val id: String, val name: String, val defaultState: Any) : Filter(id, name, defaultState)

        companion object {
            const val STATE_IGNORE = 0
            const val STATE_INCLUDE = 1
            const val STATE_EXCLUDE = 2
        }
    }
```
I do not know if it's worth it, considering that all the equals/hashcode must be implemented manually (since it is impossible to use data classes). Let me know is `states` should be removed for checkboxes.

I leave the genres as separate checkbox for now, however another idea could be to have generic filter groups (whit a title/header).

`Uri.parse("$baseUrl/search").buildUpon().appendQueryParameter(...).appendQueryParameter(...).toString()` should be used in all the catalogs to replace string concatenation, right?

inorichi:
I don't think we will need equality for filters anyways.

After some more thinking, he way I see it now is the following, with 3 layers:

* Base class `Filter`, the class exposed to the view (the view has to know the name to display).
* Subclasses of `Filter` (it doesn't need to be a sealed class, but preferable). These subclasses are used by the view to know how to render them.
* Subclasses of the previous layer. These are never exposed to the view, but they are needed because when you send the selected filters back to the source implementation, it has to know what kind of filter it's in order to build the query.

So, let's take the previous classes I wrote as examples (they are implemented in `OnlineSource` and `open` so that they can be inherited):
```kt
class Checkbox(/** common + specific attrs **/): Filter(/** common attrs **/)
class TriState(...) : Filter(...)
class List(...) : Filter(...)
class Text(...) : Filter(...)
```

Now, Mangasee for example will create its own filter classes according to the implementation of the site.

```kt
class Genre(...) : TriState(...)
class Status(...) : List(...)
```

And in the query builder, we should be able to do something like this:

```kt
override fun searchMangaInitialUrl(query: String, filterStates: List<FilterState>): String {
  val builder = /* base url */

  // For example, when the same query is used multiple times
  filterStates.forEach {
    when (it.filter) {
        is Genre -> builder.appendQueryParameter("genre", /* the value */)
        is ...
    }
  }

  // Or if you need them as a single parameter
  val genres = filterStates.filterIsInstance(Genre::class.java).joinToString(",")
  builder.appendQueryParameter("genres", genres)

  return builder.toString()
```

And this is the theory, in practice I don't know if I missed something that makes this approach invalid.

inorichi:
The sidebar can wait, this PR is big enough already.

And yes, we should use a builder if the query grows a lot. For simple queries like the popular manga it shouldn't be required.

You can either use `android.net.Uri` or `okhttp3.HttpUrl`, but I think the later doesn't let you start from a base url (you have to set scheme, host, path...)

inorichi:
And I think with that approach we don't need the `FilterState` class. State can be stored in the second layer and modified by the view, and the third one (implementations) could provide a default value in the constructor. Of course every call to `getFilters()` will return a new instance so that everything is reset.

Any thoughts?

paronos:
I have removed the `FilterState` class.

Filters class/subclasses are abstract (OnlineSource.kt)
```kotlin
    abstract sealed class Filter<T>(val name: String, var state: T) {
        abstract class List<V>(name: String, val values: Array<V>, state: Int = 0) : Filter<Int>(name, state)
        abstract class Text(name: String, state: String = "") : Filter<String>(name, state)
        abstract class CheckBox(name: String, state: Boolean = false) : Filter<Boolean>(name, state)
        abstract class TriState(name: String, state: Int = STATE_IGNORE) : Filter<Int>(name, state) {
            companion object {
                const val STATE_IGNORE = 0
                const val STATE_INCLUDE = 1
                const val STATE_EXCLUDE = 2
            }
        }
    }
```
The implementation in in the catalogs (Mangasee.kt)
```kotlin
    private class Sort(name: String, values: Array<SortOption>, state: Int = 0) : Filter.List<SortOption>(name, values, state)
    private class Genre(name: String, val id: String = name.replace(' ', '_')) : Filter.TriState(name)
    private class TextField(name: String, val key: String) : Filter.Text(name)
    private class ListField(name: String, val key: String, values: Array<String>, state: Int = 0) : Filter.List<String>(name, values, state)

    private data class SortOption(val name: String, val keys: Array<String>, val values: Array<String>) {
        override fun toString(): String = name
    }
```
 `searchMangaInitialUrl` uses them to build the search url
```kotlin
    override fun searchMangaInitialUrl(query: String, filters: List<Filter<*>>): String {
        val url = HttpUrl.parse("$baseUrl/search/request.php").newBuilder()
        if (!query.isEmpty()) url.addQueryParameter("keyword", query)
        var genres: String? = null
        var genresNo: String? = null
        for (filter in filters) {
            when (filter) {
                is Sort -> filter.values[filter.state].keys.forEachIndexed { i, s ->
                    url.addQueryParameter(s, filter.values[filter.state].values[i])
                }
                is ListField -> if (filter.state != 0) url.addQueryParameter(filter.key, filter.values[filter.state])
                is TextField -> if (!filter.state.isEmpty()) url.addQueryParameter(filter.key, filter.state)
                is Genre -> when (filter.state) {
                    Filter.TriState.STATE_INCLUDE -> genres = if (genres == null) filter.id else genres + "," + filter.id
                    Filter.TriState.STATE_EXCLUDE -> genresNo = if (genresNo == null) filter.id else genresNo + "," + filter.id
                }
            }
        }
        if (genres != null) url.addQueryParameter("genre", genres)
        if (genresNo != null) url.addQueryParameter("genreNo", genresNo)
        return url.toString()
    }
```

inorichi:
It looks very good! And a lot more extensible.

Do you need to access `state` in `Filter`? You could get rid of the type parameter `<T>` (and the wildcard in the parameters) if filter only has a `name`.

paronos:
I use `Filter.state` in `CatalogueFragment.showFiltersDialog()` in order to check if any filter state has changed with respect to the original `state` specified in the catalog.

inorichi:
Alright, it makes sense, because you have to send an empty list if everything is default, otherwise the popular query would never be used.

We could also add a dummy filter `Header` only used by the view to categorize them, like the headers in the library filters: Filter, Sort, Display.

I prefer the current implementation rather than abstract functions 👍 .

paronos:
I added the dummy filter `Header` as `open class`. Also I'm not sure to keep the other subclasses abstract o make them `open` (in some catalogs they may suffice as they are).

Batoto
batoto

Mangasee
mangasee

inorichi:
I would keep them abstract, it makes the list of filters self explanatory, even if every catalog has to declare the classes. `Genre("Action")` is better than `TriState("Action")`

paronos:
The English catalogs have been completed, I do not think I will make other changes or add more features .
If there are no other issues left, I think you can review/merge.

inorichi:
Ok, I think I'll make a new release (probably tomorrow) before merging any PR and start working in the extensions.

There's something bugging me with these new changes: the `popularManga` methods. They seem unneeded now and I'm wondering if it would be safe to remove these methods and consider everything a search (it would probably avoid having to compare each value to send an empty list). The only problem I can think of is that some sites require some time between search calls, while browsing the directory (usually a list of popular manga) doesn't.

paronos:
On http://kissmanga.com the generic search returns all the manga in alphabetical order (a slow 10 MB html file, and it is not even possible to change the sorting).
The `popularManga` method can be removed, but in this case some catalogs (the ones in which it is not possible to order research result by popularity or in which a full search is too slow) will need to check in the `searchMangaInitialUrl` method if all the filter are set to the default state and if so switch to the popular manga url.

inorichi:
We'll keep the `popularManga` methods. Simple sites with a few manga may not have a search functionality, and checking whether you used one url or another (in your Kissmanga example) is uglier than the current implementation.

I made some local changes where I included https://github.com/sevar83/indeterminate-checkbox

This lib shows the checked icon tinted, but uses a `Boolean?` for states, where the excluded value is a `null`. It also has animations, but only when showing the usual checked or unchecked state.

I don't know if it's better to change our `Filter.TriState` value from `Int` to `Boolean?` or map the values. 

Currently, it's mostly used like this:

```kt
is Status -> add("status", arrayOf("", "Completed", "Ongoing")[filter.state])
```

If we change it, index can't be used and the difference between `false` and `null` is not clear:

```kt                    
is Status -> add("status", when (filter.state) {
    false -> ""
    true -> "Completed"
    null -> "Ongoing"
})
```

To make it look better, we can change the companion object to:

```kt
companion object {
    const val STATE_IGNORE = false
    const val STATE_INCLUDE = true
    val STATE_EXCLUDE: Boolean? = null // can't be const
}
```

But it makes the `when` clause non exhaustive:

```kt
is Status -> add("status", when (filter.state) {
    Filter.TriState.STATE_IGNORE -> ""
    Filter.TriState.STATE_INCLUDE -> "Completed"
    Filter.TriState.STATE_EXCLUDE -> "Ongoing"
    else -> throw Exception()
})
```

Another way is to use `when` first:

```kt
is Status -> when (filter.state) {
    Filter.TriState.STATE_INCLUDE -> add("status", "Completed")
    Filter.TriState.STATE_EXCLUDE -> add("status", "Ongoing")
}
```

Thoughts?

paronos:
I firstly tried https://github.com/sevar83/indeterminate-checkbox but desisted because:
- the default icons checkboxs are in my opinion not really suitable for ignore/include/exclude (genre).
- It is not possible for the user to return to the 'indeterminate' state (a custom OnCheckedChangeListener must be added to change this behavior).
> Note that when user clicks an checkbox in 'indeterminate' state it always becomes 'checked' no matter of its current 'checked' state. Its 'checked' state is not getting toggled, it becomes 'checked'. Clicking it again and it's becomes 'unchecked'. Then another click toggles it back to 'checked' state and so on. In other words user does not toggle all three states one after another. With normal operations (clicks, keypresses) it's not possible. 
- There are no animated transitions from/to the 'indeterminate' state. And so after a adding the ability to return to the indeterminate state the only animation is checked→uncheked (no animations for the other two transitions).

Even if we use this library I think the null value for a state is ugly (STATE_EXCLUDE can't even be const), so it is better to hide it. In the FilterAdapter class we can convert the the internal checkbook state to the value used by Filter.TriState (i prefer Int states for the easy array mapping, but even an enum is better than null).

inorichi:
Current view:

qemu-system-x86_64_2017-01-02_16-29-39

The buttons at the top (they do nothing yet) are always visible, the rest is scrollable.

Checkboxes use the native android checkbox (with animation). Tri-state uses a ` CheckedTextView` with a compound drawable (no animation).

Tri-state still uses integers as values.

Input text uses `TextInputLayout` from the support library, to keep the hint even after writing.

qemu-system-x86_64_2017-01-02_16-36-26

NoodleMage:
Looks awesome. I don't know if it's too performance heavy but isn't it enough/more natural if the filters are real time applied and add the top you only add a clear/reset option?

inorichi:
The filters aren't calculated locally. Every change requires a new network call, and some sites require you to wait N seconds before performing a new search, and it's not nice to send a query for every change. That's why it needs a confirmation.

NoodleMage:
Hmm. Okay than apply is okay. Maybe add a check when the filter screen is dismissed -> apply filter. To prevent accidental misslicks / back press.

inorichi:
They aren't lost after closing the drawer (but not applied). It's not a big deal, open again then apply. I've renamed it to search btw.

paronos:
The UI looks better, especially text fields (TextInputLayout) and the sidebar with button on top. There is no longer the horizontal separator before the "Genre" header and "Order by", but their presence is a matter of personal preference.

A thought came to me after seeing the search button, we could also put at top the "title" text field and basically have a complete "advanced search" screen. Any thoughts?

inorichi:
Yep, I also had this idea, but should it open the keyboard in the title search when the filter icon is clicked? If not, it's one step more for a simple search (the most common operation). If yes, half the drawer is hidden by the keyboard when it's opened, it may not be bad, but a bit weird IMO.

One big advantage of doing that, is that the search query can be kept after changing the active source.

paronos:
Originally I thought to not immediately show the keyboard, since there is 🔍 button for a simple search. You were thinking of leaving only one search function?

I also noticed that on the text fields the keyboard shows the ⏎ key, it is probably better to show the done button (in my version I used `editText.setImeOptions(EditorInfo.IME_ACTION_DONE)`).

Flat:
image

Shows the confirm button to read "Apply" however the current code uses the action_search string (
https://github.com/inorichi/tachiyomi/blob/c25af3d5adf8873259bccb629da52a7c6690b034/app/src/main/res/layout/catalogue_drawer_content.xml#L17)

Is this intentional? I quite like apply better than search for filtering.

inorichi:
> You were thinking of leaving only one search function?

Yes. If the title search functionality is moved outside the toolbar, you can change the source while keeping the query. When you don't find something, with the current implementation you have to remove the query, change source, and write the query again. That's really annoying.

@Flat, yes, I said that in a previous comment. I changed it to search to replicate the labels shown in the advanced search of websites (and also avoid a new string resource), but I can change it again if most people prefer apply.

Flat:
Ah, apologies. I skipped over that comment I guess. Although if search is implemented to the drawer it may as well be kept search I think.

paronos:
>>You were thinking of leaving only one search function?
>
>Yes. If the title search functionality is moved outside the toolbar, you can change the source while keeping the query. When you don't find something, with the current implementation you have to remove the query, change source, and write the query again. That's really annoying.

I think it's a good way to improve this use case, it happened to me only a few times but it is annoying. Regarding the opening of the keyboard I think it is good both ways.

After switching the catalog you thought of keeping the user choice also for the other fileds (limited to those with the same name/values in the two catalogs)? If so we could define some strings (in strings.xml) for the more common filed and use those in the catalogs in order to have a greater uniformity. As example for the "Type" field:
- Batoto: Any, Manga (Jp), Manhwa (Kr), Manhua (Cn), Artbook, Other.
- Mangafox: Any, Japanese Manga, Korean Manhwa, Chinese Manhua.
- Mangahere: Any, Japanese Manga (read from right to left), Korean Manhwa (read from left to right).
- Mangasee: Any, Doujinshi, Manga, Manhua, Manhwa, OEL, One-shot.
- Readmangatoday: All, Japanese Manga, Korean Manhwa, Chinese Manhua.
Common strings may be defined for the most common Manga/Manhwa/Manhua.

inorichi:
I only had in mind to keep the query. I think synchronizing everything else won't be easy.

PR #616, opened 2016-12-28
Added circular thumbnails to the catalogue list view, like the ones in the library list view

blackgin:
**Screen**

ex


M2ck:
Only for Catalogues list view? If yes could you make the same thing for Latest updates list view please.

blackgin:
Sure, I'll look into that!

inorichi:
Just a note, in list mode manga isn't initialized. If you search something, everything (that wasn't already initialized) will be a white/empty cover.

blackgin:
@inorichi Is it still the case after my second commit?

inorichi:
One of the points of the list view was to avoid a lot of network calls by not initializing anything. 

So, the options we have are:
* Always initialize them like in the grid view.
* Initialize them only over wifi.
* Don't initialize, but the cover is only shown if the manga was initialized before.
* Discard all the changes.

I don't have any preference since I only use the grid view. I will only say that there won't be an option to configure this.

I guess we will select the most voted one.

inorichi:
Only with wifi then.

blackgin:
So if there are no issues left, I think you can merge

app/src/main/res/layout/item_catalogue_list.xml
```diff
@@ -1,17 +1,34 @@
 <?xml version="1.0" encoding="utf-8"?>
```
NoodleMage: This layout should be merged with `item_catalogue_list`, same as `item_catalogue_grid`.

blackgin: Do you mean `item_library_list`?

NoodleMage: yeah.... you're correct. It's better to keep it as one item because it's basically the same thing.

app/src/main/res/layout/item_catalogue_list.xml
```diff
@@ -17,18 +17,40 @@
...
-        android:paddingEnd="@dimen/material_component_lists_right_padding"
         android:paddingLeft="@dimen/material_component_lists_text_left_padding"
-        android:paddingRight="@dimen/material_component_lists_right_padding"
         android:paddingStart="@dimen/material_component_lists_text_left_padding"
-        tools:text="Manga title"/>
+        android:paddingRight="@dimen/material_component_lists_right_padding"
+        android:paddingEnd="@dimen/material_component_lists_right_padding">
+
+        <TextView
+            android:id="@+id/unread_text"
```
NoodleMage: Please set this one to android:visibility="gone" it's redundant to load this one for the catalogue_view. Rest looks good to me 👍

PR #618, opened 2016-12-30
Added add to library dialog when downloading from catalogue

### LINKED ISSUE #448: Automatically add manga to library when downloaded
I see nobody answer me in another issue. I will ask again:

> Then maybe it would be good if we will automatically add to bookmarks manga which we downloaded? Very often I download some chapters for reading later, but forgetting to add this manga to bookmarks, and then I need find this manga in catalogue again for reading, but if there's no internet connection I can't do that.

NoodleMage:
Closes #448

screenshot_1483125815_framed


Kamui-Solo:
Yeah, finally! It looks very good, especially checkbox to remember choice or all mangas (need option to disable for those who tap it accidentally). And maybe we can do same dialog when unbookmarking manga, dialog will ask for deleting downloaded chapters?

inorichi:
This is not important enough to require a setting and adds unnecessary complexity.

A snackbar with an action to bookmark should be more than enough.

blackgin:
I support the idea of the snackbar. It avoids blocking the user (even if it may be only the first time).
I'd prefer a LENGTH_INDEFINITE snackbar, though: it appears after queuing for download and can be dismissed with a swipe

NoodleMage:
No idea why I didn't think of snackbar. Like it as well. Will change it to snackbar with add option if there is more animo. Downside is that it is never automatic.

Kamui-Solo:
Need to try snackbar, but idea looks great now.

inorichi:
Can you cleanup the code from the previous implementation?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoFragment.kt
```diff
@@ -55,6 +55,22 @@ class MangaInfoFragment : BaseRxFragment<MangaInfoPresenter>() {
...
+
+    /**
+     * Set a hint to the system about whether this fragment's UI is currently visible
+     * to the user. This hint defaults to true and is persistent across fragment instance
+     * state save and restore.
+     *
+     * @param isVisibleToUser true if this fragment's UI is currently visible to the user (default),
+     * *                        false if it is not.
+     */
+    override fun setUserVisibleHint(isVisibleToUser: Boolean) {
```
inorichi: What's the purpose of this?

NoodleMage: It's to refresh the bookmark icon. Otherwise you add it with the snackbar swipe to the left and it isn't updated. This function is called when the view is visible to the user and it calls the refresh manga method. I tested to see if it had any effect on the CPU (which wasn't the case). 

But maybe it's better to change this to a subscription on the manga.favorite value? or do you have another idea?

PR #628, opened 2017-01-05
Initial support for external sources

inorichi:
This PR is quite a big refactor to prepare Tachiyomi for external sources/catalogues loading. In the future the app won't ship **any** catalog, and every user will have to install an additional apk for each one he wants. A catalog manager should be included to ease the installation and updates.

Many of the changes in this PR were made to provide a cleaner API for the external sources. I'll try to list all the changes:

* Interfaces `SChapter` and `SManga`. Previously the sources were directly creating models from the database package, but they shouldn't know anything about database fields (`manga_id`, `source_id`, ...).
* Source Id is now stored with a `Long`. By default now the source id is autogenerated from the md5 hash (first 64 bits) of a mix of the language and catalog name. Conflicts were already a low probability (given that there won't be a lot of them), but it's safer now.
* Changes to `MangasPage`. This class was used to store a list of manga and the url of the next page. In most cases, returning the url is not needed, because it can be generated given the page, but it was used to know if there was a next page. Now it only has to return the list of manga and if there's a next page (boolean).
* Due to the previous change (not returning the next page url), `*InitialUrl` methods were removed, now you have to provide a `Request` given a page number.
* Method signatures in `OnlineSource`. Previously, most of the objects (manga, list of chapter, list of pages) were created in this class and passed down to the implementations to be filled, now they have to be created in the implementations, it's a bit more verbose but a better practice.

Known issues (there are probably more):

* ~~Mangasee search doesn't work. I haven't looked too much into it.~~
* Mangachan has commented code in the searchParse implementation.

TODO:

- [x] Fix the documentation after the method changes.
- [x] Create all the requests inside an `Observable`. It will avoid crashing the app when something is wrong.
- [ ] Move packages `source` and `network` outside `data`. These two packages (and some classes from `util`) will be the only ones exposed to the API. UPDATE: It will be done in another commit/PR.

Discussion of the API:

Should we add an optional catalog icon to the API? It could be nice to show the icon in the info tab and maybe somewhere in the catalog/latest updates screen.

If you can think of any change/improvement to the source classes, please comment it before it's too late, we won't be able to change it without breaking compatibility.


Gilfar:
Mangasee problem is just the selector, so nothing major. It should be
`override fun searchMangaSelector() = "div.requested > div.row"`

The problem with the icon could be that not all sources provide them. Also if you go by logo, then they have different styles and sizes which may clash and look out of place. This is a very good idea, however I have no clue how to make it look good and consistent.

paronos:
Some initial random tough about the API:
- `interface Source` and `abstract class OnlineSource : Source`. I like the interface / abstract implementation design, but it is not really this case since the `Source` interface is not complete and some methods of `OnlineSource` (`fetchPopularManga, fetchSearchManga, fetchLatestUpdates, prepareNewChapter, getFilterList` I think) are used in the rest of the software. One option is to add these methods to `Source` and "hide" `OnlineSource` (used only by the catalogs to write less code). To handle sources without a meaning for all of these new methods boolean flags can be added (or the catalogs implementation can return null or Unsupported exception) or "marker" interfaces like LoginSource can be used. Moreover also an "OfflineSource" will probably need most of those methods.
- The `LoginSource` may be extended to become a `SourceWithSettings`. A `PrivateServer` source will probably require the user to fill also a _server url_ field (besides username/password), or a `Dropbox` source will require a _manga path_ field.
- There may be some source for witch it is better to not cache? If so the methods `final override fun fetchPageList` and `final override fun fetchImage` make this not straightforward.
- A question: the `OnlineSource` is suitable for a generic protocol network source (http, ftp, torrent?) or is really an HTTPSource?

inorichi:
@Gilfar thanks, fixed.

I was thinking of using the favicon, since that has a fixed size. It's very small, but I really doubt we would need a big icon.

@paronos:

* Even though currently there are only `OnlineSource` implementations, the `Source` interface is used by the library, because it doesn't need any of the methods related to an online source (popular, search, latest updates). A local/folder based source wouldn't have those methods, so I think the interface has its place here.

* I was thinking of providing a `getUsername()`, `getPassword()` and `saveCredentials(username, password)` methods to the `LoginSource` interface, but you have a point for additional fields like *manga path*. I like the `SourceWithSettings` interface with an `android.content.SharedPreferences`. `LoginSource` could implement the interface. Another option is to just provide the shared preferences to `OnlineSource` and let the implementation do whatever it wants.

* Injecting the cache here comes from a very old design. I want to remove the chapter cache from this class, and many unneeded methods. I can't think of a case where caching isn't useful though (in an online/HTTP source).

* It was designed for HTTP, since all the calls use an `OkHttpClient`. Other protocols like FTP would have to implement the `Source` interface.

* That's a valid point. We could append another value, like a `versionId: Int`, initialized by default to 1. If the urls change and it's needed to restart the source, you would only need to increase this number.

paronos:
If I'm not wrong only `OnlineSource`s are show in the "Catalogues" view. I was thinking that also a local source could appear there: search and latest updates (i.e. last files added to the phone) have still meaning in this case (and popular can be used to show the list of local mangas).
Of course for a local source it is also possible to bypass the "Catalogues" view and add all the manga directly to the library (may be too invasive?), but it the case of a non HTTP network source that directly implements `Source` this would not be possible.

inorichi:
Okay, I've added the `CatalogueSource` interface and hidden `OnlineSource` in the catalogue view. I had to add the language string for now, but it could be possible to remove it later (it's used for hidden catalogues, and this feature will be removed once catalogues are external).

I also want to remove the `filters` property to sources. It's supposed to be a read only list with the default filters, but nothing prevents you from modifying it (by mistake), and it's mostly used for comparing if the filters have changed. I think we should wrap the list in a new class and override the `equals` method.

My idea for a local source was selecting a folder and register a new `Manga` for each folder, adding it to the catalog view would be hard to integrate.

Flat:
Will these changes break currently bookmarked manga, or will there be some kind of database update to map the current sources to the new external sources? And if so, are there plans to recreate the current sources in the new format?

inorichi:
Implementations are the same, so it won't break compatibility. The included sources will stay for a while, but you can install the external source for the same site and it will use that one. If there's any change in the website, you won't have to wait for a new release from the app, just install the update for the external catalog and that's it.

Once everything is rune fine and maybe after an UI for managing the sources is implemented, then all the sources will be removed from the app (again, the implementation is the same, so no migration is needed).

inorichi:
The chapter cache is no longer a member of `OnlineSource`. The solution is still bad, and `Page` is also ugly, but it will have to wait.

I think I'll be merging this PR soon before it gets too big, even if it's not totally finished, furthermore these changes shouldn't affect the user at all.

paronos:
Another change (anticipating sources removal) could be to move the source ID declarations from SourceManager to inside the catalog classes. Another option is to replace the current IDs (1..10) with those generated by OnlineSource (in association with a database upgrade).

inorichi:
I prefer to override the id for those. Otherwise I would also have to handle the mapping in the backups.

paronos:
I was thinking, with the current model it is feasible for the catalog to modify the manga pictures?
E.g. to crop the "credits" at the bottom of Mangafox pictures.

inorichi:
I can see two options:

1) Creating a network interceptor and overriding `fetchImage` (it's not `open` though), read the stream, decode the bitmap, process the pixels, and setting a new `ResponseBody` with `Response::newBuilder` containing the processed bitmap.
2) Same process but using `Observable` chaining.

However I don't know how well it would play with the progress interceptor (used to notify the download progress in the reader).

paronos:
For JPG lossless crop I was thinking of this library https://github.com/bkhall/AndroidMediaUtil (I did not try, but I think that the memory requirements are low).
Steps for Mangafox:
- crop the image so that only the bottom remains
- decode (the image should be small at his point) and check if is the "credits" text
- if so crop the bottom from original image

Working directly on the Response seems a bit hacky, but since it is an operation that is unlikely to be performed by catalogs it is not worth thinking about a cleaner solution.

inorichi:
I didn't know you could crop jpg without decoding. After some reading, I've found some limitations on it, but they don't affect our case. The top-left most pixel has to be a multiple of the MCU (usually 8), but since the watermarks are at the bottom it's not a problem. If mangafox only uses jpg then it's perfect.

I prefer this hacky solution rather than hooks on tachiyomi's end because it increases complexity and the other catalogs wouldn't benefit from it.

Issue #573, opened 2016-12-12
Add a "change orientation" button


ywecur:
If you have screen orientation locked (common for many people) and a wide page appears, you have to manually unlock it, then flip the phone. Then on the next page flip the phone again and lock it again. 

Including a button, similar to how YouTube does it, that toggles between portrait and landscape would be a lot faster and an overall better experience.

blackgin:
I was looking into the code trying to understand how this can be implemented and I think there are two options here, when the users taps on the button to switch the orientation:

1. The preference `pref_rotation_type_key` is set to 3 (lock portrait) or 4 (lock landscape) depending on the current orientation of the activity.
2. The orientation is forced only on the current activity instance. In this case, we have to track that the user  pressed the button so that we don't resubscribe to the observable of the preference mentioned above.

What do you guys think?

arkon:
This is technically built into Android P+.

arkon:
I feel like this should really just be handled outside of the app, so going to close this.

PR #632, opened 2017-01-10
Added "change orientation" button in reader (#573)

blackgin:
Implemented the feature requested in #573 following one way (2) I described there:
> The orientation is forced only on the current activity instance. In this case, we have to track that the user pressed the button so that we don't resubscribe to the observable of the rotation preference.

The button as action:

ex

Critiques? Suggestions?


app/src/main/res/menu/reader.xml
```diff
@@ -9,6 +9,12 @@
         app:showAsAction="ifRoom" />
 
     <item
+        android:id="@+id/action_switch_orientation"
+        android:icon="@drawable/ic_screen_rotation_white_24dp"
```
NoodleMage: Please use [vector drawables](http://stackoverflow.com/a/28704026) (xml) when possible

inorichi:
There's no simple way to just ignore the system rotation lock. You would have to subscribe to the accelerometer or implement [OrientationEventListener](https://developer.android.com/reference/android/view/OrientationEventListener.html#onOrientationChanged(int)) which returns the orientation between 0-359 and implement your custom logic to decide when the device is in portrait or landscape.

The system uses [WindowOrientationListener](https://github.com/android/platform_frameworks_base/blob/master/services/core/java/com/android/server/policy/WindowOrientationListener.java) but that class is quite complex, private and its implementation might differ across API levels.

IMO having two ways to control the orientation within the app where one is deactivating the other is ugly.

blackgin:
So we just give up on that feature request, I guess

inorichi:
Or we remove the other implementation or find a better solution.

A fix to not break the current implementation is making the button change the setting to 3 or 4 (force portrait/landscape). That way you can set again free and lock in current.

blackgin:
> Or we remove the other implementation

I wonder how much the rotation preference is actually used today

> A fix to not break the current implementation is to make the button change the setting to 3 or 4 (force portrait/landscape). That way you can set again free and lock in current.

Yeah, but the thing is that this way we change the setting for the app in general, not only the current instance of the reader. Is it a problem? I don't know

paronos:
I find the button useful, maybe it can be used to switch between the four rotation setting (or maybe only free/force portrait/force landscape). This way even if the global setting is changed it is easy to change it back.
Possible icons:
Free free
Lock Lock
Force portrait portrait
Force landscape landscape

blackgin:
> maybe the button can be used to switch between the four rotation setting

I personally don't like this approach very much. I think the button on the reader should offer a quick way to switch orientation. By cycling between the four states this simplicity may be lost.

ywecur:
What I had in mind is that the button forces either landscape or portrait, not that it toggles wether the orientation is locked or not.

lpchaim:
I don't know how feasible this is, but what if the button just toggled a rotation lock and nothing else? Much like the CyanogenMod implementation (at least I think it's not from upstream), where the quick setting tile makes the screen remain as it were when activated and restores normal behavior otherwise.

blackgin:
@ywecur forcing an orientation still implies locking _at least_ the current instance of the reader (without changing the preference). And this is the current implementation in this PR.

@lpchaim it is feasible, but it still means inhibiting the preference when locking the current orientation, which wasn't liked in the first place. But maybe your idea is better than my current implementation, since tapping the button again restores the usual behaviour (listening to the preference value).
What do you think about it @inorichi ?

inorichi:
I think your original approach and allowing to read from the preference again with a long tap (it may need some ugly reflection though) is the best we can do.

blackgin:
> allowing to read from the preference again with a long tap

Uhm, to me this seems a little too "hidden". IMHO the user is generally not used to long tap buttons. If we choose this path, to avoid breaking the default action behaviour we also have to move the button from the actionbar to the "overlay" (which may be better in any case).

inorichi:
Yes, it's hidden, but at least you can restore it. We could even show a toast the first time it's used (backed by a preference so it's only shown once per installation) where we tell the user he can restore it with a long tap. 

What do you mean by overlay?

blackgin:
> What do you mean by overlay?

I mean that it is shown in the bottom right corner "overlaying" the displayed page, just above the bottom bar. Like most apps do with the fullscreen button. 

e.g.

lpchaim:
Speaking of overlays and hiding stuff, it just came to me that there's also the way the Pocket app does it. It's an outlier as far as I know, but just to throw this out there, it briefly shows an overlay after orientation changes which locks the current position upon being tapped then disappears. If locked, the locked icon briefly reappears upon a physical orientation change so that the user can then disable it.

blackgin:
Changed implementation to what @inorichi proposed:

- Moved button to "overlay"
- A toast is shown the first two times the user toggles the orientation
- The color of the button depends on the reader theme

inorichi:
I tried the implementation a while ago and I don't really like the overlay, IMO it's a lot better in the toolbar. 

I'm still not sure about how to handle this. I think the best thing we can do is to list what other readers do and pick the best approach. If anyone has enough free time, please do the research.

Horoneru:
Based on some other readers I saw from friends using them, I'd believe the best approach would be in the toolbar really.
Those that do have a rotation toggle or option from the viewer are placing it on the toolbar, and IMO if there should be one, it should only be there.
Unfortunately, I don't know the names of those, I only saw a glimpse of their UIs.
Still, I'd favor the option only being in the settings, as I find the rotation mode toggle has a pretty niche use for users when it's an action inside the viewer.

blackgin:
I did a little research and tried some of the other manga apps (most of them suck btw). None of the apps I tried have a preference for the rotation, only a toggle in reader be it toggling between free/locked or portrait/landscape. Except Aizoban that has both preference and toggle, but the preference isn't multi-option, it is just an orientation lock.

These are the apps :
Mangarock, Aizoban, ZingBox, Mangazone, MangaDLR

I think that we should just decide if we prefer the preference or the toggle and avoid having the two ways coexisting.

inorichi:
Sorry I've been off these days with the border crop feature and forgot about everything else.

The force portrait/landscape was something I thought it could be interesting to have when coding it. But most likely nobody uses them (?). If that's the case, I don't mind removing support for those.

What I at least want is a rotation lock even if it's unlocked in the system: I always leave it locked when reading, but I never lock the system's.

If it's decided to remove the current implementation in the settings, I vote for a toggle (in the toolbar) that behaves the same as the system's. It would work like this:

System unlocked + App unlocked = **Can rotate**
System unlocked + App locked = Can't rotate
System locked + App unlocked = Can't rotate
System locked + App locked = Can't rotate

A button that directly changes the rotation is not useful in my opinion, since you are going to rotate the screen after that anyways.

blackgin:
I like your idea, but the problem that was explained in the original issue (#573) remains unsolved:

> If you have screen orientation locked (common for many people) and a wide page appears, you have to manually unlock it, then flip the phone. Then on the next page flip the phone again and lock it again.

inorichi:
Oops, you're right. I totally forgot what the original issue was about and overlooked that use case. Thinking about it again, my solution doesn't make much sense.

I think we are going to keep the current implementation for three reasons:

1. It's the one that provides more customization.
2. By looking at the comments and reactions, people (including me) prefer the current implementation.
3. Even if an icon in the toolbar rotates the screen (not just a lock toggle), you still have to tap the screen to show the toolbar, click the icon, and hide the toolbar again. That's just one step (in most ROMs at least) further than enabling/disabling the system orientation: drag twice to open the system bar and click the orientation lock, then click outside it. If changing the orientation is required too often, it should be considered to just keep the orientation unlocked.

paronos:
Maybe a rotation gesture could be added. It would only rotate the current page (by 90 degrees steps), similar to the zoom gesture which does not affect the following pages.

inorichi:
I agree, it's probably the best solution. However it might be hard to integrate with the image library. And when the gesture is triggered, if the system rotation is unlocked, it must set the app rotation to locked. Otherwise it would rotate again to its original position.

blackgin:
So the differences with respect to the starting implementation would be:
1. The rotation is triggered by gesture, not button
2. The rotation would affect only the single image, not the activity, nor other images

inorichi:
1. Yes.
2. I like that. Probably better than rotating the activity.

inorichi:
I'm closing this PR because as discussed before we won't add an icon for rotating, it should be done with a rotation gesture.

PR #633, opened 2017-01-11
Add Sort filter [Catalogs]

paronos:
This PR adds a sorting catalog filter with an appearance similar to the library filter view.
It occupies more screen space compared to a Spinner, if despite this it is still ok I will update the other catalogs (for now only MangaFox is updated).
screen

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/source/model/Filter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/source/model/Filter.kt
@@ -2,6 +2,7 @@ package eu.kanade.tachiyomi.data.source.model
 
 sealed class Filter<T>(val name: String, var state: T) {
     open class Header(name: String) : Filter<Any>(name, 0)
+    open class Separator(name: String = "") : Filter<Any>(name, 0)
     abstract class List<V>(name: String, val values: Array<V>, state: Int = 0) : Filter<Int>(name, state)
     abstract class Text(name: String, state: String = "") : Filter<String>(name, state)
     abstract class CheckBox(name: String, state: Boolean = false) : Filter<Boolean>(name, state)
@@ -9,10 +10,16 @@ sealed class Filter<T>(val name: String, var state: T) {
         fun isIgnored() = state == STATE_IGNORE
         fun isIncluded() = state == STATE_INCLUDE
         fun isExcluded() = state == STATE_EXCLUDE
+
         companion object {
             const val STATE_IGNORE = 0
             const val STATE_INCLUDE = 1
             const val STATE_EXCLUDE = 2
         }
+    }
+
+    abstract class Sort<V>(name: String, val values: Array<V>, state: Selection? = null)
+        : Filter<Sort.Selection?>(name, state) {
+        data class Selection(val index: Int, val ascending: Boolean)
     }
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/source/online/english/Batoto.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/source/online/english/Batoto.kt
@@ -107,6 +107,10 @@ class Batoto : ParsedOnlineSource(), LoginSource {
                     val sel = if (filter.state) filter.valTrue else filter.valFalse
                     if (!sel.isEmpty()) url.addQueryParameter(filter.key, sel)
                 }
+                is OrderBy -> {
+                    url.addQueryParameter("order_cond", arrayOf("title", "author", "artist", "rating", "views", "update")[filter.state!!.index])
+                    url.addQueryParameter("order", if (filter.state?.ascending == true) "asc" else "desc")
+                }
             }
         }
         if (!genres.isEmpty()) url.addQueryParameter("genres", genres)
@@ -288,6 +292,9 @@ class Batoto : ParsedOnlineSource(), LoginSource {
     private class TextField(name: String, val key: String) : Filter.Text(name)
     private class ListField(name: String, val key: String, values: Array<ListValue>, state: Int = 0) : Filter.List<ListValue>(name, values, state)
     private class Flag(name: String, val key: String, val valTrue: String, val valFalse: String) : Filter.CheckBox(name)
+    private class OrderBy() : Filter.Sort<String>("Order by",
+            arrayOf("Title", "Author", "Artist", "Rating", "Views", "Last Update"),
+            Filter.Sort.Selection(4, false))
 
     // [...document.querySelectorAll("#advanced_options div.genre_buttons")].map((el,i) => {
     //     const onClick=el.getAttribute('onclick');const id=onClick.substr(14,onClick.length-16);return `Genre("${el.textContent.trim()}", ${id})`
@@ -298,8 +305,9 @@ class Batoto : ParsedOnlineSource(), LoginSource {
             ListField("Type", "type", arrayOf(ListValue("Any", ""), ListValue("Manga (Jp)", "jp"), ListValue("Manhwa (Kr)", "kr"), ListValue("Manhua (Cn)", "cn"), ListValue("Artbook", "ar"), ListValue("Other", "ot"))),
             Status(),
             Flag("Exclude mature", "mature", "m", ""),
-            ListField("Order by", "order_cond", arrayOf(ListValue("Title", "title"), ListValue("Author", "author"), ListValue("Artist", "artist"), ListValue("Rating", "rating"), ListValue("Views", "views"), ListValue("Last Update", "update")), 4),
-            Flag("Ascending order", "order", "asc", "desc"),
+            Filter.Separator(),
+            OrderBy(),
+            Filter.Separator(),
             Filter.Header("Genres"),
             ListField("Inclusion mode", "genre_cond", arrayOf(ListValue("And (all selected genres)", "and"), ListValue("Or (any selected genres) ", "or"))),
             Genre("4-Koma", 40),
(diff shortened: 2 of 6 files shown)


NoodleMage:
I personalty like this one better, but like you said it occupies more space so I can understand if people prefer the way it's implemented now.

Because it's a discussion for the filter menu, I do think that the buttons should be moved to the bottom (always visible) like this, opinions?:
image
[source ](https://dribbble.com/shots/2692225-Android-App-Filtering-Menu/attachments/543446)

inorichi:
I also prefer that to a spinner and a checkbox for ascending/descending. Regarding the implementation, you are calling addView` in `onBindViewHolder` without removing old views (is it placing one above the other?).

I don't have any preference about the buttons at the top/bottom, so I won't join that discussion.

paronos:
I also prefer this view (however the checkbox for ascending/descending may be removed by adding to the spinner for each sorting order both the ascending/descending option).

Regarding the implementation I have added a `removeAllViews()`.
To allow no initial sorting selection `Sort.Selection` is declared optionally `null`, the same effect could be achieved by using a `Sort.Selection.index = -1`. There is a preference on this?

Regarding the top/bottom buttons I think that if the filter view effectively replaces the search function it is better to have the buttons on top (since on the bottom they will be hidden by the soft keyboard that will appear when entering the manga title).

inorichi:
What about a expandable header? Something like https://github.com/davideas/FlexibleAdapter/raw/master/screenshots/secondary_functionalities.png with [FlexibleAdapter](https://github.com/davideas/FlexibleAdapter). We're still using v4, but I want to begin introducing v5 since it's almost done. By default they would be expanded with a view similar to header with the arrow icons to expand/collapse at the end. We could also have sticky headers, so that you can collapse a long list (genres) even at the bottom. 

Advantages of using that lib is that it's easy to wire. The downside is that we have to create more classes in the catalogue view because the items have to implement an interface or a class. Also genres would need to be wrapped in a list in the sources.

I prefer a null selection. It's easier to provide a default value with `?:`.

paronos:
Although not essential it looks like a nice thing to add.
Regarding genres it is probably better to allow generic filter grouping by adding to `Filter` a class like
`class Group(name_header: String, val values: Array<Filter<*>>, state_expanded: Boolean = true) : Filter<Boolean>(name_header, state_expanded)`

I updated the remaining catalogs. I do not know if you want to merge this PR considering that introducing FlexibleAdapter will change many things, if not the last commit also fix a bug in Mangasee `Genre` filters (you may have to fix it in the master branch).

inorichi:
Yes, it would be a class like that, but with the expanded state stored/handled in the frontend (catalogue).

It would be implemented separately. I will have to fork the project and change the package name to a different one, because it's the same as v4 and I don't want to migrate the other adapters yet.

One last question, is it really needed to do `if (filters.isEmpty()) getFilterList() else filters` in the search request? They usually don't care if there are missing parameters.

paronos:
In general it is not needed, but:
- some source requires some parameter in order to work (e.g. if I remember correctly Readmangatoday does not work if the "type" and "status" parameters are not set);
- in the base url the sort order is not set (the default filters have instead popularity/descending).

The cleanest way would be to have a different base url / initial parameters if `filters.isEmpty()`

inorichi:
I've implemented the group like this: 
```kt
abstract class Group<V>(name: String, state: List<V>): Filter<List<V>>(name, state)
```

I'm not sure if I should make V extends a Filter: `<V : Filter<*>>` or allow anything.

I overrode the `equals` and `hashCode` with the `name` and `state` to the `Filter` sealed class (not its implementations) . When the equals is called for a group, it will know if any of its filter has changed.

qemu-system-x86_64_2017-01-14_10-57-56

I've reimplemented the adapter and now we have sticky and collapsible headers (not going to use that theme, I copied it from the flexible adapter example app). I have to cleanup the implementation yet.


When creating the search request, you now have to control the group, ex:
```kt
when (filter) {
    is Status -> url.addQueryParameter(filter.id, filter.state.toString())
    is GenreList -> {
        filter.state.forEach { genre ->
            url.addQueryParameter(genre.id, genre.state.toString())
        }
    }
}
```

paronos:
I looked at the new version, it seems a good result. Regarding `V`, for an object that is not a Filter the UI will just skip the value?

Non exactly related to the filters, but I made a little test one the implementation of a catalog that reads local .zip/.cbz files: https://github.com/paronos/tachiyomi/tree/local/app
The class eu.kanade.tachiyomi.data.source.LocalSource directly implements CatalogueSource. From this test I noticed:
- The Catalogues view shows as name the class name (_eu.kanade.tachiyomi.data.source_) instead of the catalog name (_LocalSource_).
- The catalog is not showed in Setting->Sources. If the catalog needs some setting (username/password/etc) there must be a place to set them.
- The catalog need to have access to `context`. I simple added it as constructor parameter, but is this still possible for the external .apk sources?

inorichi:
Yes, it would be ignored in the UI.

* The view uses the `toString()` method, you can just override it.
* What is your concept of `LocalSource`? LAN? In my mind that name suggests a directory based source (everything located in the device).
* At first the context was needed, but became unused after changing the dependency injection library (dagger -> injekt) so I removed it. If possible I would prefer to avoid context. What do you need it for?

paronos:
Yesterday I forgot to commit/push the changes on the repository (the source file is [LocalSource.kt](https://github.com/paronos/tachiyomi/blob/local/app/src/main/java/eu/kanade/tachiyomi/data/source/LocalSource.kt)).
The source shows manga stored on the local device (one separate directory for each manga, each directory must contain a .zip/.cbz file for each manga chapter).
In this case the source need `context` for `context.cacheDir` and `context.getString(R.string.app_name)`.

If settings are available for the source they can be used to let the user chose the directory(s) that contain the local mangas.

inorichi:
I prefer to keep  `name` and maybe remove `.toString()` in the future, but for now it's going to be like this.

As a suggestion, the local source should read either a zip, cbz or a directory. The later wouldn't need the cache directory and should be easy to implement.

The local source will be included in the main app. I don't see any reason why it should be a separate apk, so it's fine to instantiate the source from the manager and give the context as parameter.

paronos:
I changed the source to read also directories. I also tried to set `SManga.thumbnail_url` with a `file://...` url, but it does not seem to work.

paronos:
Besides the thumbnail problem, I was also thinking that the zip extraction (in the cache directory) is not the best thing to do, since ZipInputStream could be directly used to read each image (or the `jar:` protocol, e.g. `BufferedReader(InputStreamReader(java.net.URL("jar:file:///storage/sdcard0/Tachiyomi/mangas/m1/c1.zip!/i1.jpg").openStream()))`.
This is not possible with the current implementation, given that `Page` takes as a parameter an `Uri`.

inorichi:
Yeah, yesterday I was thinking the same. The problem is Subsampling also works with `Uri`. It may not be easy to support.

The covers not loading is probably because of [this file](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaModelLoader.kt).

inorichi:
We can build the `Uri` with:

```kt
val file = File(/* path to .zip */)
val zip = ZipFile(file)
val entries = zip.entries().toList().map { entry ->
    Uri.parse("jar:file://$file!/$entry")
}
```

And modify every decoder in my subsampling fork to process `Uri` starting with `jar:` differently:

```kt
val uriString = uri.toString()
if (uriString.startsWith("jar:")) {
    val inputStream = URL(uriString).openStream()
    // rest of the decoder work
}
```

My fork of rapid decoder will also need similar changes.

paronos:
I had a doubt, `jar:` works also with `content://` (besides `file://` and `http://`)?

Anyway, having to change the subsampling library maybe it is better to add direct Stream support. In this way other formats (e.g. .rar files) can be supported without the further changes to the library.

To implement this without too many changes an `abstract class UriOpenStream extendes Uri` with an `openStream()` abstract method could be added, the decoders modified with
```kotlin
if (uri is UriOpenStream) {
    val inputStream = uri.openStream()
    // rest of the decoder work
}
```
and the source whit the implementation of `openStream()` based on the file type (.zip, .rar).

inorichi:
If it's coming from the Storage Access Framework it won't work. Otherwise, probably. Http could also work. Hopefully `content://` won't be required, because IIRC Android is not too restrictive for read only content on the SD.

Rar files would have to be decompressed anyways (AFAIK). 

Extending `Uri` would allow to not change the method signature, but I prefer to work with the platform classes and not introduce new types (yet).

paronos:
With `content://` not required `URL.openStream()` may be a good choice.

Of course .rar are not supported natively, but given that with a library it is possible to get the stream for a single file, it may be possible to add custom `rar:` (instead of `jar:`)
A `RarURL(uriString).openStream()` could wrap a code similar to:
```java
import com.github.junrar.*;

// Open an archive
Archive myRAR = new Archive(new File("myRAR.rar"));

// Get all the files
List<FileHeader> files = myRAR.getFileHeaders();

// The input stream of the file
InputStream ins;

// Iterate and get the file with your name
for (FileHeader hd : files)
{
    if (hd.getFileNameW().equals("myfileinrar"))
    {
        ins = myRAR.getInputStream(hd);
        break;
    }
}
```

paronos:
> The covers not loading is probably because of [this file](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaModelLoader.kt).

It was indeed [this file](https://github.com/paronos/tachiyomi/blob/local/app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaModelLoader.kt). I had to add a `private val baseFileLoader = Glide.buildModelLoader(Uri::class.java, InputStream::class.java, context)` and `if (url!!.startsWith("file:/")) return FileLoader(baseFileLoader).getResourceFetcher(File(url.substring(6)), width, height)`.

Should I make a PR for this source? I do not know if you plan to integrate a source like this or wait for a different/better implementation (anyway it was only started to test the CatalogueSource interface).

inorichi:
Sure, this source will be bundled. I'll review the implementation in the PR.

PR #648, opened 2017-01-23
Local manga in zip/cbz/folder format

paronos:
This source allows to read locally stored mangas in zip/cbz/folder format.
The zip/cbz/folder files of each manga must be stored in a subdirectory of `/sdcard/Tachiyomi/mangas/`. As example:
```
/sdcard/Tachiyomi/mangas/MANGA_TITLE_1/chapter_1.zip
/sdcard/Tachiyomi/mangas/MANGA_TITLE_1/chapter_2.zip
/sdcard/Tachiyomi/mangas/MANGA_TITLE_2/chapter_1.cbz
/sdcard/Tachiyomi/mangas/MANGA_TITLE_2/cover.jpg
/sdcard/Tachiyomi/mangas/MANGA_TITLE_3/chapter_1_folder
/sdcard/Tachiyomi/mangas/MANGA_TITLE_3/chapter_2_folder
...
```
If a `cover.jpg` file is present it is used as cover for the manga.

The source extracts the zip/cbz files to the chache directory . A better option would be to directly read the image files from the .zip by using `java.net.URL("jar:file://sdcard/Tachiyomi/mangas/m1/c1.zip!/i1.jpg").openStream()`, but it requires changes in the `subsampling-scale-image-view` and `RapidDecoder` libraries.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
@@ -10,14 +10,16 @@ import eu.kanade.tachiyomi.util.ZipContentProvider
 import net.greypanther.natsort.CaseInsensitiveSimpleNaturalComparator
 import rx.Observable
 import java.io.File
+import java.net.URLConnection
 import java.util.*
+import java.util.zip.ZipEntry
 import java.util.zip.ZipFile
 
 class LocalSource(private val context: Context) : CatalogueSource {
-    private class OrderBy() : Filter.Sort("Order by", arrayOf("Title", "Date"), Filter.Sort.Selection(1, false))
+    private class OrderBy() : Filter.Sort("Order by", arrayOf("Title", "Date"), Filter.Sort.Selection(0, true))
 
     private val fileProtocol = "file://"
-    private fun isImage(name: String) = name.endsWith(".jpg", true) || name.endsWith(".jpeg", true) || name.endsWith(".png", true) || name.endsWith(".gif", true)
+    private fun isImage(name: String) = URLConnection.guessContentTypeFromName(name).orEmpty().startsWith("image/")
 
     override val id = 0L;
     override val name = "LocalSource"
@@ -49,13 +51,16 @@ class LocalSource(private val context: Context) : CatalogueSource {
         if (chapFile.isDirectory)
             return Observable.just(chapFile.listFiles()
                     .filter { !it.isDirectory && isImage(it.name) }
-                    .sortedWith(Comparator<File> { t1, t2 -> CaseInsensitiveSimpleNaturalComparator.getInstance<String>().compare(t1.nameWithoutExtension, t2.nameWithoutExtension) })
+                    .sortedWith(Comparator<File> { t1, t2 -> CaseInsensitiveSimpleNaturalComparator.getInstance<String>().compare(t1.name, t2.name) })
                     .mapIndexed { i, v -> Page(i, fileProtocol + v.absolutePath, fileProtocol + v.absolutePath, Uri.fromFile(v)).apply { status = Page.READY } })
         else
             return Observable.just(ZipFile(chapFile).entries().toList()
-                    .filter { !it.isDirectory && isImage(it.name) }.map { entry ->
-                "content://${ZipContentProvider.PROVIDER}${chapFile.absolutePath}!/${entry.name}"
-            }.mapIndexed { i, v -> Page(i, v, v, Uri.parse(v)).apply { status = Page.READY } })
+                    .filter { !it.isDirectory && isImage(it.name) }
+                    .sortedWith(Comparator<ZipEntry> { t1, t2 -> CaseInsensitiveSimpleNaturalComparator.getInstance<String>().compare(t1.name, t2.name) })
+                    .mapIndexed { i, v ->
+                        val path = "content://${ZipContentProvider.PROVIDER}${chapFile.absolutePath}!/${v.name}"
+                        Page(i, path, path, Uri.parse(path)).apply { status = Page.READY }
+                    })
     }
 
     override fun fetchPopularManga(page: Int) = fetchSearchManga(page, "", getFilterList())
--- a/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
@@ -46,8 +46,8 @@ class LocalSource(private val context: Context) : CatalogueSource {
         }
 
         private fun getBaseDirectories(context: Context): List<File> {
-            return DiskUtil.getExternalStorages(context)
-                    .map { File("${it.absolutePath}/Tachiyomi/local") }
+            val c = File.separator + context.getString(R.string.app_name) + File.separator + "local"
+            return DiskUtil.getExternalStorages(context).map { File(it.absolutePath + c) }
         }
     }
 
--- a/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
@@ -5,10 +5,10 @@ import android.net.Uri
 import android.os.Environment
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.source.model.*
+import eu.kanade.tachiyomi.util.ChapterRecognition
 import net.greypanther.natsort.CaseInsensitiveSimpleNaturalComparator
 import rx.Observable
 import java.util.*
-import java.util.regex.Pattern
 import java.util.zip.ZipInputStream
 import java.io.*
 
@@ -16,7 +16,6 @@ class LocalSource(private val context: Context) : CatalogueSource {
     private class OrderBy() : Filter.Sort("Order by", arrayOf("Title", "Date"), Filter.Sort.Selection(1, false))
 
     private val fileProtocol = "file://"
-    private val floatPattern = Pattern.compile("[0-9]+[.]?[0-9]*")
     private fun isImage(name: String) = name.endsWith(".jpg", true) || name.endsWith(".jpeg", true) || name.endsWith(".png", true) || name.endsWith(".gif", true)
 
     override val id = 0L;
@@ -38,11 +37,7 @@ class LocalSource(private val context: Context) : CatalogueSource {
                         val chapNameCut = chapName.replace(manga.title, "", true)
                         name = if (chapNameCut.isEmpty()) chapName else chapNameCut
                         date_upload = chapterFile.lastModified()
-                        var chapNum = "0"
-                        val matcher = floatPattern.matcher(name)
-                        while (matcher.find())
-                            chapNum = matcher.group()
-                        chapter_number = chapNum.toFloat()
+                        ChapterRecognition.parseChapterNumber(this, manga)
                     }
                 }
         return Observable.just(chapters.sortedByDescending { it.chapter_number })
--- a/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
@@ -6,11 +6,12 @@ import android.os.Environment
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.source.model.*
 import eu.kanade.tachiyomi.util.ChapterRecognition
+import eu.kanade.tachiyomi.util.DiskUtil
 import eu.kanade.tachiyomi.util.ZipContentProvider
 import net.greypanther.natsort.CaseInsensitiveSimpleNaturalComparator
 import rx.Observable
 import java.io.File
-import java.net.URLConnection
+import java.io.FileInputStream
 import java.util.*
 import java.util.zip.ZipEntry
 import java.util.zip.ZipFile
@@ -19,7 +20,6 @@ class LocalSource(private val context: Context) : CatalogueSource {
     private class OrderBy() : Filter.Sort("Order by", arrayOf("Title", "Date"), Filter.Sort.Selection(0, true))
 
     private val fileProtocol = "file://"
-    private fun isImage(name: String) = URLConnection.guessContentTypeFromName(name).orEmpty().startsWith("image/")
 
     override val id = 0L;
     override val name = "LocalSource"
@@ -48,19 +48,21 @@ class LocalSource(private val context: Context) : CatalogueSource {
 
     override fun fetchPageList(chapter: SChapter): Observable<List<Page>> {
         val chapFile = File(chapter.url.substring(7))
-        if (chapFile.isDirectory)
+        if (chapFile.isDirectory) {
             return Observable.just(chapFile.listFiles()
-                    .filter { !it.isDirectory && isImage(it.name) }
+                    .filter { !it.isDirectory && DiskUtil.isImage(it.name, { FileInputStream(it) }) }
                     .sortedWith(Comparator<File> { t1, t2 -> CaseInsensitiveSimpleNaturalComparator.getInstance<String>().compare(t1.name, t2.name) })
                     .mapIndexed { i, v -> Page(i, fileProtocol + v.absolutePath, fileProtocol + v.absolutePath, Uri.fromFile(v)).apply { status = Page.READY } })
-        else
+        } else {
+            val zip = ZipFile(chapFile)
             return Observable.just(ZipFile(chapFile).entries().toList()
-                    .filter { !it.isDirectory && isImage(it.name) }
+                    .filter { !it.isDirectory && DiskUtil.isImage(it.name, { zip.getInputStream(it) }) }
                     .sortedWith(Comparator<ZipEntry> { t1, t2 -> CaseInsensitiveSimpleNaturalComparator.getInstance<String>().compare(t1.name, t2.name) })
                     .mapIndexed { i, v ->
                         val path = "content://${ZipContentProvider.PROVIDER}${chapFile.absolutePath}!/${v.name}"
                         Page(i, path, path, Uri.parse(path)).apply { status = Page.READY }
                     })
+        }
     }
 
     override fun fetchPopularManga(page: Int) = fetchSearchManga(page, "", getFilterList())
--- a/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
@@ -1,0 +1,138 @@
+package eu.kanade.tachiyomi.source
+
+import android.content.Context
+import android.net.Uri
+import android.os.Environment
+import eu.kanade.tachiyomi.R
+import eu.kanade.tachiyomi.source.model.*
+import net.greypanther.natsort.CaseInsensitiveSimpleNaturalComparator
+import rx.Observable
+import java.util.*
+import java.util.regex.Pattern
+import java.util.zip.ZipInputStream
+import java.io.*
+
+class LocalSource(private val context: Context) : CatalogueSource {
+    private class OrderBy() : Filter.Sort("Order by", arrayOf("Title", "Date"), Filter.Sort.Selection(1, false))
+
+    private val fileProtocol = "file:/"
+    private val floatPattern = Pattern.compile("[0-9]+[.]?[0-9]*")
+    private fun isImage(name: String) = name.endsWith(".jpg", true) || name.endsWith(".jpeg", true) || name.endsWith(".png", true) || name.endsWith(".gif", true)
+
+    override val id = 0L;
+    override val name = "LocalSource"
+    override val lang = "en"
+    override val supportsLatest = false
+
+    override fun toString() = context.getString(R.string.local_source)
+
+    override fun fetchMangaDetails(manga: SManga) = Observable.from(arrayOf(manga))
+
+    override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
+        val chapters = File(manga.url.substring((6))).listFiles()
+                .filter { it.isDirectory || it.extension.equals("zip", true) || it.extension.equals("cbz", true) }
+                .map { chapterFile ->
+                    SChapter.create().apply {
+                        url = fileProtocol + chapterFile.absolutePath
+                        val chapName = if (chapterFile.isDirectory) chapterFile.name else chapterFile.nameWithoutExtension
+                        val chapNameCut = chapName.replace(manga.title, "", true)
+                        name = if (chapNameCut.isEmpty()) chapName else chapNameCut
+                        date_upload = chapterFile.lastModified()
+                        var chapNum = "0"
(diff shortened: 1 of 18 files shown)


wopian:
I think using `Tachiyomi/local` would be more suitable than `Tachiyomi/mangas`, as then you have the following directory structure which is self-explanatory to what they include.

```
Tachiyomi
├─downloads
├─local
└─parsers
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+import rx.Observable
+import java.util.*
+import java.util.regex.Pattern
+import java.util.zip.ZipInputStream
+import java.io.*
+
+class LocalSource(private val context: Context) : CatalogueSource {
+    private class OrderBy() : Filter.Sort("Order by", arrayOf("Title", "Date"), Filter.Sort.Selection(1, false))
+
+    private val fileProtocol = "file:/"
```
inorichi: Shouldn't the protocol be `file://`? And absolute paths have three slashes

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+    private fun isImage(name: String) = name.endsWith(".jpg", true) || name.endsWith(".jpeg", true) || name.endsWith(".png", true) || name.endsWith(".gif", true)
+
+    override val id = 0L;
+    override val name = "LocalSource"
+    override val lang = "en"
+    override val supportsLatest = false
+
+    override fun toString() = context.getString(R.string.local_source)
+
+    override fun fetchMangaDetails(manga: SManga) = Observable.from(arrayOf(manga))
```
inorichi: `Observable.just(manga)`

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+                        name = if (chapNameCut.isEmpty()) chapName else chapNameCut
+                        date_upload = chapterFile.lastModified()
+                        var chapNum = "0"
+                        val matcher = floatPattern.matcher(name)
+                        while (matcher.find())
+                            chapNum = matcher.group()
+                        chapter_number = chapNum.toFloat()
+                    }
+                }
+        return Observable.from(arrayOf(chapters.sortedByDescending { it.chapter_number }))
```
inorichi: `Observable.just(chapters.....)`

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+                    }
+                }
+                zipIn.closeEntry()
+                entry = zipIn.nextEntry
+            }
+            zipIn.close()
+            pages = images.sortedWith(Comparator<Pair<String, Uri>> { t1, t2 -> CaseInsensitiveSimpleNaturalComparator.getInstance<String>().compare(t1.first, t2.first) })
+                    .mapIndexed { i, v -> Page(i, chapter.url + "!" + v.first, chapter.url + "!" + v.first, v.second).apply { status = Page.READY } }
+        }
+        return Observable.from(arrayOf(pages))
```
inorichi: `Observable.just(pages)`

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+                        url = fileProtocol + chapterFile.absolutePath
+                        val chapName = if (chapterFile.isDirectory) chapterFile.name else chapterFile.nameWithoutExtension
+                        val chapNameCut = chapName.replace(manga.title, "", true)
+                        name = if (chapNameCut.isEmpty()) chapName else chapNameCut
+                        date_upload = chapterFile.lastModified()
+                        var chapNum = "0"
+                        val matcher = floatPattern.matcher(name)
+                        while (matcher.find())
+                            chapNum = matcher.group()
+                        chapter_number = chapNum.toFloat()
```
inorichi: I'm not sure if this is better than delegating to the recognizer.

app/src/main/res/values/strings.xml
```diff
@@ -224,6 +224,7 @@
     <string name="select_source">Select a source</string>
     <string name="no_valid_sources">Please enable at least one valid source</string>
     <string name="no_more_results">No more results</string>
+    <string name="local_source">Local mangas</string>
```
inorichi: Without the `s`. In english the plural is also `manga`. I widely use `mangas` in the code to know at first glance that it's a list, but that's not visible to the user 😅

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+import java.util.regex.Pattern
+import java.util.zip.ZipInputStream
+import java.io.*
+
+class LocalSource(private val context: Context) : CatalogueSource {
+    private class OrderBy() : Filter.Sort("Order by", arrayOf("Title", "Date"), Filter.Sort.Selection(1, false))
+
+    private val fileProtocol = "file:/"
+    private val floatPattern = Pattern.compile("[0-9]+[.]?[0-9]*")
+    private fun isImage(name: String) = name.endsWith(".jpg", true) || name.endsWith(".jpeg", true) || name.endsWith(".png", true) || name.endsWith(".gif", true)
```
inorichi: I would use the mimetype. I have a similar approach [here](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt#L404).

Something like

```kt
private fun isImage(file: File): Boolean {
    return context.contentResolver.getType(file)
        ?: file.openInputStream().buffered().use {
            URLConnection.guessContentTypeFromStream(it)
        }
        .orEmpty()
        .startsWith("image/")
}
```

In the unlikely case the file has no extension, it would still be detected.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    override fun fetchPageList(chapter: SChapter): Observable<List<Page>> {
+        val chapFile = File(chapter.url.substring(6))
+        val pages: List<Page>
+        if (chapFile.isDirectory) {
+            pages = chapFile.listFiles()
+                    .filter { !it.isDirectory && isImage(it.name) }
+                    .sortedWith(Comparator<File> { t1, t2 -> CaseInsensitiveSimpleNaturalComparator.getInstance<String>().compare(t1.nameWithoutExtension, t2.nameWithoutExtension) })
+                    .withIndex().mapIndexed { i, v -> Page(i, v.value.absolutePath, v.value.absolutePath, Uri.fromFile(v.value)).apply { status = Page.READY } }
```
inorichi: I think `withIndex()` is redundant here.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+        } else {
+            val destDir = File(context.cacheDir.absolutePath + File.separator + chapFile.path.replace(File.separator, "_"))
```
inorichi: I'll see if I can support this in subsampling soon.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -0,0 +1,138 @@
...
+    override fun fetchPopularManga(page: Int) = fetchSearchManga(page, "", getFilterList())
+
+    override fun fetchSearchManga(page: Int, query: String, filters: FilterList): Observable<MangasPage> {
+        val mangaBaseDir = File(Environment.getExternalStorageDirectory().absolutePath +
+                File.separator + context.getString(R.string.app_name), "mangas")
+        var mangaDirs = mangaBaseDir.listFiles()
+                .filter { it.isDirectory && it.name.contains(query, true) }
+        if (!filters.isEmpty()) {
+            val state = (filters.get(0) as OrderBy).state
+            when (state?.index) {
```
inorichi: I'm starting to think we should reconsider the nullability of the sort state. Would it ever make sense? In the UI you can't unselect them, there's always one active (unless it starts with null, but you can't set it back to null after that without the reset button).

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
paronos: Since `contentResolver.getType` works only with `content://` in the case of zip files this approach will always require to open the stream from the zip file. To me this seems a too heavy (it is really improbable that local image files miss the extension).
Using `MimeTypeMap.getSingleton().getMimeTypeFromExtension` is an alternative, but I do not know if it is really useful when not all the `image/*` formats are supported.

paronos:
I made some minor changes and as suggested rename the manga directory "local".

Regarding `isImage()`, since `contentResolver.getType` works only with `content://` in the case of zip files this approach will always require to open the stream from the zip file. To me this seems a too heavy (it is really improbable that local image files miss the extension).
Using `MimeTypeMap.getSingleton().getMimeTypeFromExtension` is an alternative, but I do not know if it is really useful when not all the `image/*` formats are supported.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
paronos: I need the chapter number now to apply `chapters.sortedByDescending { it.chapter_number }` and `ChapterRecognition.parseChapterNumber(chapter: Chapter, manga: Manga)` cannot be used since it requires `Chapter` and `Manga`.

inorichi:
I've made a few tests subclassing a `ContentProvider` to handle zips. It allows to open the input stream through the content resolver, and `getType` would work too. 

With the content provider, no change would be required to subsampling.

AndroidManifest.xml:
```xml
<provider
    android:name=".ZipContentProvider"
    android:authorities="${applicationId}.zip-provider"
    android:exported="false">
</provider>
```

ZipContentProvider:

```kotlin
class ZipContentProvider : ContentProvider() {

    companion object {
        const val PROVIDER = "${BuildConfig.APPLICATION_ID}.zip-provider"
    }

    override fun onCreate(): Boolean {
        return true
    }

    override fun getType(uri: Uri): String? {
        val ext = MimeTypeMap.getFileExtensionFromUrl(uri.toString())
        return MimeTypeMap.getSingleton().getMimeTypeFromExtension(ext)
    }

    override fun openAssetFile(uri: Uri, mode: String): AssetFileDescriptor? {
        try {
            val url = "jar:file://" + uri.toString().substringAfter("content://$PROVIDER")
            val input = URL(url).openStream()
            val pipe = ParcelFileDescriptor.createPipe()
            thread {
                val output = ParcelFileDescriptor.AutoCloseOutputStream(pipe[1])
                input.use {
                    output.use {
                        input.copyTo(output)
                        output.flush()
                    }
                }
            }
            return AssetFileDescriptor(pipe[0], 0, -1)
        } catch (e: IOException) {
            return null
        }
    }

    // rest of the required methods throw an exception
}
```

The source should build the `Uri` like: `content://$PROVIDER${absPathToZip}!/$zipEntry`

Though I still haven't tested if this behaves properly.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
paronos: I changed `Chapter` to `SChapter` and and `Manga` to `SManga` in `ChapterRecognion` in order to use `parseChapterNumber()`.

paronos:
The catalog now include `ZipContentProvider`.
The images are displayed without problems, but an exception (`java.io.IOException: write failed: EPIPE (Broken pipe)`) is aways thrown by `ZipContentProvider.openAssetFile` on my device (Android 4.1):
```kotlin
                        try {
                            input.copyTo(output)
                        } catch (e: IOException) {
                            e.printStackTrace()
                        }
```
By replacing `input.copyTo(output)` with a classic `write()` cycle  I noticed that the exception is always thrown by the last write (the one that would write the last file chunk). Since the images are displayed without problems probably nothing relevant is lost.

inorichi:
I spent quite a lot of time yesterday trying to figure out why that happened. It happens with Image and Rapid, Skia is always fine.

The reason is actually simple, the reader closes the stream. Image first tries to find the dimensions of the image, when they are retrieved, it closes the streams and returns, then that exception is thrown because the pipe is closed.

In a commit I made yesterday to subsampling, it doesn't happen anymore with Image because I'm using the decoder's `getWidth` and `getHeight` methods (I don't know why I did it by opening two input streams before).

So, that error can be just ignored (still needs to be caught to avoid breaking the app).

Btw, instead of always creating a thread, we should use a thread pool.

```kotlin
class ZipContentProvider : ContentProvider() {

    private val pool by lazy { Executors.newCachedThreadPool() }
}
```

and replace `thread` with `pool.execute`.

Also, catch the entire thing in the thread. `use` can still throw exceptions. And in `query` return a `null` cursor. Subsampling was requesting the exif through a cursor and the exception was crashing the app.

In `PageView` you may need to remove the `if (!file.exists()) { page.status = Page.ERROR }`. I still have to think of a better alternative to load the pages into the reader (`OnlineSourceFetcher` I'm looking at you).

paronos:
I was trying to use `context.contentResolver.getType(uri)` to check if the file is an image, but it returns `null` with some file names (e.g. `10[1].jpg`).
This is caused by `MimeTypeMap.getFileExtensionFromUrl(uri.toString())` in `ZipContentProvider.getType()`. http://stackoverflow.com/questions/14320527/android-should-i-use-mimetypemap-getfileextensionfromurl-bugs/14321470

I was thinking of using `MimeTypeMap.getSingleton().getMimeTypeFromExtension("ext")` directly with `URLConnection.guessContentTypeFromStream()` as fallback.

inorichi:
Alright, you can implement that in `getType`. Remember that the stream must be buffered for `guessContentTypeFromStream` to work.

paronos:
`ZipContentProvider.getType()` now uses `URLConnection.guessContentTypeFromName()`.

I just discovered that the [Android implementation](https://android.googlesource.com/platform/libcore/+/android-6.0.1_r78/luni/src/main/java/java/net/URLConnection.java) of `guessContentTypeFromStream()` does not recognize `jpg` and `png` files (and for `gif` files only checks only the first two byes _("GI")_ instead of the first six _("GIF87a" or "GIF89a")_).
For this reason I just updated the `isImage(name: String)` function by using  `URLConnection.guessContentTypeFromName(name).orEmpty().startsWith("image/")`.

Before this discovery, since `context.contentResolver.getType(uri)` does not work for "file://", I had used:
```kotlin
private fun isImage(file: File): Boolean {
    return if (file.isDirectory)
        false
    else
        (URLConnection.guessContentTypeFromName(file.name) ?:
                FileInputStream(file).buffered().use {
                    URLConnection.guessContentTypeFromStream(it)
                })
                .orEmpty().startsWith("image/")
}
```
For zip files instead, to take advantage of the fact that `ZipEntry` and `ZipFile` are already available, I had used:
```kotlin
private fun isImage(zipEntry: ZipEntry, zipFile: ZipFile): Boolean {
    return if (zipEntry.isDirectory)
        false
    else
        (URLConnection.guessContentTypeFromName(zipEntry.name) ?:
                zipFile.getInputStream(zipEntry).buffered().use {
                    URLConnection.guessContentTypeFromStream(it)
                })
                .orEmpty().startsWith("image/")
}
```

inorichi:
Hm... weird, I remember checking the implementation before including it and I saw png and jpg also recognized. In fact, I just checked again and this is some of the implementation in Nougat:

studio64_2017-01-25_22-33-47

It must have been introduced in Java 8. I'm going to further investigate it and copy the implementation of Nougat in a util class if possible.

paronos:
The [openjdk6 implementation](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/net/URLConnection.java) was already better than the Android one (taken from Apache HarmonI guess). 
With Nougat Google had replaced the old implementation with [one](https://android.googlesource.com/platform/libcore/+/android-7.0.0_r27/ojluni/src/main/java/java/net/URLConnection.java) from openjdk8: This class is under _GPL2+Classpath exception_, so it is not possible to directly copy the implementation (unless you specify that the Apache 2.0 license does not apply to that specific file).

inorichi:
[This](http://256stuff.com/sources/simplemagic/) is a ~140kb dependency. Should we use this or copy the method from openjdk? To be honest I would prefer to not include GPL code, and for such a simple function.

paronos:
This code from [here](https://github.com/appcelerator/titanium_mobile/blob/master/android/titanium/src/java/org/appcelerator/titanium/TiBlob.java) is under Apache 2.0 license. It only recognize gif/png/jpg, are other image types supported by the viewer?
```java
	/**
	 * Check for additional content type reading first few characters from the given input stream.
	 * 
	 * @return the guessed MIME-type or null if the type could not be determined.
	 */
	private String guessAdditionalContentTypeFromStream(InputStream is)
	{
		String mt = null;

		if (is != null) {
			try {

				// Look ahead up to 64 bytes for the longest encoded header
				is.mark(64);
				byte[] bytes = new byte[64];
				int length = is.read(bytes);
				is.reset();
				if (length == -1) {
					return null;
				}
				if (bytes[0] == 'G' && bytes[1] == 'I' && bytes[2] == 'F' && bytes[3] == '8') {
					mt = "image/gif";
				} else if (bytes[0] == (byte) 0x89 && bytes[1] == (byte) 0x50 && bytes[2] == (byte) 0x4E
					&& bytes[3] == (byte) 0x47 && bytes[4] == (byte) 0x0D && bytes[5] == (byte) 0x0A
					&& bytes[6] == (byte) 0x1A && bytes[7] == (byte) 0x0A) {
					mt = "image/png";
				} else if (bytes[0] == (byte) 0xFF && bytes[1] == (byte) 0xD8 && bytes[2] == (byte) 0xFF) {
					if ((bytes[3] == (byte) 0xE0)
						|| (bytes[3] == (byte) 0xE1 && bytes[6] == 'E' && bytes[7] == 'x' && bytes[8] == 'i'
							&& bytes[9] == 'f' && bytes[10] == 0)) {
						mt = "image/jpeg";
					} else if (bytes[3] == (byte) 0xEE) {
						mt = "image/jpg";
					}
				}
			} catch (Exception e) {
				Log.e(TAG, e.getMessage(), e, Log.DEBUG_MODE);
			}
		}
		return mt;
}
```

inorichi:
I think Skia supports bmp and webp, but nobody uses bmp, and I haven't seen many webp yet. I'm happy with supporting the most common formats, and that method is enough IMO 👍

paronos:
I added a `isImage` function that use `URLConnection.guessContentTypeFromName(name)` and fallback to a code similar to the one above (I also added WEBP).
I placed `isImage` is in DiskUtil, that's okay or should be placed in a new/different object?

inorichi:
Yeah, it's fine there, better than adding more files.

I think the "cover.jpg" is not going to exist most of the times. Maybe we should try to copy (or extract if it's a zip) the first image in the case the file is missing. Also allowing to change the cover from within the app would be nice. The implementation is [here](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt#L347), something like:

```kotlin
if (manga.source == 0) {
    val mangaDir = File(...)
    File(mangaDir, "cover.jpg").outputStream().use {
        inputStream.copyTo(it)
    }
}
```

I'm not sure what the id of the source should be btw, `-1`, `0` or continue with `11`. Any specific reason you chose `0`? I'm a bit worried because it's the default id, though nothing bad should happen.

paronos:
`cover.jpg` is now created (if missing) by taking the first image from the first chapter.
The cover can be changed directly from from app (both from the library or from the reader).

There is no particular reason for `id=0`. I did not like to use 11 because it is a "different/unique" source. If clarity is better to avoid `0` i would prefer `-1`.

paronos:
I was thinking of implementing `fetchLatestUpdates` by showing the folder updated in the last _X_ days. Seven days is a good threshold?

inorichi:
I added support for removable SD in 85a75ea. 

Manga stores the relative url instead of the full path, and chapters the full path (but without `file://`, it's not needed).

So, `searchManga` now looks for folders in both the external and removable storages. You can also mix the chapters in the external and removable folders as long as the folder name matches, for example
`/storage/emulated/0/Tachiyomi/local/some manga/ch1` 
and 
`/storage/18F5-2C11/Tachiyomi/local/some manga/ch2`

The `cover.jpg` file is always stored in the external storage (`/emulated/0`) to avoid requiring [SAF](https://developer.android.com/guide/topics/providers/document-provider.html?hl=id) and prompting the user. The folder will be created if needed.

I have to further test my implementation though.

paronos:
I briefly tried your version and did not notice problems. It is a good idea to support removable SD. When source settings will be implemented maybe we could allow user selection of additional directories.

I just added `fetchLatestUpdates` with a 7 days threshold. Since there are two versions of this source, what do you prefer that I do?

inorichi:
My commit is based on the commit before. Since your last commit is small, I think you could add it on top of mine. I can do it too if you want but I would have to create another PR.

paronos:
I merged your changes and did some tests.

I reread this thread and the following things are still pending:
- > In PageView you may need to remove the if (!file.exists()) { page.status = Page.ERROR } **→ It does not seem to give problems**
- remove nullability of the sort state?
- change source ID (to -1?)
- change the cover filename (from `cover.jpg` to something like `cover` or `cover.dat`) since the file is not always a jpeg?
- read the manga metadata from a  `description.property.txt` file (if present) instead of leaving them blank?  **→ I think that it would be hardly used**

inorichi:
* Cool then, I want to remove that anyways with a different approach for loading pages. Also removing the need to use `status = Page.READY` in the source implementation. But it won't be changed soon/in this PR.
* Yes, I think it should have a default value, but you're right, different PR.
* I'll make the default value of `Manga::source` -1 instead, so it's fine.
* IMO it should be `cover.jpg`, even if it's a png/whatever at some point. By using a different extension or none, it could make some file managers not show a preview or not allow to open it in a gallery etc. Those are nice to have. Saving it in the extension of the file is also a possibility, but in my opinion not worth to implement (it would be required to list all files, find something starting with `cover`, update the manga with the new path, etc...)
* I really doubt people would write that file, it's also not worth IMO.

paronos:
Handling different cover file extension it is also in my opinion not worth the hassle.
The metadata file can be also omitted.

inorichi:
I'm also making the chapter url relative. It'll help if for whatever reason the SD path changes, or when changing the device. I've also made a few changes to the glide module. 

I'm merging and then adding my changes. Thanks for your work!

paronos:
Glad I could work on this good software. Moreover, your changes have really improved the source.

Issue #399, opened 2016-07-20
Restore functionnality doesn't update progress on existing manga


jscti:
Hi

If I restore my db from a JSON file but already have the same manga db (but in an older state : a few new mangas and existing ones but with progress updated), the manga progress doesn't get updated.

Use case : I have two devices (a master/tablet and my phone that I use less) and wanted to fully update my phone by restoring a fresh tablet backup ..

Version: v0.2.2-1
Phone: Nexus 4
Android version: 6.0.1

PR #650, opened 2017-01-24
Rewrote Backup

### LINKED ISSUE #613: Automatic/Routine Backup
I usually use a tablet to read, but occasionally I wind up with free time I didn't expect, and just use my phone. The only problem is, I have to manually backup and restore my library if I want to continue from where I was. 

I would like to request, since syncing across devices isn't available, the option to do routine/automatic backups. 

Something along the lines of an automatic backup check box, and a time (or multiple) everyday that the backup would occur. Alternatively/additionally, the backup could be done at Library Update too. Then you would set the preferred app to upload, ie Dropbox, Drive, and the location to keep them. 

This should be a fairly simple hold-over until full sync becomes an option, yea? 

Version: r1002 or v0.4.0

NoodleMage:
# Implementation:
Closes #613 #399 
###
### Backup
Url,Title,Source,Viewer,Chapter_Flags
### Restore
1) Fetch manga from source
2a) Manga in database get id and update with backup.
2b) Manga not in database add manga.
### Backup
Url, Read, Bookmark, Last_read
Only add if value's are changed
### Restore
1) Fetch chapters from source
2) Update known chapters (by url) with backup information.
### Backup
Url(chapter), lastRead
Only add if lastRead != 0
### Restore
1. Retrieve chapter id by URL
2a. Add history
2b. Update last_read by max(backup,library)
### Backup
Title, Sync, Remote, Last_Read
### Restore
1) check if sync_id already known for manga
2) Update remote idea if differenct
3) Update last_chapter read by max(backup,library)
### Backup
Name, Order
### Restore
Adds category if not yet in database.
TODO update order.

# TODO

- [x] See TODO in code ;).
- [x] Make backup automatic intervals longer than the library ones.
- [x] Don't fetch chapters when already in library.
- [x] Create error log for errors during restore.
- [x] Fix Local source not found error 

# Information
Tested my own library (small 20 manga/ 7 unread):
Old backup: 634.67 kb
New backup: 144.65 kb (will be a bit bigger if history is included)

If someone has idea's to safe even more space please let me know ;).

# SCREENS
I tried to make the backup/recovery process as simple as possible without the need of 3th party software. If someone has any suggestion please let me know.

screenshot_1485257618_framed

screenshot_1485257622_framed

screenshot_1485259546_framed

screenshot_1485257641_framed

screenshot_1485257654_framed


wopian:
Can this be a notification too? So I don't have to check back on the app when it's done - a backup of my manga library is **HUGE** and takes ages with the current implementation.

NoodleMage:
I can change it to a notification for the backup.  
Not for the restore though, I tried the restore while being able to access the library. It works but I got some weird glitches on the view so will keep that as a uncancable dialog.

Is it possible for you to send me a JSON file of your backup? Want to test a big library but don't have the dedication to add 200+ manga and mark some of the as read.

wopian:
DM'd you on Slack :octocat:

NoodleMage:
Edit on my earlier comment. I tested @wopian's library and it went from ~5mb (old backup) to ~167kb(new backup) with a creation time of ~1 second so as of now I will keep the backup dialog like it is.

inorichi:
I would prefer a `BackupManager` instance inside a `BackupService` to not tie the implementation to the android framework. I think @null-dev uses the BackupManager and it probably complicates things for him. Would having it in a separate class be an issue?

Not saving the metadata is definitely a better approach. 👍 

Almost 3k additions, this one will be hard. 😅 I hope to start reviewing it after merging local sources.

NoodleMage:
For the backup creation it shouldn't be as this is just normal function calling. As for the restore, this is one big subscription and I rather not mix observables with a separate class. 

Take your time, I'm still finding small mistakes + adding new things as of now . If the local gets merged I will try to implement it into the new backupservice

null-dev:
I'm currently using the `BackupManager` for both backup and restore during syncing only because it is an easy way to transfer manga between JSON and the DB format.

That said, I haven't really had much time to work on the TachiWeb and the sync framework because of a ton of IRL issues (exams :cry:).

It should be OK to remove `BackupManager` completely now as I plan on replacing it with a custom serialization mechanism in the sync framework anyways.

app/src/main/java/eu/kanade/tachiyomi/Constants.kt
```diff
@@ -7,4 +7,19 @@ object Constants {
     const val NOTIFICATION_DOWNLOAD_CHAPTER_ERROR_ID = 4
     const val NOTIFICATION_DOWNLOAD_IMAGE_ID = 5
 
+
+    /*JSON values*/
+    object JSON {
```
inorichi: If these are used only inside the backup package, put them there. (I haven't checked it, just reviewing from github)

app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/CategoryTypeAdapter.kt
```diff
@@ -0,0 +1,30 @@
+package eu.kanade.tachiyomi.data.backup.serializer
+
+import com.github.salomonbrys.kotson.typeAdapter
+import com.google.gson.TypeAdapter
+import eu.kanade.tachiyomi.data.database.models.CategoryImpl
+
+class CategoryTypeAdapter {
```
inorichi: This should be an `object`. Same for the others

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceMap.kt
```diff
@@ -0,0 +1,179 @@
...
+import android.content.Context
+import android.net.Uri
+import android.os.Environment
+import android.preference.PreferenceManager
+import com.f2prateek.rx.preferences.RxSharedPreferences
+import eu.kanade.tachiyomi.R
+import java.io.File
+import java.util.*
+
+class PreferenceMap(val context: Context) {
```
inorichi: What is this used/needed for?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedFragment.kt
```diff
@@ -108,6 +108,7 @@ class SettingsAdvancedFragment : SettingsFragment() {
                 .onPositive { dialog, which ->
                     (activity as SettingsActivity).parentFlags = SettingsActivity.FLAG_DATABASE_CLEARED
                     db.deleteMangasNotInLibrary().executeAsBlocking()
+                    db.deleteHistoryNoLastRead().executeAsBlocking()
```
inorichi: Is this really needed? Chapters with a foreign key on the manga are deleted, and so should history entries of those chapters

app/src/main/java/eu/kanade/tachiyomi/util/ContextExtensions.kt
```diff
@@ -122,7 +122,7 @@ fun Context.sendLocalBroadcastSync(intent: Intent) {
  *
  * @param receiver receiver that gets registered.
  */
-fun Context.registerLocalReceiver(receiver: BroadcastReceiver, filter: IntentFilter ){
+fun Context.registerLocalReceiver(receiver: BroadcastReceiver, filter: IntentFilter){
```
inorichi: While you are at it, add a space after `)` 😄

app/src/main/java/eu/kanade/tachiyomi/util/FileExtensions.kt
```diff
@@ -23,11 +25,49 @@ fun File.getUriCompat(context: Context): Uri {
...
 }
+
+/**
+ * Deletes file if exists
+ *
+ * @return success of file deletion
+ * @throws [IOException]
+ */
+@Throws(IOException::class)
+fun UniFile?.deleteIfExists(): Boolean {
```
inorichi: I have to check the implementation, but this should be the same as calling `?.delete()`

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceMap.kt
NoodleMage: It was used to get the default preference values. I have removed preference from new commit because I agree with:

> I'm not sure if preferences should be included, they aren't that many and I often reset them by changing the key when there's some incompatibility.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedFragment.kt
NoodleMage: I added it because the last read data of manga still in the library isn't removed. I added it partly  because of testing the backup function so it can be removed before merge.

NoodleMage:
Sorry for inactivity last few days. Started new job so not a lot of time ;)

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupCreateService.kt
```diff
@@ -0,0 +1,182 @@
...
+        // Create category array
+        val settingsEntries = JsonArray()
+
+        // Add value's to root
+        root.add(INFORMATION, information)
+        root.add(MANGAS, mangaEntries)
+        root.add(CATEGORIES, categoryEntries)
+
+        // Get manga from database
+        val mangas = backupManager.getFavoriteManga()
```
inorichi: Do these database queries (until line 129 if I'm right) inside a transaction so that another thread can't update the db while creating a backup. It's very unlikely but still possible.

inorichi:
Btw, everything else is looking fine. At most only a few changes should be required, but that's for the next step, when I actually try the implementation.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupFragment.kt
```diff
@@ -59,9 +59,9 @@ class SettingsBackupFragment : SettingsFragment() {
         internal const val EXTRA_ERROR_FILE = "$ID.$INTENT_FILTER.EXTRA_ERROR_FILE"
 
 
-        private const val BACKUP_CREATE = 101
-        private const val BACKUP_RESTORE = 102
-        private const val BACKUP_DIR = 103
+        private const val BACKUP_CREATE = 201
+        private const val BACKUP_RESTORE = 202
+        private const val BACKUP_DIR = 203
```
inorichi: Some of these are also declared in the downloads fragment. It's not a problem but I prefer to give them a unique id.

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt
```diff
@@ -50,28 +45,13 @@ class BackupManager(val context: Context) {
...
      * Version of parser
      */
-    private var version = 1
+    var version: Int = version
+        private set
 
     /**
      * Json Parser
      */
-    var parser: Gson = GsonBuilder().create()
```
inorichi: This was creating a new parser every time `parser` was called.

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt
```diff
@@ -119,14 +103,17 @@ class BackupManager(val context: Context) {
...
+        entry[MANGA] = parser.toJsonTree(manga)
 
         // Check if user wants chapter information in backup
         if (options and BACKUP_CHAPTER_MASK == BACKUP_CHAPTER) {
             // Backup all the chapters
             val chapters = databaseHelper.getChapters(manga).executeAsBlocking()
             if (!chapters.isEmpty()) {
-                entry.add(CHAPTERS, parser.typedToJsonTree(chapters))
+                val chaptersJson = parser.toJsonTree(chapters)
+                if (chaptersJson.asJsonArray.size() > 0) {
```
inorichi: This avoids creating the key if no chapter is actually serialized (there are conditions in the type adapter for not creating chapters if they aren't read or bookmarked)

app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/ChapterTypeAdapter.kt
```diff
@@ -24,11 +24,11 @@ object ChapterTypeAdapter {
                     value(it.url)
                     if (it.read) {
                         name(READ)
-                        value(it.read)
+                        value(1)
```
inorichi: Not really needed but `1` is shorter than `true` or `false` so it saves a few bytes

app/src/main/java/eu/kanade/tachiyomi/util/FileExtensions.kt
```diff
@@ -20,10 +19,3 @@ fun File.getUriCompat(context: Context): Uri {
     return uri
 }
 
-/**
- * Write line with \n
- */
-fun BufferedWriter.writeLn(line: String) {
```
inorichi: Removed because it's not really a file extension, and adding `\n` isn't too bothersome.

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/Backup.kt
```diff
@@ -1,15 +1,23 @@
...
-    const val VERSION = "version"
+package eu.kanade.tachiyomi.data.backup.models
+
+import java.text.SimpleDateFormat
+import java.util.*
+
+/**
+ * Json values
+ */
+object Backup {
```
inorichi: JSON wasn't too descriptive, I think `Backup` fits better here

PR #701, opened 2017-03-09
Category-specific auto download

### LINKED ISSUE #700: Category-specific auto download
An option for auto downloading new chapters from user defined categories would be very useful. I have created 3 categories : Reading, Plan to read, and Read. I want Tachiyomi to auto download new chapters of the manga only in the "Reading" category.

Version: v0.5.0

NoodleMage:
Closes #700 

Added option to select which categories use auto download function.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -288,13 +288,13 @@ class LibraryUpdateService : Service() {
                     if (newUpdates.isEmpty()) {
                         cancelNotification()
                     } else {
-     			showResultNotification(newUpdates, failedUpdates)
                         if (preferences.downloadNew().getOrDefault()) {
                             if (hasDownloads) {
                                 DownloadService.start(this)
                                 hasDownloads = false
                             }
                         }
+                        showResultNotification(newUpdates, failedUpdates)
                     }
                 }
     }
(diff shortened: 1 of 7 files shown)


sbstratos79:
Thanks! When will the changes be live in stable release?

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -254,10 +262,13 @@ class LibraryUpdateService : Service() {
                                 Pair(emptyList<Chapter>(), emptyList<Chapter>())
                             }
                             // Filter out mangas without new chapters (or failed).
-                            .filter { pair -> pair.first.size > 0 }
+                            .filter { pair -> pair.first.isNotEmpty() }
                             .doOnNext {
-                                if (preferences.downloadNew()) {
-                                    downloadChapters(manga, it.first)
+                                if (preferences.downloadNew().getOrDefault()) {
```
inorichi: Cache this value like you are doing for `categoriesToDownload`.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -57,9 +57,16 @@ class LibraryUpdateService : Service() {
...
 
+    /**
+     * Download Manager
+     */
     val downloadManager: DownloadManager by injectLazy()
 
     /**
+     * Boolean to determine if [DownloadManager] has downloads
+     */
+    private var hasDownloads = false
```
inorichi: This can be a local variable, it's only used by 1 method.

PR #711, opened 2017-03-15
Prevent some manga breaking the download notifier

### LINKED ISSUE #709: Downloader: Incorrectly nested parentheses in regexp patt...
Version 0.5.0

That's what the message says since I can't expand it from the notification.

This happened when it discovered a new chapter(ch.92) for A-Channel (batoto source) 
It's listed as  "A-Channel (A Channel)

Chapter has downloaded fine and is perfectly readable.

null-dev:
Escape the manga title before using it in the regex statements in the download notifier.

Fixes #709


null-dev:
Actually, yeah, it probably does, I don't see regexes used anywhere else! I've also just tried to  download '+ Tic Neesan' from Mangafox using a build with this fix and it downloads properly.

PR #724, opened 2017-03-27
Add feature mentioned in issue #625 in inorichi/tachiyomi

lifeweaver:
#625 

Please take a look and let me know if anything needs changed.

PR #737, opened 2017-04-05
Add manga straight into a category from catalogues

lifeweaver:
Closed the pull request #724, and opened this one since I had opened the old one on my master

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
@@ -109,4 +109,6 @@ class PreferenceKeys(context: Context) {
 
     val lang = context.getString(R.string.pref_language_key)
 
+    val defaultLongpressCategory = context.getString(R.string.pref_default_longpress_category_key)
+
 }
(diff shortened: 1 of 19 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/CatalogueFragment.kt
```diff
@@ -530,23 +539,62 @@ open class CatalogueFragment : BaseRxFragment<CataloguePresenter>(),
...
-                    when (which) {
-                        0 -> {
-                            presenter.changeMangaFavorite(manga)
-                            adapter.notifyItemChanged(position)
+        val defaultLongPressCategory = categories.find { it.id == preferences.defaultLongpressCategory()}
+        if(defaultLongPressCategory != null) {
+            if(!manga.favorite) {
+                presenter.changeMangaFavorite(manga)
+            }
+            presenter.moveMangaToCategory(defaultLongPressCategory, manga)
```
NoodleMage: Show something like a toast. As of now it's not clear that the manga is added to library

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
```diff
@@ -109,4 +109,6 @@ class PreferenceKeys(context: Context) {
 
     val lang = context.getString(R.string.pref_language_key)
 
+    val defaultLongPressCategory = context.getString(R.string.default_long_press_category_key)
```
NoodleMage: I know you just refactored them all, so sorry ;(. But please rename this (and all occurrences) to `defaultCategory `because it's no longer long press.

Issue #453, opened 2016-09-06
Webtoon mode doesn't retain last read page


BlackSynder:
Webtoon does note what number it is but when I click a chapter I started to read, it just goes back to the start. Is this a known issue? Doesn't look too bad to fix but I can understand why it isn't if it's known.

wopian:
#71. Adding this functionality [requires](https://github.com/inorichi/tachiyomi/issues/84#issuecomment-176204009) making a custom view for the webtoon viewer so will likely not be fixed in a long time.

inorichi:
I tried in the past to restore it with no success. There were some bugs with the scroll in the RecyclerView (and with a ListView was even worse) and I eventually gave up.

PR #738, opened 2017-04-06
Retain last read page when using the webtoon mode

lifeweaver:
See issue #453.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonReader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonReader.kt
@@ -7,7 +7,9 @@ import android.view.ViewGroup.LayoutParams.MATCH_PARENT
 import android.view.ViewGroup.LayoutParams.WRAP_CONTENT
 import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.ui.reader.ReaderChapter
+import eu.kanade.tachiyomi.ui.reader.ReaderEvent
 import eu.kanade.tachiyomi.ui.reader.viewer.base.BaseReader
+import eu.kanade.tachiyomi.util.SharedData
 import eu.kanade.tachiyomi.widget.PreCachingLayoutManager
 import rx.subscriptions.CompositeSubscription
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonReader.kt
```diff
@@ -107,6 +108,28 @@ class WebtoonReader : BaseReader() {
         return recycler
     }
 
+
+    /**
+     * Uses two ways to scroll to the last page read.
+     */
+    private fun scrollToLastPageRead() {
+        val last_page_read = SharedData.get(ReaderEvent::class.java)?.chapter?.last_page_read?.minus(1) ?: 0
```
inorichi: I'd prefer not to read from this "hack" I came up with. You can add an argument to this function and use that index.

In `onChapterSet` you get the index you have to show, but it can be updated when `onPageChanged` is called. If needed, you can save the page as an additional variable.

Btw, my previous attempts made the images jump when the first gesture is to go upwards. Does it also happen in this PR?

lifeweaver: Thanks for taking a look!

> Btw, my previous attempts made the images jump when the first gesture is to go upwards. Does it also happen in this PR?

I notice the jump on upwards scrolling in both the current 0.5.1 version and this PR, I'll see if I can find the cause and fix it when I incorporate your comment about not using the SharedData.

lifeweaver: With some testing I think I found that it's the hiding of the progress view and the loading of the previous view that's causing the jumps.

To test this I:
1. Put `recycler.setItemViewCacheSize(20)` in the WebtoonReader
2. debugged the app and viewed ~20 pages normally
3. When I scrolled up I didn't get the jumping until I run out of cached pages

Of course this doesn't solve the case where you want more than 20 or if you had just opened it.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonReader.kt
```diff
@@ -146,6 +163,7 @@ class WebtoonReader : BaseReader() {
      */
     override fun onChapterSet(chapter: ReaderChapter, currentPage: Page) {
         this.currentPage = currentPage.index
+        scrollToLastPageRead(this.currentPage)
```
inorichi: `recycler` might be null at this point, that's why there's a `if (view !=null)` below. Call this method there.

PR #761, opened 2017-04-18
Allow multiple sources in each extension source apk

null-dev:
Add the ability to specify multiple source classes in a single extension apk.

Source classes should be separated by `;`. Spaces are permitted between source classes. Old sources are backwards compatible.

SourceFactories can also be defined after this patch. SourceFactories allow extensions to specify a class that can create multiple sources when loaded. (Requires PR: https://github.com/inorichi/tachiyomi-extensions-lib/pull/1)

**Not sure if we should bump the version number of the source library? I considered this because new extension sources that contain multiple source classes will not be compatible with old versions of Tachiyomi.**

Also did a bit of code cleanup and increased verbosity of error messages

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
@@ -126,8 +126,13 @@ open class SourceManager(private val context: Context) {
         }
 
         val classLoader = PathClassLoader(ext.appInfo.sourceDir, null, context.classLoader)
-        return ext.sourceClasses.map {
-            Class.forName(it, false, classLoader).newInstance() as Source
+        return ext.sourceClasses.flatMap {
+            val obj = Class.forName(it, false, classLoader).newInstance()
+            when(obj) {
+                is Source -> listOf(obj)
+                is SourceFactory -> obj.createSources()
+                else -> throw Exception("Unknown source class type!")
+            }
         }
     }
 
@@ -140,7 +145,7 @@ open class SourceManager(private val context: Context) {
         const val EXTENSION_FEATURE = "tachiyomi.extension"
         const val METADATA_SOURCE_CLASS = "tachiyomi.extension.class"
         const val LIB_VERSION_MIN = 1
-        const val LIB_VERSION_MAX = 1
+        const val LIB_VERSION_MAX = 2
     }
 
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
@@ -74,7 +74,7 @@ open class SourceManager(private val context: Context) {
                     val map = file.inputStream().use { yaml.loadAs(it, Map::class.java) }
                     sources.add(YamlHttpSource(map))
                 } catch (e: Exception) {
-                    Timber.e("Error loading source from file. Bad format?")
+                    Timber.e("Error loading source from file. Bad format?", e)
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


paronos:
I was also thinking that it might be useful to allow more source per .apk. If this change has to be done, I think it is best to allow the definition of the sources at runtime.
In this way, for example, a `PrivateServerSources.apk` could define a source for each private server defined in a configuration file. In a similar way a `YAMLSources.apk` could define a source for each .yml file.

paronos:
Also, if the version number of the source library has to be changed it is probably better to make other breaking changes at the same time.

One thing is the addition of a `SourceWithSettings` interface. I have not investigated the problem, but perhaps the source could define the configuration screen directly as an Activity or Fragment (so there would be few changes to make in Tachiyomi).

Something that also had been thought is to remove the nullability for `state` of `Filter.Sort`.

null-dev:
@paranos I would love a "SourceWithSettings" interface but it is a major task and I'm currently not sure how to implement it. It would also require implementing some sort of persistent storage for sources and I have no idea how we could implement a preference UI builder (what if a source wanted custom preferences?).

I would also like to avoid having extensions depend on too many Android components and Android versions (it would be quite a mess if we have to include the support library in extensions).

Filter.Sort's nullability looks like it is intended. It has three states:
- no sort (null)
- ascending (ascending: true)
- descending (descending: true)

**I will put off implementing these suggestions for now.** I will however, look into implementing runtime source definitions.

paronos:
Of course `SourceWithSettings` is out of the scope of this PR. But before increasing the version number of the source library we can wait for its addition (if its implementation is expected).

inorichi:
Including this patch doesn't break compatibility with the currently installed sources, so I don't think we should make it look like an incompatible change. Also no change is made at all in the lib by including this patch.

A multi-apk source installed in an older version should (not tested) just silently fail.

About preferences, I was thinking of providing each source a unique `SharedPreferences` where they can do whatever they want.

For building preferences, I can think of two options:

1) Each apk declares its own activity in the manifest that we can launch. I'm not sure if we can use that, given that the `Context` is different and we may not be able to share the preferences file (making it public is meh).

2) Each source implementing that interface has a method that returns a `PreferenceScreen` we can inflate in tachiyomi. This should be easier to implement with simple preferences but less configurable (we will probably have to provide a generic login preference)

I have plans to create a simple DSL like Anko but only for preferences. It could be nice if we implement the option 2.

app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
```diff
@@ -140,7 +145,7 @@ open class SourceManager(private val context: Context) {
         const val EXTENSION_FEATURE = "tachiyomi.extension"
         const val METADATA_SOURCE_CLASS = "tachiyomi.extension.class"
         const val LIB_VERSION_MIN = 1
-        const val LIB_VERSION_MAX = 1
+        const val LIB_VERSION_MAX = 2
```
inorichi: We'll stay in major version 1.

Also, is SourceFactory really needed? Mind explaining where this would help?

null-dev: @paronos recommended I add something to allow for source definition at runtime.

Presumably, it could be used to implement other source formats outside of YAML and Java sources and have the loader for these formats installable as a regular extension.

It feels quite limited to me currently and I think it may be more useful if `Context` was passed to `createSources` (so sources could at least access the app's external data directory).

It may also cause the app to become too modular however, I don't think having 20+ APKs installed for Tachiyomi sounds like a really great idea.

Mistero5050: Yeah, 20 apk will be too much

Issue #686, opened 2017-02-16
Deleting categories hides manga until switch from and to library


j2ghz:

jshap70:
Hi, can you checkout [this guide on submitting issues](https://github.com/inorichi/tachiyomi/blob/master/.github/CONTRIBUTING.md) and provide a bit more information? I'm not really sure what "switch from and to library" means

j2ghz:
The one I wrote :D I sent the issue from gitskarios, for some reason the description vanished.

1. Have a library with some manga in a category, some in Default
2. Delete the category
3. There will be only manga from default category, the ones that were in the deleted one are not there
4. Switch to catalogue, latest updates,...
5. Switch back to library
6. All manga will be there

Version: r1105

PR #769, opened 2017-04-25
Deleting categories hides manga until switch from and to library #686

CarlosEsco:
This fixes bug #686 

forces a resubscribe of the library if a category is deleted.  This allows for a full refresh


inorichi:
Hm... I think the database should notify category updates to the library query instead.

CarlosEsco:
I will continue researching, there doesn't seem to be a simple solution with the way everything is setup currently.

inorichi:
Adding the category table [here](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/database/queries/MangaQueries.kt#L28) should be enough, probably.

CarlosEsco:
yea that did it.  I think i looked and tried everything except add the table.  Makes sense now that you pointed it out, was trying to figure out how it was updating, since it was only updating library on switch to library and when adding a new category with MangaCategory.  I've fixed and pushed the new build for the pull request

inorichi:
Cool. Please cleanup the new lines in the PR and keep only the change to the db.

CarlosEsco:
done thanks for the help

Issue #773, opened 2017-04-27
Add images and 'view chapters' to library updates


Akeboshiwind:
### Problem 1:
I don't tend to remember what I'm reading by name, so I find it hard to work out what's been updated on the Library Updates page.

### Suggested Fix:
Add an image to the left of the title. This also makes the page more consistent with the look of the Library page (you may also want to add dividers to the Library page, but that's your call).

A problem with this could be the same manga being updated multiple times and the image appearing many times, but I don't think this looks too bad.
<img src="https://cloud.githubusercontent.com/assets/8889986/25480059/034d5a20-2b3f-11e7-92a2-05324d775f62.png" width="300">
If you don't like that, you could do what messaging clients do, which is have the image appear only on the highest in a group of messages.

### Problem 2:
Sometimes when I see an update for a manga I'm not reading anymore, I decide to pick it up again. If I'm not completely caught up then I don't want to open the page that has just been opened, so I have to go to the Library page and then find the manga in the list.

### Suggested Fix:
To fix this, you could add another option to the ellipsis menu which opens that manga's chapter list in the Library view.
<img src="https://cloud.githubusercontent.com/assets/8889986/25480471/c6d778e4-2b40-11e7-8d8e-8b3d55377652.png" width="300">

PR #785, opened 2017-05-06
Add images and 'view chapters' to library updates

CarlosEsco:
added option to open manga directly from library update
added covers for manga in library update
added ability to click covers to open manga directly from library update

Addresses #773
image

image

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/recent_updates/RecentChaptersController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/recent_updates/RecentChaptersController.kt
@@ -6,6 +6,8 @@ import android.support.v7.view.ActionMode
 import android.support.v7.widget.DividerItemDecoration
 import android.support.v7.widget.LinearLayoutManager
 import android.view.*
+import com.bluelinelabs.conductor.RouterTransaction
+import com.bluelinelabs.conductor.changehandler.FadeChangeHandler
 import com.jakewharton.rxbinding.support.v4.widget.refreshes
 import com.jakewharton.rxbinding.support.v7.widget.scrollStateChanges
 import eu.davidea.flexibleadapter.FlexibleAdapter
@@ -15,6 +17,7 @@ import eu.kanade.tachiyomi.data.download.model.Download
 import eu.kanade.tachiyomi.data.library.LibraryUpdateService
 import eu.kanade.tachiyomi.ui.base.controller.NoToolbarElevationController
 import eu.kanade.tachiyomi.ui.base.controller.NucleusController
+import eu.kanade.tachiyomi.ui.manga.MangaController
 import eu.kanade.tachiyomi.ui.reader.ReaderActivity
 import eu.kanade.tachiyomi.util.toast
 import kotlinx.android.synthetic.main.fragment_recent_chapters.view.*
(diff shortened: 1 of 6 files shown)


inorichi:
Please follow the material design specs for lists, with a circle icon.

In https://material.io/guidelines/components/lists.html#lists-specs

Section `Two-line list`, Subsection `Avatar with text`

app/src/main/res/values/strings.xml
```diff
@@ -38,6 +38,7 @@
     <string name="action_mark_as_unread">Mark as unread</string>
     <string name="action_mark_previous_as_read">Mark previous as read</string>
     <string name="action_download">Download</string>
+    <string name="action_open_manga">Open Manga</string>
```
inorichi: I'm not sure if it would be better to change this to "Open". Not everything is a manga. Also, there are now two ways to open the manga, by clicking in the cover and the overflow.

I think keeping only the one in the cover should be enough. Thoughts? cc @NoodleMage

NoodleMage: Personally I think this line should be removed. As of now you open the "manga" pressing the cover in your library, or pressing the cover in recently read. It would be weird to suddenly give an option in the drop-down.

CarlosEsco:
got it will adjust image and remove option from three dot

CarlosEsco:
@inorichi @NoodleMage 
Is there a preference on which way to create the circle image?  You have glide transformations already 
`Glide.with(itemView.context)
                    .load(item.manga)
                    .diskCacheStrategy(DiskCacheStrategy.RESULT)
                    .bitmapTransform(CropCircleTransformation(view.context))
                    .into(itemView.manga_cover)`

but you also have the circleview library so just change the imageview to de.hdodenhof.circleimageview.CircleImageView.


but you will need to remove the glide animations according to their readme

If you use an image loading library like Picasso or Glide, you need to disable their fade animations to avoid messed up images. For Picasso use the noFade() option, for Glide use dontAnimate(). If you want to keep the fadeIn animation, you have to fetch the image into a Target and apply a custom animation yourself when receiving the Bitmap.

inorichi:
If the transformation works better and we can remove the circle image view dependency with that change, go for it.

CarlosEsco:
Updated
image
image

inorichi:
Looking better. The height of the cover should be 56dp though

CarlosEsco:
image

app/src/main/java/eu/kanade/tachiyomi/ui/recent_updates/RecentChaptersController.kt
```diff
@@ -247,6 +251,18 @@ class RecentChaptersController : NucleusController<RecentChaptersPresenter>(),
         presenter.deleteChapters(listOf(chapter))
     }
 
+    override fun onCoverClick(position: Int) {
+        val chapterClicked = (adapter?.getItem(position) as RecentChapterItem)
```
inorichi: `val chapterClicked = (adapter?.getItem(position) as? RecentChapterItem) ?: return`

Adapter can be null and the app would crash.

inorichi:
See `item_catalogue_list.xml`. I use 56dp but I add a padding so that the clickable area is bigger. With the show layout bounds setting enabled (in developer options) you can see the area:

qemu-system-x86_64_2017-05-07_15-54-53

CarlosEsco:
So this is best i am able to get it.  its matching your edit_category just doesnt have the extra clickable space.  Not sure what is missing with it.
image

One thing i noticed doing this is do you have any issues with the theme with android studios?  I just get rendering issues.  Switching themes to a standard works for preview but even then using the dimens from material-design-dimens like you are using doesnt work.  Maybe it should be moved to the internal dimens.

inorichi:
It's fine like that. I'll see if I can flatten the layout with ConstraintLayout or replace RelativeLayout with FrameLayout.

As for the preview... I don't really use it that much. Only when editing ConstraintLayout. I just run the app, which usually takes around ~6-7 seconds. There was a time a single change took 30 seconds compiling.

Thanks for your work!

inorichi:
Updated with constraint layout: 29fd574 

The item height was bigger than the specified in the spec, that's why the image had those gaps.

Issue #107, opened 2016-02-03
Show scanlation group name alongside chapter when using batoto source.


beepFTW:
This is more of a problem of seeing copies of the same chapter and being unable to differentiate between them. Different groups often use different lingo and it ends up confusing.

You can see this easily when viewing Platinum End on batoto.

wopian:
If this becomes a thing, can it be made optional in the settings?

Bloodwork:
Not sure why would it need to be optional but why not. It's a huge help for series with multiple groups scanlating them. Like Mahoutsukai no Yome's first 10 chapters: http://bato.to/comic/_/comics/mahoutsukai-no-yome-r10801

Moravuscz:
It would also be wonderful if we could filter chapters by scanlation group.
(Not sure if batoto allows empty group, but if so, using uploader's username instead could work)

PR #789, opened 2017-05-08
Add Batoto Scanlator to Chapter view

CarlosEsco:
#107 

Added Scanlator to Chapter View.  

image


j2ghz:
It seems the scanslator is close to date, and too far away from the title. Confirmed by measuring: http://i.imgur.com/pRPzji4.png
Is there a reason for it not to be in the middle?

CarlosEsco:
I'll mess with it some today the layout needs to be completly rewritten to constraint as its overly complex

CarlosEsco:
Here is updated
image

CarlosEsco:
Fixed a couple bugs and the layout
image
image

inorichi:
The field should be nullable in the database and in our models. This will take a while since we need to make sure backups still work and also release a new version of the extensions library.

CarlosEsco:
switched to allow null for scanlator in database and model

CarlosEsco:
What do you need me to do to get this merged?  Unit test backup, and a pull request for the ext?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChapterHolder.kt
```diff
@@ -1,106 +1,118 @@
...

+class ChapterHolder(

+        private val view: View,

+        private val adapter: ChaptersAdapter

+) : FlexibleViewHolder(view, adapter) {

+
```
j2ghz: Why does the whole file have added ^M at the end of each line?

wopian: They're carriage returns

CarlosEsco: Weird some reason on my merge it added new lines between all the lines.  Fixed this and  fixed the build issue though

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChapterHolder.kt
```diff
@@ -1,106 +1 @@
...
        popup.setOnMenuItemClickListener { menuItem ->
            adapter.menuItemListener.onMenuItemClick(adapterPosition, menuItem)
            true
        }

        // Finally show the PopupMenu
        popup.show()
    }

}
```
j2ghz: It doesn't look any better now. Maybe this will help: https://help.github.com/articles/dealing-with-line-endings/

PR #811, opened 2017-05-25
allow sorting by total chapters for library view

### LINKED ISSUE #796: [Feature request] Additional sort option
Sort by total chapters in my library - this would be helpful in planning to read completed manga. For example, I would first want to read a manga that is the shortest among my list, then the next, and so on. For short lists, the text (unread chapters?) above the manga is helpful, but it won't be if I happen to have a very long list in the future.

Version: v0.5.2
df99a8e4-3878-11e7-860c-d79b44432382

CarlosEsco:
Closes #796

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
@@ -105,13 +105,15 @@ class LibraryNavigationView @JvmOverloads constructor(context: Context, attrs: A
 
         private val alphabetically = Item.MultiSort(R.string.action_sort_alpha, this)
 
+        private val total = Item.MultiSort(R.string.action_sort_total, this)
+
         private val lastRead = Item.MultiSort(R.string.action_sort_last_read, this)
 
         private val lastUpdated = Item.MultiSort(R.string.action_sort_last_updated, this)
 
         private val unread = Item.MultiSort(R.string.action_filter_unread, this)
 
-        override val items = listOf(alphabetically, lastRead, lastUpdated, unread)
+        override val items = listOf(alphabetically, total, lastRead, lastUpdated, unread)
 
         override val header = Item.Header(R.string.action_sort)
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
```diff
@@ -105,13 +105,15 @@ class LibraryNavigationView @JvmOverloads constructor(context: Context, attrs: A
...
+        private val total = Item.MultiSort(R.string.action_sort_total, this)
+
         private val lastRead = Item.MultiSort(R.string.action_sort_last_read, this)
 
         private val lastUpdated = Item.MultiSort(R.string.action_sort_last_updated, this)
 
         private val unread = Item.MultiSort(R.string.action_filter_unread, this)
 
-        override val items = listOf(alphabetically, lastRead, lastUpdated, unread)
+        override val items = listOf(alphabetically, total, lastRead, lastUpdated, unread)
```
NoodleMage: I think it would be better to do this in the order: alphabetically, lastRead, lastUpdated, **total**, unread.
Two reasons:
1. It's alphabetically
2. It's sorted string based > time based > count based.

Opinions?

CarlosEsco: Cant argue with number 1.
for number 2 i was thinking
alpha and total are both orders unrelated to the user
last read, updated, unread are all based on the users 

If anything along those lines it should be
alpha (name sort)
total (chapter sort)
last updated (catalog sort)
last read (user sort)
unread (user sort)

But if @inorichi  agrees i'll modify it so total is at end.

CarlosEsco: Actually I agree now.  Make everything alphabetically sorted then there won't be some random order contributors have to follow if more sorts are added in future.

inorichi:
Use an approach similar to the last read manga sorting. Return an ordered list from the database which is retrieved lazily. Better than doing a lot of single queries to db.

CarlosEsco:
Fixed to be similar to last read sort

inorichi:
Looks good. Thanks!

Issue #779, opened 2017-05-02
[Feature Request] Save Search last between 


cd0rw0:
We have to key-in same title again if we have to search between library and catalog, 
even inside catalog, we have to rekey-in name we change source.

It will easier if, last 5/6 search listed once we click search. (similar playstore)

Regards

Version: r1149

ptoump:
@arkon Although #2885 is a duplicate from a technical standpoint (save search parameters) I was talking about a manual save process which would result in a saved Searches menu, where as this one is for a history of Searches.
Just wanted to document here the different end-user functionality proposed in the now closed #2885 .

Thanks

somebody1234:
👀

buthub:
Previous mentions of this,
https://github.com/tachiyomiorg/tachiyomi/pull/3827
https://github.com/tachiyomiorg/tachiyomi/issues/3062
https://github.com/tachiyomiorg/tachiyomi/issues/2885
have been closed.

Can't we just have an option in settings for Search History, and people can choose whether they want it or not? Retaining a search history is so basic and saves us having to retype those REALLY difficult to remember foreign titles, so I don't understand why tachi doesn't have it too. :-?  Thanks, so much for all the awesome work you guys do on this app! Love you!

PR #849, opened 2017-06-15
Global Search

### LINKED ISSUE #45: Global search (from all sources)
like MW and MR has, search in all available sources

NoodleMage:
Initial code for global search. Note that this is just a humble beginning on which we can build. Closes #45 

# Screens
screenshot_20170724-151421_framed
screenshot_1497560352_framed
screenshot_1497560358_framed
screenshot_20170724-151457_framed


inorichi:
UI looks 👌 .

A few questions/notes:

* The catalogues at the top (Mangahere) are favorite catalogues right? Those shouldn't be repeated in the other list.
* Source title in the global search looks very big. Google play's headers are smaller.
* Source languages are currently lost. Are there plans to support them?
* Round icon with the letter is better than nothing, but I think we should use the source's icon. Sadly it may not be easy to find big enough icons for every source. The website favicon would be the best one to use, but they usually don't have a big resolution variant.

NoodleMage:
> The catalogues at the top (Mangahere) are favorite catalogues right? Those shouldn't be repeated in the other list.

It's the most recent. It changes when a different on is selected. I have implemented some code for favorites (which will be visible above sources and not visible in the sources) but I haven't yet fully implemented this.

> Source title in the global search looks very big. Google play's headers are smaller.

Agree. Will fix

> Source languages are currently lost. Are there plans to support them?

Good point. Will look into it

> Round icon with the letter is better than nothing, but I think we should use the source's icon. Sadly it may not be easy to find big enough icons for every source. The website favicon would be the best one to use, but they usually don't have a big resolution variant.

Will check how favicon looks. Otherwise we can look into using the favicon to determine color of circle and do something with that.

NoodleMage:
> Source languages are currently lost. Are there plans to support them?

Perhaps this works? (I know the flags images are ugly, will find a better library or create my own ;-))

screenshot_1497638492_framed

CarlosEsco:
this might work.

https://github.com/mukeshsolanki/country-picker-android

j2ghz:
@NoodleMage is it possible to instead of doing recent and sources to split it to recent, eng, ru, de, ...

NoodleMage:
@j2ghz. I suppose it's possible. It will take some time though. And things like sort by title (ignoring language) won't be possible

CarlosEsco:
@inorichi  are you gonna do a new release before this gets merged in?

inorichi:
Yes, probably by the end of this month. I won't be adding features until then, just bugfixes.

I also prefer to categorize by language that screen. Also it would be nice to allow to disable/enable catalogues from here. Probably the easiest way is to add an toolbar button that pushes a `SettingsSourcesController`.

Also, would it make sense to merge latest updates here? Maybe we can replace the "open" button with a "latest" and "browse" (it may not look nice in small screens though).

NoodleMage:
> Also, would it make sense to merge latest updates here? Maybe we can replace the "open" button with a "latest" and "browse" (it may not look nice in small screens though).

It can make sense. 
So default -> browse / latest
Login -> Login
Update available (with extensions) -> update
It looks like a bit much though.

Btw I did already implement different cards. It's now a separate recycler thus cards are added dynamically
screenshot_1497717105_framed

inorichi:
IMO the header looks better outside the card (like before)

inorichi:
Also from 2 letter to localized language.
```
val locale = LocaleHelper.getLocaleFromString("en")
val langStr = locale?.getDisplayName(locale)?.capitalize()
```

app/build.gradle
```diff
@@ -218,7 +219,7 @@ dependencies {
 }
 
 buildscript {
-    ext.kotlin_version = '1.1.2-3'
+    ext.kotlin_version = '1.1.2-5'
```
NoodleMage: Will remove this in final commit

NoodleMage:
screenshot_1497778487_framed

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainHolder.kt
```diff
@@ -0,0 +1,80 @@
...
+class CatalogueMainHolder(view: View, val adapter: CatalogueMainAdapter) : FlexibleViewHolder(view, adapter), FlexibleAdapter.OnItemClickListener {
+    /**
+     * Adapter containing sources
+     */
+    private var adapterSource = CatalogueMainCardAdapter(this)
+
+    val prefs: PreferencesHelper = Injekt.get()
+    val controller = adapter.controller
+
+    override fun onItemClick(position: Int): Boolean {
```
inorichi: Clicks should be handled by the controller

NoodleMage: Maybe it's something I overlooked. But how would I get the item clicked in the holder withing the controller?

inorichi: You can pass `adapterPosition` as parameter. Then in the controller `adapter.getItem(position)`.

inorichi: https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/recently_read/RecentlyReadHolder.kt

Here's a good example of how I'm doing it.

NoodleMage: I'm implementing this. But how would I do the adapter.getitem(position) if the adapter is set in the holder. Your example doesn't use an adapter within a holder.

NoodleMage: Ignore what I said. It just clicked ;)

CarlosEsco:
@NoodleMage 

Regarding icons here were the highest quality icons i could find for each of the main catalog sources.  
http://imgur.com/a/HoK2F

Maybe you could do something like circle icon with logo if HQ enough or Just upper case letter.  and the below it be flag rectangle so like:

image


Alternatively if you prefer to keep separate cards for each language I think the  flag then the name of the group/language would look good.

NoodleMage:
@CarlosEsco  apologies for late response. The problem with the source icons is that it should be defined in the source itself. Maybe we will add this, but because it's an aesthetic thing I want to do it in a future commit.  The flags in the title will probably be a bit much, but perhaps if it's implemented correctly it will look right. But again future commit... ;)

Rave96:
sorry if this is the wrong place to ask this, but why do you have to login to check batoto? other apps like mangarock can access their catalogue without the need of a login

inorichi:
[Because this](http://bato.to/forums/topic/23703-moving-forward-upcoming-changes-and-schedules/). TL;DR old chapters are inaccessible.

Rave96:
oh okay, thanks for the info ^^

cd0rw0:
Hi @NoodleMage ,

ist possible to include last search  something like in mangarock 

https://imagez.to/i/UAo7Ak6g

NoodleMage:
@cd0rw0 Sure, not in this pull though (want to keep it as basic as possible). 
Could you create an issue for it? Will try to add it after this pull is reviewed and approved.

cd0rw0:
Hi @NoodleMage , 

after testing,  here are my complaint/suggestion to improve,
in global search result page, searched keyword is empty.
so we dont know what word we search and also we cant refine the search with amended spelling.

Recent catalog
in my opinion its redundant for normal user, as dont see usefulness of this as this just repeat of catalog at bottom.
i think its usefull for user with  alot of catalog.  

ist possible instead of recent, to add catalog sorting in sources page.
like sorting option in My library categories

cd0rw0:
Hi @NoodleMage 

Thanks for your interest on last search result issue ..
i have previously  created #779 on this matter

inorichi:
I don't like the idea of creating an adapter in a ViewHolder. If possible I'd like to have a single RecyclerView for everything, it should be possible but maybe not as intuitive as we'd like.

[This](https://stackoverflow.com/questions/28493207/google-play-store-like-interface-using-recycler-view) is all I could find in SO.

I'll try to find something when I get some free time.

inorichi:
After thinking it better, it's not that bad to have the adapter in the ViewHolder. However, it should be created in the constructor (along the setup, the layout manager and `isNestedScrollingEnabled`) and the binding method should only update the content. Reattaching the adapter will make those views to be recreated and that affects performance while scrolling.

The best we could get in performance would be to use a shared pool of those ViewHolders, but it's not possible if FlexibleAdapter is used, because by design the ViewHolder has a reference to an adapter so it can't be shared. But this is not a priority/required for now.

inorichi:
Tested a bit the branch, these are my thoughts on it:

* When you press and hold a source, the space around the buttons remains white. It's not a big deal but it looks ugly.
* I think sources with an empty lang should be categorized under a section "Others". `LocalSource` could be renamed to `Local` (maybe also a resource string `R.string.local_source`)
* The manga's title in a global search has a lot of padding and we are losing a few letters there.
* Going back from a manga to the global search seems to make network requests again. I'll look at the code to see what's going on.
* Changing the visible sources doesn't update the list until the next time catalogues are open. This can be easily fixed by always updating the list in `onChangeStarted` when `from` is `SettingsSourcesController`.

I'll also start a code review soon.

NoodleMage:
> When you press and hold a source, the space around the buttons remains white. It's not a big deal but it looks ugly.

True. I can put a border around the button, but this looks a bit weird with two buttons.

> I think sources with an empty lang should be categorized under a section "Others". LocalSource could be renamed to Local (maybe also a resource string R.string.local_source)

Will change.

> The manga's title in a global search has a lot of padding and we are losing a few letters there.

Will try to change this a bit.

> Going back from a manga to the global search seems to make network requests again. I'll look at the code to see what's going on.

I looked at the network refresh but couldn't figure out why. If you can find it great ;).

> Changing the visible sources doesn't update the list until the next time catalogues are open. This can be easily fixed by always updating the list in onChangeStarted when from is SettingsSourcesController.

Will look into this.

inorichi:
> True. I can put a border around the button, but this looks a bit weird with two buttons.

I was thinking of buttons with transparent background.

app/build.gradle
```diff
@@ -217,7 +217,7 @@ dependencies {
 }
 
 buildscript {
-    ext.kotlin_version = '1.1.4'
+    ext.kotlin_version = '1.1.4-2'
```
NoodleMage: -_-. Forgot to remove gradle change in commit. Will remove in future commit ;).

NoodleMage:
The buttons are now transparent. So if you click the card it will highlight the entire card. If you click the button it will highlight the button. It looks very shtitty on pre lollipop though (no ripple on buttons). But to make it look better I need to add xml drawables for all themes and make a new attribute.

Rest of the things are also implemented. Not sure if I implemented the OnChangeStarted correctly, so correct me if I did it wrong xD.

inorichi:
There's no need to add a listener to the router. There's a method `onChangeStarted` you can just override.

NoodleMage:
I know. But it only allowed to check if type.isEnter. I couldn't check if the user is returning from SourceSettingsConsutructor that way. Or does the control handler contain this information?

inorichi:
So true, I overlooked that 😞 .

We can use a custom change handler class.

At the end of the controller: `private class SettingsSourceFadeChangeHandler : FadeChangeHandler()`

When pushing the settings controller: 
```
router.pushController((RouterTransaction.with(SettingsSourcesController()))
        .pushChangeHandler(FadeChangeHandler())
        .popChangeHandler(SettingsSourceFadeChangeHandler()))
```

In `onChangeStarted`: 
```
if (!type.isPush && handler is SettingsSourceFadeChangeHandler) {
    // ...
}
```

I guess it's not a very bad approach.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainAdapter.kt
```diff
@@ -0,0 +1,12 @@
+package eu.kanade.tachiyomi.ui.catalogue.main
+
+import eu.davidea.flexibleadapter.FlexibleAdapter
+
+/**
+ * Adapter that holds the catalogue cards.
+ *
+ * @param controller instance of [CatalogueMainController].
+ */
+class CatalogueMainAdapter(val controller: CatalogueMainController) :
```
inorichi: This class seems unnecessary, you can use a `FlexibleAdapter` in the controller.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainController.kt
```diff
@@ -0,0 +1,231 @@
...
+ * [CatalogueMainCardAdapter.OnLatestClickListener] call function data on latest item click
+ */
+class CatalogueMainController : NucleusController<CatalogueMainPresenter>(),
+        SourceLoginDialog.Listener,
+        CatalogueMainCardAdapter.OnBrowseClickListener,
+        CatalogueMainCardAdapter.OnLatestClickListener {
+    /**
+     * Adapter containing sources.
+     */
+    private var adapter = CatalogueMainAdapter(this)
```
inorichi: This needs to be nullable, created in `onViewCreated` and nullified in `onDestroyView`

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainController.kt
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Boolean which is true when recent catalog is added to adapter.
+     */
+    private var isAdded = false
```
inorichi: This seems fishy. It's only updated when a login dialog is closed. Is there really a need for this variable?

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainController.kt
```diff
@@ -0,0 +1,231 @@
...
+            isAdded = false
+            presenter.loadSources()
+            presenter.loadRecentSources()
+        }
+    }
+
+    /**
+     * Called when browse or item is clicked in [CatalogueMainCardAdapter]
+     */
+    override fun OnBrowseClickListener(item: CatalogueMainCardItem) {
```
inorichi: This is a method and should start with lower case.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainController.kt
```diff
@@ -0,0 +1,231 @@
...
+            router.pushController(RouterTransaction.with(CatalogueController(null, item.source))
+                    .pushChangeHandler(FadeChangeHandler())
+                    .popChangeHandler(FadeChangeHandler()))
+        }
+    }
+
+    /**
+     * Called when latest is clicked in [CatalogueMainCardAdapter]
+     */
+    override fun OnLatestClickListener(item: CatalogueMainCardItem) {
```
inorichi: This is a method and should start with lower case.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainController.kt
```diff
@@ -0,0 +1,231 @@
...
+            searchView.setOnQueryTextListener(object : SearchView.OnQueryTextListener {
+                override fun onQueryTextSubmit(query: String): Boolean {
+                    router.pushController((RouterTransaction.with(CatalogueSearchController(query)))
+                            .popChangeHandler(FadeChangeHandler())
+                            .pushChangeHandler(FadeChangeHandler()))
+                    return true
+                }
+
+                override fun onQueryTextChange(newText: String?): Boolean {
+                    // TODO suggestions?
```
inorichi: Probably nothing else to do. It's better to not use a debouncer here and let the user type the entire title.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainHolder.kt
```diff
@@ -0,0 +1,61 @@
...
+ * Holder that binds the [CatalogueMainItem] containing catalogue cards.
+ *
+ * @param view view of [CatalogueMainItem]
+ * @param adapter instance of [CatalogueMainAdapter]
+ */
+class CatalogueMainHolder(view: View, adapter: CatalogueMainAdapter) : FlexibleViewHolder(view, adapter) {
+    /**
+     * Adapter containing sources
+     */
+    private var adapterSource = CatalogueMainCardAdapter(adapter.controller)
```
inorichi: I don't like multiple adapters in this screen. It should be possible a single adapter approach with multiple item types and view holders. I'll work on this approach.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainPresenter.kt
```diff
@@ -0,0 +1,114 @@
...
+        // Load enabled sources
+        loadSources()
+
+        // Load most recently used source.
+        loadRecentSources()
+    }
+
+    override fun onDestroy() {
+        // Unsubscribe subscriptions.
+        recentSourceSubscription?.unsubscribe()
```
inorichi: These are automatically unsubscribed. All the extension methods `subscribe*()` in the presenter unsubscribe automatically in `onDestroy`

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/card/CatalogueMainCardAdapter.kt
```diff
@@ -0,0 +1,38 @@
+package eu.kanade.tachiyomi.ui.catalogue.main.card
```
inorichi: I'd prefer to remove the package `card`. There aren't so many classes here to require more nesting.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainController.kt
NoodleMage: My idea was to load a list with title suggestion here. So if the user searches for a certain title it show an list of options an user can click.

inorichi:
You can merge my branch if you want. I'll try to review and maybe make some changes to the global search package.

I'm also going to remove the spinner in the toolbar, it's hacky and not needed anymore.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/main/CatalogueMainController.kt
inorichi: It's not going to be very useful IMO. We don't have a simple API we can query and the suggestions will probably be really slow.

inorichi:
Going back from a global search now keeps the results, however the scroll state of the recycler views inside the cards is not restored. I haven't looked into it too much.

I think it should be nice to set the toolbar title to the current global search.

`query` should be stored in the controller's bundle to survive process restart. Will change later.

Everything else seems to be working fine.

cd0rw0:
Hi guys,

from my testing here are bug found

after press & hold to bookmark from browse page -> no chapter when check in library (even after refresh)

Filter
- not updating between sources, only update once click reset in filter.
- search not working on 1st try, need to click reset 1st. (seem like related to above)
- going back from manga to filter result not working, instead going to source homepage/browse page
 
so far thats all i thing ... other than that this global search really good.
TQ

444nonymous:
Now that global search has been merged, "Clickable author and genre fields" by @QuestofIranon could be implemented https://github.com/inorichi/tachiyomi/issues/860

PR #865, opened 2017-06-27
Added Mangazuki source.

Beanow:
https://mangazuki.co/ publishes their own translations on their new site.
They are therefore way ahead of some sites that reupload their work.
For example http://kissmanga.com/Manga/Brawling-Go is currently at ch 75.
While https://mangazuki.co/series/Brawling-Go is at ch 86.

It's Patreon supported and ad-free. The reason I added the source was to have the convenience of automatic downloads and faster updates.

The RX observables confused the living daylight out of me, so this is the best I could do for the paginated chapter list. I commented most of my choices, let me know if it's an acceptable implementation.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
@@ -58,7 +58,8 @@ open class SourceManager(private val context: Context) {
             Mangachan(),
             Readmangatoday(),
             Mangasee(),
-            WieManga()
+            WieManga(),
+            Mangazuki()
     )
 
     private fun createYamlSources(): List<Source> {
--- a/app/src/main/java/eu/kanade/tachiyomi/source/online/english/Mangazuki.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/online/english/Mangazuki.kt
@@ -1,0 +1,194 @@
+package eu.kanade.tachiyomi.source.online.english
+
+import eu.kanade.tachiyomi.network.GET
+import eu.kanade.tachiyomi.network.asObservableSuccess
+import eu.kanade.tachiyomi.source.model.*
+import eu.kanade.tachiyomi.source.online.ParsedHttpSource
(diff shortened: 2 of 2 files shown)


NoodleMage:
Mangazuki is already implemented. https://github.com/inorichi/tachiyomi-extensions/pull/44

Beanow:
Well that was a day well spent wasn't it? I worked till 4am for this.

Might I suggest you mention the existence of this anywhere in the app?
Also, the installation process for those extensions is terrible. I had to build the main application from source as the available 3 (dev) builds fail to install with an "unknown error" and the non-dev one doesn't show extension sources. As well as the requirement for adding a new repo to f-droid.

wopian:
It's mentioned at the top of the [FAQ](https://github.com/inorichi/tachiyomi/wiki/FAQ), which itself is at the top of the README

Beanow:
@wopian I suggest adding it to the app instead like #867
This is what happened for me.

1. Install Tachiyomi for the first time.
2. Look at Settings > Sources.
3. Conclude Mangazuki is not there,
decide to help out and submit a pull request for it (not an issue and ask for it).
4. Look at https://github.com/inorichi/tachiyomi/tree/master/app/src/main/java/eu/kanade/tachiyomi/source/online/english to see how sources are implemented.
5. Built PR.

At which point was I supposed to know somewhere in the FAQ is a link to extension which doesn't explain installation except for the issue https://github.com/inorichi/tachiyomi-extensions/issues/37 which requires an alternative f-droid repo which doesn't have direct APK downloads as far as I can tell to make things easier to install.

NoodleMage:
The implementation with the extension is bad, because it's not yet implemented. The design from #849 will allow us to to do it design wise but we still have to implement it. 
The way it is now is just an alternative so that people who wish to load the sources can do it.

Beanow:
Right, so I got unlucky to come in during a transitional phase and the extensions are currently not intended for use by mere mortals. That means my other issues don't apply either. Damn my luck.

PR #873, opened 2017-07-02
Fix issues with Batoto

### LINKED ISSUE #872: Error on bato.to
[Image](https://imgur.com/gallery/aaYhI) only happens on bato.to, no extensions installed. Also gets http error 405 when trying to view something from batoto. The problem appeared just recently (like a few hours ago maybe).

I assume batoto updated their reader or something, I have no idea other than that.

Version: v0.5.2
Android 7 on Samsung Galaxy Tab A 10" and LG G5

ddmgy:
Some links now use https rather than http, so parsing was affected. I don't know if this is a permanent change on Batoto's end, but it does fix the issue of not being able to load any manga from Batoto in Tachiyomi (built from master branch). Fixes #872.


inorichi:
Thanks.

Issue #826, opened 2017-06-01
Feature: Downloaded chapters badge


444nonymous:
There is a badge for unread chapters (blue background & white numbers,  placed in the top left corner).

What about creating another for unread downloaded chapters?  (maybe orange background & white numbers, placed in the top right corner)

Or finding something that works better to mark what you already have to read in offline mode. Maybe "364/125" (total unread / downloaded unread). Or the opposite: "125/364" (125 downloaded out of 364).

Thanks.

PR #895, opened 2017-07-13
Add download badge, and local badge for library view

CarlosEsco:
addresses #826 as well as adds badge to distinguish local manga from network ones.

screenshot_20170713-135602
screenshot_20170713-135608


inorichi:
Showing downloaded chapters can slow a lot the library. It'll need to be configurable or totally rewrite the fetching method with a singleton (and maybe a filewatcher too).

CarlosEsco:
@inorichi   I thought it might slow, and i could load test but i dont think its gonna be any slower than say having the download filter on.  The reason being is the download filter is doing almost the same thing.  

Download filter opens source folder for every manga does a listFiles and verifies its not empty.
This opens source folder for every manga does a listFiles and does size on listFiles.

inorichi:
Yes, but downloaded filter is also optional, while this forces the lookup. 

It wouldn't be so bad if the `File` API were used, but the Storage Access Framework is painfully slow.

I think that option could be placed in the right drawer, below Display: List or Grid

CarlosEsco:
Okay I'll add the option sometime this week

CarlosEsco:
Added as an option

image

blackdevil91:
hi, ive updated tachiyomi to the latest version but i dont have the badge option. sorry for any inconvenience. would be nice if receive some assistant

CarlosEsco:
This is just a pull request that hasn't actually been added to the main library yet

blackdevil91:
noted and sorry for the question

CarlosEsco:
All good

inorichi:
Sorry I've been delaying this one a lot. I'm making it the next priority.

inorichi:
Merged manually in https://github.com/inorichi/tachiyomi/commit/d0567de4e6ce9fdc39f430ea45f04526dbe4290e and the next few commits.

Issue #926, opened 2017-08-11
Mangafox Chapter Titles


ReGtrs:
Is there a way for chapter titles and chapter numbers to show (at once) instead of the chapter number alone in Mangafox?

PR #936, opened 2017-08-15
Add Manga Fox Chapter titles if exist

CarlosEsco:
Addresses
#926 

null safe check to see if chapter title exists for a manga if so appends it to the chapter name.
Also adjusted chapter impl so that if a name changes from the source for an existing manga in library the name change will reflect in the existing chapters.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/ChapterImpl.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/ChapterImpl.kt
@@ -31,9 +31,10 @@ class ChapterImpl : Chapter {
         if (other == null || javaClass != other.javaClass) return false
 
         val chapter = other as Chapter
-        // Forces updates on manga if scanlator changes. This will allow existing manga in library
-        // with scanlator to update.
-        return url == chapter.url && scanlator == chapter.scanlator
+        // Forces updates on manga if url, scanlator, or chapter name changes.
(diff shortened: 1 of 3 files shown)


CarlosEsco:
image

inorichi:
I don't want to go through the chapter readded process if the title changes. In fact, it shouldn't do it if the scanlator changes. All that logic should go into the sync chapters with source method.

CarlosEsco:
do you want to merge and discard chapters readded code.  I'll look into and open new pull request in the sync chapter with source method

inorichi:
You can either add the mangafox part in the other PR or revert the changes to `ChapterImpl` here.

The sync method should silently update scanlator/title in the db if required (without adding them to the added/deleted chapters list).

You can use kotlin's `find` instead of `equals` with the `in` for collections if you think it's better.

CarlosEsco:
removed chapterImpl equality stuff and changed sourceChapter filter and dbChapter Filter in ChapterSourceSync

inorichi:
I was thinking of new logic once a chapter with the same url is found. You can check the title or the scanlator and store those that changed in a list `toChange` and update the db without adding them to `toAdd` or `toDelete`, because those trigger a new chapters notification.

I'll avoid changing the title though, because if the chapter is downloaded it'll be orphan.

inorichi:
c0d7b16 This is what I meant.

CarlosEsco:
yea see what you mean way more elegant solution

PR #937, opened 2017-08-15
License manga update and Manga Fox Title Update

CarlosEsco:
fixes license status for manga fox that was not being set.
does not show chapters if manga is licensed this will prevent false issues being open for manga fox, since when licensed chapters are viewable.

Also address #926

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt
@@ -202,11 +202,16 @@ abstract class HttpSource : CatalogueSource {
      * @param manga the manga to look for chapters.
      */
     override fun fetchChapterList(manga: SManga): Observable<List<SChapter>> {
-        return client.newCall(chapterListRequest(manga))
-                .asObservableSuccess()
-                .map { response ->
-                    chapterListParse(response)
-                }
+        if(manga.status  != SManga.LICENSED) {
+            return client.newCall(chapterListRequest(manga))
+                    .asObservableSuccess()
+                    .map { response ->
+                        chapterListParse(response)
+                    }
+        }else{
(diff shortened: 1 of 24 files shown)


CarlosEsco:
image

inorichi:
Can you include the mangafox chapters' title in this PR? Also fix the missing spaces etc here.

CarlosEsco:
Merged everything into this one pull request.   Fixed spacing let me know if its still not right.  If so maybe share your java code style for android studio or which schema you are using.

app/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt
```diff
@@ -197,16 +197,21 @@ abstract class HttpSource : CatalogueSource {
...
-                }
+        if (manga.status != SManga.LICENSED) {
+            return client.newCall(chapterListRequest(manga))
+                    .asObservableSuccess()
+                    .map { response ->
+                        chapterListParse(response)
+                    }
+        } else {
+            val chapters: List<SChapter> = emptyList()
+            return Observable.just(chapters)
```
inorichi: Probably better to throw an exception? `Observable.error(Exception("Licensed"))` for example

CarlosEsco: `return Observable.error(Exception("Licensed - No chapters to show"))`

Added a little more detail

inorichi:
Having a `.warning` class doesn't mean the manga is licensed. I'm reverting the changes before 0.6.2

CarlosEsco:
do you have a mangafox example where warning wasn't licensed?

inorichi:
Any 18+ rated manga, like [Berserk](http://mangafox.me/manga/berserk).

CarlosEsco:
k I will look into a better solution and create a new pull request

Issue #938, opened 2017-08-15
[Feature Request] Completed only Library filter


Yube:
A third filter in the Library section of the App, together with "Downloaded" and "Unread" like the one already present in the Catalog section.
Thank you!

PR #941, opened 2017-08-17
Add Filter by Completed for library

CarlosEsco:
#938 

image


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -110,6 +113,10 @@ class LibraryPresenter(
                 return@f false
             }
 
+            if(filterCompleted && manga.status != SManga.COMPLETED){
```
inorichi: Please use spaces.

app/src/main/res/values/strings.xml
```diff
@@ -26,6 +26,7 @@
     <string name="action_filter">Filter</string>
     <string name="action_filter_downloaded">Downloaded</string>
     <string name="action_filter_bookmarked">Bookmarked</string>
+    <string name="action_filter_completed">Completed</string>
```
inorichi: I think we could reuse the existing "completed" string for manga.

CarlosEsco: fixed and pushed

PR #954, opened 2017-08-25
batoto search/chapter issues #953 fix

### LINKED ISSUE #953: Java Error when loading a manga with Batoto
Here's a screenshot: http://i.imgur.com/MKBkOyK.png

I remember this issue was solved by updating some time ago but the Dev version has the same error.

Version: 0.6.1
Android 6.0

CarlosEsco:
#953 

looks like batoto switched from https to http


inorichi:
I think I'm going to keep using https, and update the parser to accept both http and https.

CarlosEsco:
sounds good will close.

Issue #943, opened 2017-08-20
"All" Manga Tab


cd0rw0:
All Manga

it would be convenient if all saved manga in 1 fixed place by default regardless of their categories.
this would help user 
- search saved manga ( currently after key in search, we need to changed categories 1 by 1)
- identify duplicate manga from different source

Thanks

CarlosEsco:
I like this idea.  I dont use categories but can see how its useful will look into.

ShakeBake:
Cant we already do this? Make a default catagory called all and automatically add all subsribed manga to that as default. Then you can use multiple tags to both move the manga to a desired catagory and still keep it in the 'all' catagory.

cd0rw0:
since its manual & bound to human error.

CarlosEsco:
Yea my plan was to rewrite so every manga was always in default, and you could "move" into other category but default would always have all.

inorichi:
It's easier to have a switch (probably in the overflow menu of the toolbar) and then a few changes to the library presenter to fetch the data with or without categories.

arkon:
https://github.com/tachiyomiorg/android-app/issues/10

Issue #885, opened 2017-07-08
Default Category Always Ask setting is broken.


iKzone:
The default category (always ask) setting does not work when you try to bookmark a manga. I have two categories which is Default and Plan to Read. It automatically adds the manga into my Plan to Read Category and does not give me the option to select with category to go to.

Version: v0.6.1

inorichi:
As workaround you can create a category called Default. Only one category and the default is a special case, I never thought about it.

xChickens:
Creating another category called Default still keeps the original default category and I don't think there's a way to mass move all of the old Default to new one. :/

But I found that the update r1149 to r1152 is where it stopped asking for categories

wopian:
Mass moving to a different category is possible - just no way to select all with 1 action. 

In the `default` category long hold any manga, then just tap all the other manga. Press the tag icon (1st) and chose the category to move them to

LakotaGray:
The issue is due the fact you have only created a single category, which is then assumed to be where you want series to go. Create a 2nd category to go along with default and your “plan to read” it will solve your issue

arkon:
I don't think this is still an issue?

PR #960, opened 2017-08-27
Change default category to be All category and category chooser fix

CarlosEsco:
This pull request addresses #943 

This replaces the default category with an always persistent ALL category.   While doing this code i noticed a few issue with the ChangeMangaCatalogDialog.

1.  When you have the system default category and 1 user category, the manga was always being added to the user category.  With no option to add to the default category.
This issue was referenced in #885.  I've adjusted so when favoriting a manga the dialog box will pop up with the 1 category as an option to check.  If not checked and dialog is dismissed/done  it will just be added to library with no category chosen.
2.  The way materialDialogs itemsCallbackMultiChoice works is upon clicking a checkbox the code get called and the manga is added/removed from the category. This means clicking ok, cancel, or dismissing didnt matter as checking/unchecking was adjusting categories.  This can cause some user confusion since  clicking cancel one would assume the catalog changes made would not take effect but were.  This was fixed by removing the negativebutton and renaming the positive to done.  
3.  When single(after fix 1) or multiple categories exist and no checkbox for category was clicked clicking ok/done did not add the manga to favorites.  This caused catalog view to grey the manga like it was added but going to library it was nowhere to be found.  then going back to catalog it was ungrey again.  Fixed this by adding alwaysCallMultiChoiceCallback  so dismiss/done will at very least favorite the manga to its added to the ALL category.  This allows the flow to function better.  Favoriting a manga should add it no matter what even if the user doesn't want to add to a specific category.

Unrelated to code changes when i ran android studio format it auto formatted some spacing issues that existed in CatalogueController.kt 

I manually tested a few scenarios with no issues

1.  Default add to category  with 1 or more categories.
2.  1 user category adding to catalog but not category
3.  1 user category adding to catalog and choosing category
4.  2+ user category adding to catalog but not category
5. 2+user category adding to catalog and choosing category
6.  Adding to multiple categories
7.  moving/removing to multiple categories
8.  Removing manga with no categories, single category or multiple categories
9.  Searching in ALL category
10.  filter, sort, display.  etc

basically the main difference in code is originally code was taking all manga and grouping them by category id.  new code groups them by category id but all puts them all in the default ALL category.  No category view/handling logic with exception to the change category dialog.

@inorichi regarding your comment "It's easier to have a switch (probably in the overflow menu of the toolbar) and then a few changes to the library presenter to fetch the data with or without categories."

I looked into this way but I believe the flow of control would be clunky.  You would have an all view then a switch in the overflow menu for category view.  So user breaks from their normal flow of switching between multiple categories for manga to having to select something from overflow to show ALL then back.  With this approach you can easily swipe from categories to view ALL and doesnt break up the flow.  The only downside i currently see is if user is like #885 where someone is using default in such a way they are treating it as a user created chapter.  Since upon upgrading to this the everything is added to ALL.  

No Category remains same
image
Adding a cateogry remains same just name changed
image
All Manga view (switched to list view to show more manga)
image
Category 1 view
image
Category 2 view
image

Adding Manga as Favorite or moving  with only 1 category
image

Adding Manga as favorite or moving category is the same
image


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -194,12 +194,7 @@ class LibraryPresenter(
     private fun getLibraryObservable(): Observable<Pair<List<Category>, Map<Int, List<Manga>>>> {
         return Observable.combineLatest(getCategoriesObservable(), getLibraryMangasObservable(),
                 { dbCategories, libraryManga ->
-                    val categories = if (libraryManga.containsKey(0))
-                        arrayListOf(Category.createDefault()) + dbCategories
-                    else
-                        dbCategories
-
-                    this.categories = categories
+                    this.categories = arrayListOf(Category.createDefault()) + dbCategories
```
jshap70: shouldn't you be checking if the default exists already? won't you be creating a new "ALL" and overwriting it every time with the current implementation?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -221,7 +216,28 @@ class LibraryPresenter(
...
+                            if (map.get(it.category) == null) {
+                                map.put(it.category, mutableListOf())
+                            }
+                            if (!map.get(it.category)!!.contains(it)) {
+                                map.get(it.category)!!.add(it)
+                            }
+                        }
+                    }
+                    return@map map
+                }
```
jshap70: This seems like a very poor place to calculate this tbh. It makes a lot more sense to insert them into the `ALL` category on the `Category.createDefault()` call than it does to recalculate the all group every time

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
CarlosEsco: The previous code acts the same way
`                 val categories = if (libraryManga.containsKey(0))
                       arrayListOf(Category.createDefault()) + dbCategories`
if any manga was using default it would create a new default.  This because default category is not stored in the database.  

The new code doesnt need to check because if a manga is assigned to default or to another category we always want the ALL category to show.  It keeps the remaing logic of the default/ALL category not being stored.   Also this code is just generating the headers and only once when the subscribed to the Library.

To put your concerns at ease I forgot to mention but i did verify what would have if i had a default category with multiple categories  Manga assigned to each  and then updated to this version.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
CarlosEsco: I understand your point.  The reason it was done like this was categories were already being calculated everytime with the groupBy.  This just replaced that manually to allow manga to be added to ALL category and there own category if they have one.  The ALL Category isnt stored in the database and if it was you would still need to calculate all the manga that are in categories.

cd0rw0:
Hi @CarlosEsco  
Thanks for this working on this.
From what I test, the only problem I can see is, 
since you maintaining to user bookmark without select category, later User will have problem identify this uncategorized  manga from "all" tab
I think if have filter for this like "uncategorized' in all tab will solve this issue .
This also will help user during migration from default tab to all tab.

Thanks.

CarlosEsco:
Am I understanding this right.  a user using uncategorized and category and on upgrade to all the distinction for uncategorized that a user could have been using is lost?

cd0rw0:
yes n also during normal usage, uncategorized item will be lost in "all"
so will have new report to solved this issue...
only way i can see is 
1. category selection is must (so no more uncategorized manga)
or
2. add "uncategorized" (or any other the name) filter in All tab.

CarlosEsco:
So I think it should be #1.  From my use point of the app I've never used uncategorized as a category.  It was just everything and then I was using a category for completed. If your wanting something uncategorized I think it should be put into a category of some sort

CarlosEsco:
@cd0rw0  I've pushed a change that takes all uncategorized manga pre ALL and puts them into a ORIGINAL DEFAULT category post ALL.  That way anyone who was using uncategorized as a sorta of category will not lose this distinction.

cd0rw0:
@CarlosEsco 

Understand on migration part. This will help current app user in migration.
How about  new bookmarks is category selection is a must or still option?

CarlosEsco:
@cd0rw0   it is not if you do not select a category and select done it will still add to catalogue which is change from the old behavior pre ALL.  I will change the  wording from 

Choose categories  to Choose categories (Optional) to make this more clear

lordbk:
Hi @CarlosEsco 

With recent changes, merging with latest master giving conflict error.

CarlosEsco:
@lordbk  try now  however i dont know if i will keep maintaining, it doesnt seem like this will get merged or is wanted.  Maybe a poll should be created in like they did with bottom navigation bar.

@inorichi  @NoodleMage  thoughts?

j2ghz:
Is it possible to hide/turn off the All category? For my use-case it'd just get in a way (my categories are "To read","Reading" and "On-hold").

cd0rw0:
@j2ghz the point is, to have place one where all saved manga can be viewed, searched or do some house keeping regardless of whats or how many categories we have.

CarlosEsco:
@j2ghz  @cd0rw0  it should be pretty easy to show/hide all category i can make a checkbox similar to download badges.

j2ghz:
@cd0rw0 I understand what the point is, that doesn't mean I find it useful. That's why I asked for a hide option.
@CarlosEsco Thanks

cd0rw0:
hi @CarlosEsco 

Switch is great. Thanks. but
"All" Switch conflict wt "download badge" or rather "download badge" conflict with "all"
if we tick "download badge", "all" also turn on.

<a href="https://ibb.co/kXsuC6">Screenshot_20171221_125835</a>

CarlosEsco:
@cd0rw0 there are gonna be few issues with a different merge since they both touch same code.  
fixed the issue you mentioned above

cd0rw0:
@NoodleMage @inorichi 

any love for this PR ?

inorichi:
The other PRs take precedence at the moment and this one will require some changes. It'll be done but be patient.

Issue #691, opened 2017-02-23
Zoom speed increase


ArmandBernard:
I'd like to request that the double-tap zoom speed be increased or be modifiable.
I like to read fairly quickly and on my phone and I zoom in and out all the time.
The zoom speed slows me down quite a bit.

Version: v0.4.2
Samsung S7, Android 7.0 Nougat

PR #974, opened 2017-09-10
Add option to change double tap animation speed in the reader

jshap70:
solves https://github.com/inorichi/tachiyomi/issues/691

Pretty simple, the SubsamplingScaleImageView has a builtin for double tap anim speed.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -39,7 +39,7 @@ class PreferencesHelper(val context: Context) {
 
     fun pageTransitions() = rxPrefs.getBoolean(Keys.enableTransitions, true)
 
-    fun doubleTapAnimDuration() = rxPrefs.getString(Keys.doubleTapAnimationDuration, "500")
+    fun doubleTapAnimSpeed() = rxPrefs.getInteger(Keys.doubleTapAnimationSpeed, 500)
 
     fun showPageNumber() = rxPrefs.getBoolean(Keys.showPageNumber, true)
 
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -39,6 +39,8 @@ class PreferencesHelper(val context: Context) {
 
     fun pageTransitions() = rxPrefs.getBoolean(Keys.enableTransitions, true)
 
+    fun doubleTapAnimDuration() = rxPrefs.getString(Keys.doubleTapAnimationDuration, "500")
+
     fun showPageNumber() = rxPrefs.getBoolean(Keys.showPageNumber, true)
 
     fun fullscreen() = rxPrefs.getBoolean(Keys.fullscreen, true)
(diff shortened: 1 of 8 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
```diff
@@ -39,6 +39,8 @@ class PreferencesHelper(val context: Context) {
 
     fun pageTransitions() = rxPrefs.getBoolean(Keys.enableTransitions, true)
 
+    fun doubleTapAnimDuration() = rxPrefs.getString(Keys.doubleTapAnimationDuration, "500")
```
NoodleMage: Why is doubleTapAnimDuration a string?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
```diff
@@ -62,6 +62,14 @@ class SettingsReaderController : SettingsController() {
             defaultValue = "0"
             summary = "%s"
         }
+        listPreference {
+            key = Keys.doubleTapAnimationDuration
+            titleRes = R.string.pref_double_tap_anim_duration
+            entries = arrayOf("No animation", "Fast", "Normal")
```
NoodleMage: Please make this strings.xml strings so they can be translated.

Another option is make it No animation, 0.5x, 1.0x, etc. But I like this better (no,fast,normal).

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
```diff
@@ -62,6 +62,14 @@ class SettingsReaderController : SettingsController() {
             defaultValue = "0"
             summary = "%s"
         }
+        listPreference {
+            key = Keys.doubleTapAnimationDuration
```
NoodleMage: The reader setting page is getting to crowded in my opinion. 
I think we should add some headers. If someone has some ideas for this, let me know.

app/src/main/res/values/strings.xml
```diff
@@ -143,6 +143,7 @@
     <string name="pref_fullscreen">Fullscreen</string>
     <string name="pref_lock_orientation">Lock orientation</string>
     <string name="pref_page_transitions">Page transitions</string>
+    <string name="pref_double_tap_anim_duration">Double tap to zoom animation duration</string>
```
NoodleMage: Bit long. maybe Double tap animation speed.

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
jshap70: There's some weirdness with the `listPreference` class and the `intListPreference` class. the int list preference appears to take ints but cast them to a string? anyways, this "worked", if you know how to make it work so that in `SettingsReaderController.kt` I can make the `entryValues` ints then I'll change it.

app/src/main/res/values/strings.xml
jshap70: yeah but this one rhymes :stuck_out_tongue_closed_eyes:

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
inorichi: Use `intListPreference`. Entry values are still string (because that's how ListPreference works), but you retrieve the value as an int.

jshap70:
@NoodleMage better late than never, everything should be kosher now

PR #977, opened 2017-09-11
added licensed element check for MangaFox

CarlosEsco:
@inorichi fixed issue you were mentioning about warning div.

Verifies the warning div contains licensed before assigning it as licensed.

image
vs adult non licensed
image


paronos:
I was wondering, it is true for every sources that licensed manga have no visible chapters?

On a side note, I have not found any licensed manga on mangafox, the licesed notice is based on user location?

CarlosEsco:
@paronos  so definitely location based.  Here's a sample below.  It also tricks the users because their site will still list all the chapters but when you click one no images will be available.  So in the app you get the chapter list of all chapters but then it looks like the app isnt loading the chapter correctly.   This pull request just fixes a previous pull request in that i was assuming all warnings were licensed when they also warn for adult content for manga fox.  The previous and already integrated pull request zeroes out chapters from any source when the manga is licensed based off of what the source provides.  No other source is providing licensing information/app isnt checking for it so there should be no issue.

image

PR #983, opened 2017-09-16
Added regex to strip local manga chapter names

### LINKED ISSUE #981: Request: Local manga naming
Hello! I noticed that on local manga if a chapter is called lets say "Awesomeness Chapter 1" and is placed inside a folder called "Awesomeness" that the chapter list will no longer show the full name and just start at the first non space character after the name (in this case "Chapter 1") to save on needlessly wasted space, which is really cool!

Is it possible to have it omit specific characters along with the spaces? Like if the manga was called "Awesomeness - Chapter 1" it will now show as "- Chapter 1". Not a big deal at all! But if it was something simple like adding "-" (and maybe "_" as well if it isn't already) along with spaces as characters to skip before starting the chapter name from the file name then it might be worth it since its a really common naming scheme.

Thanks!

ayuhsya:
The chapter names will now be stripped of any non-alphanumeric characters.

Closes #981


CarlosEsco:
How does this work with manga titles that contain non alphanumeric?
Like
 S.I.D
or 
3-NEN Kinmengumi
or
TOMO-CHAN WA ONNANOKO!

ayuhsya:
If the manga's name itself contains non alphanumeric characters, then it'll be removed, as it was being done before this commit. But if the chapter names (after stripping it off the manga name) contains non alphanumeric characters, it'll be removed (without affecting spaces in between words).

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -144,7 +144,9 @@ class LocalSource(private val context: Context) : CatalogueSource {
                         } else {
                             chapterFile.nameWithoutExtension
                         }
-                        val chapNameCut = chapName.replace(manga.title, "", true).trim()
+
+                        var chapNameCut = chapName.replace(manga.title, "").trim()
+                        chapNameCut=chapNameCut.replace("[^A-Za-z0-9\\s+]+".toRegex(), "").trim()
```
NoodleMage: Use """ instead of ". You can then replace \\ with \, this will prevent confusion about the regex.

You also continue after the Chapter tag. So "Awesome Manga - Chapter 3 - The man of pants" will become 
"Chapter 3  The man of pants."

If you want to format everything like Chapter 1, Chapter 2, etc. try to use the [chapter recognition](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/util/ChapterRecognition.kt) 

If you only want to remove the trailing space update your regex or use replaceFirst (not sure if this will work though).

ayuhsya: I updated the .trim() argument to trim the leading hyphens and underscores along with the whitespaces.

gradle.properties
```diff
@@ -15,4 +15,4 @@
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
\ No newline at end of file
+# org.gradle.parallel=true
```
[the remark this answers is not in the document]
ayuhsya: Yes, I already made a commit to restore the gradle.properties file, but apparently a EOF character was added. I have fixed it.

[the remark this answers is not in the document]
j2ghz: @ayuhsya The file is still changed, `git update-index --chmod=-x gradle.properties` should help.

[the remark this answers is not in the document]
ayuhsya: Thank you! Wasn't aware of this git command.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -144,7 +144,7 @@ class LocalSource(private val context: Context) : CatalogueSource {
                         } else {
                             chapterFile.nameWithoutExtension
                         }
-                        val chapNameCut = chapName.replace(manga.title, "", true).trim()
```
j2ghz: Is there a reason to no longer ignore case?

ayuhsya: It seemed redundant because the chapter names are in the same case as the manga name. Restored it back.

PR #997, opened 2017-09-26
Fix downloads from mangahere

### LINKED ISSUE #996: "unexpected url" push notification when attempting to download a chapter
When I attempt to download a chapter from the mangahere source, it will error out and send me a push notification saying "unexpected url" and the chapter will not load.

Version: 0.6.3
Android 5.1.1

Taumer:
Fix #996

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/online/english/Mangahere.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/online/english/Mangahere.kt
@@ -159,7 +159,7 @@ class Mangahere : ParsedHttpSource() {
 
         val pages = mutableListOf<Page>()
         document.select("select.wid60").first()?.getElementsByTag("option")?.forEach {
-            pages.add(Page(pages.size, it.attr("value")))
+            pages.add(Page(pages.size, "http:" + it.attr("value")))
         }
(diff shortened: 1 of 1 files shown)


samthecodingman:
Wouldn't it be more appropriate to fix this within the [Page](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/source/model/Page.kt) class itself @inorichi?

I'm not familiar enough with Kotlin to perform a PR, but it's probably better to consider this instead:
```kt
class Page(
        val index: Int,
        val url: String = "",
        var imageUrl: String? = null,
        @Transient var uri: Uri? = null
) : ProgressListener {
  init {
    // defaulting to HTTP should be upgraded to HTTPS by servers if needed
    if (url.startsWith('//')) url = 'http:' + url

    // not sure if the same fix has to occur for uri, pretty sure
    // it would have thrown a syntax error without the protocol
  }
  // rest of file
```

This would allow you to fix the '//' prefixed urls that triggered this problem for all providers simultaneously rather than just patching one.

Furthermore, the current fix proposed by @Taumer is volatile in that it will break as soon as MangaHere explicitly declare a protocol again (e.g. url = 'http:' + 'https://some-source...').

PR #1024, opened 2017-10-06
Added option to skip read chapters

### LINKED ISSUE #988: Automatically skip chapters marked as read
Some sources have multiple scanlations, it'd be convenient if after you mark the ones you want to skip as read then it automatically skipped them when reading

NoodleMage:
Closes #988

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
@@ -43,6 +43,8 @@ object PreferenceKeys {
 
     const val readWithVolumeKeysInverted = "reader_volume_keys_inverted"
 
+    const val skipReadChapters = "read_skip_read_chapters"
+
     const val portraitColumns = "pref_library_columns_portrait_key"
 
     const val landscapeColumns = "pref_library_columns_landscape_key"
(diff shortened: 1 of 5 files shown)


inorichi:
The implementation doesn't look right to me. If I understood it right, it requires that duplicate chapters are manually marked as read before opening the reader.

The use case to skip chapters is basically for Batoto, because of the multiple scanlators. So what I propose is a switch to "Restrict to scanlator". However, it could be annoying when some series change from scanlator.

NoodleMage:
It works as follows:
If the selected chapter is read it will just skip to next chapter the normal way. Thus 1 -> 2 -> 3.
If the selected chapter is unread it will skip the read ones. Thus 1 - > 4 -> 5 etc. (2 and 3 are read).

I implemented it like this because it's optional and I wanted to keep it simple (code wise). But you're probably right and it's better to do it by scanlator.

inorichi:
Somehow I thought the issue was for skipping duplicate chapters. Yes, it's exactly what I understood the code was doing, but I'm not sure if the setting is going to be useful enough.

I want to add a secondary drawer (like the one in the library) for filters, sorting and some manga specific switches, like the restrict scanlator, though I'm not sure about this one either...

NoodleMage:
I agree. Will close until we come up with a better implementation.

PR #1051, opened 2017-11-03
Icon changes, vector icons for sdk26+, Dev icon, new theme

CarlosEsco:
Changed Icon to be consistent between SDK16-SDK25 also updated so the icon has some blue like original book icon.

image

Added vectordrawable Adaptive icon for sdk 26+

image
image
image
image
image

Removed circle drawable from navigation drawer switched to vector reference verified it worked in 16,21,26
image

DEV version of icon

image

Teal Theme
image
image
image

Theme option:  
image
image


j2ghz:
Is the dev theme necessary? I like the blue much better.

CarlosEsco:
@j2ghz  was just looking for a way to distinguish when in the app if it was dev or regular.  I think what i can do is make it a new option on theme select in general.  So those who want blue in dev build it stays blue and for those like me can select different color in settings.

j2ghz:
@CarlosEsco If it can be force switched in settings, that would be okay (to have green default).
I also realised it's also a great way to let new users see the difference.

CarlosEsco:
@j2ghz  changed it  be more options under themes see latest screen shots on original post

CarlosEsco:
@inorichi  any thoughts or concerns?

inorichi:
I agree about having a different icon for release/dev builds, but I don't like having 3 themes for every accent color. The best customization would be to allow to select primary the primary and accent color (I saw a cool demo app implementing this), but we are already using a lot of colors (badges, the yellow filter in the toolbar, ...) so I'm not sure if it's feasible/worth.

@LinkCable (author of the current icon) has also made an adaptive icon that's looking better to my eyes: 

screen_shot_2017-11-19_at_10 33 20_pm

I asked him if he can make a dev variant too.

CarlosEsco:
Sounds good I will close this and look into finding a demo app that themed like your mentioning.  I'm sure it was on Reddit at some point.  @LinkCable any ETA on the icons?

CarlosEsco:
@inorichi was is something like this?

CarlosEsco:
https://github.com/garretyoder/Colorful

LinkCable:
@CarlosEsco Yeah, the icons are up in our private dev chat, I'm sure @inorichi will put them in soon.

Issue #1055, opened 2017-11-06
Page number cut off


Echophobia:
When using a newer device such as the s8 or the pixel 2 xl the page number on the bottom is cut off. It's not an important change just visual change.

PR #1056, opened 2017-11-06
page number cut off in rounded screen devices

CarlosEsco:
#1055

Not noticeable on screen shots here's a before and after on actual pixel 2 XL.

before:
image

after:
image


NoodleMage:
Jealous of your phone btw :)

CarlosEsco:
Thanks it's a great one

inorichi:
I didn't like the increased background with more padding. I placed the indicator in the bottom center, added a shadow and removed the background. If people don't like it there, it can be moved to the left edge again, but keeping the shadow.

PR #1074, opened 2017-11-22
Fix automatic backups

### LINKED ISSUE #900: Automatic backup does not seem to be working properly
Version: 0.6.1
Android version: 7.1.2
Device: Google Pixel XL, (also tested on Samsung Galaxy S8+, Google Nexus 7 (2013), Amazon Fire 8 (7th gen))

I have enabled automatic library backups, but no files are being written to the specified backup directory. Not sure how else to troubleshoot the problem. Please let me know if there is any additional information I can provide to help!

NoodleMage:
This closes #900 

~Probably needs some more testing and stuff. But tested working on Android 8.0.0.~

~Fix is using a correct URI call + creating the directory through:~
~`val autoDir = dir.createDirectory(getString(R.string.automatic_backup))`~

~If you have a cleaner solution let me know ;)~

Tested working on: API 19,26 and 27

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupCreateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupCreateService.kt
@@ -8,7 +8,6 @@ import com.github.salomonbrys.kotson.set
 import com.google.gson.JsonArray
 import com.google.gson.JsonObject
 import com.hippo.unifile.UniFile
-import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.data.backup.models.Backup
 import eu.kanade.tachiyomi.data.backup.models.Backup.CATEGORIES
 import eu.kanade.tachiyomi.data.backup.models.Backup.MANGAS
@@ -47,7 +46,7 @@ class BackupCreateService : IntentService(NAME) {
          * Make a backup from library
          *
          * @param context context of application
-         * @param path path of Uri
+         * @param uri path of Uri
          * @param flags determines what to backup
          * @param isJob backup called from job
          */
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
```diff
@@ -161,7 +160,7 @@ class SettingsBackupController : SettingsController() {
             CODE_BACKUP_DIR -> if (data != null && resultCode == Activity.RESULT_OK) {
                 val activity = activity ?: return
                 if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
-                    val uri = Uri.fromFile(File(data.data.path))
+                    val uri = data.data
```
inorichi: This if/else could be simplified with the new change

app/src/main/res/values/strings.xml
```diff
@@ -231,6 +231,7 @@
     <string name="backup_choice">What do you want to backup?</string>
     <string name="restoring_backup">Restoring backup</string>
     <string name="creating_backup">Creating backup</string>
+    <string name="automatic_backup">automatic</string>
```
inorichi: I think this should be a hardcoded string. We don't want translations on this one.

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupCreateService.kt
```diff
@@ -113,20 +114,21 @@ class BackupCreateService : IntentService(NAME) {
         try {
             // When BackupCreatorJob
             if (isJob) {
-                // Get dir of file
+                // Get dir of file and create
                 val dir = UniFile.fromUri(this, uri)
+                val autoDir = dir.createDirectory(getString(R.string.automatic_backup))
```
inorichi: You could keep the previous variable and append `.createDirectory(...)` since `dir` is not being used anywhere else.

Issue #1078, opened 2017-11-23
Cache displayed badges data to speed up library browsing


444nonymous:
I start Tachiyomi in "My Library" by default. (Not cached) it takes 1 second to display all my mangas. With "display download badges" enabled it takes 3 seconds. So far so good. The problem is if I select any manga and then I go back to "My library". I takes much more time, like 5 seconds. Every single time.

Could it be possible to cache the download badges data and only update it when the library is updated (new chapter added / new downloaded chapter / deleted-read chapter)? Instead of reloading it every single time. It's ok if tachiyomi does a full reload when it starts like it does now, just in case mangas / chapters were manually added / removed.

Thanks.

PR #1086, opened 2017-11-26
Download Badge Cache

CarlosEsco:
#1078

Badges should be cached now and whether you delete or download a new chapter in app it will be displayed correctly.
If the chapter is deleted outside of the app it will require you to perform a update library call from the over flow menu, or a swipe to refresh library.

When app leaves memory download badges are recalculated.  Time constraint can be added in future if need be.  

Here is a dropbox for anyone wanting to verify this works for them.

https://www.dropbox.com/s/gpuecwgtfede2y2/tachiyomi-dev.apk?dl=0

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
@@ -5,6 +5,7 @@ import android.webkit.MimeTypeMap
 import com.hippo.unifile.UniFile
 import com.jakewharton.rxrelay.BehaviorRelay
 import com.jakewharton.rxrelay.PublishRelay
+import eu.kanade.tachiyomi.data.database.DatabaseHelper
 import eu.kanade.tachiyomi.data.database.models.Chapter
 import eu.kanade.tachiyomi.data.database.models.Manga
 import eu.kanade.tachiyomi.data.download.model.Download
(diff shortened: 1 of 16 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -323,7 +345,7 @@ class LibraryPresenter(
      */
     fun onOpenManga() {
         // Avoid further db updates for the library when it's not needed
-        librarySubscription?.let { remove(it) }
+      //    librarySubscription?.let { remove(it) }
```
CarlosEsco: @inorichi  This is commented out now, so the library updates correctly when a chapter is deleted in the Manga view.

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -427,6 +434,9 @@ class Downloader(private val context: Context, private val provider: DownloadPro
         // Only rename the directory if it's downloaded.
         if (download.status == Download.DOWNLOADED) {
             tmpDir.renameTo(dirname)
+                 //update download count
+                download.manga.download_count++
+                db.updateDownloadCount(download.manga).executeAsBlocking()
```
CarlosEsco: @inorichi  didnt like this being here in the downloader but the way the classes and methods are set up there is no way to really mar a chapter when downloaded outside of the downloader.

CarlosEsco:
closing for now.

PR #1097, opened 2017-11-29
Restore tracking on backup

### LINKED ISSUE #767: New backup system doesn't store status of synched manga
What I mean is, if I restore the backup file to a completely new tachiyomi it won't restore the status of manga with the synced service. For example if I restore the manga "prison school" KissManga catalogue, it'll show how many chapters I've read but won't show the status is in like "competed, reading"

Version: r1144
Android 7.1.1 (Logcat can be provided)

### LINKED ISSUE #951: Backup/Restore bug with tracking
Version: latest dev version

Hi, 

I think I found a bug with tracking when restoring from backup.
Dunno if this one "works" in MAL or not, but Im using anilist.

Reproduction:

1. add manga to anilist.. it set your Tracking status as reading.
2. if you read manga, it tracking your progress, thats good.
3. make backup
4. restore to new device (clean one, without db)
5. status will disappear from tracking information = no reading status = even if you continue reading new chapters, it doesn't add anything to "chapters" in tracking

EDIT: if you manually change "chaters" count in tracking without reading status = app will crash 

inyrilx

NoodleMage:
Fixes #951 and fixes #767

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt
@@ -23,6 +23,7 @@ import eu.kanade.tachiyomi.data.database.DatabaseHelper
 import eu.kanade.tachiyomi.data.database.models.*
 import eu.kanade.tachiyomi.data.preference.PreferencesHelper
 import eu.kanade.tachiyomi.data.preference.getOrDefault
+import eu.kanade.tachiyomi.data.track.TrackManager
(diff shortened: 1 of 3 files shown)


inorichi:
Looking fine, thanks!

PS: I'm not very fond of the inline methods (in stateless classes like extensions they are fine), but IMO they get harder to read in classes. I know the linter wants you to use them but...

NoodleMage:
In that case I will defer from them in the future. Did it because the functions are one liners and Kotlin is indeed pushing it really hard ;).

PR #1123, opened 2017-12-10
Remove local broadcast receiver to prevent race conditions

NoodleMage:
Added option to set tile for extension update.

Todo: comments


CarlosEsco:
Nice once merged I'll update my pull request for extensions titles

Issue #1079, opened 2017-11-23
Advanced search in globlal search


444nonymous:
Could it be possible to add the individual catalog icon for advanced search also in global search? So, you could also global search by author. 

I understand that global search by genre is also useful for obscure genres with few mangas, or trying to find mangas using a combination of several genres. IE: Drama + Seinen + Sci-Fi + Psychological. Results: "Akira" and similars.

This also relates to the "Clickable authors / artists / genres fields in info tab" option what could open a global search -> https://github.com/inorichi/tachiyomi/issues/1077

Thanks!.

arkon:
This is dependant on the extensions' search functionality.

PR #1139, opened 2017-12-31
Information Page Improvements (click to search, copy to clipboard, etc)

### LINKED ISSUE #860: Clickable author and genre fields in info tab
New feature. it will help find mangas by the same author or genre. It can open a search in catalogs.

### LINKED ISSUE #1098: Latest Chapter Date in Info Page
It would be great if we could see the `date` of the latest chapter that has been updated for a series in it's `info` page. It could go under `last chapter`.

### LINKED ISSUE #1127: Click to copy Manga Name
Hi Guys, *[Feature Request]

Ist possible to implement click to copy manga name from manga page
This would make it easy to search in other source or to just search inside my library.

im aware there is #860 #1077 & #1079 but this NOT click to global search
this request only for "click to copy - manga name"
of course it would be great if "click to copy" can expand to others such as genre, author or artist , until click to global search implemented for genre, author or artist.

Thank You.

### LINKED ISSUE #1141: Full Manga Name
Hi

a lot of manga have long name but we cant find full name anywhere even after we open manga due limited space in title bar (refer screenshot)

since we have alot space in info page, maybe we have full name in info page 

<a href="https://ibb.co/nEod2G">Screenshot_20171231_195006</a>

QuestofIranon:
Opening up this pull request for adding in click to search and a few other info page requests I saw. If I missed any that are currently open let me know and I'll try to slip them in.

Issues:
- [x] Long Click to Copy to clipboard 
Closes #1127 
- [x] Add full title name to info pag
Closes #1141
- [x] Click to search 
Closes #860  
- [x] Last date in info page 
Closes #1098

Other improvements
- [x] Switch to using tags for the genres (ie. https://github.com/2dxgujun/AndroidTagGroup or https://github.com/Cutta/TagView)
- [x] Needs final decision on full title location
- [x] Fixes the title in library/catalogue list view so it truncates instead of running under the count badges


cd0rw0:
Hi @QuestofIranon ,
referring to my #1141 , maybe you include together in this.
you can have name in info page has click to global search. and put small copy icon at end of name for copy manga name.

QuestofIranon:
@cd0rw0 I could add that.

QuestofIranon:
This is how the info page currently looks. I replaced the "Description" label with the Title as a possible solution for #1141 
image

QuestofIranon:
Is it better to show "Unkown" for labels where data doesn't exist, or attempt to hide the labels?

LinkCable:
The wording of Date here is a little bit confusing, I would change it to "Last Updated" or something of the sort.

re: "unknown" it depends: what labels are coming up missing?

QuestofIranon:
image

image

image

This does seem to work better, I think adding more emphasis to the title might be good though. Right now I have it set to max out at two lines for the title.

LinkCable:
What about making it black text instead?

Also the date label and I think showing the label with Unknown is best.

QuestofIranon:
image

An issue I noticed with having the title up there is that if the name goes over 2 lines, and the genre list goes over 2 lines they get covered and the user must scroll down to see them.

CarlosEsco:
What if you put the manga title above the cover.  So It spans both columns

QuestofIranon:
@CarlosEsco I thought about that, but I thought it would look odd with the actual page title above it as well.

CarlosEsco:
You could put it above make it a little larger make it black and then remove it from the the page title

CarlosEsco:
Or just make it so that if you touch the page title it shows a toast of the full title and copies it

QuestofIranon:
I added in click to search. 

I tried a few different behaviors when opening the search but this way seemed the best. When you click on any information on the title page it will open the search. 

Pressing back from the search goes back to the info page, right now you can create a deep navigation stack by continuously doing searches though.

QuestofIranon:
image

Spacing and maybe the font style could use some work but I think this looks nice.

QuestofIranon:
I fixed this so that the text is Ellipsized properly, previously it would run under the floating text.
image

I think this looks better as an improvement to the title (and it will also avoid going under the favorites button)
image

QuestofIranon:
image

I forgot to change the horizontal layout, I left title on the top here since it would look awkward between genres and description.

I haven't pushed this change yet.

LinkCable:
> I also noticed that if the name goes over 2 lines and the genre list goes over 2 lines they get covered and the user must scroll down to see them. You can see this in one of the screenshots above.

I think this is acceptable behavior as there are already manga with loads of genres tagged on them that can cause this overflow to happen. 

I liked this option of yours best.

screen shot 2018-01-01 at 8 59 21 pm

I also made a mockup with the title below, we can just drop the description label as it's implicit and then make the text the title and boost the font size up to 14pt so that it's not the same size as the label descriptors. But would ideally rather go with the above screenshot, as it makes sense to have the title first in a list of all these descriptors of the series. 

manga info

I do like your latest landscape view. I think just making the text black will do. 

Also changing "latest" to "updated" I think is a clearer label.

LinkCable:
To free up space in the table view, we could also move Genres to using the tag library suggested in the original issue #860  (https://github.com/2dxgujun/AndroidTagGroup)

That makes them much easier to tap, and there's lots of space to use in the description view.

Might look something like this

tags

QuestofIranon:
Right now it just auto searches the entire string of genres. I definitely want to fix that... and I totally forgot about recommending that library in the past, I'll try working with it.

Another thing I was thinking about, would it be better to have everything on a single page instead of seperate to tabs? I was looking around in different content provider apps and almost all of them do single page info + chapters/episodes now. Going with concensus isn't always best but it might with some of the awkwardness here.

LinkCable:
>Another thing I was thinking about, would it be better to have everything on a single page instead of seperate to tabs? I was looking around in different content provider apps and almost all of them do single page info + chapters/episodes now. Going with concensus isn't always best but it might with some of the awkwardness here.

Do you mean having the chapter list further below the Info page?

If that's what you meant, in our case the top right buttons change on the context of the tab, and we can't really shoehorn in all those features into the info page (sorting, filtering, etc)

QuestofIranon:
> Do you mean having the chapter list further below the Info page?

That's what I meant, but I hadn't considered those features being an issue. Doing such a change would probably require reworking how the chapter lists are. I think it'd be doable though. 

One option might be adding those options once the user scrolls down to the chapter list, and removing them if you're at the top. It might get annoying to use if you just want to view chapters though, but it could still start at the chapter list if you're opening from the library (then you just scroll up a bit to view details).

LinkCable:
>One option might be adding those options once the user scrolls down to the chapter list, and removing them if you're at the top. It might get annoying to use if you just want to view chapters though, but it could still start at the chapter list if you're opening from the library (then you just scroll up a bit to view details).

We can think about this, but not in the scope of this PR I'd say.

QuestofIranon:
Right, I wasn't planning to add it in this one, I just wanted to throw it out as an idea (I probably should've used the Discord). I think it'd require more research.

QuestofIranon:
image

image

Looks pretty good, I went with this library instead as it is more recently maintained and has less issues reported:
https://github.com/Cutta/TagView

I'll change the colors and tag shapes, I just wanted to get it set up first. I think on portrait I'll probably move genres above the Description and maybe remove the genres label altogether.

Horizontal is a bit more questionable, I might put them under the title without the label? I haven't tested making them smaller yet.

LinkCable:
>I'll change the colors and tag shapes, I just wanted to get it set up first. I think on portrait I'll probably move genres above the Description and maybe remove the genres label altogether.

>Horizontal is a bit more questionable, I might put them under the title without the label? I haven't tested making them smaller yet.

Keep Genres below the Description, as it provides more context over the genres as to what the manga is about. Also I wouldn't have them all the way at the bottom like on your screenshot, just the standard margin between the Genre header and the description paragraph.

You can probably boost the font-size of the title on portrait by 2px or so. 

And portrait and landscape should keep the same order / content. Especially since they are both essentially a list of info about the manga the list shouldn't change order just based on orientation.

Also, that was quick!

QuestofIranon:
image
image

@LinkCable Here are some screenshots with the genres as I proposed.

QuestofIranon:
I didn't like putting them under the description in landscape because then the genres only appear if you scroll down, or I'd have to make Description a small scroll view.

Also, any color recommendations?

QuestofIranon:
I think they look good under the description too though, but without the label. Here's a quick and dirty mockup ->
screenshot

LinkCable:
>I didn't like putting them under the description in landscape because then the genres only appear if you scroll down, or I'd have to make Description a small scroll view.

That's okay, the priority is to have the manga's unique information.

The problem with having the tags upfront is that it encourages the user to click them and navigate away from the manga they are looking at, rather than it coming at the end and instead providing similar suggestions of manga by browsing that tag. By placing if before everything else/the description, it's telling readers that the genres define the manga more than it's individual story.

Also it looks like you changed the margins a bit on portrait mode, it should be the same as it is currently. (notice how the description is now getting but off by the Bookmark button) For the style, use the blue we use for buttons (#2979FF) and have an outline instead of solid as they are secondary descriptors. If you could boost the border-radius to be more pill-like that'd be great too. I used 10dp border in my mockup.

QuestofIranon:
The margins were just bad because I did a mock up with ms paint in 2 minutes...
I switched to that other library we originally discussed, it's actually less limiting.

I applied your suggestions. My main concern was trying to have information digestable from a single view, I think the layout is fine like this though with maybe a few minor tweaks.

image
image

LinkCable:
Yeah, this looks great, thanks! I would just add a little more bottom padding to the tabs so the words don't get so close to the bottom. If not I can do it after this is merged in.

Also looks like Travis is failing, so maybe some new tests will have to be made to accommodate for the change in layout.

@inorichi @j2ghz (not sure who handles Travis)

QuestofIranon:
I think the tests are failing because I modified the API version in gradle and forgot to flip it back before pushing. I'll change it when I get home and push it.

The padding looked off to me but I wasn't sure if it was just me. Tanks for confirming that!

QuestofIranon:
It looks like I'm going to need to modify the tag library itself to enable the ability to modify the bottom padding to get text centered properly.

QuestofIranon:
Just a heads up, I may not be available much for the next week. This should be ready to go though other than the padding issue.

QuestofIranon:
The next step I think would be to add this
#1079 
I missed that there was an issue open for it and had already intended to go back and add in this feature, so I'll probably do it next, I just wasn't sure if it would fall under the scope of this PR.

LinkCable:
Yeah I would say tackle that in a new PR as it addresses an entirely different screen/user flow.

LinkCable:
@QuestofIranon I would get rid of the WIP tag so we know it's ready to merge, then ping one of the maintainers

LinkCable:
@inorichi @NoodleMage if any of you guys want to do a code review

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/CatalogueController.kt
```diff
@@ -32,7 +32,7 @@ import uy.kohesive.injekt.api.get
  * [CatalogueAdapter.OnBrowseClickListener] call function data on browse item click.
  * [CatalogueAdapter.OnLatestClickListener] call function data on latest item click
  */
-class CatalogueController : NucleusController<CataloguePresenter>(),
+class CatalogueController() : NucleusController<CataloguePresenter>(),
```
inorichi: Change not needed

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/CatalogueController.kt
```diff
@@ -99,6 +99,7 @@ class CatalogueController : NucleusController<CataloguePresenter>(),
         recycler.layoutManager = LinearLayoutManager(view.context)
         recycler.adapter = adapter
         recycler.addItemDecoration(SourceDividerItemDecoration(view.context))
+
```
inorichi: Change not needed

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/CatalogueController.kt
```diff
@@ -185,10 +186,11 @@ class CatalogueController : NucleusController<CataloguePresenter>(),
         // Create query listener which opens the global search view.
         searchView.queryTextChangeEvents()
                 .filter { it.isSubmitted }
-                .subscribeUntilDestroy {
-                    val query = it.queryText().toString()
-                    router.pushController(CatalogueSearchController(query).withFadeTransaction())
-                }
+                .subscribeUntilDestroy { performGlobalSearch(it.queryText().toString())}
```
inorichi: Add an space before `}`

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersPresenter.kt
```diff
@@ -91,6 +93,10 @@ class ChaptersPresenter(
                     // Emit the number of chapters to the info tab.
                     chapterCountRelay.call(chapters.maxBy { it.chapter_number }?.chapter_number
                             ?: 0f)
+
+                    // Emit the upload date of the most recent chapter
+                    lastUpdateRelay.call( Date(chapters.maxBy { it.date_upload }?.date_upload ?: 0))
```
inorichi: Remove extra space before `Date`

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
```diff
@@ -79,6 +87,20 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
 
         // Set SwipeRefresh to refresh manga data.
         swipe_refresh.refreshes().subscribeUntilDestroy { fetchMangaFromSource() }
+
+        manga_full_title.longClicks().subscribeUntilDestroy{ copyToClipboard(view.context.getString(R.string.title), manga_full_title.text.toString()) }
```
inorichi: These lines are too long (at least the ones calling `copyToClipboard`). Introduce line breaks after `{` and before `}`. Also fix missing spaces before `{`

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -143,6 +144,12 @@ class MangaController : RxController, TabbedController {
         textView.compoundDrawablePadding = if (visible) 4 else 0
     }
 
+    fun performGlobalSearch(query: String){
```
inorichi: This method should go inside `MangaInfoController` instead.

Use `val router = parentController?.router ?: return` there.

Fix blank lines here.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
```diff
@@ -79,6 +87,20 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
...
+        manga_full_title.clicks().subscribeUntilDestroy{ performSearch(manga_full_title.text.toString()) }
+        manga_artist.longClicks().subscribeUntilDestroy { copyToClipboard(manga_artist_label.text.toString(), manga_artist.text.toString()) }
+        manga_artist.clicks().subscribeUntilDestroy{ performSearch(manga_artist.text.toString()) }
+        manga_author.longClicks().subscribeUntilDestroy { copyToClipboard(manga_author.text.toString(), manga_author.text.toString()) }
+        manga_author.clicks().subscribeUntilDestroy{ performSearch(manga_author.text.toString()) }
+        manga_summary.longClicks().subscribeUntilDestroy { copyToClipboard(view.context.getString(R.string.description), manga_summary.text.toString()) }
+
+        //todo: make an rxbinding extension for this?
+        manga_genres_tags.setOnTagClickListener { tag -> performSearch(tag) }
```
inorichi: Or call `manga_genres_tags.setOnTagClickListener(null)` in `onDestroyView` (before super).

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
```diff
@@ -122,30 +145,40 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
     private fun setMangaInfo(manga: Manga, source: Source?) {
         val view = view ?: return
 
+        //update full title TextView.
+        manga_full_title.text = when(manga.title.isBlank()){
```
inorichi: Use if/else instead: http://kotlinlang.org/docs/reference/coding-conventions.html#if-versus-when

Also use an space after `when`, `if`, `for`, `while`... and before `{`

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
```diff
@@ -122,30 +145,40 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
...
+        manga_source.text = when(source == null){
+            false -> source.toString()
+            true -> view.context.getString(R.string.unknown)
         }
 
-        // Update genres TextView.
-        manga_genres.text = manga.genre
-
-        // Update status TextView.
-        manga_status.setText(when (manga.status) {
```
inorichi: Is this now unsupported?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
```diff
@@ -278,6 +315,13 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
         swipe_refresh?.isRefreshing = value
     }
 
+    /**
+     * Perform global search
+     * @param query string that is searched
+     */
+    private fun performSearch(query: String) =
```
inorichi: As said before, `val router = parentController?.router ?: return`, then push the controller.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
```diff
@@ -377,6 +421,21 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
...
+     * @param label Label to show to the user describing the content
+     * @param content the actual text to copy to the board
+     */
+    private fun copyToClipboard(label: String, content: String){
+        if(content.isBlank()) return
+
+        val clipboard = activity?.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+        clipboard.primaryClip = ClipData.newPlainText(label, content)
+
+        activity?.toast("Copied ${content.truncateCenter(20)} to clipboard!", Toast.LENGTH_SHORT)
```
inorichi: This should be a resource. Also, shouldn't the label be enough? cc @LinkCable

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
```diff
@@ -377,6 +421,21 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
...
     }
 
+    /**
+     * Copies a string to clipboard
+     *
+     * @param label Label to show to the user describing the content
+     * @param content the actual text to copy to the board
+     */
+    private fun copyToClipboard(label: String, content: String){
+        if(content.isBlank()) return
```
inorichi: Since you'll be using activity a lot here, you can do `val activity = activity ?: return` at the beginning.

app/src/main/java/eu/kanade/tachiyomi/util/StringExtensions.kt
```diff
@@ -11,3 +13,17 @@ fun String.chop(count: Int, replacement: String = "..."): String {
         this
 
 }
+
+/**
+ * Replaces the given string to have at most [count] characters using [replacement] near the center.
+ * If [replacement] is longer than [count] an exception will be thrown when `length > count`.
+ * Is not exact
```
inorichi: "Is not exact"? Is this incomplete or can it be removed?

app/src/main/res/layout/manga_info_controller.xml
```diff
@@ -197,7 +232,7 @@
                     app:layout_constraintBaseline_toBaselineOf="@+id/manga_source_label"
                     app:layout_constraintLeft_toRightOf="@+id/manga_source_label"
                     app:layout_constraintRight_toRightOf="parent"/>
-
+                <!--
```
inorichi: Just remove the block

app/src/main/res/layout/manga_info_controller.xml
```diff
@@ -217,25 +252,52 @@
                     app:layout_constraintTop_toBottomOf="@+id/manga_genres_label"
                     app:layout_constraintLeft_toLeftOf="parent"
                     app:layout_constraintRight_toRightOf="parent"/>
+                -->
 
             </android.support.constraint.ConstraintLayout>
 
         </android.support.v4.widget.NestedScrollView>
 
+        <!--
```
inorichi: Remove them

app/src/main/res/layout/manga_info_controller.xml
```diff
@@ -217,25 +252,52 @@
...
             android:layout_marginStart="8dp"
-            android:layout_marginTop="16dp"
-            app:layout_constraintBottom_toBottomOf="parent"
+            android:layout_marginTop="8dp"
             app:layout_constraintLeft_toLeftOf="parent"
             app:layout_constraintRight_toRightOf="parent"
-            app:layout_constraintTop_toTopOf="@+id/guideline">
+            app:layout_constraintTop_toBottomOf="@+id/guideline">
+
+            <!--app:layout_constraintTop_toBottomOf="@id/manga_full_title_layout">-->
```
inorichi: Same here

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
LinkCable: I think including the fact that it copied is better, as otherwise you might just assume it's showing you the full name of the manga without really knowing why.

inorichi: I mean, instead of `Copied <manga title> to clipboard!`, something like `Title/Author/Whatever copied to clipboard!`. Shorter and faster to read, especially useful since the toasts disappear quickly.

LinkCable: Ohh that's what you meant. Yes I agree. Also don't think we need the exclamation mark, but that's more a nitpick

inorichi: I don't particularly like it either. Just kept it because it was in the PR.

LinkCable: Yeah, but since @QuestofIranon is gonna go in and change the text anyway, mind taking out the
 exclamation mark? 😛

app/src/main/java/eu/kanade/tachiyomi/util/StringExtensions.kt
QuestofIranon: I don't think it's needed.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
QuestofIranon: It's replaced by
`
        // Update genres list
        if(manga.genre.isNullOrBlank().not()){
            manga_genres_tags.setTags(manga.genre?.split(", "))
        }
`
on line 172

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
QuestofIranon: I added it after the content is blank check, so this isn't automatically detecting that there was a change.

QuestofIranon:
@inorichi I have big fingers and use an Xperia Z5 compact, I was concerned about the ability to select specific fields too. However, I haven't really noticed it being a problem while testing.

I left the specific field value instead of the label for the copied to clipboard text. I think it makes more sense to show the specific text being copied, but if @inorichi or @LinkCable still disagree I don't mind changing it.

LinkCable:
@QuestofIranon ...I actually misinterpreted what Inorichi said. Yeah, keep the text that is actually being copied in the toast, you right on this one.

QuestofIranon:
@LinkCable Yeah I was just using an app earlier today that had the hold down to copy function for some text and it just toasted with "Copied to clipboard". I was kind of irked by it because of the ambiguity.

LinkCable:
@QuestofIranon 

> I thought it'd be interesting to maybe have it copied to the user's image folder perhaps though, but it seemed like that would be unintuitive.

Yeah I would not just stick images into user's folders without permission. I think copying the title as well is acceptable behavior.

LinkCable:
@inorichi want to review @QuestofIranon's comments? hoping we can get this merged in soon

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
inorichi: I meant the status. It's never updated now.

PR #1148, opened 2018-01-06
Fix padding on RecyclerViews

null-dev:
As you can see below, when scrolling, the RecyclerViews are clipping to the padding, bringing my OCD great distress.

Left = Before
Right = After
I have circled the points of interest in red.

triggered

Notice that outrageous black line that cuts off the manga covers. It brings me great pain and I request to immediately abolish it through this PR!

*On a more serious note:* If that's actually the intended behavior, then feel free to close this PR.


null-dev:
Oops, messed up the git history, one sec...

e-shl:
And corrected the bottom?
img_20180106_100329

null-dev:
Yep!

Issue #1122, opened 2017-12-09
Backing out of manga during a library search will cancel the search


null-dev:
**Version:** v0.6.6
**Steps to reproduce:**
1. In "My library", tap on the search icon in the top right icon and search for any manga in your library
2. Open the manga by tapping on it
3. Press the back button
4. Notice that your search is cancelled

inorichi:
This is already fixed but I forgot to close it.

PR #1150, opened 2018-01-06
Fix lost library search

fvcaputo:
Adresses https://github.com/inorichi/tachiyomi/issues/1122

Saves the current search on the presenter when a manga is opened, return the saved search when back to the library view.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
@@ -322,6 +322,11 @@ class LibraryController(
         val searchItem = menu.findItem(R.id.action_search)
         val searchView = searchItem.actionView as SearchView
 
+        if(!presenter.query.isEmpty()){
+            query = presenter.query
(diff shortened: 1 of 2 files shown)


inorichi:
I don't like the way the presenter is used to temporarily store the query. You could just use a local variable in the controller to do just that. The subscribe until detach looks weird but maybe it's not so weird since the library view is retained now.

I'll test it and come back later. Thanks

fvcaputo:
@inorichi Indeed, seems like I was overcomplicating things by saving the query on the presenter, it could be saved on the controller. The subscribe until detach is what keeps the query from being cleared as the manga opens.

inorichi:
I ended up using a more robust approach (https://github.com/inorichi/tachiyomi/commit/1292c0eceac913532c093784f33fa42754ce6560) by ignoring events when the controller is not at the top. For example you could also open the edit categories screen while you had an active query, so you'd have to save the `backupQuery` there too.

Also fixed subscribing many times to the search view, as the options menu doesn't follow a controller's lifecycle.

Issue #1154, opened 2018-01-08
Change Add to Library icon and add Toast Message feedback


LinkCable:
Currently, the process of adding a manga to your library for new users is confusing due to the icon and lack of feedback given when pressing the Bookmark button.

Two suggested changes to alleviate this:

- Change the Add to Library icon to include a plus to be more straightforward with its function
- Add a toast message "Added to your library" for immediate feedback on what happens when tapping

![add_to_library](https://user-images.githubusercontent.com/620506/34656520-54024408-f3e9-11e7-82da-c2946aba450d.png)

Possible svgs include the one I made:
https://cdn.discordapp.com/attachments/350729100526092289/397565694696357888/add_to_library.svg

or this one from the Material Design Icon site we've been referencing lately:
https://materialdesignicons.com/icon/bookmark-plus-outline

EDIT: Icons in the screenshot should be outlined, not filled. My b. Will update the svgs to match.

LinkCable:
Closed in PR https://github.com/inorichi/tachiyomi/pull/1160

PR #1160, opened 2018-01-08
Address #1154

CarlosEsco:
Address fixes requested in #1154 also adds the toast message when long click manga in the catalog view.
image
image
image
image
image
image


LinkCable:
I would avoid using "Manga", since Webtoons are present on the platofrm and we also support ReadComicsOnline which is just normal Western comics. 

We could either use just "Added to your library" / "Removed from your library", or use the title of what you're adding to your library, but that could get hairy with long titles that those LN adaptations are well known for.

CarlosEsco:
Good idea will remove manga from string

CarlosEsco:
Good idea will remove manga from string

PR #1170, opened 2018-01-09
Handle ActivityNotFoundException

Flat:
When using `Intent.ACTION_CREATE_DOCUMENT` on SDK >= Lollipop there is no
guarantee that the ROM supports the built in file picker such as MIUI

A discord user had this issue which caused Tachiyomi to crash when trying to start a backup
activitynotfoundexception

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
@@ -117,19 +117,25 @@ class SettingsBackupController : SettingsController() {
 
                 onClick {
                     val currentDir = preferences.backupsDirectory().getOrDefault()
+                    val customPicker = Intent(activity, CustomLayoutPickerActivity::class.java)
+                            .putExtra(FilePickerActivity.EXTRA_ALLOW_MULTIPLE, false)
+                            .putExtra(FilePickerActivity.EXTRA_ALLOW_CREATE_DIR, true)
+                            .putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
+                            .putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
 
                     val intent = if (Build.VERSION.SDK_INT < 21) {
                         // Custom dir selected, open directory selector
-                        val i = Intent(activity, CustomLayoutPickerActivity::class.java)
-                        i.putExtra(FilePickerActivity.EXTRA_ALLOW_MULTIPLE, false)
-                        i.putExtra(FilePickerActivity.EXTRA_ALLOW_CREATE_DIR, true)
-                        i.putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
-                        i.putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
-
+                        customPicker
                     } else {
                         Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
                     }
-                    startActivityForResult(intent, CODE_BACKUP_DIR)
+                    //Fall back to custom picker on error
+                    try{
+                        startActivityForResult(intent, CODE_BACKUP_DIR)
+                    } catch (e: ActivityNotFoundException){
+                        startActivityForResult(customPicker, CODE_BACKUP_DIR)
+                    }
+
                 }
 
                 preferences.backupsDirectory().asObservable()
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
@@ -27,10 +27,7 @@ import eu.kanade.tachiyomi.data.preference.getOrDefault
 import eu.kanade.tachiyomi.ui.base.controller.DialogController
 import eu.kanade.tachiyomi.ui.base.controller.popControllerWithTag
 import eu.kanade.tachiyomi.ui.base.controller.requestPermissionsSafe
-import eu.kanade.tachiyomi.util.getUriCompat
-import eu.kanade.tachiyomi.util.registerLocalReceiver
-import eu.kanade.tachiyomi.util.toast
-import eu.kanade.tachiyomi.util.unregisterLocalReceiver
+import eu.kanade.tachiyomi.util.*
 import eu.kanade.tachiyomi.widget.CustomLayoutPickerActivity
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
@@ -117,23 +114,18 @@ class SettingsBackupController : SettingsController() {
 
                 onClick {
                     val currentDir = preferences.backupsDirectory().getOrDefault()
-                    val customPicker = Intent(activity, CustomLayoutPickerActivity::class.java)
-                            .putExtra(FilePickerActivity.EXTRA_ALLOW_MULTIPLE, false)
-                            .putExtra(FilePickerActivity.EXTRA_ALLOW_CREATE_DIR, true)
-                            .putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
-                            .putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
-
-                    val intent = if (Build.VERSION.SDK_INT < 21) {
-                        // Custom dir selected, open directory selector
-                        customPicker
-                    } else {
-                        Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
-                    }
-                    //Fall back to custom picker on error
                     try{
+                        val intent = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+                        // Custom dir selected, open directory selector
+                        preferences.context.getFilePicker(currentDir)
+                        } else {
+                          Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
+                        }
+
                         startActivityForResult(intent, CODE_BACKUP_DIR)
                     } catch (e: ActivityNotFoundException){
-                        startActivityForResult(customPicker, CODE_BACKUP_DIR)
+                        //Fall back to custom picker on error
+                        startActivityForResult(preferences.context.getFilePicker(currentDir), CODE_BACKUP_DIR)
                     }
 
                 }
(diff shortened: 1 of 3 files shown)


inorichi:
If I remember correctly, the documents provider is still there, but disabled by default.  On later versions they found out they were creating many apps and discarded their own implementation. If we merge this PR we also break SD card saving on Lollipop.

Maybe not a big deal, but I'd double check if he could actually enable the documents provider since this is a Xiaomi only issue

inorichi:
Errr.. I misread the diff. Quirks of reading on mobile. Somehow I though you changed the API check.

Then I guess it's ok to merge.

Flat:
Missed the import for the exception on accident so just pushed that. It should still use the same checks for API. Only added an exception handler that on catch uses the built-in picker instead as a backup.

Flat:
Found two other intents that made use of the Android documents, handled them in a similar fashion.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
```diff
@@ -116,19 +117,25 @@ class SettingsBackupController : SettingsController() {
 
                 onClick {
                     val currentDir = preferences.backupsDirectory().getOrDefault()
+                    val customPicker = Intent(activity, CustomLayoutPickerActivity::class.java)
+                            .putExtra(FilePickerActivity.EXTRA_ALLOW_MULTIPLE, false)
+                            .putExtra(FilePickerActivity.EXTRA_ALLOW_CREATE_DIR, true)
+                            .putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
+                            .putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
 
                     val intent = if (Build.VERSION.SDK_INT < 21) {
```
NoodleMage: val intent = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {

no clue why this wasn't already done like this xD

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadController.kt
```diff
@@ -150,17 +151,22 @@ class SettingsDownloadController : SettingsController() {
...
-            i.putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
-            i.putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
+        val customPicker = Intent(activity, CustomLayoutPickerActivity::class.java)
+                .putExtra(FilePickerActivity.EXTRA_ALLOW_MULTIPLE, false)
+                .putExtra(FilePickerActivity.EXTRA_ALLOW_CREATE_DIR, true)
+                .putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
+                .putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
 
-            startActivityForResult(i, DOWNLOAD_DIR_PRE_L)
+        if (Build.VERSION.SDK_INT < 21) {
```
NoodleMage: if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
```diff
@@ -204,25 +211,33 @@ class SettingsBackupController : SettingsController() {
...
-        // If API lower as KitKat use custom dir picker
-        val intent = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
-            // Get dirs
-            val preferences: PreferencesHelper = Injekt.get()
-            val currentDir = preferences.backupsDirectory().getOrDefault()
-
-            Intent(activity, CustomLayoutPickerActivity::class.java)
+        // Setup custom file picker intent
+        // Get dirs
+        val preferences: PreferencesHelper = Injekt.get()
```
NoodleMage: Can be removed.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
```diff
@@ -116,19 +117,25 @@ class SettingsBackupController : SettingsController() {
...
-                        i.putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
-                        i.putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
-
+                        customPicker
                     } else {
                         Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
                     }
-                    startActivityForResult(intent, CODE_BACKUP_DIR)
+                    //Fall back to custom picker on error
+                    try{
```
NoodleMage: Move try to top. Thus:

```
try {
                        val intent = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
                            // Custom dir selected, open directory selector
                            customPicker
                        } else {
                            Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
                        }
                        startActivityForResult(intent, CODE_BACKUP_DIR)
                    } catch (e: ActivityNotFoundException) {
                        //Fall back to custom picker on error
                        startActivityForResult(customPicker, CODE_BACKUP_DIR)
                    }
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
```diff
@@ -204,25 +211,33 @@ class SettingsBackupController : SettingsController() {
...
         } else {
             // Use Androids build in file creator
             Intent(Intent.ACTION_CREATE_DOCUMENT)
                     .addCategory(Intent.CATEGORY_OPENABLE)
                     .setType("application/*")
                     .putExtra(Intent.EXTRA_TITLE, Backup.getDefaultFilename())
         }
-        startActivityForResult(intent, CODE_BACKUP_CREATE)
+        // Handle errors where the android ROM doesn't support the built in picker
+        try {
```
NoodleMage: Move try to top. Thus:

```
        try {
            // If API is lower than Lollipop use custom picker
            val intent = if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
                customPicker
            } else {
                // Use Androids build in file creator
                Intent(Intent.ACTION_CREATE_DOCUMENT)
                        .addCategory(Intent.CATEGORY_OPENABLE)
                        .setType("application/*")
                        .putExtra(Intent.EXTRA_TITLE, Backup.getDefaultFilename())
            }
            startActivityForResult(intent, CODE_BACKUP_CREATE)
        } catch (e: ActivityNotFoundException) {
            // Handle errors where the android ROM doesn't support the built in picker
            startActivityForResult(customPicker, CODE_BACKUP_CREATE)
        }
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
```diff
@@ -116,19 +117,25 @@ class SettingsBackupController : SettingsController() {
 
                 onClick {
                     val currentDir = preferences.backupsDirectory().getOrDefault()
+                    val customPicker = Intent(activity, CustomLayoutPickerActivity::class.java)
```
NoodleMage: I don't really like that this val is created three times. Perhaps you can create a Extension like e.g.
```
fun Context.getFilePicker(currentDir : String) : Intent{
    return  Intent(this, CustomLayoutPickerActivity::class.java)
            .putExtra(FilePickerActivity.EXTRA_ALLOW_MULTIPLE, false)
            .putExtra(FilePickerActivity.EXTRA_ALLOW_CREATE_DIR, true)
            .putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
            .putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
}
```


}

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadController.kt
```diff
@@ -150,17 +151,22 @@ class SettingsDownloadController : SettingsController() {
...
+            startActivityForResult(customPicker, DOWNLOAD_DIR_PRE_L)
         } else {
             val i = Intent(Intent.ACTION_OPEN_DOCUMENT_TREE)
```
NoodleMage: Not your code.  But change val i to val intent?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
Flat: I agree, I'm not a huge fan of the duplication as well. I wasn't sure the best way to break this out, will likely do what you've suggested.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsBackupController.kt
Flat: Is there a specific reason for try  to be at the top of that block? Right now it will only catch `ActivityNotFoundException` which is thrown by `startActivityForResult`. Are there other exceptions we need to catch or is it just for style?

NoodleMage: 1) Mostly style. It looks cleaner in my opinion.
2) Perhaps in the future more exceptions needs to be added which can obfuscate the code. But this is also style issue, I think... :)

Flat: 👌Was just curious on the reasoning. Will make these changes today.

Flat:
Should be all requested changes there, let me know if I missed anything

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadController.kt
```diff
@@ -151,20 +152,15 @@ class SettingsDownloadController : SettingsController() {
...
-                .putExtra(FilePickerActivity.EXTRA_ALLOW_MULTIPLE, false)
-                .putExtra(FilePickerActivity.EXTRA_ALLOW_CREATE_DIR, true)
-                .putExtra(FilePickerActivity.EXTRA_MODE, FilePickerActivity.MODE_DIR)
-                .putExtra(FilePickerActivity.EXTRA_START_PATH, currentDir)
-
-        if (Build.VERSION.SDK_INT < 21) {
-            startActivityForResult(customPicker, DOWNLOAD_DIR_PRE_L)
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+            startActivityForResult(preferences.context.getFilePicker(currentDir), DOWNLOAD_DIR_PRE_L)
```
NoodleMage: Still create a val for this. Otherwise you have to create the Intent twice.

So:

```
val customPicker = context.getFilePicker(currentDir)
...
startActivityForResult(customPicker), DOWNLOAD_DIR_PRE_L)
```

Flat:
Added version checks so the `Intent` is never added more than once to the heap, and if an exception isn't thrown and the phone is using the document picker it is never added at all.

Issue #1158, opened 2018-01-08
Improve Tracking Search results


LinkCable:
This one is definitely most up for discussion, but we've had issues (#1102, #422) regarding the results of the MAL search (and likely other trackers) being less-than-ideal for picking the correct manga to sync.

These changes depend on what sort of data we can get from the API, and how much dev time we want to devote, and also what other devs/users think.

1. Show the manga with it's cover page, title, author, and first lines or so of the description to make a more informed decision of what you're picking to sync
2. Just recycle the list view we already use for Catalogues, which at least adds the manga image

![mal](https://user-images.githubusercontent.com/620506/34656850-7391e806-f3ed-11e7-8cad-710158e1416d.png)

Again these are just initial designs and not necessarily final.

Flat:
Anilist and Kitsu provide fields for cover images through their API. According to their documentation MAL does not.

LinkCable:
>According to their documentation MAL[3] does not.

Dang not even the author, stingy.

PR #1178, opened 2018-01-11
Improve tracking search results

CarlosEsco:
Addresses #1158 

Also added click to logo to launch the tracking websites url for the manga

Needs some minor ui adjustments as well as contemplating click to show all of summary?

image

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
@@ -1,5 +1,6 @@
 package eu.kanade.tachiyomi.data.database.models
 
+import eu.kanade.tachiyomi.data.track.model.TrackSearch
 import java.io.Serializable
 
 interface Track : Serializable {
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
@@ -22,6 +22,8 @@ interface Track : Serializable {
 
     var status: Int
 
+    var tracking_url: String
+
     fun copyPersonalFrom(other: Track) {
         last_chapter_read = other.last_chapter_read
         score = other.score
@@ -33,6 +35,11 @@ interface Track : Serializable {
         fun create(serviceId: Int): Track = TrackImpl().apply {
             sync_id = serviceId
         }
+
+        fun createTrackSearch(serviceId: Int): TrackSearch = TrackSearch().apply {
+            sync_id = serviceId
+        }
+
     }
 
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
@@ -1,6 +1,5 @@
 package eu.kanade.tachiyomi.data.database.models
 
-import eu.kanade.tachiyomi.data.track.model.TrackSearch
 import java.io.Serializable
 
 interface Track : Serializable {
@@ -32,15 +31,9 @@ interface Track : Serializable {
     }
 
     companion object {
-
         fun create(serviceId: Int): Track = TrackImpl().apply {
             sync_id = serviceId
         }
-
-        fun createTrackSearch(serviceId: Int): TrackSearch = TrackSearch().apply {
-            sync_id = serviceId
-        }
-
     }
 
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
@@ -22,6 +22,8 @@ object TrackTable {
 
     const val COL_TOTAL_CHAPTERS = "total_chapters"
 
+    const val COL_TRACKING_URL = "url"
+
     val createTableQuery: String
         get() = """CREATE TABLE $TABLE(
             $COL_ID INTEGER NOT NULL PRIMARY KEY,
@@ -33,9 +35,12 @@ object TrackTable {
             $COL_TOTAL_CHAPTERS INTEGER NOT NULL,
             $COL_STATUS INTEGER NOT NULL,
             $COL_SCORE FLOAT NOT NULL,
+            $COL_TRACKING_URL TEXT NOT NULL,
             UNIQUE ($COL_MANGA_ID, $COL_SYNC_ID) ON CONFLICT REPLACE,
             FOREIGN KEY($COL_MANGA_ID) REFERENCES ${MangaTable.TABLE} (${MangaTable.COL_ID})
             ON DELETE CASCADE
             )"""
 
+    val addTrackingUrl: String
+        get() = "ALTER TABLE $TABLE ADD COLUMN $COL_TRACKING_URL TEXT DEFAULT ''"
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
@@ -22,7 +22,7 @@ object TrackTable {
 
     const val COL_TOTAL_CHAPTERS = "total_chapters"
 
-    const val COL_TRACKING_URL = "url"
+    const val COL_TRACKING_URL = "remote_url"
 
     val createTableQuery: String
         get() = """CREATE TABLE $TABLE(
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.graphics.Color
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.data.database.models.Track
+import eu.kanade.tachiyomi.data.database.models.TrackSearch
 import eu.kanade.tachiyomi.data.preference.getOrDefault
 import eu.kanade.tachiyomi.data.track.TrackService
 import rx.Completable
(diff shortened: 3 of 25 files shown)


CarlosEsco:
@linkcable let me know your thoughts.  I for sure want to remove the grey scroll bar that shows when browsing.  I repurposed the recently read card for the list view.

LinkCable:
@CarlosEsco just want to lyk I've seen this, just haven't had the time to sit down and mock something up.

I can attach my original mockup as a basis, but I like how you've included more of the description.

mock

sharedkestrel:
If I can suggest something, instead of  a short snippet of the plot, I think the pusblishing year and the status (ongoing, completed, etc.) can be better for picking the right manga.

sharedkestrel:
If I can suggest something, instead of  a short snippet of the plot, I think the pusblishing year and the status (ongoing, completed, etc.) can be better for picking the right manga.

sharedkestrel:
If I can suggest something, instead of  a short snippet of the plot, I think the pusblishing year and the status (ongoing, completed, etc.) can be better for picking the right manga.

wopian:
For Kitsu and AniList, the subtype (Manga, Novel etc) would also be useful. They usually have the same synopsis so would still be  hard to differentiate 👍

CarlosEsco:
Would have to double check but think tachiyomi is ignoring none manga results from tracker

inorichi:
Yes, some results like novels are filtered out.

LinkCable:
@CarlosEsco can you add a latest screenshot? Will work off of that to make design suggestions.

CarlosEsco:
@LinkCable 
.

CarlosEsco:
@LinkCable if any of the fields returned back from tracker are null the field is removed for the display see Brilliant Magic below.

.

LinkCable:
So... I really wanted to figure out a way for this not to involve card on cards (as it's bad practice), but couldn't come up with a good way to intuitively have a manga in the list be intuitively selectable and also show it's selected status. Here's one concept based on [Material Design selection](https://material.io/guidelines/patterns/selection.html#selection-item-selection)

mal v5

But I think for now it's okay to use the card approach, I will revisit this whole Tracking flow once we start the rewrite mentioned in Dev. Here's a mock for that:

mal v3 copy_card

CarlosEsco:
@LinkCable  none of the trackers have author provided on their search 😞

CarlosEsco:
@LinkCable here is a standard list view with layout tweaks in your sample.  This would work if you dont want card on card. 
.
.

LinkCable:
@NoodleMage Really, no author? Dang.

I think it's ok to keep it just "Finished" btw, no need for the publishing. Unless that's what's returned from the search API?

And as we discussed, let's keep it card on card. They are just Material "guidelines" after all.

CarlosEsco:
@LinkCable the status is whatever the tracker returns

CarlosEsco:
let me know if you want anything else tweaked 

.

LinkCable:
@CarlosEsco I would add the divider line back above the Ok/Cancel options.

Also move the first card up a bit closer to the search term, right now there's a bit too much space.

Is it possible to chop off the 00-00, like write a quick check stating if the month is not 1-12 just display the year?

CarlosEsco:
image

LinkCable:
It looks like the search bar/line is hitting the edge, I would place it in line with the cards width, and also move a card about 1/2 of the space still available above it. (~20dp)

CarlosEsco:
image

LinkCable:
@CarlosEsco perf 😊

NoodleMage:
Looks 👌. 

Do we need to include the TrackingURL in the backup or is the current backup information enough to retrieve this url?

CarlosEsco:
Technically we don't.  It pops a toast up when tracking url is missing and asks to remove and re add that tracker.  This was because people that use that tracker updating to this version would have all urls missing

LinkCable:
@CarlosEsco good to merge?

CarlosEsco:
Yes

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
```diff
@@ -33,6 +35,11 @@ interface Track : Serializable {
         fun create(serviceId: Int): Track = TrackImpl().apply {
             sync_id = serviceId
         }
+
+        fun createTrackSearch(serviceId: Int): TrackSearch = TrackSearch().apply {
```
inorichi: Avoid referencing the search model in the DB model. I'd move `TrachSearch` to `data/track/model` instead.

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
```diff
@@ -49,31 +50,31 @@ class Anilist(private val context: Context, id: Int) : TrackService(id) {
 
     override fun getScoreList(): List<String> {
         return when (preferences.anilistScoreType().getOrDefault()) {
-            // 10 point
+        // 10 point
```
inorichi: Damn I hate AS enforcing this alignment

app/src/main/java/eu/kanade/tachiyomi/data/track/kitsu/KitsuApi.kt
```diff
@@ -27,25 +28,25 @@ class KitsuApi(private val client: OkHttpClient, interceptor: KitsuInterceptor)
         return Observable.defer {
             // @formatter:off
             val data = jsonObject(
-                "type" to "libraryEntries",
```
inorichi: Even turning off the formatter, it moved everything. Sad :(

I can't wait for the day I change all the spacing to 2

app/src/main/java/eu/kanade/tachiyomi/data/track/kitsu/KitsuModels.kt
```diff
@@ -10,19 +10,29 @@ open class KitsuManga(obj: JsonObject) {
...
+        publishing_status = this@KitsuManga.status
+        publishing_type = type
+        start_date = startDate.orEmpty()
     }
 }
 
 class KitsuLibManga(obj: JsonObject, manga: JsonObject) : KitsuManga(manga) {
     val remoteId by obj.byInt("id")
-    val status by obj["attributes"].byString
+    override val status by obj["attributes"].byString
```
inorichi: Why is the override needed?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackController.kt
```diff
@@ -79,6 +81,10 @@ class TrackController : NucleusController<TrackPresenter>(),
         swipe_refresh?.isRefreshing = false
         activity?.toast(error.message)
     }
+    override fun OnLogoClick(position: Int) {
```
inorichi: First word lower case + empty line before the method

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackPresenter.kt
```diff
@@ -80,6 +84,14 @@ class TrackPresenter(
                         TrackController::onSearchResultsError)
     }
 
+    fun openTrackerUrl(item: Track, service: TrackService) {
+        if(item.tracking_url.isNullOrBlank()){
+            context.toast(R.string.url_not_set)
+        }else{
+        context.startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(item.tracking_url)))
```
inorichi: Never use the presenter to handle UI stuff. Either make the UI ask the presenter for the data, or use one of the delivery methods (DeliverFirst) if the method requires some IO or is slow (it's not in this case).

app/src/main/res/values/strings.xml
```diff
@@ -465,4 +465,10 @@
     <string name="channel_library">Library</string>
     <string name="channel_downloader">Downloader</string>
 
+    <string name="url_not_set">Manga url is not set please click title and select manga again</string>
```
inorichi: Move these to the screen where they are used

app/src/main/java/eu/kanade/tachiyomi/data/track/kitsu/KitsuModels.kt
CarlosEsco: not needed anymore so reverted

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
inorichi: Move this method to the `TrackSearch` class.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackController.kt
```diff
@@ -80,6 +84,17 @@ class TrackController : NucleusController<TrackPresenter>(),
         activity?.toast(error.message)
     }
 
+    override fun onLogoClick(position: Int) {
+        val item = adapter?.getItem(position) ?: return
```
inorichi: ` val track = adapter?.getItem(position)?.track ?: return` so `?.let` isn't needed.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
inorichi: This one is still not fixed

app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
```diff
@@ -22,6 +22,8 @@ object TrackTable {
 
     const val COL_TOTAL_CHAPTERS = "total_chapters"
 
+    const val COL_TRACKING_URL = "url"
```
inorichi: I'd call it "remote_url", because we already have "remote_id" and it helps a bit when trying to understand the columns. Also I'd change the string from "url" to "remote_url" too.

I want to have unique column names, to avoid issues when using JOINs statements, because you have to alias columns with the same names (or bugs will bite you).

CarlosEsco: fixed this and removed the createTrackSearch.  I had already removed all references just forgot to remove the method.

PR #1185, opened 2018-01-12
Added custom download option

NoodleMage:
Because there aren't enough pull request yet.

The extra option looks bloated though. So thinking about removing one of the options, not sure which one.

screenshot_1516653320_framed

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
@@ -210,7 +210,7 @@ class ChaptersController : NucleusController<ChaptersPresenter>(),
         }
     }
 
-    fun fetchChaptersFromSource() {
+    private fun fetchChaptersFromSource() {
         swipe_refresh?.isRefreshing = true
         presenter.fetchChaptersFromSource()
     }
@@ -272,18 +272,18 @@ class ChaptersController : NucleusController<ChaptersPresenter>(),
         actionMode?.invalidate()
     }
 
-    fun getSelectedChapters(): List<ChapterItem> {
+    private fun getSelectedChapters(): List<ChapterItem> {
         val adapter = adapter ?: return emptyList()
         return adapter.selectedPositions.mapNotNull { adapter.getItem(it) }
     }
 
-    fun createActionModeIfNeeded() {
+    private fun createActionModeIfNeeded() {
         if (actionMode == null) {
             actionMode = (activity as? AppCompatActivity)?.startSupportActionMode(this)
         }
     }
 
-    fun destroyActionModeIfNeeded() {
+    private fun destroyActionModeIfNeeded() {
         actionMode?.finish()
     }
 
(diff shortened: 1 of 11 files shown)


j2ghz:
Maybe if you remove the repeating `download`?

Flat:
Maybe remove the the `Download next 5` and `Download next 10` and instead make the dialog have a decrement by 5 button on the left and an increment by 5 button on the right with the current amount in the middle - still editable by keyboard

inorichi:
As @j2ghz pointed out, the repeating `Download` should be removed (only kept in the title).

I think a increasing/decreasing button in intervals of 5 would look weird but I'm not against it. cc @LinkCable you might have a better idea or some design in mind.

LinkCable:
Off the top of my head and looking around, what about a slider that moves in steps of 5?

Do you think people would put specific amounts (like 27, or 12, basically anything not a multiple of 5)

LinkCable:
Or perhaps something like this could work too: 
dialog

cd0rw0:
Hi guys 

how about something like below, i seen in some app but forgot which one. 
Double arrow add/reduce by 10, single arrow add/reduce by 5
if really want custom, click middle box and key in desired no...

Just a MS paint work :-) ,forgive if look ugly.

<a href="https://imgbb.com/">aaa</a>

e-shl:
And if so

and the maximum number depends on the number of unloaded chapters

e-shl:
And if so
and the maximum number depends on the number of unloaded chapters

e-shl:
And if so
and the maximum number depends on the number of unloaded chapters

e-shl:
And if so
and the maximum number depends on the number of unloaded chapters

e-shl:
If so
and the maximum number depends on the number of unloaded chapters

e-shl:
If so
and the maximum number depends on the number of unloaded chapters

NoodleMage:
Though I appreciate the suggestions, you guys are aware that instead of simply filling in a number your suggestions are:

- One weird ass slider which almost looks like my traffic bag lock xD.
- A clickable chapter increaser without any confirmation option (Though I know this is paint and can be improved, the downside is that it needs a custom view which will probably look ugly..
- A slider, for selecting a very specific number :)

I will look into clickable increaser. But I still don't really see the problem with just typing in the desired amount.

LinkCable:
@NoodleMage it's just that normally a dialog shouldn't have a keyboard present IMO, just touch events. But we're already breaking that paradigm in Tracking, so ¯\_(ツ)_/¯

NoodleMage:
@linkCable. Looks awesome will try to implement. Thanks

LinkCable:
@NoodleMage let me know if you need the svg's, I just used the material arrows, but had to make my own version of the "double" left and rights.

NoodleMage:
Svg would be appreciated :)

LinkCable:
https://materialdesignicons.com/icon/chevron-right
https://materialdesignicons.com/icon/chevron-left

Github doesn't support svg upload, so here's a zip

[double-arrows.zip](https://github.com/inorichi/tachiyomi/files/1639998/double-arrows.zip)

NoodleMage:
Perhaps centering looks better?

If someone has some time somewhere. Quick code review is appreciated :)

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
```diff
@@ -439,29 +441,45 @@ class ChaptersController : NucleusController<ChaptersPresenter>(),
...
         }
     }
 
+    private fun showCustomDownloadDialog() {
+        DownloadCustomChaptersDialog(this, presenter.chapters.size).showDialog(router)
+    }
+
+
+    override fun downloadChapters(choice: Int) {
+        if (choice == 3) {
```
inorichi: Include this in the `when` and use `return` to avoid calling `downloadChapters` later.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/DownloadCustomChaptersDialog.kt
```diff
@@ -0,0 +1,57 @@
...
+import eu.kanade.tachiyomi.ui.base.controller.DialogController
+import eu.kanade.tachiyomi.widget.DialogCustomDownloadView
+
+/**
+ * Dialog used to let user select amount of chapters to download.
+ */
+class DownloadCustomChaptersDialog<T>(bundle: Bundle? = null) : DialogController(bundle)
+        where T : Controller, T : DownloadCustomChaptersDialog.Listener {
+
+    private var maxChapters = 0
```
inorichi: Probably not a big deal but this will be reset to 0 on process death (not saved in bundle).

app/src/main/java/eu/kanade/tachiyomi/widget/DialogCustomDownloadView.kt
```diff
@@ -0,0 +1,108 @@
...
+    var amount: Int = 0
+        private set
+
+    /**
+     * Minimal value of custom download chooser.
+     */
+    private var min = 0
+
+    /**
+     *
```
inorichi: 🤔

inorichi:
Centered definitely looks better (comparing with @LinkCable's sketch).

PR #1191, opened 2018-01-17
Added select all above and below functions

thisismaky:
There are times when I really want to be able to select a number of chapters that I've already read, but I can't select all since I haven't finished the manga yet. With these 2 new options, you'll be able to easily select all the chapters above or below a given chapter. I think this would be really useful to others to save time (in case they want to select 200+ out of 300 chapters for example).

screenshot_20180116-233950
screenshot_20180116-233956


e-shl:
Selects all the chapters, even those that were read?

Deterio:
I'd still like to see this implemented.

AmineI:
Is it still relevant with #1641 ? Wouldn't range select somehow remove the need of "select above/below", or not quite ?

thisismaky:
If #1641 works as it sounds then yes, that would make this feature obsolete. Seeing as this pull request was about a year ago, surprised mine or range select hasn't been implemented yet. This makes selecting and downloading specific chapters much easier.

arkon:
I think we'll just go with range selection (in some form).

PR #1198, opened 2018-01-20
Fixes and Tweaks to Info Page

### LINKED ISSUE #1194: Full Name Error
A few examples of full name errors from the latest commit r1321.

screenshot_2018-01-19-00-25-58-216_eu kanade tachiyomi debug 01
screenshot_2018-01-19-00-26-10-651_eu kanade tachiyomi debug 01

### LINKED ISSUE #1196: Genre tags bug
Latest dev
Descriptions that are too long force genre tags to be cut off instead of scrolling the field.
image

This source has issues with tags being just one 
image

QuestofIranon:
Addresses issues related to pr #1139 as well as some other minor tweaks.

Todo:
- [x] fixes #1194 
- [ ] fixes #1196 (partially fixed by #1215)
- [ ] Switch field searches to a source filter search as discussed in https://github.com/inorichi/tachiyomi/issues/1196#issuecomment-359329958

### DIFF
--- a/app/src/main/res/layout-land/manga_info_controller.xml
+++ b/app/src/main/res/layout-land/manga_info_controller.xml
@@ -65,7 +65,6 @@
                     style="@style/TextAppearance.Medium.Title"
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
-                    android:ellipsize="end"
                     android:maxLines="2"
                     android:textIsSelectable="false"
                     app:layout_constraintTop_toTopOf="parent"
--- a/app/src/main/res/layout-land/manga_info_controller.xml
+++ b/app/src/main/res/layout-land/manga_info_controller.xml
@@ -69,7 +69,11 @@
                     android:maxLines="2"
                     android:textIsSelectable="false"
                     app:layout_constraintTop_toTopOf="parent"
-                    app:layout_constraintLeft_toLeftOf="parent"/>
+                    app:layout_constraintLeft_toLeftOf="parent"
+                    app:autoSizeTextType="uniform"
+                    app:autoSizeMinTextSize="12sp"
+                    app:autoSizeMaxTextSize="100sp"
+                    app:autoSizeStepGranularity="2sp"/>
 
                 <TextView
                     android:id="@+id/manga_author_label"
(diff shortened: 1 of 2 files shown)


QuestofIranon:
With the auto size text view to fix #1194 

image

QuestofIranon:
Also this really needs more testing on different screens. I couldn't seem to accurately reproduce the problem and the sizes still seem inconsistent.

Sparh4wk:
is it possible to add genres to the top right under source? its kinda pain to look down just for them.. I picking up mangas by genres only..

LinkCable:
@Sparh4wk they need to be separated from the view so they are clickable to search. You can always filter by genre in the sources if you want to look for specific genres.

Sparh4wk:
@LinkCable well its not the same.. I usually searching for manga, that are being scanlated via catalogue and latest... there you cant filter it by genre or can you?

CarlosEsco:
@Sparh4wk you can't filter in latest cause 99% of sources have a non filterable latest page.  If you want to filter by genre if the source supports click browse and you will have filter options.

Sparh4wk:
@CarlosEsco ye, I know, thats why I wanted to genres be at top like it used to be, cause its much better when you picking mangas from latest...

CarlosEsco:
@sparh4wk searchable genre tags overcrowd the top.  I don't know what device you are using but a quick glance bottom shouldn't change use pattern that much

QuestofIranon:
What if I put the genre tags above the description but as a single row that scrolls left/right?

CarlosEsco:
I like that idea @LinkCable thoughts?

LinkCable:
>I like that idea @LinkCable thoughts?

Too little discoverability (for any genres off screen), still small tap targets as they need to fit in the table now, and now you're getting really close to the bookmark button.

QuestofIranon:
I think I'm going to leave the genres at the bottom under the description, but make it where they won't overflow off the screen (at least in portrait mode). In this case the scrollable view for the description will be nested between the top half and the genres.

QuestofIranon:
just wanted to reference #1201 here

CarlosEsco:
@QuestofIranon dont worry about #1196 I will fix the source

QuestofIranon:
Ok, currently all other sources provide the genres as csv. I've pretty much fixed the style issue though.

inorichi:
Should we merge this as it's and leave the others for another release? I'd like to push an update soon.

inorichi:
Merging the fix for the first issue

Issue #1157, opened 2018-01-08
Change visuals of Library Filters for better UX


LinkCable:
Currently, the Library Filter menu differs visually from the Categories Filter menu, when they should really both use the same design.

Changes include:

- Using the same collapsable headers as Category Filter uses
- Fixing the bug of the collapsable sections to allow them to actually work (mentioned in #1155)
- Changing the sort/order icon to use Up and Down arrows instead of reusing the dropdown arrows  (mentioned in #1155)

![filter_library](https://user-images.githubusercontent.com/620506/34656667-539d3cc8-f3eb-11e7-847f-a374ab514cc0.png)

These changes will also make it easier to fit in filtering by tracking as mentioned in #1147 as you could just collapse to the tracking section.

CarlosEsco:
This is gonna be a little harder then the other one as it requires extensive rewrites.  I will look into more next week

LinkCable:
Thanks for tackling these @CarlosEsco!

Issue #1147, opened 2018-01-05
Visual Reference to Tracking Service in "My Library"


lordbk:
Dear All, [feature Request]

i would like to request simple visual reference in "my library" for synced tracking service. 
i did simple MS mock up of my visualization of this request, maybe others can come up with better ideas.
i did this way to make at as simple as as possible without cluttering cover. 
Hope any dev/contributor can consider this request.

Colour Reference ( as per their logo colour)
Blue - MAL
Red - Kitsu
white & Blue - Anilist

<a href="https://ibb.co/nH3RBb"><img src="https://preview.ibb.co/mjTtrb/Screenshot_20180105_100738_1.jpg" alt="Screenshot_20180105_100738_1" border="0"></a>

zunjae:
In general I'd advice against color coding functionality due to UX reasons.

wopian:
I don't see this working for the list view either. 

Would be better to be able to filter your library by trackers (None, AniList, Kitsu and MAL)

cd0rw0:
hi @zunjae  @wopian 
this is just an idea...
List view - I dont see reason why it wont work in list view. it can have 3 dot next to unread badge.
filter by tracker (4 option) will make filter option long. i dont think dev will agree to that.

LinkCable:
Hmm, off the top of my head I would think using the favicons would be good, but would probably get too low res. Another option could be shapes (triangle, circle, square) with the colors, but it'd still be iffy how to communicate to the user which means which.

j2ghz:
What is the use case for this? I can only think of going through all your manga that e.g. doesn't have a binding to MAL, to bind them. For that filters would be much better I think.

cd0rw0:
@j2ghz im not against filter option. what i mean was, filter list will become too long if have 3/4 option as per above suggestion. (AniList, Kitsu and MAL)

LinkCable:
>@j2ghz im not against filter option. what i mean was, filter list will become too long if have 3/4 option as per above suggestion. (AniList, Kitsu and MAL)

I don't think it will get too long, I mean if you look at the filters for various catalogues there are so many genres and it's still fine.

Also we plan to make the filters in your library collapsable by section like the filters in Catalogues, so you could just collapse all until you get to the "tracking" section or something.

inorichi:
I don't see any good reason to have this visual reference in the library.

as280093:
[...]

PR #1211, opened 2018-01-24
Library filter UI change

CarlosEsco:
#1157 

@LinkCable  styled it same as Catalog is expandable/collapsible necessary?  If so will have to roll our own custom solution for this area. 
image


LinkCable:
@CarlosEsco I would say it would be preferable, both so it would match the Categories filter list and so that we could incorporate stuff like https://github.com/inorichi/tachiyomi/issues/1147 without lots of scrolling down.

If it's a hassle though I don't wanna put more work on you.

CarlosEsco:
I think for now just having similar visuals will help.  Adding  expand/collapse can be done later

inorichi:
Please fix the commented code, unless they are there for a reason, but I believe they were left there while testing.

CarlosEsco:
@inorichi removed them.  I was waiting on LinkCables thoughts of expandable/collapsible.  Also should i remove DrawerSwipeCloseListener?  Its not used anymore anywhere.

LinkCable:
@CarlosEsco I'd like it, but like you said we can add it later.

Flat:
I think for consistency this filter menu should also be openable by sliding it out from the right, the same way the menu works for catalogues.

CarlosEsco:
It does now

Flat:
Ah I missed that. Cheers

LinkCable:
> It does now

Does it interfere with swiping across categories at all in your testing?

CarlosEsco:
Just checked and no like the catalog swipe the start has to be right on the edge to open drawer.   The category swipe can be anywhere else

LinkCable:
@inorichi @NoodleMage code review/merge?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
```diff
@@ -177,11 +176,8 @@ class LibraryController(
 
     override fun createSecondaryDrawer(drawer: DrawerLayout): ViewGroup {
         val view = drawer.inflate(R.layout.library_drawer) as LibraryNavigationView
-        drawerListener = DrawerSwipeCloseListener(drawer, view).also {
-            drawer.addDrawerListener(it)
-        }
         navView = view
-        drawer.setDrawerLockMode(DrawerLayout.LOCK_MODE_LOCKED_CLOSED, Gravity.END)
+        drawer.setDrawerLockMode(DrawerLayout.LOCK_MODE_UNLOCKED, Gravity.END)
```
NoodleMage: Keep this LOCK_MODE_LOCKED_CLOSED. 
LOCK_MODE_UNLOCKED gives can give weird behavior when switching categories.

app/src/main/java/eu/kanade/tachiyomi/widget/ExtendedNavigationView.kt
```diff
@@ -193,7 +192,7 @@ open class ExtendedNavigationView @JvmOverloads constructor(
         @CallSuper
         override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): Holder {
             return when (viewType) {
-                VIEW_TYPE_HEADER -> HeaderHolder(parent)
+                VIEW_TYPE_HEADER -> HeaderHolder(parent, onClick)
```
NoodleMage: What's the reason for onClick?.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
CarlosEsco: So locking it you lose the consistency between this and the filter drawer in catalog.   I verified this and it only slides the drawer out when your right near the edge of the screen.

app/src/main/java/eu/kanade/tachiyomi/widget/ExtendedNavigationView.kt
CarlosEsco: not needed anymore

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
NoodleMage: I will merge with LOCK_MODE_UNLOCKED. If it turns out it bothers people we revert it back to LOCK_MODE_LOCK_UNLOCKED, good with you?

NoodleMage:
```:app:compileStandardDebugKotline: /home/travis/build/inorichi/tachiyomi/app/src/main/java/eu/kanade/tachiyomi/widget/ExtendedNavigationView.kt: (209, 28): Unresolved reference: title```

You broke something :)

CarlosEsco:
fixed

Issue #1202, opened 2018-01-21
[Feature request] E-reader Page-up/down button support


Nelthorim:
Simple request: Android e-readers exist, and this app works amazingly in them. However, it would be nice to be able to use the page-up/down buttons available on them to pass the pages in addition of swiping the e-ink screen, which isn't as responsive as it should be.

I'm not sure how hard this would be to implement, but didn't see it suggested here yet, so I'll bring it to your attention.

CarlosEsco:
What ereader do you use?

Nelthorim:
Energy System Pro HD. Most of the other ebook apps work with the side buttons, so I was hoping this one would too :(

CarlosEsco:
Are you using dev or stable?

Nelthorim:
I've tried with both the stable and dev version. Unfortunately, that .apk gives error to install.

If it's useful, I'll attach the output of KeyEvent for those keys.

[keyevent_2018-01-24-205848+0000.txt](https://github.com/inorichi/tachiyomi/files/1661660/keyevent_2018-01-24-205848.0000.txt)

PR #1212, opened 2018-01-24
Add page down/page up hardware detection

CarlosEsco:
#1202

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
@@ -206,18 +206,6 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
                         return true
                     }
                 }
-                KeyEvent.KEYCODE_PAGE_DOWN -> {
-                    if (event.action == KeyEvent.ACTION_UP) {
-                        if (!volumeKeysInverted) viewer?.moveDown() else viewer?.moveUp()
-                    }
-                    return true
-                }
-                KeyEvent.KEYCODE_PAGE_UP -> {
-                    if (event.action == KeyEvent.ACTION_UP) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -206,6 +206,18 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
                         return true
                     }
                 }
+                KeyEvent.KEYCODE_PAGE_DOWN -> {
+                    if (event.action == KeyEvent.ACTION_UP) {
+                        if (!volumeKeysInverted) viewer?.moveDown() else viewer?.moveUp()
```
inorichi: I think this should be in `onKeyUp` a few lines below. Volume keys had to be handled here to not trigger the system's volume too. Does it make sense to use the inverted volume keys preference too (just asking)?

CarlosEsco: Will remove and if in future somebody wants add a new invert page property.  Moved to onKeyUp

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -414,16 +416,16 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
 
     private fun setRotation(rotation: Int) {
         when (rotation) {
-            // Rotation free
+        // Rotation free
```
inorichi: Can you revert these changes?

CarlosEsco: done

CarlosEsco:
should be good now

PR #1217, opened 2018-01-28
Hide Latest button in catalog when source doesnt support latest

CarlosEsco:
Currently clicking latest pops up toast message of whatever the writer of the source put in the exception.  This hides it instead
image


NoodleMage:
Nice one 👍

Issue #1220, opened 2018-01-29
Enlarge the 3-dot icon click space in chapter lists


paulo27ms:
I know this has been modified in the dev version already (I have r1335 right now) but the space you have to click the icon is simply far too small. With this change the icon is also now vertical instead of horizontal so there's effectively no change as far as I can see outside of it not looking as bad.
I have a 720x1280 screen and it's pretty evenly split between opening the chapter, jumping to another location with the scroll bar and actually opening the menu when I try.
Please either make the space to click it significantly bigger or add an option to disable the scroll bar.
Thank you.

paulo27ms:
I'm not sure I follow. To the right? That brings up the scroll bar and moves me somewhere else.
I'm asking to make it bigger, like, double its current size added to the left.
It does not work like this for me if this was the final product:

![35238868-d3f150d4-ff7c-11e7-95dc-941544631d67](https://user-images.githubusercontent.com/13546823/35538140-b36c4f62-0544-11e8-9869-d59b3c40182b.png)

I am noticing that it looks quite a bit different on my end though:

![screenshot_2018_01_29_22_33_01](https://user-images.githubusercontent.com/13546823/35538174-cfd6b87c-0544-11e8-91b8-f8b2d21b5edb.png)

CarlosEsco:
Sorry meant left and that's weird.  Is that tablet can you try on phone? Mine is spaced out

paulo27ms:
This is an Android phone. Wiko CINK FIVE, 5'', 4.1.2 Jelly Bean Android, 720p resolution. Not sure what else could be interfering with this.
Landscape mode doesn't change anything. 
I'll have access to a tablet tomorrow and I'll try to install Tachiyomi on it to see.

Sparh4wk:
I think after update in dev version, its ok that way..

CarlosEsco:
@Sparh4wk  It works for you right?  What version of android you using?
@paulo27ms What version of dev

CarlosEsco:
I found issue

paulo27ms:
I see. I'll wait for an update then. Thank you.

PR #1221, opened 2018-01-29
Adjust 3 dot chapter menu to support Android 16

CarlosEsco:
Start/End not supported for android 16. #1220

### DIFF
--- a/app/src/main/res/layout/chapters_item.xml
+++ b/app/src/main/res/layout/chapters_item.xml
@@ -68,6 +68,9 @@
         android:paddingStart="24dp"
         android:paddingEnd="16dp"
         android:layout_marginEnd="8dp"
+        android:paddingLeft="24dp"
+        android:paddingRight="16dp"
+        android:layout_marginRight="8dp"
         android:paddingBottom="24dp"
         android:paddingTop="12dp"
         android:contentDescription="@string/description_cover"/>


LinkCable:
Hey @CarlosEsco can you revert the list overflow menu back to horizontal, while keeping the same touch area?

screen shot 2018-01-29 at 10 15 29 pm

This looks very odd from a visual design perspective, there's no sense of hierarchy with both the header and list items using the same overflow menu.

CarlosEsco:
So that image isn't quite right the vertical is more left however I changed it from horizontal cause that's what @inorichi requested

LinkCable:
Hmm, I'd say Horizontal is better in this context, both to have a hierarchy under the top Overflow menu and that the content in the list is all horizontal as it's text, so a horizontal overflow menu fits in with the reading flow.

wopian:
Tachiyomi is the only app I've seen use the horizontal overflow in list views (or at all) with the vertical still being used in the header, so this change is more consistent with the general usage

CarlosEsco:
This is what v17+ looks like 
image

LinkCable:
Hmm, I want to revisit this cell layout at some point, as there is really big empty space between the chapter name and the date/download/page info. 

But for this PR it's not necessary.

PR #1241, opened 2018-02-08
Update README.md

Flat:
Include CONTRIBUTING.md guidelines as collapseable elements.
Remove white background from screenshots
Optimize app-icon.png and screens.png using `optipng -o2`


wopian:
Wouldn't a link to CONTRIBUTING.md be better instead of duplicating the content?

Flat:
My hope is that people will actually read it if they don't have to navigate away from the page.

PR #1257, opened 2018-02-17
Chapter Metadata update

CarlosEsco:
change chapter update to refresh on any metadata change
namely the two missing were date uploaded and chapter number

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/util/ChapterSourceSync.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/util/ChapterSourceSync.kt
@@ -52,7 +52,7 @@ fun syncChaptersWithSource(db: DatabaseHelper,
         } else {
             //this forces metadata update for the main viewable things in the chapter list
             ChapterRecognition.parseChapterNumber(sourceChapter, manga)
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/util/ChapterSourceSync.kt
```diff
@@ -49,13 +49,16 @@ fun syncChaptersWithSource(db: DatabaseHelper,
...
-                dbChapter.name != sourceChapter.name) {
-
-            dbChapter.scanlator = sourceChapter.scanlator
-            dbChapter.name = sourceChapter.name
-
-            toChange.add(dbChapter)
+        } else {
+            //this forces metadata update for the main viewable things in the chapter list
+            ChapterRecognition.parseChapterNumber(sourceChapter, manga)
+            if (dbChapter.scanlator != sourceChapter.scanlator || dbChapter.name != sourceChapter.name || dbChapter.date_upload != sourceChapter.date_upload || dbChapter.chapter_number != sourceChapter.chapter_number) {
```
inorichi: This line is too long. I'd create a private method in this file (also using line breaks after 100 lines).

PR #1262, opened 2018-02-19
allow copyFrom to copy url from manga

CarlosEsco:
@inorichi 
Edit:  I might close this as i need to find another solution since this breaks restore.  Unless your fine with SMangaImpl being defaulted to blank instead of lateinit

this allows updating of the manga url when the source sets it.
 99% of time no source will set it.  I have to set it for Mangadex since they changed urls at some point from
/manga/153/

to 

/manga/153/detective-conan

the old url works, but when they added paging of chapters 
/manga/153/detective-conan/200 works
/manga/153/200 does not

so all the old urls need updated on the metadata refresh


CarlosEsco:
closing for now supposedly gonna be a json response soon from mangadex so this might be irrelevant

Issue #1201, opened 2018-01-21
[Feature Request] Adding "SEE ALL" in the global search result per source.


Skcyte:
As the clickable tag/author is being develop, the sideway scrolling is a bit troublesome when the results are way too many. Adding see all option per source would make the results are easy to browse. Basically it's the same page where in the browse catalogue mode with the filter in it.

CarlosEsco:
Maybe something as simple as clicking the source name relaunches same search on just that source

Skcyte:
That would be fine as well. I'd more concern on tag/author one since you basically can do name search on just one source right.

arkon:
Closed by #3265

PR #1264, opened 2018-02-19
More option in catalog

NoodleMage:
This should implement #1201 . Problem is that the pager generates a new page which means the source gets added to stack. Thus:

Globals search > Catalog > search Catalog

When you go back it puts the catalog on the stack which means the user needs two back clicks to go from catalog search to global search. 

@inorichi @CarlosEsco idea how to fix this behavior perhaps?


CarlosEsco:
Haven't looked too much in it but what if when you global search you take more than 10 from a source, but  only show ten in global view.  Then upon clicking it loads a new page which is just the recyclerview of the results instead of relaunching the catalog.

inorichi:
I was thinking the same. You can either create a new controller or use the same one with a `ViewState`/`render` approach (see the `MigrationController`, which has two RV adapters).

The first one is easier to create, but it has the problem of restoring state after process death (honestly I wish Conductor had a feature to not persist certain controllers in the backstack), and the second one has the problem of creating a nice `render` method that handles everything properly.

NoodleMage:
Gonna focus on new release

Issue #1265, opened 2018-02-20
Can't restore backup from main app to dev


briggsi:
Using 0.6.8 and r1347 with extensions for Mangadex, Dynasty, and MerakiScans (both installed from the FDroid repository and via the Dev extensions option, which have different numbered versions).

After making a backup with the main app (with all options checked), I tried restoring it to the dev app in order to use some of the newer features.  It fails without completing the first manga with this error:

![screenshot_20180219-221856](https://user-images.githubusercontent.com/8586304/36407987-1f1f8da8-15c8-11e8-8c1a-da1e1bbf64c1.png)

The first manga it runs into is from Dynasty, and it still happens after removing that one such that the first is from Kissmanga.

Last note while testing, unchecking "History" and "Tracking" from the create backup options, allows the dev version to continue on the backup, but everything from Mangadex and Dynasty return "Source not found" in the log.

briggsi:
And now after uninstalling the dev version, FDroid won't let me download for some reason, so I downloaded r1357 from github.  It has all the extensions still installed, and the backup is currently going but very slowly (as expected from Mangadex right now) with the original backup with "History" and "Tracking".  I'll close since the problem seems to have cleared up.

briggsi:
And now that it's finished doing it, I noticed I was not signed into MAL and all the tracking failed.  Trying to use the backup again proceeded to fail the same way as previously stated.  And after uninstalling and reinstalling r1357, it fails to use the backup whenever I am signed into MAL.

CarlosEsco:
@briggsi 
Try this build let me know 

https://www.dropbox.com/s/bv4u4fikf0ohzid/tackiyomi-restore-fix.apk?dl=0

briggsi:
@CarlosEsco

I'm getting a "package appears to be corrupt" error when installing that .apk.

CarlosEsco:
[app-dev-debug.zip](https://github.com/inorichi/tachiyomi/files/1740700/app-dev-debug.zip)

CarlosEsco:
[app-dev-debug.zip](https://github.com/inorichi/tachiyomi/files/1740904/app-dev-debug.zip)

Sparh4wk:
well I have the same problem, but both on dev.. I have backup like from version 1345-1350 and now I cant restore it to 1353 for some reason. have the same error as OP.

Flat:
There is a fix pending merge. #1269

PR #1269, opened 2018-02-21
fix restore from old backup to updated trackimpl.

CarlosEsco:
addresses #1265.
Fixes restores from old backup into new updated trackimpl with tracking url.
adds tracking url to backup like i should have done originally @NoodleMage  😄

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/TrackTypeAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/TrackTypeAdapter.kt
@@ -14,6 +14,7 @@ object TrackTypeAdapter {
     private const val REMOTE = "r"
     private const val TITLE = "t"
     private const val LAST_READ = "l"
+    private const val TRACKING_URL = "u"
 
     fun build(): TypeAdapter<TrackImpl> {
         return typeAdapter {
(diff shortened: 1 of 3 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/TrackTypeAdapter.kt
```diff
@@ -27,6 +28,8 @@ object TrackTypeAdapter {
                 value(it.remote_id)
                 name(LAST_READ)
                 value(it.last_chapter_read)
+                name(TRACKING_URL)
```
NoodleMage: This will probably crash older backups.  Will try to look at it tonight. No promises because stuff and such :)

inorichi:
Shouldn't it be able to recover the url after the restoration process?

CarlosEsco:
when it was lateinit and backup was before the url was there it was getting never initialized

inorichi:
But given that it's not lateinit now and doesn't crash, does it need to be part of the backup file or can it be restored [here](https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt#L390)?

CarlosEsco:
isnt that method just comparing db tracks vs backup file tracks?  So if its not in db it goes to the tracksfor update but since cover isnt in the file in your scenario  the track gets inserted into db with no cover url.

PR #1270, opened 2018-02-21
Option to use Tracking Services covers at a global level

CarlosEsco:
This adds ability to select a specific tracker as the main cover url source.

If manga is not linked to tracker it will default to the sources cover.  
If it is linked to the selected tracker the cover gets updated on a manga info refresh.

image

once #1269 gets merged i will fix any merge issues and repush this

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -129,6 +129,7 @@ dependencies {
 
     // Network client
     implementation "com.squareup.okhttp3:okhttp:3.9.1"
+    implementation 'com.squareup.okhttp3:logging-interceptor:3.9.1'
     implementation 'com.squareup.okio:okio:1.14.0'
 
     // REST
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -129,7 +129,6 @@ dependencies {
 
     // Network client
     implementation "com.squareup.okhttp3:okhttp:3.9.1"
-    implementation 'com.squareup.okhttp3:logging-interceptor:3.9.1'
     implementation 'com.squareup.okio:okio:1.14.0'
 
     // REST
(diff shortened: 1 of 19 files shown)


bbtora:
Never thought of this but it's an incredible idea.

Rinto-kun:
I think it might be a good idea to add [MangaUpdates](https://www.mangaupdates.com/index.html) to this list because their covers usually update more often.

inorichi:
While it's not a bad idea, it has a few big problems:

* The library covers cache is never used now, which means that they will be deleted when Glide's cache is filled with the covers from the catalogues and will have to be redownloaded.

* The cover requests now don't use the network client of the source. If one source requires some cookies, it won't be able to load any cover (Kissmanga for example won't be able to resolve cloudflare's challenge)

In my opinion this gives more troubles than improvements.

CarlosEsco:
For number 1 maybe I missed it but I thought I was adding the new cover to the cache.

For number 2 it still uses the source client. After all info is pulled from source including cover. It checks if it should be using the tracker and at that point it changes the url.  Or am I missing something?

CarlosEsco:
I just wanted a way to use better covers cause kissmanga and other sources have some very LQ covers..

CarlosEsco:
Latest push should fix not using library cover caches, and using the network client for source.

Only thing it changes it reverts load(manga.thumbnail_url) back to load(manga)  to go through custom ModelLoader, and it now skips the glide memory cache so that it forces glide to go through the modelLoader and either pull the new image, or load from custom cover cache

inorichi:
`skipMemoryCache` shouldn't be added. It's the best optimization for a smooth scroll on RVs. Why do you want to skip cache? If it's for invalidation, there are other approaches to achieve that.

CarlosEsco:
I didn't see anyway to invalidate just their in memory cache, and when using the custom cache passing in the manga.  That's why originally I was passing url only cause it forced invalidation.

Deterio:
Anything new on this? This is a feature I'd like to see.

enwokoma:
@CarlosEsco . What if you use something like `.signature(new StringSignature(String.valueOf(System.currentTimeMillis())))` instead of `skipMemoryCache` so as to invalidate the cache as well as retain a smooth scroll on RVs

PR #1316, opened 2018-03-18
Enable TLS 1.1 and TLS 1.2 on Android KitKat (and older)

paronos:
This fixes connection errors with some servers on Android 4.1-4.4. In particular I noticed this with the _xkcd_ extension.

More info on https://blog.dev-area.net/2015/08/13/android-4-1-enable-tls-1-1-and-tls-1-2/

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt
@@ -1,9 +1,22 @@
 package eu.kanade.tachiyomi.network
 
 import android.content.Context
+import android.os.Build
 import okhttp3.Cache
 import okhttp3.OkHttpClient
 import java.io.File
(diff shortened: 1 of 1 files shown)


inorichi:
Please move the instantiation of the trust manager to a private function in that same file. There's a lot of indention there.

paronos:
Maybe I did not understand correctly, but if I move the trust manager's instantiation into a separate function, I do not think that the number of indention.
Maybe I should instead move the TLSSocketFactory class immediately after `class NetworkHelper(context: Context) {`?

inorichi:
It does. The client builder adds 8 spaces, the apply 4, and the sdk version check another 4.

If you move what's inside to a method it'll be easier to read, besides having a much shorter and readable client builder.

You can also make a private extension method of the builder and avoid the `apply` call.

paronos:
I have replaced the `apply` with an extension function.

PR #1334, opened 2018-04-02
[1.x] Initial Theme / Layout rewrite

NoodleMage:
Removes some of the bulk created over the time in the values and themes. 

~Note that this is made for API 21+ so will crash on API 19.~ fixed 
If we implement API 19 I think I will only allow the main theme (on API 19) to avoid code overhead.

TODO:

- [x] Amoled theme
- [ ] Style guide: https://github.com/NoodleMage/tachiyomi/wiki/Style-Guide
- [ ] Rewrite XML
- [ ]  Cleanup


CarlosEsco:
Thoughts on using this as part of theming. 

https://github.com/garretyoder/Colorful

NoodleMage:
Looks interesting. Will consider implementing it. Thanks :)

inorichi:
**PR:**
Nice! Resource cleanup is always welcomed!

I'd really like to have minSdk 21 but it's not the time [yet](https://twitter.com/minSdkVersion), sadly. For now we'll stick to 19.

But I'm fine with limiting themes or other features to 21. We may give our users another reason to upgrade.

**Code comments:**
Now that the rewrite is taking some shape, I'll start documenting the parts that most probably won't be changed. I'm still trying to figure out a few UI things when following the MVI pattern but it shouldn't be a big deal.

**Colorful:**
I've seen this before but never digged too much into it. I don't know how it'll play with our hardcoded colors (badges and the yellow filter color in the toolbar as far as I can remember, but I think I'm missing some)

CarlosEsco:
yea just looked at you cant even use custom colors for colorful.  There has to be a library out there that fits the needs of Tachi

CarlosEsco:
https://github.com/52inc/Scoops this one might suit Tachi better

paronos:
If it's possible, I'd like there to be one last release with minSdk 16 before this PR merge (in order to have TLS 1.2 on my old smartphone).

paronos:
My bad, I only now noticed that there is a rewrite in progress.

LinkCable:
Would you want a style guide for this?

NoodleMage:
Style guide would be awesome 😊

paronos:
Maybe it is time for minSdk=21 https://twitter.com/minSdkVersion/status/988538738639765505

zunjae:
@paranos let me quote Jake Wharton

>Devices are not users and users are not your customers and your customers are not even the subset of customers which provide value. Making decisions based on the platform distribution is equally as stupid as making decisions based off of a random (albeit amazing) Twitter account. Some apps will have 30% pre-21. Others will have effectively 0% pre-21.

In short. Don't make decisions on a Twitter account.

CarlosEsco:
Min sdk is gonna raise at some point though. And while I agree with what Jake says I don't think it applies to open source free app

inorichi:
Yep, I agree with @zunjae here. Although I don't have numbers, from many issues I can imagine a lot of our users are using old tablets, mainly because they still work fine and tachiyomi doesn't need a high-end device, not to mention the tablet market is almost dead at the moment. I'm still using the Nexus 7 (2013) and I don't like any of the current alternatives...

So, for now it'll probably stay at minSdk 19, but sooner or later it's going to be upgraded to 21. This doesn't tachiyomi would stop working for them, but they just won't receive new features, and they should still receive extension updates.

NoodleMage:
Closing this because of the insane amount of commits done since this pull. Will create new smaller pulls for code cleanup / theme rewrite.

BTW. I'm back (at least I think so :))

PR #1341, opened 2018-04-05
Update README.md

vchen30:
thought it would be cool to have hyperlinks to the sites


inorichi:
I'm fine with linking to MAL, AL and Kitsu, but not to aggregators, I prefer just the text. Can you change that?

PR #1343, opened 2018-04-06
Creating a pull request for a new pink theme

Camology:
Had to create a pull request with a feature upgrade for a course I am in. I also don't know anything about android dev but I made this little pink theme that doesn't look too bad actually. Never figured out where the onscreen button background is set though.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -30,7 +30,7 @@ ext {
 
 android {
     compileSdkVersion 27
-    buildToolsVersion "27.0.2"
+    buildToolsVersion '27.0.3'
     publishNonDefault true
 
     defaultConfig {
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.0'
         classpath 'com.github.ben-manes:gradle-versions-plugin:0.17.0'
         classpath 'com.github.zellius:android-shortcut-gradle-plugin:0.1.2'
         // NOTE: Do not place your application dependencies here; they belong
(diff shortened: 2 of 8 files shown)


Sparh4wk:
I thought that 1st of april was gone, but as I can see I was wrong :)

CarlosEsco:
Screens?

NoodleMage:
This pull won't be merged. Mostly because we don't want too much different themes (If we allow pink, people want different colors as well).

You say it is for a course. Is the pull enough or do you want a code review as well?

Camology:
Code review would also be useful for future purposes. My initial goal was to create a new menu with sliders so the entire theme menu would be custom and people could export/import an XML with themes but seemed a bit outside the time I have. Maybe over the summer

app/build.gradle
```diff
@@ -30,7 +30,7 @@ ext {
 
 android {
     compileSdkVersion 27
-    buildToolsVersion "27.0.2"
+    buildToolsVersion '27.0.3'
```
paronos: Typically a pull request should not contain unrelated changes.

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.0'
```
paronos: same here

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Oct 25 23:17:30 CEST 2017
+#Thu Apr 05 21:09:14 AKDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
paronos: same here

Camology:
Ah okay, yeah also new to Android studio so wasn't sure what was needed and wasn't necessary :p

Issue #762, opened 2017-04-19
Warn about missing sources


j2ghz:
Extension sources make it easy to add and remove sources. When you remove a source now, any manga you have from the source will not be shown until you add it again.
It would be good to show a warning which manga are affected and which sources are missing.

Enhancement:
* Instead of a warning show dialog that can be used to delete the affected manga or "Remind me later"

inorichi:
For the record, if a source is missing, we can't tell which one is. We only know the id, a 64 bits number not useful for the user.

So this can only act as a way to tell the user he needs to install more sources.

j2ghz:
Is the id not a string? Or are you hashing the id?

But still, such warning is better than nothing.

null-dev:
The I'd is a long.

j2ghz:
With #1101 and https://github.com/inorichi/tachiyomi/issues/1151#issuecomment-355854003 this should be changed. New proposal:
When a manga's extension does not exist, as much info as possible should be shown, the rest replaced by warnings. Not sure what the design limitations on this are.

inorichi:
Closing since manga is not hidden anymore.

On 1.x we could create another index (always keeping the history) to map source ids to names. It'd be hard to script it on the current version because the source api requires tachiyomi to run.

Issue #1333, opened 2018-04-01
Update to 0.7.0 hides the ReadComicOnline comcis


jotumn:
**App version:**
0.7.0

**Issue/Request:**
When I updated to 0.7.0, all my comics from ReadComicOnline disappeared (added through the old yaml method). However, when I press the update library button I can see their names displayed in the progress bar so they must exist in the database. I tried to install the extension but that didn't help. Selecting/unselecting any filters also does not do anything.

CarlosEsco:
Yml has not been supported for a very long time. Switch to the internal extension

jotumn:
I did but the problem is that there are invisible titles stuck in the rotation.

LakotaGray:
Did you uninstall all the old extensions before trying to install new ones or internal ones? Are the old ones you have from Fdroid? Just to make sure everything is in proper order, uninstall all your extensions, and then install them again from within the extensions section of 0.7.0

jotumn:
I did not have any old extensions except that YAML file that I downloaded from here (back when it had a link): https://github.com/inorichi/tachiyomi/wiki/YAML-Sources

Uninstalling/Reinstalling the extensions and removing that YAML file does not help the update still shows there are ghost comicbooks.

inorichi:
Fixed on master.

PR #1345, opened 2018-04-06
Show manga with no installed source

paronos:
I think it would be better to show a manga even when the corresponding extension is not installed. This would also be a solution for #1333 (would allow migration or deletion).

For now a stub source is added for each source id found in the database (since I do not know if this PR will be merged, for now there are no optimizations).


inorichi:
Isn't it easier to return a default source in the `get` method?

Anyways this would need a lot of testing before merge.

paronos:
Now the get method return a default source (placed in a map to avoid multiple incantations).
Also I removed the `Bototo` source since it was almost equal to `StubSource`. Maybe the other sources could also be removed (leaving everything to the extensions).

I did some tests (by installing and uninstalling extensions and then going to the "My library", "Library updates" and "Recently read" sections), but it can not be said to be "a lot of testing".

Sparh4wk:
just wondering, isnt better solution to store information about catalogues in backup and force uset to install extensions stored there before restore?

j2ghz:
Closes https://github.com/inorichi/tachiyomi/issues/762

fam007e:
Could you please add "www.heavenmanga.life" site into the list...
This site enables some of the otherwise licensed manga to be read.

inorichi:
Merged manually with a few modifications.

fam007e:
I didn't understand.

PR #1357, opened 2018-04-09
Remove F-Droid.org link

j2ghz:
Because of slow updates, mainly when an important bugfix is released, and subsequent user complaints, F-Droid.org is not a recommended way. Those that already use F-Droid for something else can still search for Tachiyomi in the app.

I have tried to better convey the difference between the 3 buttons for getting the app. The first one seems on edge of being too long. Ideas, or is it fine?

Preview: https://github.com/j2ghz/tachiyomi/blob/patch-1/README.md


Lionirdeadman:
F-droid should still be recommended. Otherwise, those who are not aware of it being on F-droid could be misled into downloading it from another source.

If you want, you could add a note saying that F-droid may not have the latest version and that you have no control over when an update will come.

j2ghz:
Ready to merge

paronos:
The link https://github.com/inorichi/tachiyomi/wiki/FDroid-for-dev-versions does not work.

BlackSynder:
I'ts been changed a little (extra hyphen in F-Droid):
https://github.com/inorichi/tachiyomi/wiki/F-Droid-for-dev-versions

PR #1383, opened 2018-04-23
Implement Anilist API v2 (closes #1159)

### LINKED ISSUE #1159: [Tracking] Anilist API v1 is Deprecated
Anilist's v1 API is [deprecated](https://anilist-api.readthedocs.io/en/latest/) and will be disabled May 1st, 2018. The new api documentation is here: https://github.com/AniList/ApiV2-GraphQL-Docs

They've switched to GraphQL which will require different structures and query changes.

Flat:
Switches to using the Anilist v2 API.
Login is now done by implicit grant and tokens are good for one year.
Users will need to login again after token expiration.
"clientId" on line 289 of AnilistApi.kt should be changed to Tachiyomi's
own client ID number.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/TrackTypeAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/TrackTypeAdapter.kt
@@ -1,5 +1,6 @@
 package eu.kanade.tachiyomi.data.backup.serializer
 
+import android.telecom.DisconnectCause.REMOTE
 import com.github.salomonbrys.kotson.typeAdapter
 import com.google.gson.TypeAdapter
 import com.google.gson.stream.JsonToken
@@ -11,7 +12,8 @@ import eu.kanade.tachiyomi.data.database.models.TrackImpl
 object TrackTypeAdapter {
 
     private const val SYNC = "s"
-    private const val REMOTE = "r"
+    private const val MEDIA = "m"
+    private const val LIBRARY = "ml"
     private const val TITLE = "t"
     private const val LAST_READ = "l"
     private const val TRACKING_URL = "u"
@@ -24,8 +26,10 @@ object TrackTypeAdapter {
                 value(it.title)
                 name(SYNC)
                 value(it.sync_id)
-                name(REMOTE)
-                value(it.remote_id)
+                name(MEDIA)
+                value(it.media_id)
+                name(LIBRARY)
+                value(it.library_id)
                 name(LAST_READ)
                 value(it.last_chapter_read)
                 name(TRACKING_URL)
@@ -43,7 +47,8 @@ object TrackTypeAdapter {
                         when (name) {
                             TITLE -> track.title = nextString()
                             SYNC -> track.sync_id = nextInt()
-                            REMOTE -> track.remote_id = nextInt()
+                            MEDIA -> track.media_id = nextInt()
+                            LIBRARY -> track.library_id = nextLong()
                             LAST_READ -> track.last_chapter_read = nextInt()
                             TRACKING_URL -> track.tracking_url = nextString()
                         }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/TrackTypeAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/TrackTypeAdapter.kt
@@ -12,7 +12,7 @@ import eu.kanade.tachiyomi.data.database.models.TrackImpl
 object TrackTypeAdapter {
 
     private const val SYNC = "s"
-    private const val MEDIA = "m"
+    private const val MEDIA = "r"
     private const val LIBRARY = "ml"
     private const val TITLE = "t"
     private const val LAST_READ = "l"
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
@@ -10,7 +10,7 @@ object TrackTable {
 
     const val COL_SYNC_ID = "sync_id"
 
-    const val COL_MEDIA_ID = "media_id"
+    const val COL_MEDIA_ID = "remote_id"
 
     const val COL_LIBRARY_ID = "library_id"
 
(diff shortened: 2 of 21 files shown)


inorichi:
Client id is 385. Is the client secret no longer required?

I'll try to review the code later today.

Flat:
Correct, since using the implicit grant flow no client secret is required. Implicit grant is the recommendation when the client secret can't be secured such as in JavaScript webapps and in this case, open source.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/AnilistLoginActivity.kt
```diff
@@ -23,13 +23,14 @@ class AnilistLoginActivity : AppCompatActivity() {
         val view = ProgressBar(this)
         setContentView(view, FrameLayout.LayoutParams(WRAP_CONTENT, WRAP_CONTENT, CENTER))
 
-        val code = intent.data?.getQueryParameter("code")
-        if (code != null) {
-            trackManager.aniList.login(code)
+        val regex = "(?:access_token=)(.*?)(?:&)".toRegex()
```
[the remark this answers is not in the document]
Flat: The token is provided in the URL fragment, my research didn't produce any easy ways to parse the URI fragment from Java/Kotlin aside from regex. `tachiyomi://ani-auth/#access_token=eyJ0eXAiOiJKV1QiLCJhbmIzMDM4MWJmOGViN2I4ZDI3YmFlMzMwOTExYTYzNDc5NzQxNTIzNDJjMDgzIn0.eyJhdWQiOiIzODciLCJqdGkiOiIyZmUyNWM2NzViNzQ0NDZlMzEzZmQzNWIxOTNlM2I1ZjFiZWZiMzAzODFiZjplJyrRH4LKM4FYnVy_E3MDMfiEWNoKYskue9tdLHGPs7hWKia2zimNf33P72UjW0t_7y4E&token_type=bearer&expires_in=31536000` I deleted a large portion of the token so it is not valid.

Flat:
Should be all the requested changes. Let me know if I missed anything.

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt
```diff
@@ -1,167 +1,276 @@
...
+                .asObservableSuccess()
+                .map { netResponse ->
+                    val responseBody = netResponse.body()?.string().orEmpty()
+                    netResponse.close()
+                    if (responseBody.isEmpty()) {
+                        throw Exception("Null Response")
                     }
+                    val response = parser.parse(responseBody).obj
+                    val id = response["data"]["SaveMediaListEntry"]["id"].asLong
+                    track.id = id
```
inorichi: `id` in the track model belongs to our local id. This should be `remote_id`

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt
```diff
@@ -1,167 +1,276 @@
...
+            mutation UpdateManga(${'$'}listId: Int, ${'$'}progress: Int, ${'$'}status: MediaListStatus, ${'$'}score: Int) {
+                        SaveMediaListEntry (id: ${'$'}listId, progress: ${'$'}progress, status: ${'$'}status, scoreRaw: ${'$'}score) {
+                            id
+                            status
+                            progress
+                        }
                     }
+            """
+        val variables = jsonObject(
+                "listId" to track.id,
```
inorichi: Same here, `remote_id`.

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistInterceptor.kt
```diff
@@ -20,24 +20,20 @@ class AnilistInterceptor(private var refreshToken: String?) : Interceptor {
...
-        if (oauth == null || oauth!!.isExpired()) {
-            val response = chain.proceed(AnilistApi.refreshTokenRequest(refreshToken!!))
-            oauth = if (response.isSuccessful) {
-                Gson().fromJson(response.body()!!.string(), OAuth::class.java)
-            } else {
-                response.close()
-                null
-            }
+        if (oauth!!.isExpired()) {
+            anilist.logout()
```
inorichi: I guess we should throw an exception here, since the requests are still going through.

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt
inorichi: Also it's already included in the request so I doubt this needs to be changed here.

Flat: The issue, and why I changed the I'd is that to update a manga we now need the list id from findLibManga, not only the manga id. Should I update the schema to include another id?

app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/TrackTypeAdapter.kt
```diff
@@ -11,7 +12,8 @@ import eu.kanade.tachiyomi.data.database.models.TrackImpl
 object TrackTypeAdapter {
 
     private const val SYNC = "s"
-    private const val REMOTE = "r"
+    private const val MEDIA = "m"
```
inorichi: For compatibility reasons, I'd call this `val MEDIA = "r"`

app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
```diff
@@ -10,7 +10,9 @@ object TrackTable {
 
     const val COL_SYNC_ID = "sync_id"
 
-    const val COL_REMOTE_ID = "remote_id"
+    const val COL_MEDIA_ID = "media_id"
```
inorichi: `const val COL_MEDIA_ID = "remote_id"` so that there's no need to do a table migration.

app/src/main/java/eu/kanade/tachiyomi/data/database/tables/TrackTable.kt
```diff
@@ -46,4 +46,37 @@ object TrackTable {
 
     val addTrackingUrl: String
         get() = "ALTER TABLE $TABLE ADD COLUMN $COL_TRACKING_URL TEXT DEFAULT ''"
+
```
inorichi: Wow, this was harder than it should have been :sweat_smile: .

If you keep the column name as "remote_id" and just add a new column for "library_id", there's no need to copy the entire table, just an ADD COLUMN clause.

Flat: Yeah, sqlite doesn't support `alter table rename column` :cry:

Issue #1434, opened 2018-05-27
Can't track webtoons with AniList


Azureum:
**App version:** 0.7.4

**Issue/Request:** Issue

**Steps to reproduce (if applicable)**

 1. Add a webtoon like Tower of God, Lookism, etc. to your library.
 2. Go to the tracking tab
 3. The AniList details will be empty.
 4. Selecting on edit button and then searching returns no results.

**Other details:**
Tracking with AniList for webtoons from webtoons.com is not possible. Here is an example screenshots of a webtoon called Lookism.
![screenshot_20180527-143619](https://user-images.githubusercontent.com/20842882/40584360-9a1cb176-61bc-11e8-9ccf-378979582cdb.jpg)
![screenshot_20180527-143628](https://user-images.githubusercontent.com/20842882/40584365-a3646436-61bc-11e8-9f1d-8fd20459f6c9.jpg)
The corresponding entry is present in the AniList database.
![screenshot_20180527-143720](https://user-images.githubusercontent.com/20842882/40584374-bf43dbdc-61bc-11e8-8752-cb7a99bb5bf6.jpg)

The same issue is present for all webtoons from webtoons.com. For e.g., Tower of God, Noblesse, unOrdinary, Yumi's Cells, etc.

ddmgy:
That's odd, I'm not having any issue tracking any of the webtoons you mentioned.

For example:
![](https://i.imgur.com/44CDj54.png)

Are you able to track anything in AniList, with any extension?

RED-404:
I gave it a shot on r1414 and no joy.![](https://i.imgur.com/xTDb2T2.png)

ddmgy:
Playing around with the search query Tachiyomi uses, I think I've found the problem. Running the query without argument 'format: MANGA' returns the correct result. I'll make the change in Tachiyomi and see if it works when I'm able to access my computer.

xKeiro:
This used to work on the development version but the last few days it can't find webtoons again.

Akihisa17:
This issue still not solve. Not only in Webtoons but in Manhua as well (e.g Soul Land, Soul Land II, Panlong, Spirit Blade Mountain and etc.

PR #1435, opened 2018-05-27
Change AniList search query to show some previously hidden entries.

ddmgy:
Addresses issue #1434.

For some reason, AniList API v2 was not returning the correct search results when query arguments `type: MANGA` and `format: MANGA` were both present. Removing `format: MANGA` makes the query return the correct results. Specifically, it returns Korean manhwa as well as Japanese manga, where it was not always returning manhwa before. For instance, Noblesse (manhwa) and One Piece (manga) were both showing fine, but Lookism (manhwa) was not.

I changed the format argument to `format_not_in: [NOVEL]` because there are only three manga related MediaFormats: MANGA, NOVEL, and ONE_SHOT. Filtering out NOVEL entries leaves manga and one-shots, both of which are relevant to Tachiyomi, and which allows all manga and manhwa entries (that I have so far tested) to show in the search results.

Example of Lookism correctly showing in tracking search:
screenshot_20180527-151350 1

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt
@@ -91,7 +91,7 @@ class AnilistApi(val client: OkHttpClient, interceptor: AnilistInterceptor) {
         val query = """
             query Search(${'$'}query: String) {
                   Page (perPage: 25) {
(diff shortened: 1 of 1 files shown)


CarlosEsco:
Haven't looked at API, but does this skip anime and ovas?

ddmgy:
Yes, with `type: MANGA` argument, the only applicable formats are MANGA, ONE_SHOT, and NOVEL. No items with type ANIME are returned.

Issue #1436, opened 2018-05-27
[Bug] Anilist total chapter count disappears after manual tracking update.


eldrGungnir:
Tachiyomi (dev) 0.7.4-1417 

AniList tracking seems to 'forget' the total chapter count for manga with a complete chapter list when manually updated by either the specific manga tracking tab or by refreshing the tracking metadata in advanced settings. To fix this re-add the manga through the tracking tab and the total chapter count will re-appear

Here is a video showing the issue:
https://gfycat.com/RespectfulSlushyAfricanmolesnake

To reproduce:
1. Add tracking to manga with complete chapter list on AnilList.
2. Manually update through the specific manga tracking tab or the advanced settings refresh tracking metadata option.
3. Re-add the manga to tracking as done in step 1 and the problem will be fixed.

On Android 5.0

BaronKrause:
I can verify this on Android 8 and 8.1 as well. Even simply changing the status from "Reading" to "Plan to Read" after adding a new manga will make it lose its total count.

PR #1481, opened 2018-06-11
Update AnilistModels.kt

Sklan:
Closes the issue #1436


inorichi:
I think the query should include the total chapters instead of removing the field, because it can change over time.

Sklan:
You are correct. It's a better approach to instead update total chapters which was not being updated the originally, giving a null value.

Issue #1022, opened 2017-10-05
"Always Ask" for Default Category no longer working, it goes straight to a man-made category (not Default)


thesidejoe:
Pixel XL on Oreo 8.0.0
Tachiyomi version 0.6.3

https://gfycat.com/chiefimpressivegalapagostortoise

Whenever I bookmarked a new manga, it used to always ask which Category I wanted to put it in. Now, even with the Default category set to "always ask," it puts it in my only man-made category, "Finished." The only way to get it back into Default category is by selecting the manga, moving it, and unchecking "Finished." I was previously on version 0.6.1 when this was occurring, and I updated to 0.6.3 to see if it would fix, but it did not get fixed when I updated. The gif shows the process and the problem.

CarlosEsco:
Yea i noticed this when i rewrote this portion in my pull request.  #960 .  If that does end up getting approved it will be fixed if not and @inorichi  rejects  I can look into this.

LakotaGray:
The issue comes about due to you having created only 1 category. If you create an additional category, so you’d have default, finished, and “X” it will then return to standard operation. The best analogy I can give is that by only creating one category, Tachi assumes you want all manga added to library put into your created category. By creating 2+, it then seeks clarification and will default if none is given

arkon:
I don't think this still happens.

PR #1495, opened 2018-06-18
Fix category chooser behavior when adding a manga

kappa7194:
When there is only one user-defined category the app always place new manga there, even when the user asked for confirmation every time. By checking if the categories list contains zero elements instead of one this problem should be prevented.

Should fix issues #885 and #1022.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
@@ -378,9 +378,13 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
         if (manga.favorite) {
             val categories = presenter.getCategories()
             val defaultCategory = categories.find { it.id == preferences.defaultCategory() }
+            // If a default category has been set, place the manga there.
+            // If there are no categories defined, place the manga in the default one.
+            // If there is at least one category defined, ask the user where the manga should go.
+            // See issues #885 and #1022.
             when {
                 defaultCategory != null -> presenter.moveMangaToCategory(manga, defaultCategory)
-                categories.size <= 1 -> // default or the one from the user
+                categories.isEmpty() ->
                     presenter.moveMangaToCategory(manga, categories.firstOrNull())
                 else -> {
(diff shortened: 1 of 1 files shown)


kappa7194:
Tests that I manually ran:

- Add a manga when there are no categories defined: the manga is placed in the system "default" category.
- Add a manga when there is one category defined without a default category set: the app asks the user where the manga should be placed.
- Add a manga when there is one category defined with a default category set: the app places the manga in the chosen default category.

The next step should be listing the system "default" category in the default category chooser, at the moment only "Always ask" and user-defined categories are listed.

kappa7194:
@inorichi any chance of merging this PR? The issue is still present in the current version of the app.

hXtreme:
What happens when you add a manga when there is one category defined without a default category set and then click cancel in the category selection? Is that the intended behaviour?

kappa7194:
> What happens when you add a manga when there is one category defined without a default category set and then click cancel in the category selection? Is that the intended behaviour?

I have no idea. This pull request is over a year old and no longer up to date with master, I don't know if something has changed since then.

However, since this change only changes a single quantity check (from <= 1 to == 0 basically) I'd say that it shouldn't affect how the cancel button works.

What is the problem? Was it working correctly before this change?

hXtreme:
I just need to figure out the exact behaviour that is intended for #2192 and then maybe make a pr to solve it.

bbtora:
Will this PR also fix/include the scenario when 
Someone downloads chapter from non-library manga > the app gives a prompt to add manga in library at the bottom. 
If you tap on add, it adds the manga in Default category regardless of whether you have a default category set up or set it as ask to choose.

kappa7194:
> Can you make the change then?

Done.

kappa7194:
> I just need to figure out the exact behaviour that is intended for #2192 and then maybe make a pr to solve it.

This behavior probably affects other places where the categories are used (e.g. "Categories to include in global update" in Settings → General).

FlaminSarge:
This may be superseded by https://github.com/inorichi/tachiyomi/pull/2292

inorichi:
The other PR seems to handle more cases so I merged that one.

PR #1511, opened 2018-06-25
Fixed shit.

az4521:
Can you dont the app add exclamation points and words when I want to copy the link thanks?


az4521:
Even if it's a share function, it should just share the manga, not "CHEKC OUT MANGA!!! AT LINK"
Look at how other android apps do it, YouTube as an example, it just shares the link without extra text.

inorichi:
Fine by me, but the string resource is unneeded now. I'm not sure it'd work if you delete the first argument.

The change should be made in the kotlin code that uses this resource. 

Do you want to change it? Otherwise I'll but I'd have to close this PR. Also please use a better commit message, no need to be rude.

inorichi:
Merged manually 5152891

Issue #1302, opened 2018-03-12
Old Dark Theme


sleey:
With the new update we lost the old dark theme, why not make a new option rather than replace the old one.
I think some people will prefer the old one, like me.

dkanada:
I actually like the new dark theme, except for the new navigation bar color. It blends in with the background color and looks odd in my opinion. Usually I don't mind small changes to a vanilla material design look but I definitely think that the navigation bar, toolbar, and status bar should stay a shade darker for contrast. Lots of apps will also merge the toolbar and status bar into one large bar with the same color, which I also think looks odd.

inorichi:
I don't mind a new theme 'Dark blue' to restore the old toolbar color. Open for pull request.

InFlames03:
@dkanada I have an issue with that as the creator of the theme myself. It should be a simple fix, the chapter list background needs to be changed to a lighter shade instead of having it use the same colour as the navbar/status bar. I'd have fixed it already if I wasn't in the middle of my semester (and currently a week from finals)

PR #1527, opened 2018-07-02
add legacy dark theme with more color

dkanada:
As mentioned in #1302 here is a dark theme with more color. I didn't notice the name you wanted (dark blue) so I named it dark and named the other theme gray. Which do you prefer? Also, I sorted the themes from light to dark, which places the new one second in the list. This would forcibly change the theme if someone was using dark or amoled before though, which might not be ideal. Should I move it to the end of the list or keep it where it is? I can fix those before you merge, just wanted to get the PR opened.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
@@ -48,8 +48,9 @@ class MainActivity : BaseActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         setTheme(when (preferences.theme()) {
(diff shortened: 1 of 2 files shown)


inorichi:
I didn't notice the name you wanted (dark blue) so I named it dark and named the other theme gray. Which do you prefer?

I don't really mind which, both fit the purpose.

> Should I move it to the end of the list or keep it where it is? 

Yes, move it to maintain the order of themes.

dkanada:
Fixed the theme order and left the names, works fine on my phone. Let me know if there is anything else that needs to be done.

dkanada:
I am closing my fork, I will leave the pull request open for now though.

inorichi:
I added the theme on commit 2c4fd34

PR #1535, opened 2018-07-07
Update manga.last_update on ChapterSourceSync.syncChaptersWithSource()

FlaminSarge:
Update manga.last_update on ChapterSourceSync.syncChaptersWithSource() rather than only during LibraryUpdateService.updateChapterList()

Viewing a manga's info page for the first time forces a chapter sync.
Prior behavior would cause new chapters to be retrieved for that manga, but with manga.last_update remaining at 0 (until a library update occurred in which chapters were added).
New behavior updates last_update any time the chapters are changed via syncChaptersWithSource().


FlaminSarge:
I was debating only updating last_update if [toAdd - readded] wasn't empty, but ultimately felt that last_update should reflect any updates, not just additions.

inorichi:
Makes sense, thanks.

Issue #1331, opened 2018-03-30
Filename / Directory containing % will make Local Catalogues error


rahadiancs:
Version: 0.7

Issue: If we put directory / file with % on it's name, the local Catalog will error. On my case I put Ichigo 100% doujinshi on it. After removing the %, it will works as usual.

Steps to reproduce:

 1. Copy directory / zip files containing % on it's name
 2. Browse Local Catalog
 3. Nothing showed, and there's only an empty notification and retry button on the bottom of the screen.

inorichi:
What happens with .rar or folders? If it's specific to .zip I'm afraid it won't be fixed anytime soon, as it's using the built-in zip API.

rahadiancs:
It's not specific to zip because it will trigger the error when you put a folder / directory with %.

PR #1536, opened 2018-07-08
Fixes for loading local files

jshap70:
Fix bug where archives containing '%' or '#' failed to load. (#1331)

Allow files in a chapter directory to load as the chapters themselves if it is the only item in the directory. ie: a folder contains a zip with the chapter's files, load the zip instead of the folder itself. Previously the user would receive a "Page list is empty" message.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
@@ -180,6 +180,14 @@ class LocalSource(private val context: Context) : CatalogueSource {
     private fun getLoader(file: File): Loader {
         val extension = file.extension
         return if (file.isDirectory) {
+            // if folder only contains one entry, check if it is a valid entry (a subdirectory, zip, epub, etc)
+            val contents = file.listFiles()
+            if (contents.size == 1) {
+                try {
+                    return getLoader(contents[0])
+                } catch (e: Exception) {
+                }
+            }
             DirectoryLoader(file)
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -315,4 +323,16 @@ class LocalSource(private val context: Context) : CatalogueSource {
...
-}
\ No newline at end of file
+}
+
+/**
+ * Returns a URI readable string for a local file.
+ *
+ * The .zip/.rar/.epub ContentProvider has issues when URIs contain certain symbols, so they have to be URL encoded
+ * before being passed. java.net.URLEncoder tends to mangle paths, such as spaces are replaced with "+",
+ * so symbols are replaced on an individual bases.
```
msm595: typo: "bases" -> "basis"

inorichi:
I'll hold this PR for now, because I've decided to rewrite the reader from scratch to fix issues like these.

The content providers for ZIP/RAR won't be needed anymore. I also want to add an extra page between chapters to notify the title of the current and next/previous chapter. It might be a bit hard to implement on the webtoon reader, though.

I'll try to keep the old reader and include a toggle to use the new one. I don't know how much time it's going to take, hopefully I won't find many obstacles.

inorichi:
Content providers are not needed anymore in the new reader and there's no need to merge.

Thanks for bringing this issue to light.

PR #1542, opened 2018-07-16
For migration, put the selected source at the top of the search list instead of excluding it

FlaminSarge:
Verified that this does not affect the order of the other elements.

I've also verified that both copy and migrate work for same-source migration, even if you select the manga you've already got selected.


FlaminSarge:
The other option: putting it at the bottom using `sortedBy`. Verified that that also does not affect the order of the other elements.

FlaminSarge:
Here's how it looks with all three themes:
Main
Dark
AMOLED

PR #1593, opened 2018-08-30
Translations

j2ghz:
A batch of weblate translations. I was hoping that weblate would be able to do this automatically, as presented in the documentation, but apparently not.


inorichi:
Can you rebase? Also there seems to be an invalid directory format, last travis build failed with `> /home/travis/build/inorichi/tachiyomi/app/src/main/res/values-zh-rCMN-HANS: Error: Invalid resource directory nam`

PR #1637, opened 2018-09-30
Allow pausing downloads from progress notification

FlaminSarge:


app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
```diff
@@ -38,6 +38,11 @@ class NotificationReceiver : BroadcastReceiver() {
             ACTION_DISMISS_NOTIFICATION -> dismissNotification(context, intent.getIntExtra(EXTRA_NOTIFICATION_ID, -1))
             // Resume the download service
             ACTION_RESUME_DOWNLOADS -> DownloadService.start(context)
+            // Pause the download service
+            ACTION_PAUSE_DOWNLOADS -> {
+                DownloadService.stop(context)
+                downloadManager.pauseDownloads()
```
inorichi: Isn't it enough with the previous line?

Issue #1293, opened 2018-03-06
Range select, more extensions and others


AbdulrahmanHadz:
Hi, i would like to request for the app to have chapter range select so that you can select from chapter 13-45 for example. I would also like to request an extension for lhscans.com for raw manga, chapter grid view and to have a resume button that lets you read for the chapter you were on last even if the chapter has been marked as read.

Thank you,
and keep up the hard work

Flat:
Requests for additional sources do not belong in this repository. They go here https://github.com/inorichi/tachiyomi-extensions

Please elaborate on a range select. You can already long-tap to select a range of chapters.

A resume button already exists and should pick up where you left off.

What is the rationale of a grid view for chapters? What information would this include?

Madidas:
> Please elaborate on a range select. You can already long-tap to select a range of chapters.

Long-tap allows user to select multiple chapters but selecting those chapters must be done one chapter at the time (unless it's changed). Lots of tapping and scrolling if you need to select lots of chapters (like 15+ chapters at once).
 
I guess @AbdulrahmanHadz would like to have feature to select two different chapters and also have every chapter between those two selected chapters to be also selected. Like extra option in the "three dot menu". Or just a text field/box/menu option to type where the selection of chapters starts and ends (Select everything from Chapter 13 to Chapter 45). Could be useful for manipulation of large set of chapters (set read/unread, download/delete certain chapters).

That is how I interpreted that "chapter range select" suggestion.

AbdulrahmanHadz:
What i meant is what @Madidas explained, the continue reading only works with unread chapters, i want to be able to resume reading a chapter that i have already read before and the app has marked as read (e.g. re-reading a series). The resume button gives a float notification that next chapter is not found. 
The grid view would help when navigating a series that has a lot of chapters and you want to reach the chapter quicker than needing to scroll.
Thanks, and next time i'll put this thread in the correct repository.

Flat:
Regarding resume button: https://github.com/inorichi/tachiyomi/issues/1186#issuecomment-366460272

Tl;dr working as intended, mark them un-read if you're re-reading. 

The range select is probably doable if someone has time to implement it.

As for the grid view, how would this be different than the list view? Do you mean just take the current list of chapters and cut the width in half to have two entries in one row? Images for each chapter?

AbdulrahmanHadz:
For the grid view:
https://imgur.com/a/843wk

[...]

PR #1641, opened 2018-10-01
Implement "Range Select" for Manga Chapters

### LINKED ISSUE #1493: Download Feature Request: Range Selection
**App version:**
0.7.4
**Issue/Request:**

[REQUEST]  Can we have the range selection to download manga chapters. 
Eg. Select starting chapter then select ending chapter?
Then all the chapters between those will be downloaded

**Other details:**

I used this feature in other manga app and is very convenient to have. 

screenshot_2018-06-17-15-27-19-666

Flat:
Closes #1493 and partially #1293

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
@@ -353,20 +353,15 @@ class ChaptersController : NucleusController<ChaptersPresenter>(),
     private fun selectBetween() {
         val adapter = adapter ?: return
         if (selectedItems.size == 2) {
-            val first = adapter.items.indexOf(selectedItems.first())
-            val last = adapter.items.indexOf(selectedItems.last())
+            var first = adapter.items.indexOf(selectedItems.first())
+            var last = adapter.items.indexOf(selectedItems.last())
             if (first > last) {
-                selectedItems.clear()
-                selectedItems.addAll(adapter.items.subList(last, first + 1))
-                selectedItems.forEach { adapter.addSelection(adapter.items.indexOf(it)) }
-                adapter.notifyDataSetChanged()
-
-            } else {
-                selectedItems.clear()
-                selectedItems.addAll(adapter.items.subList(first, last + 1))
-                selectedItems.forEach { adapter.addSelection(adapter.items.indexOf(it)) }
-                adapter.notifyDataSetChanged()
+                first = last.also { last = first }
             }
+            selectedItems.clear()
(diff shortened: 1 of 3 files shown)


BakaSai:
A much wanted feature of mine as well, I wonder why it hasn't been merged yet 🤔

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
```diff
@@ -349,6 +350,27 @@ class ChaptersController : NucleusController<ChaptersPresenter>(),
         actionMode?.invalidate()
     }
 
+    private fun selectBetween() {
+        val adapter = adapter ?: return
+        if (selectedItems.size == 2) {
```
arkon: You could potentially enable this only when there's 2 items selected, in `onPrepareActionMode`.

hXtreme: Seems like a nice idea, I could try to implement it but I'm not sure if I know how to do it.

My idea on how to implement it:  
1. Select Between option starts out as hidden.
2. Whenever a chapter was selected check to see if exactly 2 chapters are selected.
    a. If there are then set visibility to visible
    b. Else set it to hidden
    c. Finally do something to tell the UI to update.

arkon: Sounds about right. I had it as a disabled option in my mind, although I'm not entirely sure if menu items can be set as enabled/disabled, or if that's the same as visible/hidden.

You _should_ be able to invalidate the menu as needed to get it to update.

hXtreme: I do not know what this "invalidate the menu" means.

arkon: https://developer.android.com/reference/android/app/Activity#invalidateOptionsMenu()

hXtreme: Thanks! I'll take a look at it.

joseph619:
Is this still being worked on or is it ready to merge?

Flat:
It's ready to merge. Could be improved but more effort towards 1.x would be better.

hXtreme:
I'd say it's pretty much ready unless any reviewers want some changes to be made. All the previously requested changes have been implemented.

arkon:
Went with #2087

Issue #698, opened 2017-03-04
Metadata for local manga


shogunreaper:
Can you add the ability to grab metadata for local manga, maybe by searching titles on sites?

i'd like to have the plot/genres and other stuff in my local manga as well.

inorichi:
I won't code this. We discussed reading metadata from a file but decided not to include it since it's not critical. I can afford reading from a file since is easy (an external tool could generate them), but looking online is something I don't want to maintain here, at least yet.

Anyways, if anyone wants to implement this and it's not a lot of code, I will consider it.

shogunreaper:
but shouldn't you just be able to use the same way that you grab information for any other manga?

inorichi:
I grab that information from the url of the manga in the site. A local manga doesn't have an online url.

wopian:
The metadata could potentially be populated via the tracker its synced to?

shogunreaper:
well if it grabs the info by the url couldn't you just use that same url for the info?

like say if your folder is named naruto

it will search the manga sites for naruto the same way any other catalog searches (maybe with a popup if multiple hits come up) then just use the url you picked to grab the info.

inorichi:
> The metadata could potentially be populated via the tracker its synced to?

That's feasible, but limited to english only.

> it will search the manga sites for naruto the same way any other catalog searches (maybe with a popup if multiple hits come up) then just use the url you picked to grab the info.

And that's what I don't want to (and won't) do.

shogunreaper:
why not?

the only thing that needs to be maintained are what you are already maintaining for the other sites

inorichi:
It's totally not the same. If you want, you can fork the repo and do it yourself.

shogunreaper:
how is it not the same?

what else would need to be maintained?

paronos:
There are a few ways to include metadata within `cbr/cbz` files. One option is to manage them like the covers (read a local file if present, otherwise try to read them from the first `cbz/cbr`).

The format that I found are (more info at http://www.denvog.com/comet/comet-specification/ and https://wiki.mobileread.com/wiki/CBR_and_CBZ#Metadata):
- CoMet (uses a CoMet.xml file)
- ComicRack (uses a ComicInfo.xml file)
- ComicBookLover (zip header)
- ComicBookInfo (zip header)
- Advanced Comic Book Format (xml)

The most popular format could be used (although I do not know which is between them)

Moravuscz:
[...]

PR #1658, opened 2018-10-12
JSON metadata for local manga

ChrisJAllan:
Simple loading of metadata from a JSON file for local manga. Implement issue #698.


app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -101,6 +104,8 @@ class LocalSource(private val context: Context) : CatalogueSource {
                     }
                 }
 
+                fetchMangaDetails(this)
```
inorichi: I think it'd be better to remove this and set initialized to false, so that details are initialized lazily.

samyak-jain:
Any idea when this will be merged?

PR #1669, opened 2018-10-18
Handle empty manga thumbnail_urls in the cover cache

### LINKED ISSUE #1663: Cant change cover when cover does not exist
So i cant change a cover in a library when the cover does exist. 
For example the manga inside of mangazuki some dont have cover because the wevsite does not have one but chapter loads fine. Here the error that show up when trying to change cover.
screenshot_20181015-023425_tachiyomi

ChrisJAllan:
Fixes #1663 

Uses the source and manga's url as a unique value for hashing.

Example of it wirking with the JapScan source, which has no thumbnails: screenshot_1539830104

~~This does cause FileNotFound exceptions in FileFetcher when there is no cover set, though.~~

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaModelLoader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaModelLoader.kt
@@ -7,7 +7,6 @@ import com.bumptech.glide.load.model.*
 import eu.kanade.tachiyomi.data.cache.CoverCache
 import eu.kanade.tachiyomi.data.database.models.Manga
 import eu.kanade.tachiyomi.network.NetworkHelper
-import eu.kanade.tachiyomi.source.LocalSource
 import eu.kanade.tachiyomi.source.SourceManager
 import eu.kanade.tachiyomi.source.online.HttpSource
 import uy.kohesive.injekt.Injekt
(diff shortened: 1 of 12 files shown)


simakover:
Hello, there are still souces, that have that issue. And looks like nobody merge youre fix into actual version(
tried make apk myself, but im not familiar with android studio, and cant build apk properly- too many errors

ChrisJAllan:
Now tested to work on Russian "henchan" source.

app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaModelLoader.kt
```diff
@@ -80,13 +80,8 @@ class MangaModelLoader : ModelLoader<Manga, InputStream> {
...
     override fun buildLoadData(manga: Manga, width: Int, height: Int,
                                options: Options): ModelLoader.LoadData<InputStream>? {
-        // Check thumbnail is not null or empty
         val url = manga.thumbnail_url
-        if (url == null || url.isEmpty()) {
-            return null
-        }
-
-        if (url.startsWith("http")) {
+        if (url != null && url.startsWith("http")) {
```
tcyrus:
```suggestion
        if (url?.startsWith("http") ?: false) {
```

PR #1672, opened 2018-10-19
 Dependencies management with gradle-kotlin-dsl-libs

jmfayard:
Hello inorichi!

I wrote a plugin that simplified Gradle Dependencies management

Have a look here https://github.com/jmfayard/gradle-kotlin-dsl-libs
and tells me if you like it.


app/build.gradle
```diff
@@ -102,137 +102,127 @@ android {
...
-    implementation "com.android.support:recyclerview-v7:$support_library_version"
-    implementation "com.android.support:preference-v7:$support_library_version"
-    implementation "com.android.support:support-annotations:$support_library_version"
-    implementation "com.android.support:customtabs:$support_library_version"
-
-    implementation 'com.android.support.constraint:constraint-layout:1.1.2'
-
-    implementation 'com.android.support:multidex:1.0.2'
-
-    standardImplementation 'com.google.firebase:firebase-core:11.8.0'
```
jmfayard: Note: I have a strange error

```
 /Users/jmfayard/Dev/try/tachiyomi/app/src/main/res/values/themes.xml: AAPT: error: resource style/PreferenceThemeOverlay.Material (aka eu.kanade.tachiyomi.debug:style/PreferenceThemeOverlay.Material) not found.
```

Is that something you are familiar with? Did I remove a dependency?

build.gradle
```diff
@@ -1,20 +1,22 @@
...
-        classpath 'com.github.ben-manes:gradle-versions-plugin:0.17.0'
         classpath 'com.github.zellius:android-shortcut-gradle-plugin:0.1.2'
         classpath 'com.google.gms:google-services:3.2.0'
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
 }
+plugins {
+   id("jmfayard.github.io.gradle-kotlin-dsl-libs") version "0.2.4"  // $ ./gradlew syncLibs
+}
```
jmfayard: I removed the `gradle-versions-plugin` not because it isn't good, but because my own plugin already includes and extends it!

app/build.gradle
inorichi: Yes, this one: `implementation 'com.github.inorichi:conductor-support-preference:27.0.2'` points to `conductor`.

inorichi:
How do you handle updates if the files are autogenerated?

The project looks cool but I don't want to touch the build system of the `master` branch, but I may include it on `1.x`. And I'll probably use kotlin scripts for gradle. Do you know if they are stable enough yet?

jmfayard:
@inorichi 
> The project looks cool but I don't want to touch the build system of the master branch, but I may include it on 1.x.
Sure, no pb

> And I'll probably use kotlin scripts for gradle. Do you know if they are stable enough yet?

I use Kotlin Gradle scripts for half of my projects.
They are usable and point to a direction that I like,
but both Gradle itself and the Android Studio of it have rough edges. 
And it's a much larger step to take that to just migrate the build logic to the `buildSrc` module like I'm doing here. 
IMHO I wouldn't do it for now and would wait until you are doing some larger changes to your Gradle build, like if you want to split your codebase in multiple repositories. 
Then yes, you will have to do a lot of editing and the Gradle Kotlin DSL is worth it. 

> How do you handle updates if the files are autogenerated?

That's a good question!
The `Versions` file is *both auto-generated and manually edited*

This is possible because since the plugin is applied to your build, it knows and will never change automatically the dependency version that you are using. 
I think that updating the dependency should only be done manually by the developper. 

But if you re-run the script anytime, it will be regenerated in the sense that the comment will be updated to tell you what is the newest version available.

jmfayard:
@inorichi this is how I would get started! https://github.com/inorichi/tachiyomi/pull/1690

PR #1690, opened 2018-10-29
Setup buildSrc repository with shared build logic

jmfayard:
Follow-up from this pull-request
https://github.com/inorichi/tachiyomi/pull/1672

See TODOs in ./build.gradle

Also update Gradle and the Android Studio Gradle plugin to the latest stable version


buildSrc/src/main/kotlin/Libs.kt
```diff
@@ -0,0 +1,220 @@
+import kotlin.String
```
inorichi: Is this import really needed?

buildSrc/src/main/kotlin/Libs.kt
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * Generated by [gradle-kotlin-dsl-libs](https://github.com/jmfayard/gradle-kotlin-dsl-libs)
+ *
+ * Run again
+ *   `$ ./gradlew syncLibs`
+ * to update this file */
+object Libs {
+    /**
+     * [appcompat website](http://developer.android.com/tools/extras/support-library.html) */
+    const val appcompat: String = "androidx.appcompat:appcompat:" + Versions.appcompat
```
inorichi: `: String` shouldn't be needed.

inorichi:
I didn't ask last time... Does autocompletion work on the default groovy scripts?

buildSrc/src/main/kotlin/Libs.kt
jmfayard: No, looks like a small issue in https://github.com/kotlintest/kotlintest

PR #1713, opened 2018-11-08
Add 'new' label to library view

PusherOfCommits:
Adds a 'new' label to manga in the library view if it has a chapter uploaded in the past week
Label disappears when latest chapter is older than configurable timeframe or latest chapter has been read
Allows sorting library by latest upload

I made this change as I wanted to easily be able to see when a series I'm following has a new chapter. Now they are clearly marked and show up at the top of the library page. Sort by latest update didn't suit this as a manga is considered updated even if a chapter is deleted.

image
image
image
image


CarlosEsco:
Reason color is same as downloaded badge color?

PusherOfCommits:
> Reason color is same as downloaded badge color?

Red suits the label best in my opinion, it stands out
I'd rather change the download badge colour than the new badge, but I'm open to different opinions

hXtreme:
Perhaps change the download badge colour to green?

allbombson:
This makes my library take ages to load.

arkon:
"Last updated" now behaves differently, so I'm going to close this.

cf. https://github.com/inorichi/tachiyomi/commit/ee8a53188c77d60b8105d1e6dcaedf35c2260ec8

PR #1714, opened 2018-11-09
1.x - add new manga status

CarlosEsco:
add publishing finished status for sources that indicate this but the manga is not completed.

### DIFF
--- a/source-api/src/main/java/tachiyomi/source/model/MangaInfo.kt
+++ b/source-api/src/main/java/tachiyomi/source/model/MangaInfo.kt
@@ -20,5 +20,6 @@ data class MangaInfo(
     const val ONGOING = 1
     const val COMPLETED = 2
     const val LICENSED = 3
+    const val PUBLISHING_FINISHED = 4
   }
 }


inorichi:
Do you have examples of sites implementing this?

CarlosEsco:
Mangadex is only one i know.
Here is a sample.  Publishing is complete but scanlators have not scanned all chapters. 
https://mangadex.org/title/17291

inorichi:
I can only see "completed". Are you calculating it by parsing chapter number and comparing with the reading progress field?

CarlosEsco:
so completed on the site is only for publishing status.  when a manga is actually completely scanned it has a tag on the chapter.  So for example.
https://mangadex.org/title/558/fairy-tail

if we marked it as complete like in 0.x  the setting of dont update completed manga would ignore the Red Manga even if new chapters were coming out.  

In 1.x  once a chapter is tagged and the publishing status is completed we could mark it to the traditional completed.

inorichi:
I see. In the meantime you could mark it as ongoing.

AmineI:
I think a few mangas have missing chapters on mangadex as well (for example when a scanlator group picks up a dropped manga and skip chapters), and sometimes the last chapter is available but not many previous ones. Is the "END" tag displayed in this case ? If so, wouldn't that break the detection if you only use the tag to check that out @CarlosEsco ?

CarlosEsco:
@AmineI It would be broke but we cant be 100% all the time

AmineI:
> @AmineI It would be broke but we cant be 100% all the time

Ok ! At least it would greatly reduce the number of cases where you need to disable "only update ongoing mangas"

CarlosEsco:
maybe 1.x should have an option to select individual manga to check for updates besides categories.

AmineI:
> maybe 1.x should have an option to select individual manga to check for updates besides categories.

So manually marking a manga as "PUBLISHING_FINISHED" for sites that unlike MangaDex don't support end tags ? (and eventual special cases like the one mentioned above)
A switch on the manga's page when it's status is "COMPLETED" maybe ?

PR #1715, opened 2018-11-09
1.x - add DateAdded to Manga

CarlosEsco:
add dateAdded field.  This will allow future library view to sort by added date.

Should this be lateint?  to make setting it mandatory?  Or does the interactor ChangeMangaFavorite need to create a new manga object with the date when added to library.


inorichi:
Merging for now, I'll consider later how to save this field.

PR #1726, opened 2018-11-12
Attempt 2 on fixing bottom layout and updating text

NoodleMage:


NoodleMage:
Why does it suddenly work xD. The amazing power of Git

PR #1772, opened 2018-12-18
Update ranking

hXtreme:
The app checked for updates in lexicographic order of title names. This is fine when the update list is small but, as soon as the list becomes long, series later in the lexicographic order say (Zannen Jokanbu Black General-san) starting with the letter 'Z' takes a lot of time as the previous implementation has to go through hundreds of other manga that don't even get updates that frequently/are completed before it can get to 'Z'.

This pull request adds a heuristic to decide the update order of the manga to update active series before checking updates to inactive ones. Furthermore, it provides an easy way to, in the future use more complicated heuristics.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
@@ -1,0 +1,41 @@
+package eu.kanade.tachiyomi.data.library
+
+import java.util.Comparator
+
+import eu.kanade.tachiyomi.data.database.models.Manga
+
+/**
+ * This class will provide various functions to Rank mangas to efficiently schedule mangas to update.
+ */
+internal class LibraryUpdateRanker {
+    companion object {
+        /**
+         * Provides a total ordering over all the Mangas.
+         *
+         * Assumption: An active [Manga] mActive is expected to have been last updated after an
+         * inactive [Manga] mInactive.
+         *
+         * Using this insight, function returns a Comparator for which mActive appears before mInactive.
+         * @return a Comparator that ranks manga based on relevance.
+         */
+        fun relevanceRanking(): Comparator<Manga> {
+            return Comparator { mangaFirst: Manga,
+                                mangaSecond: Manga ->
+                compareValues(mangaSecond.last_update, mangaFirst.last_update)
+            }
+        }
+
+        /**
+         * Provides a total ordering over all the Mangas.
+         *
+         * Order the manga lexicographically.
+         * @return a Comparator that ranks manga lexicographically based on the title.
+         */
+        fun lexicographicRanking(): Comparator<Manga> {
+            return kotlin.Comparator { mangaFirst: Manga,
+                                       mangaSecond: Manga ->
+                compareValues(mangaFirst.title, mangaSecond.title)
+            }
+        }
+    }
+}
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
@@ -9,6 +9,9 @@ import eu.kanade.tachiyomi.data.database.models.Manga
  */
 internal class LibraryUpdateRanker {
     companion object {
+        val rankingScheme = listOf(
+                (this::lexicographicRanking)(),
+                (this::relevanceRanking)())
         /**
          * Provides a total ordering over all the Mangas.
          *
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
@@ -1,44 +1,43 @@
 package eu.kanade.tachiyomi.data.library
 
-import java.util.Comparator
-
 import eu.kanade.tachiyomi.data.database.models.Manga
 
 /**
  * This class will provide various functions to Rank mangas to efficiently schedule mangas to update.
  */
-internal class LibraryUpdateRanker {
-    companion object {
-        val rankingScheme = listOf(
-                (this::lexicographicRanking)(),
-                (this::latestFirstRanking)())
-        /**
-         * Provides a total ordering over all the Mangas.
-         *
-         * Assumption: An active [Manga] mActive is expected to have been last updated after an
-         * inactive [Manga] mInactive.
-         *
-         * Using this insight, function returns a Comparator for which mActive appears before mInactive.
-         * @return a Comparator that ranks manga based on relevance.
-         */
-        fun latestFirstRanking(): Comparator<Manga> {
-            return Comparator { mangaFirst: Manga,
-                                mangaSecond: Manga ->
-                compareValues(mangaSecond.last_update, mangaFirst.last_update)
-            }
-        }
+object LibraryUpdateRanker {
 
-        /**
-         * Provides a total ordering over all the Mangas.
-         *
-         * Order the manga lexicographically.
-         * @return a Comparator that ranks manga lexicographically based on the title.
-         */
-        fun lexicographicRanking(): Comparator<Manga> {
-            return kotlin.Comparator { mangaFirst: Manga,
-                                       mangaSecond: Manga ->
-                compareValues(mangaFirst.title, mangaSecond.title)
-            }
+    val rankingScheme = listOf(
+            (this::lexicographicRanking)(),
+            (this::latestFirstRanking)())
+
+    /**
+     * Provides a total ordering over all the Mangas.
+     *
+     * Assumption: An active [Manga] mActive is expected to have been last updated after an
+     * inactive [Manga] mInactive.
+     *
+     * Using this insight, function returns a Comparator for which mActive appears before mInactive.
+     * @return a Comparator that ranks manga based on relevance.
+     */
+    fun latestFirstRanking(): Comparator<Manga> {
+        return Comparator { mangaFirst: Manga,
+                            mangaSecond: Manga ->
+            compareValues(mangaSecond.last_update, mangaFirst.last_update)
         }
     }
+
+    /**
+     * Provides a total ordering over all the Mangas.
+     *
+     * Order the manga lexicographically.
+     * @return a Comparator that ranks manga lexicographically based on the title.
+     */
+    fun lexicographicRanking(): Comparator<Manga> {
+        return Comparator { mangaFirst: Manga,
+                                   mangaSecond: Manga ->
+            compareValues(mangaFirst.title, mangaSecond.title)
+        }
+    }
+
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
@@ -11,7 +11,7 @@ internal class LibraryUpdateRanker {
     companion object {
         val rankingScheme = listOf(
                 (this::lexicographicRanking)(),
-                (this::relevanceRanking)())
+                (this::latestFirstRanking)())
         /**
          * Provides a total ordering over all the Mangas.
          *
@@ -21,7 +21,7 @@ internal class LibraryUpdateRanker {
          * Using this insight, function returns a Comparator for which mActive appears before mInactive.
          * @return a Comparator that ranks manga based on relevance.
          */
-        fun relevanceRanking(): Comparator<Manga> {
+        fun latestFirstRanking(): Comparator<Manga> {
             return Comparator { mangaFirst: Manga,
                                 mangaSecond: Manga ->
                 compareValues(mangaSecond.last_update, mangaFirst.last_update)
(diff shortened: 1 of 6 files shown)


metroid112:
This looks extremely useful and I like the heuristic (Since the vast majority of manga updates on a consistent basis, weekly or monthly). Plus with auto-download on, you get the benefit of checking manga faster and start the download earlier (Example: If you have 100 manga and the only one with consistent updates is the last one lexicographically, like Zannen Jokanbu Black General-san, it would take 99 checks for updates before the download even starts). Also the worst case scenario is that it takes the same amount of time updating.

Also as much as I love watching this git I'm not well versed in kotlin, what's the expected result of the comparator when the last update dates are the same?

hXtreme:
> Also as much as I love watching this git I'm not well versed in kotlin, what's the expected result of the comparator when the last update dates are the same?

I too have almost no prior experience with kotlin/android development so I can't be super sure but to the best of my knowledge, dates are stored to the accuracy of seconds so, last update dates are almost guaranteed to not be the same.

What I know as a fact is that in the off chance that two dates are the same, comparator wold return a 0, indicating no particular preference between those two mangas. However, I believe that on sorting, the item appearing earlier in the lexicographic order would be processed first (if kotlin's sort function is implemented the same as Java's).

inorichi:
I'm not sure it'll be a big change. If I understood this correctly, the next update will start with the manga that have been recently updated, and most probably those won't have a new update. IMO categories should be used for manga that almost never update, also the setting to only look for updates for ongoing manga should be enabled.

hXtreme:
The main idea behind Relevance Ranking scheme is that there would be hundreds of non-completed manga that get released sporadically sometimes not for months, there would be a small set of manga that gets updates weekly or bi-weekly. Furthermore there are a number of manga that don't get updates and have a not complete status because the source didn't update the status after it was axed or it was dropped by a scanlation group.

So what the relevance Ranking scheme does is have everything that was updated recently (say in the previous week) be checked before everything that didn't receive updates in the last few months/years.

hXtreme:
I would recommend you to test it out yourself for a few weeks, it works pretty well.

We could also just set it up to use the lexicographical ranking scheme and just leave the framework for people to come up with even smarter ranking scheme. Which I am thinking about already but, nothing is as clean as relevance Ranking scheme and would require storing more information about past few update dates.

inorichi:
I didn't have time to check this, but I'm okay with merging as long as it's opt-in. It shouldn't be too much work to add a setting.

hXtreme:
I might be able to add the settings option in the week after, currently I have a bunch of exams and assignments so won't be able to do it right away.

hXtreme:
@inorichi I've added the settings UI. The last commit sets the latestFirstRanking scheme (previously called relevanceRanking) is the default option with the option to opt-out. If you feel that latestFirstRanking should not be made default then we could just roll-back the last commit and then merge it.

I would also really appreciate if someone could go through and review the changes for stylistic consistency, I've tried to follow the style I saw in other commits and the codebase but as I may have mention multiple times, this is pretty much my first time working with Kotlin.

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
```diff
@@ -9,6 +9,7 @@ import com.f2prateek.rx.preferences.RxSharedPreferences
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.data.track.TrackService
 import eu.kanade.tachiyomi.source.Source
+import kotlinx.coroutines.experimental.channels.NULL_VALUE
```
inorichi: Weird import

app/src/main/res/values/strings.xml
```diff
@@ -129,6 +129,10 @@
     <string name="update_monthly">Monthly</string>
     <string name="pref_library_update_categories">Categories to include in global update</string>
     <string name="all">All</string>
+    <string name="pref_library_update_prioritization">Prioritize library updates</string>
```
inorichi: Hm... I think this is quite misleading. IMO it should be something like "Library update sorting". And instead of A-Z/Latest First I'd go with the already defined strings as found in the library's drawer (Alphabetically/Last updated)

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -205,6 +206,9 @@ class LibraryUpdateService(
         subscription = Observable
                 .defer {
                     val mangaList = getMangaToUpdate(intent, target)
+                            .sortedWith(
+                                    rankingScheme[preferences.libraryUpdatePrioritization().getOrDefault()]
```
inorichi: Small nitpick: add a local variable (before `mangaList`) and you won't need those line breaks.

app/src/main/res/values/strings.xml
hXtreme: While "Library update sorting" makes a lot more sense internally, because we are simply sorting things. However, on the goal/user perspective what we are doing is setting a prioritization on what is to be checked for updates first and, because this is part of the UI element I went for "Prioritize library updates".

For the time being I'll make the changes for A-Z and Latest First but wait for your final call on "Prioritize library updates"

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
```diff
@@ -0,0 +1,44 @@
...
+         * Provides a total ordering over all the Mangas.
+         *
+         * Assumption: An active [Manga] mActive is expected to have been last updated after an
+         * inactive [Manga] mInactive.
+         *
+         * Using this insight, function returns a Comparator for which mActive appears before mInactive.
+         * @return a Comparator that ranks manga based on relevance.
+         */
+        fun latestFirstRanking(): Comparator<Manga> {
+            return Comparator { mangaFirst: Manga,
```
arkon: Why is one a Java comparator and the other a Kotlin comparator?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
```diff
@@ -159,6 +159,22 @@ class SettingsGeneralController : SettingsController() {
...
+            // ../../data/library/LibraryUpdateRanker.kt
+            entriesRes = arrayOf(
+                    R.string.lexicographic,
+                    R.string.latest_first
+            )
+            entryValues = arrayOf(
+                    "0",
+                    "1"
+            )
+            defaultValue = "1"
```
arkon: Why is the default here `1`, but the default in the helper class `0`?

app/src/main/res/values/strings.xml
arkon: "Library update order" might make more sense.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * This class will provide various functions to Rank mangas to efficiently schedule mangas to update.
+ */
+internal class LibraryUpdateRanker {
+    companion object {
+        val rankingScheme = listOf(
+                (this::lexicographicRanking)(),
+                (this::latestFirstRanking)())
+        /**
```
arkon: Nitpick: add an empty line before this.

hXtreme:
Should I refactor all places I'm using rank/ranking to instead use prioritize/prioritizing ? I that case should I also rename the file LibraryUpdateRanker.kt to LibraryUpdatePrioritizer?

app/src/main/res/values/strings.xml
hXtreme: I like "Library update order".

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
hXtreme: Fixing that, this was mostly because I was confused by all the stuff my android studio was telling me during my first try at coding in Kotlin.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
hXtreme: Fixing that!

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
hXtreme: Adding a line before and after the definition of rankingScheme

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
hXtreme: New commit should fix this.

app/src/main/res/values/strings.xml
hXtreme: Now using preexisting strings.

However, uses the name "Library update order" instead of "Library update sorting".

Let me know if this isn't alright.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.Comparator
+
+import eu.kanade.tachiyomi.data.database.models.Manga
+
+/**
+ * This class will provide various functions to Rank mangas to efficiently schedule mangas to update.
+ */
+internal class LibraryUpdateRanker {
+    companion object {
```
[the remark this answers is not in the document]
hXtreme: Done!

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
hXtreme: Removed.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
```diff
@@ -159,6 +159,22 @@ class SettingsGeneralController : SettingsController() {
                             selectedCategories.joinToString { it.name }
                     }
         }
+        intListPreference{
+            key = Keys.libraryUpdatePrioritization
+            titleRes = R.string.pref_library_update_prioritization
+            // The following arrays are to be lined up with the list rankingScheme in:
```
arkon: This feels like code smell. I think it'd make more sense if the preference helper actually returned the comparator based on more meaningful entry values instead of arbitrarily relying on list orders.

app/src/main/res/values/strings.xml
```diff
@@ -129,6 +129,8 @@
     <string name="update_monthly">Monthly</string>
     <string name="pref_library_update_categories">Categories to include in global update</string>
     <string name="all">All</string>
+    <string name="pref_library_update_prioritization">Library update order</string>
+    <string name="pref_library_update_prioritization_summary">Select the order in which Tachiyomi checks for update</string>
```
arkon: Shouldn't the summary actually be the chosen item, similar to other the other preferences?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
hXtreme: I was thinking about the same thing and was wondering if I should make `rankingScheme` a key value type of data structure that way we wouldn't have to keep things in order everywhere.

app/src/main/res/values/strings.xml
hXtreme: The reason I went for giving an explanation instead of the chosen value is because just the name "Library update order"(now) seems inadequate as a description of what this option means, but for the sake of consistent UI I will change it promptly.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
inorichi: The preferences helper always return simple types so I'd prefer not to change that. If needed you could make it an additional class or add a method in the Ranker object to convert an int (from the preference) to the comparator.

app/src/main/res/values/strings.xml
inorichi: That's fine.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
hXtreme: @inorochi my idea was to make preference helper return a string and turn rankingScheme into a map where the keys would be strings (returned by the preference) and values would be the comparison functions.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
inorichi: Please change the defaults to 0. Let the users opt-in to this alternative instead of forcing everyone.

inorichi:
Merging once you change the default preference value.

Issue #1720, opened 2018-11-10
[Feature Request]  Swipe Side of Page to Change Brightness


coltaku:
It would be awesome if you could add the ability to change the brightness inside the app with swiping up and down on the left (or right) side of the page, on the edge of the screen. This would be super convenient and would honestly work perfectly with this app on my opinion. This feature would make this app nearly perfect.

dragneelfps:
This looks interesting.

that-one:
But please, make it optional, maybe a toggle in the brightness& colour window or in the reader settings.

coltaku:
Yeah, a toggle switch in the brightness/color filter tab would be perfect. I'm glad other people would also enjoy this feature being added.

PR #1777, opened 2018-12-26
Add brightness change by swiping the left edge.

Karkoon:
This is related to issue https://github.com/inorichi/tachiyomi/issues/1720
Added brightness change by swiping the left edge in reader activity.
Added enable/disable switch in reader_color_filter.xml

The brightness controls work on the bar on the left.

image

This is how the color filter sheet looks now. The added switch is disabled when custom brightness is disabled. I'm not sure if I should mix their code like I did.

image

I'm sorry for the PreferenceKeys file. I don't why it's marked as if I deleted everything and added it again.
I'm not sure what to do with the translations. I only added the english version.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
@@ -23,6 +23,8 @@ object PreferenceKeys {
 
     const val customBrightnessValue = "custom_brightness_value"
 
+    const val customBrightnessBySwiping = "custom_brightness_by_swiping"
+
     const val colorFilter = "pref_color_filter_key"
 
     const val colorFilterValue = "color_filter_value"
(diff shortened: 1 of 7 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
```diff
@@ -1,122 +1,124 @@
...
+
+    const val fullscreen = "fullscreen"
+
+    const val keepScreenOn = "pref_keep_screen_on_key"
+
+    const val customBrightness = "pref_custom_brightness_key"
+
+    const val customBrightnessValue = "custom_brightness_value"
+
+    const val customBrightnessBySwiping = "custom_brightness_by_swiping"
```
Karkoon: This is the only change in this file

inorichi:
It's probably because you changed the line endings. About translations don't worry about it.

arkon:
@Karkoon Would you be able to update this PR?

Issue #1774, opened 2018-12-21
Show List of manga from anilist & kitsu 


as280093:
**App version:**

**Issue/Request:**
   Request

**Steps to reproduce (if applicable)**

 1. Show all my Manga list from anilist & kitsu in different page
 2. Click on list then search through manga sources
 3. then add to library 

**Other details:**
i have made some images in photoshop 
side panel  : [https://i.imgur.com/D9x4nFE.png](https://i.imgur.com/D9x4nFE.png)
main screen : [https://i.imgur.com/qGWMOE4.png](https://i.imgur.com/qGWMOE4.png)

this will help to view all the list from anilist.co and kitsu.io without adding each one by one to library.  it will help for backup too (partial backup)

![screenshot_2018-12-21-20-44-30-337_eu](https://user-images.githubusercontent.com/25240839/50353380-1cd87500-056e-11e9-9f5d-fee2986ef98d.png)
![screenshot_2018-12-21-20-47-24-964_com1](https://user-images.githubusercontent.com/25240839/50353381-1cd87500-056e-11e9-8cd7-faadbdcaa3db.png)

Flat:
This seems out of the scope of a manga reading application. Tachiyomi is not a kitsu/anilist/mal client.

Flat:
Even if we show the list in the app you will still have to manually track and add each one, adding the additional step of having to back out of the list to search a catalog.

wopian:
Yes, but it removes the need of having to get the name of, then search by pasting/typing the name for hundreds of manga. It's why I still don't have all of what I (was) reading added to Tachiyomi over 2 years later.

as280093:
@CarlosEsco :joy: no need for that this will be the end 
its not posting app only manga reading app 
so only manga content

CarlosEsco:
But why can I not post on the manga page for anilist or kitsu about  the manga I am reading?

wopian:
That would be a separate issue that can be discussed/closed for being out of scope. As it stands this is relevant as it has the potential of making adding multiple manga to Tachiyomi from tracking services so much quicker, as users would no longer need to search for each individual manga in their library.

I don't know if the navigation drawer is the best place for it, but of course they could only be listed if you have an account linked.

Z3ROGRAV1TYx:
I think this would be a nice feature to have

as280093:
so anyone agreed on adding this feature

as280093:
[...]

PR #1787, opened 2019-01-03
Added search intent handler and Google Search Action, for the global search

AmineI:
- [x] Allows to launch a global search for a manga, from an external app/shortcut/command, with the standard search intent. 

     It can be used to speed up/automate manga searches from a known list, with automation apps (like Tasker/Automate). I thought of this with #1774, it could help mass import tracking lists - you could automate the search of items of your AniList for example.

- [x] Also added support for Google Search Actions
    Anyone can now say or type "Search *manga* in Tachiyomi", either to Google Assistant or directly on the Google app, and be redirected to the search results in Tachiyomi

Note about future UI improvements, 1.x or later :
Currently Tachiyomi's "global search" is made to search items in catalogs and not in the library. I believe it would be more relevant for a Google Search to show a search activity with Library results on top, and then Catalog results - basically, a global "global search to satisfy both the users searching something in their library, and the ones looking to add a new manga.
Since I think you've mentioned some UI changes for 1.x, I'll wait until then to start thinking about that

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -29,8 +29,10 @@
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.SEARCH" />
+                <action android:name="com.google.android.gms.actions.SEARCH_ACTION"/>
+                <category android:name="android.intent.category.DEFAULT"/>
             </intent-filter>
-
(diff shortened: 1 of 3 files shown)


KingRecycle:
Came to see if this feature exists after wishing tachiyomi would look at my plan to read and download the manga. Hope this gets merged in.

PR #1816, opened 2019-01-25
add sdcard1 path to local library path if exist. Because some custom roms mount external sdcard in path /storage/sdcard1 as R/W

Fahad1998:


app/src/main/java/eu/kanade/tachiyomi/util/DiskUtil.kt
```diff
@@ -27,6 +27,18 @@ object DiskUtil {
         return size
     }
 
+    fun altExtStoragePath(): String {
+        val extdir = File("/storage/sdcard1")
+        var path = ""
+        if (extdir.exists() && extdir.isDirectory) {
+            val contents = extdir.listFiles()
```
inorichi: Why do you need to list files?

app/src/main/java/eu/kanade/tachiyomi/util/DiskUtil.kt
```diff
@@ -27,6 +27,18 @@ object DiskUtil {
         return size
     }
 
+    fun altExtStoragePath(): String {
```
inorichi: I don't like this being part of the public API when it won't be used anywhere else.

app/src/main/java/eu/kanade/tachiyomi/util/DiskUtil.kt
```diff
@@ -44,7 +56,15 @@ object DiskUtil {
                     }
                 }
 
-        if (Build.VERSION.SDK_INT < 21) {
+        var altExt = altExtStoragePath()
+        if (altExt !in directories.toString()){
```
inorichi: If that method returns a `File?` you don't have to do this ugly check, which is creating a string of a list and comparing with that.

app/src/main/java/eu/kanade/tachiyomi/util/DiskUtil.kt
```diff
@@ -44,7 +56,15 @@ object DiskUtil {
...
+
+            if (altExt !== "") {
```
inorichi: Likewise, not needed if that returns a `File?`. Also strings should be compared with `==` not `===`. In this case it may have worked because the empty string is a special case on the VM and may use a unique instance.

inorichi: btw, directories is a `Set`, so you don't have to check if it has been added already.

PR #1842, opened 2019-02-08
[1.x] Implement ratelimiting for sources

Flat:
This will allow sources to specify a rate limit to chain. Requests that are over the rate limit will be dropped and an error returned.


common/src/main/kotlin/http/RatelimitInterceptor.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package tachiyomi.core.http
+
+import okhttp3.Interceptor
+import okhttp3.Response
+import prefs.RateBucketPreferences
+import tachiyomi.core.di.AppScope
+
+class RatelimitInterceptor(
```
inorichi: Make this class receive a `PreferenceStore` as parameter, then you can pass the store assigned to each source through `deps.preferences`

core/src/main/kotlin/prefs/RateBucketPreferencesProvider.kt
```diff
@@ -0,0 +1,34 @@
...
+import android.content.Context
+import tachiyomi.core.prefs.SharedPreferencesStore
+import toothpick.ProvidesSingletonInScope
+import javax.inject.Inject
+import javax.inject.Provider
+import javax.inject.Singleton
+
+@Singleton
+@ProvidesSingletonInScope
+internal class RateBucketPreferencesProvider @Inject constructor(
```
inorichi: You won't need this class anymore.

common/src/main/kotlin/utils/RateBucket.kt
```diff
@@ -0,0 +1,73 @@
...
+ *
+ */
+
+package utils
+
+import java.util.Calendar
+
+class RateBucket(internal var capacity: Int, internal var refillRate: Long) {
+  internal var tokens: Int = capacity
+  internal var refillTime = Calendar.getInstance().timeInMillis
```
inorichi: Better use `System.currentTimeMillis()`.

common/src/main/kotlin/utils/RateBucket.kt
```diff
@@ -0,0 +1,73 @@
...
+      false
+    }
+  }
+
+  fun getTokens(): Int {
+    return this.tokens
+  }
+
+  private fun refill() {
+    val now = Calendar.getInstance().timeInMillis
```
inorichi: `System.currentTimeMillis()`

common/src/main/kotlin/utils/RateBucket.kt
```diff
@@ -0,0 +1,73 @@
...
+ *
+ *  * Copyright (C) 2018 The Tachiyomi Open Source Project
+ *  *
+ *  * This Source Code Form is subject to the terms of the Mozilla Public
+ *  * License, v. 2.0. If a copy of the MPL was not distributed with this
+ *  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ */
+
+package utils
```
inorichi: `package tachiyomi.core.utils`

common/src/main/kotlin/prefs/RateBucketPreferences.kt
```diff
@@ -0,0 +1,29 @@
...
+ *
+ *  * Copyright (C) 2018 The Tachiyomi Open Source Project
+ *  *
+ *  * This Source Code Form is subject to the terms of the Mozilla Public
+ *  * License, v. 2.0. If a copy of the MPL was not distributed with this
+ *  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ */
+
+package prefs
```
inorichi: `package tachiyomi.core.prefs`. Although I'm not sure this should belong to prefs. Maybe move both RateBucket and this to `http`.

common/src/main/kotlin/utils/RateBucket.kt
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.Calendar
+
+class RateBucket(internal var capacity: Int, internal var refillRate: Long) {
+  internal var tokens: Int = capacity
+  internal var refillTime = Calendar.getInstance().timeInMillis
+
+  constructor(capacity: Int, refillRate: Long, tokens: Int, refillTime: Long) : this(
```
inorichi: You can use a single constructor with default parameters instead.

```kt
class RateBucket(
  internal var capacity: Int,
  internal var refillRate: Long,
  internal var tokens: Int = capacity,
  internal var refillTime: Long = System.currentTimeMillis()
)
```

inorichi:
Can you fix the license headers? There are double asterisks.

It should be:

```
/*
 * Copyright (C) 2018 The Tachiyomi Open Source Project
 *
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 */
```

Flat:
Fixed. Not sure how that happened. Gonna just blame android studio.

PR #1913, opened 2019-03-20
[Cloudflare] Fix 503 due to missing value in js challenge.

DitFranXX:
Related issues: inorichi/tachiyomi-extensions#951

From: Anorov/cloudflare-scrape#200


DitFranXX:
just add a note here. 

A hidden value `s` was appears long ago but it is not need to solve a challenge until today.

CarlosEsco:
sneaky sneaky

buzmeg:
Kissmanga is still returning a 503 even with dev build r1471

einstein95:
@buzmeg Did you clear your cookies

Vanaks:
Hi all, after I installed dev build I'm having this error

Screenshot_20190321-142341

I tried to clear cookies, cache, going directly to readcomicsonline and coming back to the app and isn't working, besides that, the app is great

PR #1919, opened 2019-03-22
[Cloudflare] Fix recent CF JS Challenge error that calls DOM

DitFranXX:
From Anorov/cloudflare-scrape#206
Related issues: inorichi/tachiyomi-extensions#951

Recent CF changes research: https://github.com/inorichi/tachiyomi-extensions/issues/951#issuecomment-475467259

Ps. [`atob` package in NPM reaches ~780K downloads/week](https://www.npmjs.com/package/atob)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
@@ -14,6 +14,8 @@ class CloudflareInterceptor : Interceptor {
 
     private val sPattern = Regex("""name="s" value="([^"]+)""")
 
+    private val kPattern = Regex("""k\s+=\s+'([^']+)';""")
+
     private val serverCheck = arrayOf("cloudflare-nginx", "cloudflare")
 
     @Synchronized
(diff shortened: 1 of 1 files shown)


DitFranXX:
Seems they reverted it or not enabled on my side.. I didn't get any `div` now

app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
```diff
@@ -80,4 +90,42 @@ class CloudflareInterceptor : Interceptor {
         }
     }
 
+    companion object {
+        // atob() is browser API, Got this from NPM package, `atob`.
```
arkon: Doesn't Kotlin or Java have their own base64 encoding/decoding functions?

Peuuuur-Noel: You can maybe use Kotlin base64 decode function in Javascript with https://github.com/square/duktape-android#calling-java-from-javascript

DitFranXX: Updated on `eba3e52`

inorichi: I think Okio has a base64 decode function. Please use that if possible. I prefer that to android dependencies

DitFranXX: Updated on de8e950

PR #1927, opened 2019-03-23
Fix bug with update lib and parse chapters

Pavkazzz:
Hey, i found a bug. If source has custom prepareNewChapter -  this method: https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/util/ChapterSourceSync.kt#L54 will override this in library update.


inorichi:
Fine, but the recognition should always run (no `else`)

Pavkazzz:
@inorichi fixed

Issue #1932, opened 2019-03-25
[Design Discrepancy] Download Queue


hXtreme:
Of all the items in the Nav Bar under the group "@+id/group_feature" the option Download queue is treated differently in the following manner:

1. It is not check-able
2. In MainActivity.kt it is treated differently leading to different transition animation and different back button behavior.

While it is the same with settings, it makes a lot more sense to have different behavior for settings as it is a completely different thing than reading/managing manga.

inorichi:
As explained in the PR, this will still use an activity. Feel free to submit another PR to override the default transition with a fade, if it's too ugly.

PR #1933, opened 2019-03-25
Fix the Design discrepancy with download queue

hXtreme:
Fixes [this issue](https://github.com/inorichi/tachiyomi/issues/1932)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
@@ -83,9 +83,7 @@ class MainActivity : BaseActivity() {
                     R.id.nav_drawer_recently_read -> setRoot(RecentlyReadController(), id)
                     R.id.nav_drawer_catalogues -> setRoot(CatalogueController(), id)
                     R.id.nav_drawer_extensions -> setRoot(ExtensionController(), id)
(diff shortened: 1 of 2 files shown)


hXtreme:
Please let me know if you have any questions or queries.

inorichi:
This was due to the downloads notification opening multiple fragments. With the activity it was easier to only allow one instance, plus easier to handle the intent.

If the transition is a concern, we can override the activity's transition with a fade.

hXtreme:
More so than the transition, the back button behavior is a bit more irritating. For all other options, back button would take you to My Library but not for Download Queue. Also, I couldn't understand what you meant by multiple fragments. If yo could tell me more about it then I may be able to work a bit on that, hopefully leading to something.

lenyer21:
Hey newbie here....Can anyone can tell me why the KissAnime and mangafox are not functioning in tachiyo

hXtreme:
@lenyer21  this is not exactly the place for this question I'd suggest that you go to tachiyomi-extensions however, if I have to take a guess it's probably because of some new security feature they implemented.

inorichi:
There's an intent handler that opens the download activity when you tap the downloads notification. I believe that'll be broken with this patch, as the intent now has to be handled by the `MainActivity`.

inorichi:
I'll keep this the current way, it might be ugly but more changes than this are required.

PR #1936, opened 2019-03-27
add tachiyomi://search/... intent-filter to prepopulate searchbox

dexmanga:
This gets used by inorichi/tachiyomi-extensions#974

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -28,6 +28,16 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
 
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data
+                    android:scheme="tachiyomi"
+                    android:host="search"
+                    android:pathPattern="/..*/..*"/>
+            </intent-filter>
+
             <!--suppress AndroidDomInspection -->
             <meta-data android:name="android.app.shortcuts" android:resource="@xml/shortcuts"/>
         </activity>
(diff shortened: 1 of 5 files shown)


inorichi:
On 1.x I added a loading screen that directly opens the `MangaController` once the manga is added to db.

Using the search function works but tbh I don't like that a tap on a manga in the webpage opens the search screen... I'm wondering whether to port that feature here, because it's a nice addition and 1.x is going to take a while.

inorichi:
Another thing I forgot. Since an extension provides  multiple sources, you have to select the language. We'd need another  screen to select the source that should handle the link... 1.x doesn't  have this issue since extension-sources are 1-1.

dexmanga:
Thoughts on tachiyomi::/search/en/MangaDex/query ?

inorichi:
But then we're assuming everybody uses the english source. Unless only one language is enabled, we have to let the user select the language. That's one of the reasons I hate the 1-N relationship on the current extensions. :disappointed:

AmineI:
One way to avoid the languages issues in the current extensions would be to be able to do a "limited global search", basically global but only with the requested sources, without changing user settings - for example launch a global search in all the enabled mangadex sources only (no matter the language), and then the user would select his own as usual.

I actually had something like that in mind in #1787, but using a [search intent ](https://developer.android.com/reference/android/content/Intent.html#ACTION_SEARCH) instead of a custom url like tachiyomi://search, and that would trigger a global search and not a specific extension's search.
(I think your way with an URL would be easier for users to manually call than an intent though, didn't think of that at the time.)

inorichi:
That could work. Probably the best solution we could have before 1.x

inorichi:
I'll accept this PR if it's implemented as we discussed before, by using the global search.

The extension's activity handle should send to the main process its own package name, which can use later to filter results for that specific extension, and of course only the enabled languages/sources will show results.

This is how the redirecting activity is implemented in 1.x: https://github.com/inorichi/tachiyomi-extensions/blob/1.x/deeplink/src/main/java/tachiyomix/deeplink/SourceDeepLinkActivity.java
Note the `Intent.EXTRA_REFERRER`

If anyone wants to try this go ahead, if no one has assigned this by next week, I'll work on it.

dexmanga:
Sounds like you want to change the `Source` interface to add the package?

inorichi:
There's no need for that, you can check against the `ExtensionManager`.

inorichi:
Anyone working on this? If not I'll probably implement it tomorrow.

inorichi:
Closing in favour of 5619543

Issue #1416, opened 2018-05-12
Color banding issue


Deumiankio:
**App version:**
tachiyomi-dev 0.7.3-1411

**Issue:**
I've noticed an issue with color banding which is most apparent in background gradients.
The issue happens with all tested sources and also occurs when viewing local files.

Testing shows that the issue appears on multiple image formats (tested .jpg and .png) and with all selectable image decoders (Image, Rapid and Skia).
Custom color filter was disabled in each case.

![Color banding in Tachiyomi](https://i.imgur.com/d6wiedZ.jpg)

**Steps to reproduce**

View any image with color gradient in Tachiyomi (for example this gradient image from [here](https://imgur.com/a/CkGSz8R)).

**Other details:**

inorichi:
Banding on the reader won't have a fix because they are decoded using RGB565 and using ARGB8888 is too memory aggresive, which leads to crashes due to out of memory errors.

Maybe this could be a setting on Android Oreo, because as of this version pixels are stored on the native heap so OOM errors are not problem. I'll keep it in mind but for now it won't be added.

Deumiankio:
I can understand that in order to make the app usable on a large number of android devices with a multitude of different specs, some cutting of corners might be necessary.

However, AFAIK ARGB8888 approximately doubles the memory usage compared to RGB565.
On current-gen mobiles with 4GB+ RAM this increase is not a big price to pay for better picture quality.
(Even when reading large high-resolution chapters while downloading others in the background i've never seen tachiyomi use more than 500MB of memory).

I was curious to see how large the performance impact would be.
Seeing as Glide defaults to DecodeFormat.ARGB_8888 since v4, i tried changing the default RequestOptions in TachiGlideModule.kt to PREFER_ARGB_8888.
Unfortunately this did not change anything, presumably because of implied format conversions elsewhere (while downloading/downsampling/saving/cropping the images)?

Anyhow, the best way to approach this would probably be as you said, a setting with which you could switch between RGB565 and ARGB8888.
Then tachiyomi could be used by a large number of budget devices while providing better quality on devices that support a larger memory footprint.

I'm looking forward to seeing a feature like this in a future release.
Until then tachiyomi is still the best manga reader for Android for all series without gradients.
Thank you for creating it.

inorichi:
> ARGB8888 approximately doubles the memory usage compared to RGB565.

Yes, as the numbers say it's 32 bit vs 16 bit per pixel.

[...]

PR #1941, opened 2019-03-27
Add 32-bit color mode to reader settings

Deumiankio:
This adds the option in reader settings to enable ARGB_8888 (24/32-bit color) instead of RGB_565 (16-bit color) and fixes [this issue](https://github.com/inorichi/tachiyomi/issues/1416).

I've tested this change in daily use for a few weeks on a Samsung S8 running Android 9 without a single error or crash.
The previously noticeable color banding is gone, while performance remained largely unchanged.
(Only quick flicks scrolling multiple screens per second in webtoon mode seem a little choppier on 32-bit color compared to 16-bit).

Since images in 32-bit color mode take twice the space in ram compared to before, enabling this option can lead to OutOfMemoryErrors in old android versions and low memory devices, thus the setting is disabled by default.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
@@ -15,6 +15,8 @@ object PreferenceKeys {
 
     const val showPageNumber = "pref_show_page_number_key"
 
+    const val trueColor = "pref_true_color_key"
+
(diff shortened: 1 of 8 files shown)


inorichi:
This setting should only be accessible to devices running Oreo or later. Previous versions stored the bitmap pixels in the heap memory, and that's going to produce many crashes.

Also, to make things easier, I'd say to only enable it from the settings screen, I don't think you'll be changing it often enough to be added to the reader.

PR #1949, opened 2019-03-29
[WIP] new Cloudflare challenge workaround

### LINKED ISSUE #1945: Duktape error for Cloudflare protected catalogues (Japscan, Kissmanga, etc.)
Version : latest (0.8.2)
OS : MIUI 10 (Androïd 8.1)

Description of the error : Duktape error when accessing to Japscan and Kissmanga : com.squareup.duktape.DuktapeException : TypeError : undefined not callable.

DitFranXX:
Seems they have a cycle. `Japscan`'s CF challenge uses now. 
(But site itself is unstable so can't test it properly. Wait for other site's cycle.)

fixes #1945 when workaround works.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
@@ -18,12 +18,12 @@ class CloudflareInterceptor : Interceptor {
 
     private val serverCheck = arrayOf("cloudflare-nginx", "cloudflare")
 
-    private interface IBase64 {
-        fun decode(input: String): String
+    private interface INativeFunctions {
+        fun DecodeBase64(input: String): String
     }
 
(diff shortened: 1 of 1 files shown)


inorichi:
I think we're going to end up needing a WebView to solve the challenges. The problem is that we need an UI component, and that's going to be tricky. I'll probably need to request the permission to draw over other apps and load a 0x0 pixels WebView, but I'm still not sure if it'd work, plus a lot of code rewrite due to things turning async...

I also have a local patch where we could download the challenge resolver from github (we'd push a `cloudflare.jar` to this repo) and we save it to local storage and load the jar from the app.

I really want to avoid the WebView, but I think they're now getting serious with their antibot system (I honestly don't know why they weren't up to now).

inorichi:
If f1f6a2b works fine, this should be a solution really hard to break.

I'm still not sure it's bug free so I'll wait for issues.

inorichi:
Closing since this is not needed anymore.

PR #1975, opened 2019-04-13
add JS injection support to mangaWebViewController

az4521:
this would allow JS injection into the webview from extensions  
i have a thing planned that this would be useful for  
it doesnt break anything


az4521:
ive realised this is absolutely useless and i'm an idiot

PR #1999, opened 2019-04-20
Bugfix. Sharing images with very long name

Pavkazzz:
Fix sharing images, which generate very long image name for sharing. 
Currently ocures on readmanga parser - this manga: http://readmanga.me/neveroiatnye_prikliucheniia_djodjo_chast_8__djodjolion__cvetnaia_versiia_

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
@@ -434,7 +434,7 @@ class ReaderPresenter(
 
         // Build destination file.
         val filename = DiskUtil.buildValidFilename(
-                "${manga.title} - ${chapter.name}".dropLast(200)
+                "${manga.title} - ${chapter.name}".take(225)
         ) + " - ${page.number}.${type.extension}"
 
         val destFile = File(directory, filename)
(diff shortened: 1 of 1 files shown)


AviKav:
"I don't think that function means what you think it means."

Pavkazzz:
@avikav doesn't get what you mean. Now we have a bug with images with very long names, this pull request fix it

AviKav:
```kotlin
val string = "1234567890"
println(string.dropLast(6)) // 1234
```

Pavkazzz:
Yes, in practice you cannot create file with more then 255 length in android

CarlosEsco:
Function drop last n

CarlosEsco:
Meaning it's gonna drop 200 characters Everytime. If you don't believe me check neko save as I had pulled this already.

Pavkazzz:
@carlosesco you absolutely right, what should i use to reduce big string to short one?

AviKav:
@Pavkazzz 
1. `.dropLast(n)` means that the string will be empty if it's length is less than or equal to `n`.
2. There are better workarounds for filename limits that involve breaking the file name onto the path
3. Dropping characters over if the first string's length is over 200 would still leave at least 46 characters left to fill

inorichi:
Use `.take(225)`. That should leave enough room for the page number and extension (we should consider the limit to be 240)

Pavkazzz:
@inorichi That's what i want, thank you!

Issue #2002, opened 2019-04-22
[REQUEST] Multiply Blending for Color Filter


sitalkio:
Just a quick request for the color filter.

Values like  R138 G44 B22 A126 now give this result:
https://i.imgur.com/6qq29Dw.jpg

It would be cool, instead of just applying a semi-transparent color, to use a multiply blend mode and preserve the original blacks like this:
https://i.imgur.com/bJOaqxt.jpg

Deumiankio:
This seems like a good idea.

On older Android versions with rooted devices CF.Lumen used to work fine for this to customize system-wide color filters with various blend modes, but AFAIK there is no way to do this on generic un-rooted devices.
Most color filter apps use overlays similar to the color filter currently implemented in tachiyomi, since IIRC Android does not allow views blending over different contexts.

Allowing the user to configure the blend mode for the color filter allows for better customization as well as better image quality while using the color filter e.g. by preserving blacks as shown in the original post.

sitalkio:
I don't know how is technically done in Android but a similar manga app on F-Droid (Mi Manga Nu) has currently such a filter feature that blends the overlay to preserve blacks.
Of course this it's not so important, personally I'll always use and prefer Tachiyomi over all the other apps, but would be quite a nice touch.

Deumiankio:
The PR has been merged and the new blend modes will be available in the next release.
If you want to try it before that use the [dev version](http://tachiyomi.kanade.eu/latest).

Details and example images are in the linked PR.

To get an effect similar to your second image try "Burn / Darken" with R255 G0 B0 A100" (only available if you're on Android Pie or newer) or "Multiply" with R255 G150 B150 A255 (on all Android versions).

sitalkio:
Thank you so much, it works great!

PR #2013, opened 2019-04-28
Add color filter blend modes

Deumiankio:
This commit adds an user-configurable option to change the color filter blend mode.

In addition to the current default, it adds the following PorterDuff modes:
Overlay, Multiply, Screen, Lighten, Darken

Closes request/issue [#2002](https://github.com/inorichi/tachiyomi/issues/2002).

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
@@ -29,6 +29,8 @@ object PreferenceKeys {
 
     const val colorFilterValue = "color_filter_value"
 
+    const val colorFilterMode = "color_filter_mode"
+
     const val defaultViewer = "pref_default_viewer_key"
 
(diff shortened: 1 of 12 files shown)


CarlosEsco:
You have sample pics of each mode?

Deumiankio:
Sure, [have some pics](https://imgur.com/a/n4eCfqe).

Deumiankio:
After further testing it seems only the Xfermodes Default, Multiply and Screen work as expected on Android API levels 16 to 28, while the modes Darken, Lighten and Overlay behave like Default on levels 16 to 27 but work fine on 28.

It turns out this happens because hardware acceleration support for those modes [was only added in API level 28](https://developer.android.com/guide/topics/graphics/hardware-accel#unsupported).

The easiest solution to that is to only show the modes supported by the currently used API level.

inorichi:
This is cool, thanks!

The base arrays.xml still contains all the xfer modes, though.

Deumiankio:
Thanks for spotting that, it seems I committed an older file by mistake.

PR #2020, opened 2019-05-04
Single tap up/down scrolls up/down on webtoon viewer.

unnurq:
Makes it more intuitive.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonViewer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonViewer.kt
@@ -88,10 +88,10 @@ class WebtoonViewer(val activity: ReaderActivity) : BaseViewer {
             }
         })
         recycler.tapListener = { event ->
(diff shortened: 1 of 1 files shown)


inorichi:
Long ago I was thinking about something like this:
```
+---+---+---+
| P | P | P |   P: Previous
+---+---+---+
| P | M | N |   M: Menu
+---+---+---+
| N | N | N |   N: Next
+---+---+---+
```

How does it sound?

hXtreme:
Kindle devices have something similar so it might be better to mimic them. I'm assuming that they used their massive amounts of usage data to fine tune their UI.

inorichi:
@unnurq, what do you think about my proposal?

0450:
i think having the middle with all menu would be better.

0450:
maybe the L shaped can be an option too.

hXtreme:
Can you please wait a bit to merge this? I've opened a [PR](https://github.com/unnurq/tachiyomi/pull/1) on @unnurq 's fork with a number of additions.

If you (@inorichi ) merges this i can also just open a PR directly to master

arkon:
Closing in favour of #2152

PR #2029, opened 2019-05-11
Change versionCode on debug version

j2ghz:
Would this work?  If so it'd make my job much easier, because my build infrastructure was breaking all the time, to the point I discontinued the fdroid server. This way I could just get the apk's from http://tachiyomi.kanade.eu/latest let's say once an hour and be done with it. I can't do that now, because fdroid requires the version code to change with updates.

The interest in autoupdated dev versions seems to be higher than I thought: https://discordapp.com/channels/349436576037732353/411962173598990340/563415246530805771

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -5,7 +5,7 @@ git fetch --unshallow #required for commit count
 cp .travis/google-services.json app/
 
 if [ -z "$TRAVIS_TAG" ]; then
-    ./gradlew clean assembleStandardDebug
+    ./gradlew clean assembleFdroid-devDebug
 
     COMMIT_COUNT=$(git rev-list --count HEAD)
     export ARTIFACT="tachiyomi-r${COMMIT_COUNT}.apk"
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -77,7 +77,7 @@ android {
             minSdkVersion 21
             resConfigs "en", "xxhdpi"
             dimension "default"
-            versionCode getCommitCount()
+            versionCode getCommitCount().toInteger()
         }
     }
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -79,7 +79,7 @@ android {
         dev {
             minSdkVersion 21
             resConfigs "en", "xxhdpi"
-            dimension "default"            
+            dimension "default"
         }
     }
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -38,7 +38,6 @@ android {
         minSdkVersion 16
         targetSdkVersion 27
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
-        versionCode 41
         versionName "0.8.4"
 
         buildConfigField "String", "COMMIT_COUNT", "\"${getCommitCount()}\""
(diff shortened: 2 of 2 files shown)


arkon:
You can't override the version code like that per flavour, unfortunately. You could modify is so that the top level version code increments in some manner, like generated based on the timestamp.

CarlosEsco:
@arkon you can have flavored versions.  I think you need to remove the high level one.  I have this in Neko and it works fine

https://github.com/CarlosEsco/Neko/blob/master/app/build.gradle#L66-L79

Here's the logcat result
2019-05-11 15:08:45.654 8654-8654/tachiyomi.mangadex.debug D/App: ESCO 1

so you should be able to do something like above but use commit count as the version code

arkon:
> I think you need to remove the high level one.

Ah that makes more sense. 👍

inorichi:
I know it's misleading, but that "dev" flavor isn't meant to be used as the dev version. You'd have to create another flavor, but keep this in mind: https://developer.android.com/studio/build/optimize-your-build#use_static_build_properties

inorichi:
This has the problem that you can't downgrade. If I mess up one commit you've got to uninstall in order to downgrade (or try with `adb install -d`, but it didn't always work for me). 

Would faking the version code with some changes on the fdroid server work? I don't think it'd be too difficult. IIRC it parses the apk with `aapt dump badging`?

j2ghz:
I don't think it would. I wanted this to simplify the fdroid setup because it was breaking too often. The workflow with fdroid server is put the apk into the repo folder and run `fdroid update`. I don't see any space to insert a version chagning logic. The result of that are index files. The problem is there are 4 of them, one xml, one json, and two jars (all equivalent, which one is used depends on the version of fdroid client).
I looked into changing the versionCode of the apk after it was built but it didn't look easy either.

inorichi:
What I meant was updating FDroid's server code just after the dump badging and if the package name matches the debug version, override the parsed version code with the provided in the apk name. Then it'll generate the files the metadata files thinking the apk is version code 15xx while it's still 40 or so. The apk shouldn't be tampered at all.

j2ghz:
Then the user will be prompted to install an update all the time because F-Droid will see a versionCode 40 installed and versionCode 1511 available. So the user updates but the versionCode is still 40, and an update is still available.

Other ideas:
- static versionCode, no autoupdate, but still fdroid so it's easy to update manually
- make versionCode from time
- think of some other versionCode source that will always increment

inorichi:
True, I missed that.

What about supporting updates for dev builds? No Fdroid needed. This was requested long ago but I never thought it was worth it. Now I think I was wrong.

j2ghz:
That would be the best solution

inorichi:
Okay. Anyone wants to work on it? I don't think it's going to be hard.

A query to https://tachiyomi.kanade.eu/latest will return a 302 with this header: 
`Location: tachiyomi-r1512.apk`

For the /latest request, you'd need to build a new OkHttpClient (based on the default one with `.newBuilder()` and set `followRedirects` to false.

Then you parse the number from the header and check it with `BuildConfig.COMMIT_COUNT`.

If it's bigger, you can proceed with the download (you can maybe share some common code with the release's version downloader)

arkon:
I can try taking a stab at it.

inorichi:
Supeseded by #2128

PR #2031, opened 2019-05-13
Auto fit page charset when use Response.asJsoup.

feilongfl:
Fixed the use of asJsoup method to cause garbled characters in non-utf-8 encoded web pages when adapting to non-English websites (eg: GBK)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/util/JsoupExtensions.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/util/JsoupExtensions.kt
@@ -22,5 +22,18 @@ fun Element.attrOrText(css: String): String {
  * @param html the body of the response. Use only if the body was read before calling this method.
  */
 fun Response.asJsoup(html: String? = null): Document {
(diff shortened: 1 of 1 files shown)


inorichi:
I'm a bit worried about merging this one. Do you have some examples of what is happening and what should?

feilongfl:
If you access a web page which not encoded by utf-8, no processing will cause page content garbled. Therefore, this PR adds the processing of the page algorithm by identifying the `charset` setting of the page.

image

More detailed examples I will give later.

feilongfl:
sorry for late response.

this one is a webpage with `gb2312` codec and without this patch.

https://github.com/feilongfl/tachiyomi-extensions/commit/9eba82545fd4c97f7e86b8ed9e63ad3b620b5c54

screen

inorichi:
Ah, I see. Can't the extensions themselves use the appropriate encoding instead of having to parse the charset on any request?

arkon:
I believe this can be done within the extensions as needed.

Issue #65, opened 2016-01-24
Bit of a problem with MAL search and short titled manga


Bloodwork:
Series with short names like Rin or Magi are impossible to find in the integrated search feature. While the Magi has the affix "The Labyrinth of Magic" in MAL, Rin has nothing like it and currently it is impossible to find and update it through the app.

Rin on MAL: http://myanimelist.net/manga/46894/Rin 

A long list of a bunch of stuff that is not Rin:
![screenshot_2016-01-25-00-04-08](https://cloud.githubusercontent.com/assets/7024084/12539291/eb27a852-c2f7-11e5-9645-5d461cf10349.png)

inorichi:
There's nothing I can do there. Those are the results returned by MAL's API. Try looking for the alternative name in the browser and try with that.

Bloodwork:
Unfortunately there is no alternative name. Maybe it is possible to make the list longer than it is currently to include more results?

inorichi:
http://myanimelist.net/modules.php?go=api#animemangasearch

I don't know if it's possible.

inorichi:
There's a way. I could show the list of the user (with a checkbox or something) and it could be selected there. I don't know how to design the dialog to do this in a friendly way though.

inorichi:
Install from F-Droid: https://f-droid.org/repository/browse/?fdid=eu.kanade.tachiyomi

inorichi:
As of 8ca0814 you can search a manga from your list like this: `my:rin`

reconman:
Can you please add some tooltip or any sort of hint so users don't have to look up this feature on github?  

For example you could display a text like "Use my:<title> to search for titles in your MAL profile"  below the search textfield all the time or when the search field is empty. When the search textfield is empty there is plenty of space in the white rectangle dedicated to the search results.

nearbaskara:
omg it helps me :lol 
would be better if you put that <my:<<manga>>> instruction on the app tho.

AmineI:
Just to be sure, regardless of the current MAL API status, was this feature removed ? It's still in the FAQ but I couldn't get it to work nor find the current code of the feature (a quick "getList" search  in the repo returns nothing else than the function 
itself available from MAL's workaround that was just merged, but I could have missed it somehow so I'd prefer to ask)

PR #2042, opened 2019-05-19
Make MAL Tracking Slightly Less Shitty

MCAxiaz:
It's been a year now with no official api in sight, so might as well keep maintaining this workaround.

Not sure what the original intention of prepending a `.` to the beginning of the host was, but it does not set cookies properly. Also issue with using host is that it strips the https scheme, which causes secure cookies to be ignored. Seems like the tracking worked regardless of that bug since I believe the critical cookie that needed to be cleared is not secure. Regardless, this should fix the issue of needing to manually clear cookies when relogging.

Also fixing the track setting showing a green check mark even if it's not actually logged in (due to clearing cookies). Kind of hacky but don't think there's an easy way around it since the preferences ui wasn't designed to account for cookies.

No idea what causes some people to not be able to login even after clearing cookies though. Though it seems like MAL sometimes does force a captcha (e.g. if your account hasn't been logged into for a long time), so most likely this will need to be redone with a webview.


inorichi:
Thanks for working on this. 

> Also fixing the track setting showing a green check mark even if it's not actually logged in (due to clearing cookies). Kind of hacky but don't think there's an easy way around it since the preferences ui wasn't designed to account for cookies.

Yep, it wasn't. But I think we should fix this the other way around. Instead of checking for the cookie, we assume the user is logged in (checking with the preference) and if the cookie isn't present we attempt to login.

MCAxiaz:
I think it would be kind of weird though. I'm thinking of this scenario:
1. I clear cookies.
2. I go to manga and see tracking is now missing.
3. I go to tracking settings, the app now tries to auto login, except I'm not aware of that, so I open the dialog and do a manual login anyways.

Maybe another option is to immediately do an auto login after cookies are cleared?

inorichi:
The idea is that the "isLogged" function only checks for the user/pass preferences. So the tab will always be there even if you clear cookies. Then we either use a network interceptor to check if login is required, or we check the cookie before any network request.

MCAxiaz:
ahh, I had actually overriden isLogged to also check for cookies. I'll revert it. One last concern is it seems like doing manual changes in the tab will show error toasts, but if the login fails during the update when you exit out of the reader, I believe it would ignore all errors? Would adding another toast along the lines of 'MAL authentication failed, please login again" be sufficient as a warning?

MCAxiaz:
Apologies for the massive diffs. Nearly all of them are just moving code around cause it was getting difficult to read. I've commented on the relevant changes.

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyanimelistApi.kt
```diff
@@ -22,89 +22,70 @@ import java.io.InputStreamReader
 import java.util.zip.GZIPInputStream
 
 
-class MyanimelistApi(private val client: OkHttpClient) {
+class MyanimelistApi(private val client: OkHttpClient, interceptor: MyAnimeListInterceptor) {
 
-    fun addLibManga(track: Track, csrf: String): Observable<Track> {
+    private val authClient = client.newBuilder().addInterceptor(interceptor).build()
+
+    fun search(query: String): Observable<List<TrackSearch>> {
```
MCAxiaz: Something I totally forgot that I had fixed, but the original pr accidentally deleted this feature https://github.com/inorichi/tachiyomi/issues/65. Adding it back in.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSearchAdapter.kt
```diff
@@ -1,79 +1,79 @@
...
+        notifyDataSetChanged()
+    }
+
+    class TrackSearchHolder(private val view: View) {
+
+        fun onSetValues(track: TrackSearch) {
+            view.track_search_title.text = track.title
+            view.track_search_summary.text = track.summary
+            GlideApp.with(view.context).clear(view.track_search_cover)
+            if (!track.cover_url.isNullOrEmpty()) {
```
MCAxiaz: Not sure if this was intentional? Seemed weird that missing a cover_url would cause all the other fields to not be updated.

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeList.kt
```diff
@@ -1,141 +1,140 @@
...
+                .toCompletable()
+    }
+
+    override fun logout() {
+        super.logout()
+        preferences.trackToken(this).delete()
+        networkService.cookieManager.remove(HttpUrl.parse(BASE_URL)!!)
+    }
+
+    val isAuthorized: Boolean
```
MCAxiaz: keep the original definition of isLogged, and use this instead

MCAxiaz:
@inorichi 
I wonder why git thinks the entire file changed 🤔 . Seems like there's some line ending inconsistencies among the files.

I'm on Windows with core.autoclrf set to true. I tried changing a bunch of random files and see what diffs git produces. Setting autoclrf to false fixes the ones that are broken, but breaks the ones that were working.

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListInterceptor.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.json.JSONObject
+import java.lang.Exception
+
+class MyAnimeListInterceptor(private val myanimelist: Myanimelist): Interceptor {
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        if (!myanimelist.isAuthorized) {
+            if (!myanimelist.isLogged) throw Exception("MAL Login Credentials not found")
+
+            myanimelist.login(myanimelist.getUsername(), myanimelist.getPassword())
```
inorichi: This call is async. You shouldn't use RxJava here but create a `Request` and execute it through `chain.proceed(request)`.

The other sites interceptors do something similar IIRC.

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeList.kt
```diff
@@ -117,13 +117,12 @@ class Myanimelist(private val context: Context, id: Int) : TrackService(id) {
...
-                !getPassword().isEmpty() &&
-                checkCookies() &&
-                !getCSRF().isEmpty()
+    val isAuthorized: Boolean
+        get() = super.isLogged &&
+                getCSRF().isNotEmpty() &&
+                checkCookies()
 
-    private fun getCSRF(): String = preferences.trackToken(this).getOrDefault()
+    fun getCSRF(): String = preferences.trackToken(this).getOrDefault()
```
inorichi: How long does the CSRF token last? Is it tied to the MALSESSIONID cookie?

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyanimelistApi.kt
```diff
@@ -22,89 +22,70 @@ import java.io.InputStreamReader
...
-                .asObservable()
-                .map { response ->
-                    Jsoup.parse(response.consumeXmlBody(), "", Parser.xmlParser())
-                }
-    }
-
-    fun findLibManga(track: Track, csrf: String): Observable<Track?> {
-        return getList(csrf)
+    fun findLibManga(track: Track): Observable<Track?> {
+        return getList()
```
inorichi: We should avoid `getList()` here, it's an expensive request and we can workaround it.

This weekend I wrote this method to find out if a manga is in the library.

```kotlin
val url = "$baseUrl/ownlist/manga/$mediaId/edit"
val response = client.get(url).await()
response.close()

// If the prior request is a redirect, the manga isn't in the list
val priorResponse = response.priorResponse()
return if (priorResponse == null || !priorResponse.isRedirect) {
  mediaId
} else {
  null
}
```

I then used this other method to parse the state (you may need to parse a few more fields, I don't recall if `findLibManga` is used for anything else besides finding out whether it's in the library) 

```kt
override suspend fun getState(entryId: Long): TrackState? {
  val url = "$baseUrl/ownlist/manga/$entryId/edit"
  val response = client.get(url).await()
  val body = Jsoup.parse(response.awaitBody(), url)

  val lastChapterRead = body.getElementById("add_manga_num_read_chapters").`val`()
  val totalChapters = body.getElementById("totalChap").text()
  val score = body.getElementById("add_manga_score").select("option[selected]").`val`()
  val status = body.getElementById("add_manga_status").select("option[selected]").`val`()

  return TrackState(
    lastChapterRead = lastChapterRead.toFloat(),
    totalChapters = totalChapters.toInt(),
    score = score.toFloatOrNull() ?: 0f,
    status = status.toInt().toStatus
  )
}
```

I used coroutines here and I return a different model, but it's not that much different with RxJava.

inorichi:
About prepending the dot to the cookies: https://stackoverflow.com/questions/9618217/what-does-the-dot-prefix-in-the-cookie-domain-mean

Are you sure they aren't cleared? I believe I tested this before.

MCAxiaz:
> 
> 
> About prepending the dot to the cookies: https://stackoverflow.com/questions/9618217/what-does-the-dot-prefix-in-the-cookie-domain-mean
> 
> Are you sure they aren't cleared? I believe I tested this before.

Yes, the result of a `getCookie(".myanimelist.net")` is always null. So based on [this reading](https://www.mxsasha.eu/blog/2014/03/04/definitive-guide-to-cookie-domains/), my understanding is that this is the expected behavior. Since the actual cookies themselves have a domain of `myanimelist.net`, so when you try to get cookies for `.myanimelist.net` or something like `www.myanimelist.net` (tested this too), they would not match and thus not return anything.

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeList.kt
MCAxiaz: It is tied to the cookie. Lasts forever (or until cookie expires) as far as I can tell, it's just returned as a tag in the html.

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListInterceptor.kt
MCAxiaz: Sorry, not too familiar with RxJava but what makes this call async? I was under the impression that it's always synchronous unless you explicitly specify `subscribeOn`.

From a behaviour standpoint this seems to work fine, wondering if you could explain why doing this isn't a good idea. Thanks!

inorichi: `.subscribe()` returns a disposable. It's true that Rx is synchronous by default (depending on which operators you use, it's not only limited `subscribeOn`/`observeOn`), but it's a code smell to use `.subscribe()` and expect it'll be synchronous. If you really want to use it in a blocking way, use `.toBlocking()` instead of `.subscribe()` but it's possible this would case deadlocks (should be fine as long as there are no schedulers involved).

That's why I avoid Rx when interacting with blocking APIs.

inorichi:
Since I'm not sure about this cookies thing, I'm just going to get and delete the cookies for both the leading dot and only the domain.

I saw your updates to the login query, well done. Now only remains the `getList` on `findLibManga`.

MCAxiaz:
[I think this is the implementation of cookieManager that's used](https://android.googlesource.com/platform/frameworks/base/+/a1140f9d7c393bf222bd89256a2fa5a116513822/core/java/android/webkit/CookieManager.java). I think the confusion here is that a cookie's domain != the url you pass into `getCookie/setCookie`. The domain is an explicitly set property of the cookie (e.g. name=; domain=;). If a cookie did want to be included in subdomains then it would've specified it. So I don't think it makes sense to be handling domains here. Should just pass urls as is to the cookieManager and let it handle it.

PR #2044, opened 2019-05-21
Update tracking sites after finishing chapter

### LINKED ISSUE #1731: Update tracking sites after finishing chapter instead of as a bulk after exiting reader
**App version:** 0.8.0

**Request:** If I read chapter 1, 2, 3 then instead of setting chapter to 3 in the tracker, update it after finishing every chapter.

This is required for maintaining activity records in trackers such as Anilist.

moka491:
Fixes #1731

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
@@ -147,10 +147,9 @@ class ReaderPresenter(
 
     /**
      * Called when the user pressed the back button and is going to leave the reader. Used to
-     * update tracking services and trigger deletion of the downloaded chapters.
+     * trigger deletion of the downloaded chapters.
      */
     fun onBackPressed() {
-        updateTrackLastChapterRead()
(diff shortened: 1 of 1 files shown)


CarlosEsco:
What happens if you have no internet connection?  does it fail silently?

moka491:
@CarlosEsco Yeah, it's documented to ignore errors:
https://github.com/inorichi/tachiyomi/blob/6d34a060a02dd84d9355a14897dfe33c711d98ff/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt#L554-L560

I also tested it specifically and it works fine.

inorichi:
If you move it there, shouldn't the other call be deleted? Otherwise we could quickly send 2 requests.

moka491:
@inorichi Yeah you're right, I added another commit. 

I also thought about wether or not it would be a good idea to only update tracking if the *next* unread chapter has been read, and not just *any* unread chapter. Right now, if you have read 20 chapters, and then suddenly read the 80th, it will update the track to 80.

On the other hand, I think we want to give the users the ability to skip chapters and still have tracking working, so I kept it at that.

Issue #1809, opened 2019-01-18
[Request] Found Manga At Top and Merged


Z3ROGRAV1TYx:
So, I think it would be a good idea: when you search for a manga using Catalog>Global Search it puts all found at top. So if I type "One Piece", if any results are found it moves it to the top instead of showing all these "Not found" per extensions and scrolling through them all. 

It also would be better if we can just see the icon and title of found books and then click them and pick a repo of choice. This will eliminate duplicate results from multiple repos and instead merge them to choose there. I think this will provide more seamless searching.

Also sidenotes: I often find myself searching for something on global search and it doesn't come up, but if I go into a specific extension and search for the same thing it actually shows up all of a sudden. Also sometimes some of the repos in global search just show a lingering buffer circle.

metroid112:
I think it's a good idea, even more so when you have installed a lot of extensions, the global search gets pretty big quickly

arkon:
https://github.com/inorichi/tachiyomi/issues/1732

Issue #1732, opened 2018-11-16
[Features Request] Add option to hide empty search results


Septolum:
**DO NOT OPEN ISSUES/REQUESTS RELATING TO EXTENSIONS/CATALOGUES IN THIS REPOSITORY. Open them at the following repository  https://github.com/inorichi/tachiyomi-extensions/**

**For all other requests Please fill out the form below and remove the first 3 lines of this template**

**App version:**
0.8.0

**Issue/Request:**
Add an option that hides sources in the search and migration screens when they have no search results

**Steps to reproduce (if applicable)**
 1. Search for a manga that not all sources have
 2. Many with have a big card saying no results found
 3. When you have a lot of extensions installed especially, these cards take up a lot of space on the list, making it hard to find actual results

**Other details:**

blackholeearth:
Maybe all start in condensed . With loading circle. İf mange exist . İnflate that site source other stay condensed or be gone if searc resulted in nothing

metroid112:
I'm totally keen on this. I have a tons of extensions installed in case I search a manga that's not in my usual extensions, and it's a bit hard to scroll when doing a global search.

arkon:
Nevermind, #2066 does fix this but it just hasn't be released in stable.

PR #2066, opened 2019-06-02
Hide Empty Search Results in Catalogues

MCAxiaz:
Solves https://github.com/inorichi/tachiyomi/issues/1732 https://github.com/inorichi/tachiyomi/issues/1809

Not sure on the exact UI (i.e. keep or hide the catalogue title), and if whether or not this should be in settings (can probably just remove the nothing_found view if not).

Also didn't see any value in showing a loading circle for catalogues that error out, so changed it to return an empty result.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/global_search/CatalogueSearchHolder.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/global_search/CatalogueSearchHolder.kt
@@ -55,14 +55,17 @@ class CatalogueSearchHolder(view: View, val adapter: CatalogueSearchAdapter) :
             results == null -> {
                 progress.visible()
                 nothing_found.gone()
+                showHolder()
             }
             results.isEmpty() -> {
                 progress.gone()
                 nothing_found.visible()
+                hideHolder()
             }
             else -> {
                 progress.gone()
                 nothing_found.gone()
+                showHolder()
             }
         }
         if (results !== lastBoundResults) {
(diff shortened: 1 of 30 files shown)


inorichi:
I just noticed this is similar to #2097. I'd probably wrap the source API call into a `Observable.defer` so that we make sure that call can never crash the app. And change onException with onError, since this handles a few more cases.

Not sure yet if I'll merge this or the other, I'd have to test which one behaves better.

inorichi:
Can you fix conflicts? Basically you don't have to update the *Presenter anymore.

Btw, if I merge this we don't need the `nothing_found` view, right?

MCAxiaz:
Yea, so I'll go ahead and just remove the the view then?

inorichi:
If it's not going to be used, yeah.

MCAxiaz:
Would the resources need to be removed as well?
I believe there's one string resource `@string/no_results` and one vector `ic_search_black_112dp` which don't appear to be used elsewhere.

inorichi:
If the app still compiles after removing them it should be fine

hXtreme:
Found this: 

httpss://www.randomlytyping.com/blog/2015/6/2/hunting-unused-android-resources-with-lint

Hope it helps.

MCAxiaz:
> 
> 
> Found this:
> 
> httpss://www.randomlytyping.com/blog/2015/6/2/hunting-unused-android-resources-with-lint
> 
> Hope it helps.

Thanks, that was interesting. Now I know there were already 100+ unused resources LOL.
Wondering if that should be cleaned up 🤔

hXtreme:
@MCAxiaz you should definitely open an issue with all the linting errors for resources.

PR #2084, opened 2019-06-15
Fix sync chapters when source deletes all chapters

MCAxiaz:
Handles case where all chapters from the source have been removed (as is the case in the recent Mangadex incident). 

Expected
I expect any local chapters to also be removed accordingly.


inorichi:
I saw this last week and I'm still not really convinced. When a source doesn't return any chapters it's most probably an error on the parser or the website... Merging this could create more issues than solve the Mangadex incident.

MCAxiaz:
Yea I was worried about that that too, I was kind of hoping that any errors encountered upstream would just be thrown instead of resolving to an empty list, though I guess there's no guarantee of that.

I'm fine with just closing this off for now then, since it is a pretty edge case scenario.

inorichi:
Yeah I'll hold it off for now. I still have to think if 1.x would benefit from this.

PR #2087, opened 2019-06-16
Implement long hold selection for Manga Chapters

### LINKED ISSUE #1493: Download Feature Request: Range Selection

zhuoyang:
Long hold is functioned as "Select until from the last selection" (Inspired by Solid Explorer)
Closes #1493, alternative solution to #1641
[Demo here](https://streamable.com/2543q)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
@@ -241,12 +241,13 @@ class ChaptersController : NucleusController<ChaptersPresenter>(),
         }
         startActivity(intent)
     }
-
+    var lastClickPosition = -1
     override fun onItemClick(position: Int): Boolean {
         val adapter = adapter ?: return false
         val item = adapter.getItem(position) ?: return false
         if (actionMode != null && adapter.mode == SelectableAdapter.Mode.MULTI) {
             toggleSelection(position)
+            lastClickPosition = position
             return true
         } else {
             openChapter(item.chapter)
@@ -256,7 +257,19 @@ class ChaptersController : NucleusController<ChaptersPresenter>(),
 
     override fun onItemLongClick(position: Int) {
         createActionModeIfNeeded()
-        toggleSelection(position)
+        if (lastClickPosition == -1) {
+            toggleSelection(position)
+        } else {
+            if (lastClickPosition > position){
+                for (i in position..lastClickPosition-1)
+                    toggleSelection(i)
+            } else if (lastClickPosition < position){
+                for (i in lastClickPosition+1..position)
+                    toggleSelection(i)
+            } else
+                toggleSelection(position)
+        }
+        lastClickPosition = position
     }
 
     // SELECTIONS & ACTION MODE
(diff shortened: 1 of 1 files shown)


inorichi:
Please fix the line endings of the file to what they were. I know there's a diff on github to exclude them but I'd rather not merge a small feature with big changes due to just line endings.

hXtreme:
I can't seem to open the demo but from just the description, this seems like a feature that users won't notice for years because probably there are no UI cues that indicate this feature. 

An example of a no UI cue feature that gets it right:

Long pressing the manga name copies it to clipboard.

I say it gets it right because of the norm that long pressing text selects(highlights) it with options for copy, cut, and paste (of of which only copy makes sense in our context).

zhuoyang:
> Please fix the line endings of the file to what they were. I know there's a diff on github to exclude them but I'd rather not merge a small feature with big changes due to just line endings.

Fixed, have you considers using LF project wide? so that git's "core.autocrlf" can handle commit from different platform without breaking line endings.



> I can't seem to open the demo but from just the description, this seems like a feature that users won't notice for years because probably there are no UI cues that indicate this feature.
> 
> An example of a no UI cue feature that gets it right:
> 
> Long pressing the manga name copies it to clipboard.
> 
> I say it gets it right because of the norm that long pressing text selects(highlights) it with options for copy, cut, and paste (of of which only copy makes sense in our context).

Agreed, however I seems to discover this features fairly quick when using Solid Explorer without UI cue.
The UI cue that I can think of is displaying toast message when user select more than 3 chapters in a row using single selection mode. What do you think of this idea?

inorichi:
> Fixed, have you considers using LF project wide? so that git's "core.autocrlf" can handle commit from different platform without breaking line endings.

Yes, 1.x is entirely LF.

Guiorgy:
> this seems like a feature that users won't notice for years because probably there are no UI cues that indicate this feature.

Well, here's a different example for inspiration: Novel Library is essentially like Tachiyomi but for novels. There they have a button in the context menu Select Between, which does exactly that, selects all chapters between the chapters you have selected, and only appears if you have at least 1 chapter selected.

Either way, I really hope some way to select in bulk is added, and this solution looks pretty elegant imo.

Guiorgy:
If #1641 is merged, wouldn't this be unnecessary?

zhuoyang:
> If #1641 is merged, wouldn't this be unnecessary?

Yes, but UI cue aside this implementation is much more convenient imo.

Guiorgy:
@zhuoyang, indeed, your's is more.. elegant? Perhaps not too bad of an idea to implement both? Like, to each their own?

arkon:
@zhuoyang Sooo I ended up making the project use LF instead. Would you be able to update your PR again? Otherwise I'll have to do it.

arkon:
Nevermind, cherry picked a fixed version from @Jays2Kings 's fork: https://github.com/inorichi/tachiyomi/commit/2d3bfa9a89e08bb7318a39ba73620ae2820d6181

PR #2095, opened 2019-06-21
Allow download manager to seek cached file first before redownloading…

### LINKED ISSUE #2094: [Feature Request] Download From cache
**App version:** 0.8.4

**Android version:** 9.0

**Issue/Request:** 
Download from cache. If a manga already exists within the cache, it makes sense to download it directly from the cache rather than request the same image from whatever source.

Seflyx:
This pull request meant to resolve #2094

Internet connection still required since the CacheChapter requires image URL as parameter for check/getting the file unless I missed something.


inorichi:
I'm not really sure this is a big improvement. What happens if both try to access the file at the (almost) same time? Will one of them end up having a corrupt image?

I'm more worried about this PR introducing an issue rather than this (in my opinion) small optimization (given you usually either left it downloading or just read online)

arkon:
Superseded by https://github.com/inorichi/tachiyomi/commit/548dbf4b78e6b3d3b5edba46b1cb974abf1891aa

PR #2121, opened 2019-07-07
Notch fix

tonarii:
Here is a notch fix to have full screen in reading mode.
Might need to check and clean, cause i'm not a dev ;)


Soitora:
What does this fix specifically? I don't notice anything wrong when reading full screen with my notched-phone.

CarlosEsco:
It add the manga image to the side of the notch as you scroll up

arkon:
This doesn't seem to work properly after the AndroidX updates (the theme references need to be updated).

That said, is this actually desirable? Seems weird to have your page have a notch cutout in it.

hXtreme:
Because of the dimensions of most phones the pages shouldn't just flow into the notch area. However things change when a user would manually zoom into a page, now the page should flow into the notch area.

CarlosEsco:
Webtoon mode it will go to side

hXtreme:
I think most people just focus towards the centre of the screen most of the time, of course when  a webtoon has a long spread people may notice the notch.

arkon:
Closing in favour of #2595.

PR #2125, opened 2019-07-08
solving cf without using webview

badrazizi:
i found a way to solve cf without relying on webview, this meant to avoid locking the thread for more than is needed which is 4 seconds.


arkon:
Tachiyomi actually did it similarly to this implementation earlier. It was switched to solving it using the Webview so that it could better handle changes on CF's side, which was happening relatively frequently. An unfortunate con to this approach is that it's more fragile for those sorts of changes, resulting in the app becoming largely unusable for most users in those scenarios without a full app update.

hXtreme:
Close this?

Issue #2126, opened 2019-07-10
[Feature Request] Categories as filters


hXtreme:
It would be really nice if there could be filters for categories just like we have filters for Download, Unread, and Completed.

This would work really well along with #943 

It would also be really nice if the categories filters were tri-state as suggested by #1814

CarlosEsco:
Not sure how feasible this is as each site is defining their own genres most are standard but some like hentai ones go based off of tags and their tags can be unique.

 I guess you would have to keep a list of seen genres from manga you see or in your library then normalize them  so Shonen and shonen end up being the same.

hXtreme:
I'm not asking for categories to be generated automatically from tags.

hXtreme:
I just want the user defined categories to also act as filter in Library.

hXtreme:
As an example here's the list of categories that I use
![cat](https://i.imgur.com/YI7o68k.png)

What I would like is a filter for each one of them in library so that I can effectively find manga that I read irregularly is short and something that gives me the feels.

CarlosEsco:
Oh i see.  I saw categories and assumed you meant the genres.

hXtreme:
No problem.

Also, the code from the gist doesn't work because I don't know how to get the list of all categories, can you suggest something for it?

CarlosEsco:
make a new sql request to read entire category table?

hXtreme:
Can you point me to somewhere that's done, I'm not well versed in SQL so I'll need that.

CarlosEsco:
I think it would be this first method
https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/database/queries/CategoryQueries.kt

hXtreme:
@CarlosEsco I'm just going to wait for #2127 to be merged it adds a few things I would need for catalogue filters.

arkon:
I don't actually understand what this request means. Please open another issue with more details if desired.

PR #2127, opened 2019-07-10
Tri-state filters

### LINKED ISSUE #1814: [Feature request] Negative filters in library
Background:
For most of the sources suported by tachiyomi, positive and negative filters for genre, completed status etc. are available. This allows a great deal of control while searching for manga.
I can fine tune my search using filters to only include for example _non-yuri_ romantic comedy manga.
Meanwhile in the library we only get positive filters for Downloaded, Unread and Completed.

Request:
To add negative filter support in the library to allow users to filter with ease in the following cases:
1. I want to find all of the manga in my library that are _not downloaded_ and are unread. (To download some stuff for a trip)
2. I want find all the manga that's completed and I've read(_not unread_) it. (To search for an old forgotten manga)
etc.

hXtreme:
Implements negative filters for Download, Unread, Completed

Closes #1814 and works towards #2126


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
```diff
@@ -48,7 +51,7 @@ class LibraryNavigationView @JvmOverloads constructor(context: Context, attrs: A
      * Returns true if there's at least one filter from [FilterGroup] active.
      */
     fun hasActiveFilters(): Boolean {
-        return (groups[0] as FilterGroup).items.any { it.checked }
+        return !((groups[0] as FilterGroup).items.all { it.state == STATE_IGNORE })
```
inorichi: I think you're looking for `.none` instead of `!.all`.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryTriStateFilter.kt
```diff
@@ -0,0 +1,12 @@
+package eu.kanade.tachiyomi.ui.library
+
+object LibraryTriStateFilter {
```
inorichi: Do we really need this class? I mean, this doesn't really add anything new to what a common TriStateFilter has (yes/no/unchecked).

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
```diff
@@ -69,33 +72,53 @@ class LibraryNavigationView @JvmOverloads constructor(context: Context, attrs: A
...
-        private val completed = Item.CheckboxGroup(R.string.completed, this)
+        private val completed = Item.TriStateGroup(R.string.completed, this)
 
         override val items = listOf(downloaded, unread, completed)
 
         override val header = Item.Header(R.string.action_filter)
 
         override val footer = Item.Separator()
 
+        // TODO: Ask if there's a better way to do this mapping & inverse mapping
```
inorichi: Well, I'd just remove the mappings, as they aren't really changing the values (maybe if they're not 0, 1 or 2 but in that case we could just coerce the values between 0 and 2.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
hXtreme: I am not sure if none would be better because filters could be either Including or Excluding.

making the clause look like:

`it.state == STATE_INCLUDE || it.state == STATE_EXCLUDE`

inorichi: Maybe `any { it.state != IGNORE }`? It has the advantage that it can return earlier (nobody will notice but it should be a better method)

hXtreme: I like `any { it.state != STATE_IGNORE }`

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryTriStateFilter.kt
hXtreme: If I remove the mapping, this class is pretty useless.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
hXtreme: On it.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -114,33 +116,37 @@ class LibraryPresenter(
...
+                else -> downloadManager.getDownloadCount(item.manga) > 0
             }
-            true
+
+            return@downloaded if (filterDownloaded == STATE_INCLUDE) isDownloaded
+                else !isDownloaded
         }
 
-        return map.mapValues { entry -> entry.value.filter(filterFn) }
+        return map.mapValues { entry ->
```
inorichi: I'm not sure about this change. Isn't it possible to use the previous function that filters just once? I don't like doing 3 filter iterations.

hXtreme: To the best of my knowledge about various functional languages, this is the defacto way to use filters.

arkon: I think he means just using a single filter instead of three separate ones. Unless it's a sequence, the filters will result in 3 iterations of the list.

inorichi: And even worse, 3 copies of the list

hXtreme: Noticing @arkon 's point, these should be made into sequences instead of list or Iterable.

Also, given that we call the class LibraryMap shouldn't it be lazy as well?

Maps are supposed to be lazy right.

arkon: I don't think there's really a benefit to using a sequence here.

> Maps are supposed to be lazy right.

Why?

hXtreme: map...filter...reduce paradigm only really works well when the operations are lazy.

For example consider functional languages such as Haskell, OCaml, Scala...

inorichi: The filter function could be something like:

```kt
val filterFn: (LibraryItem) -> Boolean = f@ { item ->
  if (filterUnread == STATE_INCLUDE && item.manga.unread == 0) return@f false
  if (filterUnread == STATE_EXCLUDE && item.manga.unread != 0) return@f false

  ...
}
```
and so on

hXtreme:
Sorry that it took this long, my semester has been too busy. I'm finishing this up as I procrastinate my assignments.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
hXtreme: Done!

I've done something similar but I used `||` short-circuiting for the early return.

hXtreme:
@inorichi I have made the changes and marked the review as resolved but github for some reason thinks that there is still a change request pending.

Jays2Kings:
This needs to use different keys and delete the old ones, since the existing keys are saved as booleans and won't convert to ints automatically.

hXtreme:
Having to delete the old keys, seems like it would become technical debt in future. Is there any way to cast old Boolean values to new integer ones?

arkon:
Casting would be technical debt too. I'd prefer just using new keys, at the risk of existing users having to set their filters again.

Jays2Kings:
Like arkon said, both are debts, and the upgrade method runs after the preferences are loaded so it can't be converted on upgrade without changing when upgrade migration happens, but that can at least be used to delete the old keys.

hXtreme:
Can you let me know where the code to delete the preference goes?

Jays2Kings:
"eu/kanade/tachiyomi/Migrations.kt" you would have update the version code in the build.grade then add an if statement like the others in the upgrade method

arkon:
@hXtreme Were you planning on finishing this?

hXtreme:
Can't work on this(or any android development) anytime soon, so if someone wants to finish it up that's perfectly fine if not I'll do it when I can.

You probably shouldn't keep a version update hanging for this change.

joseph619:
@hXtreme Are you free to work on this now?

hXtreme:
@joseph619 not really, amongst other things I don't have a machine to run android studio.

Also I have been pretty out of touch with the codebase so it might just be better if someone else does it. 

Also because of how many changes Tachiyomi has gone under, it might be best to scrap this and do it from the scratch using this as only a reference.

arkon:
Superseded by 687f3d48ea03e09c2fe33a8d8e36e1adb87bbf4b

PR #2140, opened 2019-07-23
Added FAQ in About

jogerj:
The link redirects to the wiki page


hXtreme:
I don't see why this would help.

We already have the help option in navigation bar(more accessible than About) why not just rename it to FAQ instead and update that link?

jogerj:
> We already have the help option in navigation bar(more accessible than About) why not just rename it to FAQ instead and update that link?

Not all devices have navigation bars enabled (I didn't and I wasn't even aware that was there). I don't know where else to put it either, and it seems fitting to place there

arkon:
He means the side navigation.

jogerj:
My bad I haven't updated my dev build for a while. All is well.

Issue #2047, opened 2019-05-23
[Improvement][1.x] Make viewers into extensions


hXtreme:
There are a number of issues that request more and more features from the manga viewer. 

#2010, #2046, #1997, #1977, #1894(if we move things to another viewer), #1649, #1558, #2062  

While a lot of them are interesting to have, I believe that not everyone needs them and including all the varied feature would bloat up the source code. Instead if viewers could be installed as extensions then anyone who really wants a particular feature could in theory make a viewer and use it.

arkon:
I'm not sure if there's any benefit to this from an engineering perspective.

hXtreme:
Option 1: We have one viewer with all (say 10) features which could be turned on or off.  
Here, every time a new one if added extensive amounts of testing would be required about 2^10 many tests. These tests would more or less be required because some features can conflict with others and we need to understand how those conflicts are resolved. Which is bad/tedious!

Option 2: We have multiple viewers bundled together (the current option).  
This solves the problem encountered in Option 1 but every time a new viewer is submitted inorichi (he already has a lot of stuff to do) would have to sign off a pull request and everyone would only receive it on app updates.

Option 3: Viewers are made into extensions (and maybe a few default ones are bundled together).  
This option allows viewers to be updated independently and the development of viewers could follow the same principle as the development of tachiyomi-extensions(where inorichi doesn't have to do most of the work).

metroid112:
How would you use different viewers in the app? Change them in the same menu as rotation/orientation settings while reading a chapter?

hXtreme:
Yes, I guess that would be the best way and infact kind of how it is right now with webtoons viewer.

Harsh

________________________________

inorichi:
That'd be a real pain. I prefer to have them internal and add many configurations.

Issue #2143, opened 2019-07-23
[UI BUG] ReaderSettingsSheet doesn't refresh when viewer is changed.


hXtreme:
**Version: r1521**

Check out [video](https://youtu.be/-IA2378q_g0)

There when I switch from Webtoon viewer to Default (Pager) viewer the Settings sheet continues to show Webtoon settings.

*Lazy fix*: Auto close the Settings sheet when viewer is changed.

*Ideal fix*: Have Settings sheet update the list of settings whenever viewer is changed.

Edit: fixed the video link.

PR #2152, opened 2019-07-28
Viewer navigation

hXtreme:
Extends over #2020 and adds more functionally from that was discussed in the comments there.

It's done in a way that makes it easier to add more options for viewer navigation, which is in line with what was mentioned [here](https://github.com/inorichi/tachiyomi/issues/2047#issuecomment-506951642)


app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/ViewerNavigation.kt
```diff
@@ -0,0 +1,30 @@
...
+        const val PREV = 2
+    }
+
+    abstract var menuRegion : List<RectF>
+
+    abstract var nextRegion : List<RectF>
+
+    abstract var prevRegion : List<RectF>
+
+    fun getAction(pos : Pair<Float, Float>) : Int {
```
inorichi: This could potentially be a `PointF`.

hXtreme: I've made the changes.

hXtreme:
@inorichi While testing for this i found it a bit frustrating when the location for menu changed in every mode so I was wondering if it might be a good idea to make sure that the top 10-15% (or alternatively bottom) of the screen always is for menu.

hXtreme:
The effect of bug #2143 becomes a bit more visually prominent with this pull request (at least that's when I noticed it).

I have a dirty fix that I can use to try to solve #2143 should I do that?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerConfig.kt
```diff
@@ -43,6 +43,9 @@ class PagerConfig(private val viewer: PagerViewer, preferences: PreferencesHelpe
     var doubleTapAnimDuration = 500
         private set
 
+    var navigationMode = 0
```
inorichi: I'd store either here or in the `PagerViewer` the instance of the current navigation mode. Then you won't need to retrieve the current instance on each touch but delegate it.

hXtreme: Thanks! That's pretty much what I wanted to do but it seemed like everything else in PagerConfig was of primitive type so navigationMode should also be primitive.

Now that that's been cleared, I'll get right to it.

Also, storing it in PagerViewer might not be very clean as we'd still need to update the navigationMode instance every time the preference is changed, so I'll store the instance in PagerConfig.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerNavigationHelper.kt
```diff
@@ -0,0 +1,39 @@
+package eu.kanade.tachiyomi.ui.reader.viewer.pager
+
+import android.graphics.RectF
+import eu.kanade.tachiyomi.ui.reader.viewer.ViewerNavigation
+import eu.kanade.tachiyomi.ui.reader.viewer.navigation.KindlishNavigation
+import eu.kanade.tachiyomi.ui.reader.viewer.navigation.LNavigation
+
+
+class PagerNavigationHelper {
```
inorichi: I think this class is unneeded once you store the instance in the config class, also no need to store a lazy instance, just create a new instance when the user changes the navigation mode.

Same goes for the WebtoonNavigationHelper.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/ViewerNavigation.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    abstract var nextRegion : List<RectF>
+
+    abstract var prevRegion : List<RectF>
+
+    fun getAction(pos: PointF) : Int {
+        val x = pos.x
+        val y = pos.y
+        val action = when {
+            menuRegion.any { it.contains(x, y) } -> MENU
```
inorichi: A little nitpick, but I'd apply this order: "next, prev, menu"

hXtreme:
Thanks for the review! I'll try to get to making the changes over the weekend.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/ViewerNavigation.kt
hXtreme: I totally get that "next, prev, menu" order is aesthetically pleasing and also what I initially went for. However during testing, I managed soft-lock myself into a position where I couldn't access the menu (because my co-ordinates for next were bugged), so I made the change to give menu the highest priority.

No biggie though, I can make this change.

Charuru:
How is this going?

hXtreme:
@Charuru , I've been super busy and haven't been able to work on this at all. If you want to finish this up feel free to fork from my branch and work on it.

I'll merge it quickly and then inorichi can finish up the review process.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/ViewerNavigation.kt
hXtreme: Okay here's what I'm planning to do:

A constant region across all navigation modes (which is very helpful when you forget which mode you're currently on) I'm making it fairly small to not be much of a intrusion(5% of top screen height). Beyond that first Next then Prev and then everything else is Menu. @arkon (I'm assuming that you're the one who's going to review it)

ghostbear:
@hXtreme what is left to do in this PR? Other than maybe adding away to invert tapping and fixing the conflicts

hXtreme:
Things that need to be done here:

Revert 4117b6b
Rebase and resolve conflicts

----

Inverting touch directions should probably go in a separate pr.

---

@arkon can you label this as help wanted to indicate to others that they should feel free to adopt this pr given that I won't be able to do any android development till god knows when.

ghostbear:
Gonna give this a try

hXtreme:
awesome, let me know if you want any help understanding my code here.

ghostbear:
If I understand L Shaped and Kindlish navigation this is how they should behave, right?

L Shaped Navigation:
```
+---+---+---+
| P | P | P |   P: Previous
+---+---+---+
| P | M | N |   M: Menu
+---+---+---+
| N | N | N |   N: Next
+---+---+---+
```
 
Kindlish Navigation:
```
* +---+---+---+
* | M | M | M |   P: Previous
* +---+---+---+
* | P | N | N |   M: Menu
* +---+---+---+
* | P | N | N |   N: Next
* +---+---+---+
```

hXtreme:
Yes that's right.

arkon:
Superseded by #3869

PR #2175, opened 2019-08-11
Allow setting a preferred date format

Wiles:


inorichi:
Hi, thanks for your contribution and sorry for the late reply.

First of all, I'd like to ask why you think this is a feature that should be included in the app. I don't like having a lot of settings and IMO this one isn't particularly useful. For english locales, we have british and american english. If you are missing a region for a forced language (which should use the date format you're used to), I'd prefer to add it rather than this PR.

Wiles:
I want this change because I want dates formatted in ISO8601 (yyyy-mm-dd) which is not the default for any language that I understand. This is a pretty specific requirement and wont feel bad if you reject it. 

Thank you for all your hard work creating Tachiyomi, it is a great application.

asdkant:
This is a very reasonable requirement , I also want to have ISO dates (and playing the guessing game with locales and date formats is bothersome)

arkon:
@Wiles Any chance you can resolve the merge conflicts? I'd like to merge this in.

Wiles:
@arkon Yes, I will resolve those when I have time this week.

Wiles:
@arkon Sorry for the delay should be good now.

PR #2176, opened 2019-08-11
Allow back button to navigate to previous URL in WebView

FlaminSarge:
Adds Forward, Refresh, and Close actions to webview controller menu

-----------------------------------------------
Things before this should be merged:

- Do I need to null-check 'web' since view comes from getView() instead of a param?
- How do I add a 'close' menu item for the web view controller so that if a user gets stuck in e.g. an ad-loop they can force the whole WebView to close (as pressing back used to do here)?

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaWebViewController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaWebViewController.kt
@@ -1,9 +1,7 @@
 package eu.kanade.tachiyomi.ui.manga.info
 
 import android.os.Bundle
-import android.view.LayoutInflater
-import android.view.View
-import android.view.ViewGroup
+import android.view.*
 import android.webkit.WebView
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.source.SourceManager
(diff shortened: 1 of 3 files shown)


arkon:
1. Probably not.
2. We ought to change the up button from a back arrow to a close (X) icon instead.

FlaminSarge:
Since this page is a controller wrapping a webview, I wasn't sure if it was possible to change the back button's icon.

(Is there webview implementation we can use that doesn't require this kind of wrapping?)

FlaminSarge:
So it turns out that the back button on the controller is also affected by this handleBack() change, so changing it to a close icon wouldn't be accurate unless we can swap it out entirely for a 'close' action. Trying to figure out how to do that, as I'm not too familiar with anything here.

FlaminSarge:
I've updated this to work as the Samsung Internet and Chrome webviews work for the most part, with the exception of the button in the top left which is still 'back' (because the hardware back button is tied to that, unfortunately). Close is instead in the dropdown menu, as are Forward and Refresh; Forward only shows up if webview.canGoForward() is true.

Ready for review.

joseph619:
@FlaminSarge Did you request for a review?

FlaminSarge:
As in, the GitHub feature for it? I think that's limited to those with write access to the repo.

joseph619:
@arkon This has been requested for review, but FlaminSarge doesn't have write access to formally request one

Issue #2158, opened 2019-07-31
[UI Issue] The white splash screen is very painful to look at.


hXtreme:
Currently Tachiyomi doesn't have a splash screen which leads to a white screen to flash.

Check the [video](https://youtu.be/FRRP73lOsbg)

It would be great if a splash screen with dark background and Tachiyomi logo was displayed instead.

Similar to #1223 but a theme independent splash screen.

markelgaston:
I totatly agree, it is painful to watch at night and eventually the app would certently need a splash screen too.

Ill work on it.

hXtreme:
@markelgaston have you started working on it/ are about to? If not, I want to try to take a shot at it.

markelgaston:
> @markelgaston have you started working on it/ are about to? If not, I want to try to take a shot at it.

Hi, I have actually already finished it. Im just not familiar on how to use git on android studio and how to do pull requests and stuff, so Im doing a bit of research to make sure i do it correctly hah, but yea, i will do the pull requests soon.

RYNAPU:
When I launch it, the white screen is there for less than a second. Not sure why it's taking so long for yours ?

bbtora:
It also depends on the size of your library and maybe too on the downloads size in the storage.

arkon:
Added via https://github.com/inorichi/tachiyomi/compare/4bc5f1401f57cf208c09e2655536b29e160b6732..8d3166c5febe2fcf9d53084440a9f18650a467dd

PR #2185, opened 2019-08-21
Added a app loading splash Screen

markelgaston:
As a fix to #2158

Instead of a white background, the app now starts up with a custom theme (defined in "splash_nackground.xml") until the main activity loads, after which it will update the theme as always with our choosed theme on the settings.

Result: https://streamable.com/q8urq


markelgaston:
> Add end of file newline.

oh, already fixed, thanks.



> Colour for build and Dev version should be different right? (dark blue and grey respectively)
> 
> Does anyone know how to do that?

i didnt know there were two different versions xD, ill take it a look at that later during the day.

hXtreme:
[What the two versions look like](https://i.imgur.com/wxUNUQf.jpg)

Soitora:
**Tachiyomi Stable** uses `#2E84BF` while **Tachiyomi Dev** uses `#455A64`.
What you're showing in your demonstration is for the Dev release.

hXtreme:
Alternatively we can just use the colour from stable release.

Also, how about scaling the icon a bit, it looks too small.

markelgaston:
> **Tachiyomi Stable** uses `#2E84BF` while **Tachiyomi Dev** uses `#455A64`.
> What you're showing in your demonstration is for the Dev release.

ohh i see, but where is the stable version? is it another branch?, and how do we usually convert a dev version to the stable version?

> Alternatively we can just use the colour from stable release.
> 
> Also, how about scaling the icon a bit, it looks too small.

To change the color you just have to change the color in splash_background.xml.

Yea, i also thought about that, ill scale it a bit bigger whenever i can, i have been a bit busy this days.

markelgaston:
> Latest stable release: https://github.com/inorichi/tachiyomi/releases/tag/v0.8.4
> 
> I am not sure how tachiyomi does stable releases but you should ask on Discord.

Okay thanks ^^

inorichi:
In `themes.xml` there's a `Theme.BrandedLaunch` that was never used because it didn't work. Please use that theme instead of a new one `styles.xml`. I'd also rename it to `Theme.Splash`.

About the color, I'd just use `#455A64` just because it's kind of dark (people was complaining about the white screen so...)

arkon:
Going to work on this in a branch and submit another PR.

PR #2196, opened 2019-08-30
ConfigurableSourceFactory can be configured in the same way as…

gotson:
…ConfigurableSource, enabling use-cases where the list of sources produced by the factory depends from user input.

Here is an example where the `SourceFactory` will produce multiple sources from a list of comma-separated values:
Screenshot_1567169007
Screenshot_1567169041
Screenshot_1567169074

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/model/Extension.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/model/Extension.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.extension.model
 
 import eu.kanade.tachiyomi.source.Source
+import eu.kanade.tachiyomi.source.SourceFactory
 
 sealed class Extension {
 
@@ -16,7 +17,8 @@ sealed class Extension {
                          override val versionCode: Int,
                          val sources: List<Source>,
                          override val lang: String,
-                         val hasUpdate: Boolean = false) : Extension()
+                         val hasUpdate: Boolean = false,
+                         val sourceFactory: SourceFactory?) : Extension()
 
     data class Available(override val name: String,
                          override val pkgName: String,
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/util/ExtensionLoader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/util/ExtensionLoader.kt
@@ -10,6 +10,7 @@ import eu.kanade.tachiyomi.data.preference.getOrDefault
 import eu.kanade.tachiyomi.extension.model.Extension
 import eu.kanade.tachiyomi.extension.model.LoadResult
 import eu.kanade.tachiyomi.source.CatalogueSource
+import eu.kanade.tachiyomi.source.ConfigurableSourceFactory
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.source.SourceFactory
 import eu.kanade.tachiyomi.util.Hash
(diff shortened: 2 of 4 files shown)


gotson:
Hello, I wanted to know if you would consider this pull request or not at all?

I know you may be working more on 1.0, so I would understand you don't have much time to look at that now 🙂

gotson:
Hi @arkon @inorichi just wanted to check with you if that's something you would consider? Thanks!

arkon:
The biggest issue with this is going to be compatibility with older versions of the app. We'd have to make sure they don't crash if there's an extension that depends on this but isn't there.

gotson:
It's been a long time, and i lost the code i had that implemented this on my extension.

After rebasing this PR onto the latest Tachiyomi code, and redoing the code to implement this in the extension, i could test the following:
- running this PR with extensions that don't implement it is OK
- running current Tachiyomi with extensions that implement it make the extension unavailable. It crashes in the loader and is not available at all, it's not showing in the Installed. Some debug log [here](https://gist.github.com/gotson/840e9e9b0291bb115a3bea86147ee6a2).

I think a reasonable path would be to increment the library version from `1.2` to `1.3` with this new feature. However i checked the repo history but couldn't find out what are versions `1` and `1.1` about.

We would need to enforce all extensions that use this feature to increase the base version to `1.3`, so they are ignored with previous versions of Tachiyomi.

gotson:
@arkon would you be OK with the proposed solution to increase version to 1.3 for this feature?

I have many people asking for multiple Komga servers in the extension and that's the only way to handle it. 

Thanks

arkon:
The proposal more or less makes sense, but bumping the extension library version isn't a trivial change. Moving to 1.3 would at best happen in 2 stable versions from now.

gotson:
Hi Arkon,

i understand bumping the extension library version is not trivial, i'll try to help if possible.
No problem for adding 1.3 in 2 stables from now, people can use the dev in between. What's the new schedule of releases by the way ?

I checked https://github.com/inorichi/tachiyomi-extensions-lib but couldn't see `ConfigurableSource` there, so i doubt i should add `ConfigurableSourceFactory`, as it seems to add a dependency on `androidx.preference.PreferenceScreen`, which maybe why `ConfigurableSource` is not there in the first place.

gotson:
However if it's too much hassle maybe i can hardcode a number of slots for the Komga extension, for example 3, so that people can add up to 3 servers.

arkon:
> i understand bumping the extension library version is not trivial, i'll try to help if possible.
No problem for adding 1.3 in 2 stables from now, people can use the dev in between. What's the new schedule of releases by the way ?

There isn't really a proper schedule, but I'm still working towards the next stable release (0.9.0). A majority if is UI-related work and other internal migrations. With the release of that version, the minimum extension-lib version will be 1.2, while 0.8.5 has a maximum extension-lib version of 1.2.

At that point we can start adding the APIs necessary for extension-lib 1.3 (tweaked functions, use of coroutines, what this PR is addressing, maybe moving away from multi-sources as they exist today, etc.). The app would need to retain both extension-lib 1.2 and 1.3 for a bit for compatibility. 

> I checked https://github.com/inorichi/tachiyomi-extensions-lib but couldn't see `ConfigurableSource` there, so i doubt i should add `ConfigurableSourceFactory`, as it seems to add a dependency on `androidx.preference.PreferenceScreen`, which maybe why `ConfigurableSource` is not there in the first place.

It's located in a stub module in the extensions repo, which also includes stubs for both the old support-v7 preferences and AndroidX preferences: https://github.com/inorichi/tachiyomi-extensions/blob/master/lib/preference-stub/src/main/java/eu/kanade/tachiyomi/source/ConfigurableSource.java

arkon:
@gotson Is this still desired?

gotson:
I haven't had anyone complain about this on our Discord, so I guess the current extension with 3 configurable sources for Komga is enough for most people.

arkon:
Cool. I could potentially include it, but it's niche enough that it's probably simpler to just leave it out if the current setup is sufficient.

PR #2211, opened 2019-09-08
Implemented extension search functionality.

### LINKED ISSUE #2202: [Feature Request] Add searchbar to extensions.
Hi there,
Thank you for the excellent work on this application.

My request is:
There are currently 126 extensions, it would be a good idea to add a search bar.
[https://imgur.com/a/Q2PZCkG](https://imgur.com/a/Q2PZCkG)

Thanks.

funkyhippo:
This is a quick implementation of searching for extensions. This PR adds a search menu icon on the extensions view, while also removing the extensions count per extensions group (because the count is set on ExtensionManager updates, which would be more expensive to recompute on every new search query). 

Closes #2202. 

Demo here:
https://imgur.com/rYwGVJK

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionController.kt
@@ -1,10 +1,14 @@
 package eu.kanade.tachiyomi.ui.extension
 
 import android.support.v7.widget.LinearLayoutManager
+import android.support.v7.widget.SearchView
 import android.view.LayoutInflater
+import android.view.Menu
+import android.view.MenuInflater
 import android.view.View
 import android.view.ViewGroup
 import com.jakewharton.rxbinding.support.v4.widget.refreshes
+import com.jakewharton.rxbinding.support.v7.widget.queryTextChanges
 import eu.davidea.flexibleadapter.FlexibleAdapter
 import eu.davidea.flexibleadapter.items.IFlexible
 import eu.kanade.tachiyomi.R
@@ -27,6 +31,10 @@ open class ExtensionController : NucleusController<ExtensionPresenter>(),
      * Adapter containing the list of manga from the catalogue.
      */
     private var adapter: FlexibleAdapter<IFlexible<*>>? = null
+
+    private var extensions: List<ExtensionItem> = emptyList()
+
+    private var query = ""
 
     init {
         setHasOptionsMenu(true)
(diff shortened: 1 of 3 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionController.kt
```diff
@@ -84,6 +92,30 @@ open class ExtensionController : NucleusController<ExtensionPresenter>(),
         }
     }
 
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
+        inflater.inflate(R.menu.extension_main, menu)
+
+        val searchItem = menu.findItem(R.id.action_search)
+        val searchView = searchItem.actionView as SearchView
+        searchView.maxWidth = Int.MAX_VALUE
```
inorichi: Is this really needed?

app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionController.kt
```diff
@@ -114,7 +146,19 @@ open class ExtensionController : NucleusController<ExtensionPresenter>(),
...
-        adapter?.updateDataSet(extensions)
+        this.extensions = extensions
+        drawExtensions()
+    }
+
+    fun drawExtensions() {
+        if (!query.isBlank()) {
+            adapter?.updateDataSet(
+                    extensions.filter {
+                        it.extension.name.contains(query.toRegex(RegexOption.IGNORE_CASE))
```
inorichi: You could use `.contains(query, ignoreCase = true)` instead of creating a Regex instance each time

app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionGroupHolder.kt
```diff
@@ -15,6 +15,6 @@ class ExtensionGroupHolder(view: View, adapter: FlexibleAdapter<*>) :
         title.text = when {
             item.installed -> itemView.context.getString(R.string.ext_installed)
             else -> itemView.context.getString(R.string.ext_available)
-        } + " (" + item.size + ")"
+        }
```
inorichi: I guess we could get rid of the count per category, not a big deal IMO. Thoughts @arkon @CarlosEsco ?

app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionController.kt
funkyhippo: It expands the search bar to take up the full width of the layout. It's a bit hacky, but it seems to be the solution that has the widest compatibility. You can actually see the problem in the demo; I made this change after I recorded it.

app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionGroupHolder.kt
```diff
@@ -11,6 +11,6 @@ class ExtensionGroupHolder(view: View, adapter: FlexibleAdapter<*>) :
 
     @SuppressLint("SetTextI18n")
     fun bind(item: ExtensionGroupItem) {
-        title.text = item.name + " (" + item.size + ")"
+        title.text = item.name
```
funkyhippo: Just forwarding the [old conversation](https://github.com/inorichi/tachiyomi/pull/2211/files#r334091442) over, as I updated the commit to account for the latest commit.

app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionGroupHolder.kt
arkon: Yeah, I don't really think it's all that useful.

PR #2239, opened 2019-09-21
Upgrade Kotlin (to 1.3), Coroutines, Gradle, and Android Gradle plugin

SamB:
Specifically,
 * Kotlin: 1.2.71 -> 1.3.50
 * Coroutines: 0.22.2 -> 1.3.1 (in many, many steps)
 * Gradle: 4.6 -> 5.4.1
 * Android Gradle plugin: 3.2.1 -> 3.5.0

Most of the work was:

1. Figuring out which versions to stop and fix coroutine-related deprecation warnings at
2. Waiting for builds and/or inspection runs to finish

I used lots of small commits so I wouldn't have to start all over when I made a mistake. (You know, *after* starting over maybe three times.)

They **really** weren't kidding when they said the coroutines library was experimental. Thankfully, after the upgrade, we only have one experimental symbol left in Tachiyomi.

The point of all this is that the new Gradle, Android plugin, and Kotlin versions have better work-avoidance features (such as caching, incremental processing, and diagnostic tools), but I've postponed the work of actually making sure that all of that is turned on. 

For example, incremental annotation processing will not work without upgrading to [Glide 4.9.0](https://github.com/bumptech/glide/releases/tag/v4.9.0).


arkon:
I assume extensions didn't have issues?

SamB:
@arkon I'm not entirely certain; is there anything in particular I am supposed to do to check that?

arkon:
@SamB Not particularly, just try browsing catalogues, opening manga, etc. If it all seems to be working fine, that's cool.

app/build.gradle
```diff
@@ -247,10 +245,9 @@ repositories {
     mavenCentral()
 }
 
-kotlin {
-    experimental {
-        coroutines 'enable'
-    }
+// See https://kotlinlang.org/docs/reference/experimental.html#experimental-status-of-experimental-api-markers
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.AbstractKotlinCompile).all {
+    kotlinOptions.freeCompilerArgs += ["-Xuse-experimental=kotlin.Experimental"]
```
CarlosEsco: What else being used is still experimental?  I believe if there is anything it could be refactored to the non expiremental verison.  Reason i mention is i  did not need this in Neko

inorichi:
I didn't update Kotlin because I'm not sure the bytecode of an extension built with 1.2 will work with the app using >1.3.30, as they mentioned there were some incompatibilities. If you can confirm they still work fine, I'll merge this.

Weasy666:
I haven't tested this extensively, but i built the app from this PR and used it a bit with the MangaHub extension. Worked normally, i could add Mangas to my library, view chapters and use everything else normally. Is there some other extension that should be checked?

inorichi:
I'll merge this once they release a new version of the coroutines library that fixes a long cold start issue.

arkon:
v1.3.3 of coroutines is out, although I'm not sure if the cold start issue is still there.

Regarding the bytecode issue, there might be issues with older versions of Android but we can do a large scale update of all of the extensions to bring it in line with the app. It's easier to just force users to use latest versions of everything for compatibility.

inorichi:
Yes, it's been fixed.

I think we should merge this (once conflicts are resolved) and test it on dev with the old compiler on the extensions, and if everything is ok we bump the compiler for the extensions alongside the next app release.

arkon:
Did a quick smoke test on my device and it seems to work fine. Can't really guarantee it for older versions of Android, but I'll test it a bit more after merging (and doing more updates, like coroutines).

PR #2263, opened 2019-10-03
Set glide to use the gif loop count

### LINKED ISSUE #2240: Mangadex: Gif playback
When trying to read [Ijousha no Ai Vol. 2 Ch. 9 Pg. 10](https://mangadex.org/chapter/416287/10), the page shows a flashing message via gif which doesn't continually flash on the original site.

I'm not sure if this issue should be on this repo or the main repo, but I'm submitting it here since it's related to the Mangadex extension.

Flat:
Sets Glide to use the GIF file loop count instead of always looping infinitely. Fixes #2240

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerPageHolder.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerPageHolder.kt
@@ -19,6 +19,7 @@ import com.bumptech.glide.load.DataSource
 import com.bumptech.glide.load.engine.DiskCacheStrategy
 import com.bumptech.glide.load.engine.GlideException
 import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions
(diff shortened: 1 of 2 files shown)


arkon:
Is there Glide documentation about this? I'm surprised it doesn't follow the file by default.

Flat:
https://bumptech.github.io/glide/javadocs/460/index.html

inorichi:
Does it work even if the page is offscreen? The `ViewPager` loads the adjacent pages too, so it could play the animation while it's not visible.

Flat:
From what I recall during testing the animation didn't start until the image was visible, I won't be able to confirm until sometime tomorrow though.

Flat:
From testing it appears that the animation will play when the image is loaded as the next page by the pager. This is more visible using the webtoon reader than the default reader. If this is an issue then I can look at waiting to play animations until they are the active page. I'm not sure that any manga or webtoons have non-looping images as parts of the manga/manwa though.

inorichi:
Fair enough. I'll merge this change.

PR #2275, opened 2019-10-13
Filter Extensions By Language

MCAxiaz:
add options menu item to filter extensions by languages


app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionPresenter.kt
```diff
@@ -61,9 +64,10 @@ open class ExtensionPresenter(
         val installedSorted = installed.sortedWith(compareBy({ !it.hasUpdate }, { it.pkgName }))
         val untrustedSorted = untrusted.sortedBy { it.pkgName }
         val availableSorted = available
-                // Filter out already installed extensions
+                // Filter out already installed extensions and disabled languages
                 .filter { avail -> installed.none { it.pkgName == avail.pkgName }
-                        && untrusted.none { it.pkgName == avail.pkgName } }
+                        && untrusted.none { it.pkgName == avail.pkgName }
+                        && (avail.lang in preferences.enabledLanguages().getOrDefault() || avail.lang == "all")}
```
inorichi: Save a local `val` for `preferences.enabledLanguages().getOrDefault()`

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsExtensionsController.kt
```diff
@@ -0,0 +1,49 @@
+package eu.kanade.tachiyomi.ui.setting
```
inorichi: Better use the extension package for this file.

app/src/main/java/eu/kanade/tachiyomi/ui/extension/ExtensionPresenter.kt
```diff
@@ -53,6 +56,7 @@ open class ExtensionPresenter(
     @Synchronized
     private fun toItems(tuple: ExtensionTuple): List<ExtensionItem> {
         val context = Injekt.get<Application>()
+        val activeLangs = preferences.enabledLanguages().getOrDefault()
```
funkyhippo: Is it necessarily desirable to link the catalogue languages with the extension languages?

MCAxiaz: They are one and the same. Catalogues are just installed extensions. 
If you have an extension language enabled then you must necessarily also want it to be enabled in catalogues otherwise it would be pointless.

funkyhippo: You're right that they're one and the same, but (IMO) there's value in being able to see all the extensions while having only some enabled in your catalogue. 

In addition, it branches where you can change this setting (changing it in the catalogues menu will change it in the extensions menu and vice versa), which might be a bit confusing. Also, since only `en` and their system language is shown by default, a new user might think that there aren't many extensions available.

MCAxiaz: What's a realistic use case where a user would want to see extensions available in a specific language, but want to hide them in catalogues? If they really wanted to hide specific extensions, catalogues already supports enabling/disabling specific extensions independent of language.

It does branch but the alternative is they have to go through two settings screens every time they want to change a language. This preference really shouldn't be viewed as "catalogue language" or "extension language", it's simply just the list of languages that a user is interested in seeing in the app.

As for new users thinking there aren't that many extensions available. That's really a separate issue regardless of whether you create a second preference. This is also a problem with catalogues as any non-English extension that gets installed isn't visible until it gets explicitly enabled. 
Either you enable all by default and then the user has to disable everything they don't want, or you only enable those that are likely to be used by the user. There are trade offs for either approach. Though I suppose the icon can be changed to the [language icon](http://puu.sh/EsGMu/865f781f92.png) if that makes it more obvious that there's a language setting available.

MCAxiaz: I'm actually quite partial to the idea of something like a clickable list footer that says "enable other extension languages" that just takes you to the settings.

funkyhippo: I suppose it's less about realistic use-cases and more about having two places to change the same setting, then. I was originally thinking about users who might want to check if a particular extension exists (regardless of their language preferences), but that is probably a very niche use-case. 

However, I disagree that it's a separate issue wrt having `en` and their system language shown by default. A second preference could just have all the languages enabled (as you suggested), which would mitigate the problem; I don't think there are any trade offs to this approach relative to displaying only a particular set of languages by default, since the app is multilingual. 

I'm a fan of combining the preferences so you only configure it in one place, but it'd be a challenge to do it in a clean way. Fwiw, [this](https://i.imgur.com/fq9Sdku.mp4) is the implementation I had which used a second preference. I've since tweaked it slightly so the filter icon is tinted when there's an active filter, thus mirroring the behaviour of the library and also giving a clear indicator that there are hidden extensions.

MCAxiaz: I"m saying you have to explicitly enable languages in catalogues to see them. So even if you show every extension language by default, to actually use those extensions you still have to enable them in catalogues. 
Which begs the question, if I'm able to download an extension in a certain language, why do I have to perform extra work to use it. So if you're going to display all languages in extensions then you need to change catalogues to also display all languages. Hence why it's a separate issue.

The trade off is that it puts extra work on the user. A typical user would have practically no interest in having to scroll through all the other languages, and even less interest in manually disabling every single one. So it makes more sense to just show the most commonly used ones, and they can enable any additional ones they need.

MCAxiaz:
https://github.com/inorichi/tachiyomi/pull/2275
Linking this issue since it seems relevant.

I think it's better to have most stuff disabled by default since most users only use a very small set of languages/extensions. But the key is to not have it be even more confusing than it already is for users.

joseph619:
@MCAxiaz This has already been implement in Jays2King's fork :)

arkon:
I think this is still a valid change regardless of forks.

MCAxiaz:
That was an accident oops.`

PR #2276, opened 2019-10-14
Date added

anifake7112k:
Filter- sort  - last manga show on top (time view)


anifake7112k:
I don't know vary much about kotlin but i try to made it if there any fault please correct it. Thanks.

PR #2318, opened 2019-11-08
Made 'Default' category selectable in global update settings

andrecsilva:
It wasn't possible to select the 'Default' category in the categories to include in global update.

This fixes that.


arkon:
What if the user no longer actually has the default category?

andrecsilva:
> 
> 
> What if the user no longer actually has the default category?

I'm unsure if you can actually get rid of it.  The 'Default' category is not actually included in the database (hence is not included in DatabaseHelper.getCategories()). You can't edit it via the edit categories window.

However, whenever you add a manga without selecting a category, it is added in this 'Default' (id= 0) category.

arkon:
Ah you're right. I was thinking about it being hidden if it's empty, but it can always be "re-added" like that.

PR #2332, opened 2019-11-17
Two-Pages in Landscape Mode

loopdeer:
This adds a setting to display two pages on the screen in landscape mode. When displaying two pages, tapping to turn pages "flips" forward two pages instead of one. Additionally when turning, the application will check for two-page spreads (two pages in one page according to the source). If a two page spread is detected on the next page it will flip forward one page to the two-page spread and set the application back to displaying one-page per screen. If a two-page spread is detected on the next-next page the application will try to "align" the page flipping and only advance one page instead of two.

This is my first pull request and work with Android in general and am open to feedback!

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
@@ -1,6 +1,8 @@
 package eu.kanade.tachiyomi.ui.reader.viewer.pager
 
+import android.graphics.BitmapFactory
 import android.support.v4.view.ViewPager
+import android.util.Log
 import android.view.InputDevice
 import android.view.KeyEvent
 import android.view.MotionEvent
@@ -13,6 +15,7 @@ import eu.kanade.tachiyomi.ui.reader.model.ReaderPage
 import eu.kanade.tachiyomi.ui.reader.model.ViewerChapters
 import eu.kanade.tachiyomi.ui.reader.viewer.BaseViewer
 import timber.log.Timber
+import java.io.InputStream
 
 /**
  * Implementation of a [BaseViewer] to display pages with a [ViewPager].
@@ -134,6 +137,30 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
      */
     private fun onPageSelected(page: ReaderPage, position: Int) {
         val pages = page.chapter.pages!! // Won't be null because it's the loaded chapter
+
+        val streamFn = page.stream ?: return
+
+        var openStream: InputStream? = null
+
+        val stream = streamFn().buffered(16)
+        openStream = stream
+
+        val bmp = BitmapFactory.decodeStream(openStream)
+
+        //Two-Page Spread
+        val oldBool = adapter.twoPageSpreadMode
+        if(bmp.width >= bmp.height) {
+            adapter.setBoolS(false)
+            if(adapter.twoPageSpreadMode != oldBool)
+                refreshAdapter()
+        }
+        else {
+            adapter.setBoolS(true)
+            if(adapter.twoPageSpreadMode != oldBool)
+                refreshAdapter()
+        }
+
+
         Timber.d("onPageSelected: ${page.number}/${pages.size}")
         activity.onPageSelected(page)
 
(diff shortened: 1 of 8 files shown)


hXtreme:
You may want to remove the changes from the README file.

allbombson:
It also displays 2 page spreads on the 2nd page

loopdeer:
I will investigate the landscape mode issue later today, but are you flipping pages by tapping or by scrolling (swiping)?

allbombson:
Swiping its a bit sudden but I don't mind. However u may want to look into that.

loopdeer:
On the phone I made the app for (the Axon M) the app doesn't force landscape mode. Make sure your rotation settings are set to automatic or portrait and that the settings in tachiyomi is not set to force landscape.

allbombson:
K i will try again when i get a chance, Can i add you on discord or something github issues r kinda slow.

Laaas:
Any progress on this?

allbombson:
no

arkon:
This might be interesting for foldable devices like the Surface Duo ([SDK docs here](https://docs.microsoft.com/en-us/dual-screen/android/)).

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
```diff
@@ -47,6 +47,14 @@ class SettingsReaderController : SettingsController() {
             defaultValue = "1"
             summary = "%s"
         }
+        intListPreference {
```
arkon: Use a switch preference (which gives you a boolean).

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewerAdapter.kt
```diff
@@ -98,4 +104,18 @@ class PagerViewerAdapter(private val viewer: PagerViewer) : ViewPagerAdapter() {
...
+
+        if(twoPageSpreadMode && config.landscapeMode == 2)
+            return 0.5f
+        else
+            return 1.0f
+    }
+
+    fun setBoolS(bool: Boolean){
+        twoPageSpreadMode = bool
+    }
```
arkon: You can just add a setter on the variable.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewerAdapter.kt
```diff
@@ -98,4 +104,18 @@ class PagerViewerAdapter(private val viewer: PagerViewer) : ViewPagerAdapter() {
...
     }
 
+    override fun getPageWidth(position: Int):Float{
+
+
+
+        if(twoPageSpreadMode && config.landscapeMode == 2)
+            return 0.5f
+        else
+            return 1.0f
```
arkon: This is a bit weird if the page is fitted in a non-default way (e.g. fit height). It's also a bit weird how you can zoom a single page. It might actually need to be a lot more complex that your current approach (e.g. stitching together images in memory?).

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -221,17 +248,123 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
      * Moves to the page at the right.
      */
     protected open fun moveRight() {
```
arkon: All the changes below here are very repetitive.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -134,6 +137,30 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
...
         val pages = page.chapter.pages!! // Won't be null because it's the loaded chapter
+
+        val streamFn = page.stream ?: return
+
+        var openStream: InputStream? = null
+
+        val stream = streamFn().buffered(16)
+        openStream = stream
+
+        val bmp = BitmapFactory.decodeStream(openStream)
```
arkon: I believe this logic is repeated elsewhere. Could extract it.

zurce:
I was able to rebase and port this in the latest master, if @arkon is okay, i can try to revive this feature and send a PR, but i agree that the approach should be more into creating a new page that bundles the two images together, as it would automatically support swipe and tap without extra changes. 

I'm new to kotlin (I do iOS) but i got a Galaxy fold and two page mode is a great feature for this 

fold-manga

arkon:
The issue with this implementation is that it basically just splits the screen into two. It's very awkward when the screen isn't actually big enough (easy example is a regular phone in portrait), or when you've got some zooming/panning around.

zurce:
Yes, i'm gonna look into merging them into 1 page, either by how you said, stitching together the bitmaps or by  maybe doing it a page level, 

that would require less change on the page viewer if I'm correctly, as only the page/image would be changing, so swipes, taps and zoom would work correctly

zurce:
More specifically, define a subclass of ReaderPage named DualReaderPage, that can hold 2 urls and streams, and in PagerPageHolder if the `page` sent in the constructor can be cast as DualReaderPage change `setImage()` to observe and combine the result of both streams to stitch into one page 

when the `PagerViewerAdapte.setChapters` is called i can populate the items array with `DualReaderPage` if the options is turned on 

Again, I'm not an android developer, this coming from my swift and rxswift experience

PR #2348, opened 2019-11-30
fix: Don't send newlines and whitespace in API calls

Flat:
Currently when sending requests to the AniList API newline characters are included in the JSON payload `\n` due to the usage of raw strings. This strips `\n` and trims white space.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt
@@ -24,10 +24,10 @@ class AnilistApi(val client: OkHttpClient, interceptor: AnilistInterceptor) {
 
     fun addLibManga(track: Track): Observable<Track> {
         val query = """
-            mutation AddManga(${'$'}mangaId: Int, ${'$'}progress: Int, ${'$'}status: MediaListStatus) {
-                     SaveMediaListEntry (mediaId: ${'$'}mangaId, progress: ${'$'}progress, status: ${'$'}status)
-                     { id status } }
-                     """
+            |mutation AddManga(${'$'}mangaId: Int, ${'$'}progress: Int, ${'$'}status: MediaListStatus) {
+                     |SaveMediaListEntry (mediaId: ${'$'}mangaId, progress: ${'$'}progress, status: ${'$'}status)
+                     |{ id status } }
+                     |""".trimMargin()
         val variables = jsonObject(
                 "mangaId" to track.media_id,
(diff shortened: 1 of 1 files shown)


arkon:
What's the actual impact from including the characters?

Flat:
We're sending junk data to AniList that could potentially cause errors.

arkon:
Have we actually seen this happening? I could see how trimming might help with some queries, but newlines generally don't have any impact for GraphQL queries.

Flat:
Have any issues occurred so far? Probably not. Is it still good practice to send the least data needed for your query? Yes.

hXtreme:
Realistically speaking, by how much does this change decrease the payload size by? If it's only a very small percentage then I wonder if making this optimization is worth it.

Flat:
By a few bytes. Let me ask this, what is to lose by implementing it since both yourself and arkon seem to be against this change? The other option is to not use raw strings.

arkon:
Nothing, really. If you do want to do it, maybe use `trimMargin`? You should preserve the newlines though, or you might accidentally go from something like `foo\nbar` to `foobar`, which isn't the same thing.

Flat:
From testing and looking at the raw requests in android studio that doesn't happen.
Before:
```json
{
  "query": "\n            mutation UpdateManga($listId: Int, $progress: Int, $status: MediaListStatus, $score: Int) {\n                        SaveMediaListEntry (id: $listId, progress: $progress, status: $status, scoreRaw: $score) {\n                            id\n                            status\n                            progress\n                        }\n                    }\n            ",
  "variables": {
    "listId": 85698528,
    "progress": 83,
    "status": "CURRENT",
    "score": 0
  }
}
```

After:
```json
{
  "query": "mutation UpdateManga($listId: Int, $progress: Int, $status: MediaListStatus, $score: Int) {                        SaveMediaListEntry (id: $listId, progress: $progress, status: $status, scoreRaw: $score) {                            id                            status                            progress                        }                    }",
  "variables": {
    "listId": 85698528,
    "progress": 83,
    "status": "CURRENT",
    "score": 0
  }
}
```

Although there is still excessive whitespace that I will look into removing.

hXtreme:
I don't remember the exact regex language used by kotlin but the following is probably what you want to do:

```
1. Replace '\n' with ' '
# Replace all occurrences of more than one spaces with a single space.
2. Replace ' +' with ' '
3. Trim spaces
```

arkon:
> Although there is still excessive whitespace that I will look into removing.

That's why I suggested `trimMargin`. :)

Flat:
Uses `trimMargin` now

PR #2350, opened 2019-12-01
Fixes #2322. Adds the feature to mark sources as favorite.

### LINKED ISSUE #2322: [Feature Request] Favorite Sources
**App version: 0.84**

**Android version: 9**

**Request:**
I have tons of sources (I basically download almost everyone) but I have a few that I like to see if there's something new.

Like the Manga Plus one... every now and then they release new mangas and every now and then I go there to check if I'm missing new mangas... same for other few sources I like.

I can either hide everything else or endlessly scroll to find those sources. If I'm searching for manga, then I would either have to un-hide the sources or do something else.

**Other details:**

By having favorite sources, I can check those whenever I want more easily and if I need to search new manga, those would show up first. 

Then after that could be the sources you already have manga saved.

After that all others.

It would help a lot when searching for manga or trying to see if your favorite scanlators started new projects.

Thank you all, love the app.

jatinAroraGit:
Added new database table Extensions which records the favorite extensions and all resolvers and type mappings related to it. Mark sources as favorite and the list of sources automatically gets sorted when a source is marked or unmarked as favorite. The favorite source is marked or unmarked by clicking the star shaped checkbox.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -234,7 +234,7 @@ dependencies {
 }
 
 buildscript {
-    ext.kotlin_version = '1.2.71'
+    ext.kotlin_version = '1.3.10'
     repositories {
         mavenCentral()
     }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -234,7 +234,7 @@ dependencies {
 }
 
 buildscript {
-    ext.kotlin_version = '1.3.10'
+    ext.kotlin_version = '1.2.71'
     repositories {
         mavenCentral()
     }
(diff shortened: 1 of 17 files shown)


app/build.gradle
```diff
@@ -234,7 +234,7 @@ dependencies {
 }
 
 buildscript {
-    ext.kotlin_version = '1.2.71'
+    ext.kotlin_version = '1.3.10'
```
hXtreme: Don't update the kotlin version here.

jatinAroraGit: I resolved it in the commit  https://github.com/inorichi/tachiyomi/pull/2350/commits/02045aa9532812b1adf765902be38c750f43ca92

jatinAroraGit: Resolved this in https://github.com/inorichi/tachiyomi/pull/2350/commits/ed2d8eedb25b8629519208fcb3375988a4886367

joseph619:
@jatinAroraGit Can you please post an image or a GIF as an example to show how it works?

Efreak:
Does this allow moving the 'Local' source/catalog to the top of the list?

Soitora:
How's this going?

joseph619:
@Efreak This is only for sources from extensions, Local manga is different

arkon:
This should ideally be saved into preferences, not the database. The database should be things related to the library itself (i.e. roughly what gets backed up/restored).

Issue #2358, opened 2019-12-05
[UI Bug] Unexpected behavior of the search input field


earth2305:
### Device information
* Tachiyomi version: r1534
* Android version: 5.0.2

## Steps to reproduce
1. In the search input field in any catalogue manga search, enter any text in the input field
2. Erase/delete the entered text

### Expected behavior
Should remain focus in the input field

### Actual behavior
After erasing the text a few seconds later searchbar input field disappears, and the catalogue with all the mangas shows again

### Other details
Pictures with all the steps below
![screenshot](https://user-images.githubusercontent.com/8114488/70214814-20e4af00-174d-11ea-9cbc-9dc19c66d768.jpg)

It would be strange if in the Google after deleting all the text from the search bar, focus would disappear and you would have to click on the input field again to search something

PR #2396, opened 2019-12-27
Fix catalogue search focus automatically being removed

MCAxiaz:
Fixes https://github.com/inorichi/tachiyomi/issues/2358

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
@@ -89,15 +89,16 @@ abstract class BaseController(bundle: Bundle? = null) : RestoreViewOnCreateContr
      * This method should be removed when fixed upstream.
      * Issue link: https://issuetracker.google.com/issues/37657375
      */
-    fun MenuItem.fixExpand() {
+    fun MenuItem.fixExpand(onExpand: ((MenuItem) -> Boolean)? = null, onCollapse: ((MenuItem) -> Boolean)? = null) {
         setOnActionExpandListener(object : MenuItem.OnActionExpandListener {
             override fun onMenuItemActionExpand(item: MenuItem): Boolean {
-                return true
+                return onExpand?.invoke(item) ?: true
             }
 
             override fun onMenuItemActionCollapse(item: MenuItem): Boolean {
                 activity?.invalidateOptionsMenu()
-                return true
+
+                return onCollapse?.invoke(item) ?: true
             }
         })
     }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/browse/BrowseCatalogueController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/browse/BrowseCatalogueController.kt
@@ -28,13 +28,9 @@ import eu.kanade.tachiyomi.util.*
 import eu.kanade.tachiyomi.widget.AutofitRecyclerView
 import kotlinx.android.synthetic.main.catalogue_controller.*
 import kotlinx.android.synthetic.main.main_activity.*
-import rx.Observable
 import rx.Subscription
-import rx.android.schedulers.AndroidSchedulers
-import rx.subscriptions.Subscriptions
 import timber.log.Timber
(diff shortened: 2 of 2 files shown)


MCAxiaz:
There's some weird interaction between `fixExpand` and `debounce`... makes it so the searchView can't be closed anymore.

CarlosEsco:
@MCAxiaz  
https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

arkon:
I'd rather not lose the denouncing.

CarlosEsco:
yea not a fan of this implementation.  Without debounce you lose the auto search 
so if i mistype or even delete the entire string it doesn't research or clear the search and show the new results

MCAxiaz:
If you guys really feel that strongly about having the auto search then I'll try something else.

I don't really see too much value in it though and the global search pretty much behaves like this already so it's not exactly going to be inconsistent either.

CarlosEsco:
Global it makes sense cause it could be a ton of sites all at once.   Some people have every extension installed lol.

MCAxiaz:
Makes sense here too then if the issue is excessive network requests. You'd just as well make the argument that a debounce on global search would make sense as well.

Anyhow I've put the debounce back.

MCAxiaz:
@arkon mind reviewing this again?

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
```diff
@@ -89,15 +89,16 @@ abstract class BaseController(bundle: Bundle? = null) : RestoreViewOnCreateContr
      * This method should be removed when fixed upstream.
      * Issue link: https://issuetracker.google.com/issues/37657375
      */
-    fun MenuItem.fixExpand() {
+    fun MenuItem.fixExpand(onExpand: ((MenuItem) -> Boolean)? = null, onCollapse: ((MenuItem) -> Boolean)? = null) {
```
arkon: I'm not entirely sure, but I think this might be causing duplicated "Search" and "Set filter" menu items (the two that were visible for me) to show up in the menu still even when the `SearchView` is expanded. Tapping "Search" again (when it's already expanded) makes it disappear.

MCAxiaz: Screenshot_20200110-193524_Tachiyomi
Are you referring to this?

arkon: Yeah.

MCAxiaz: Oh that happens on dev too.

arkon: Dang, you're right. 😔

MCAxiaz: I think if you change the menu item's attributes from `app:showAsAction="ifRoom"` to `always` that might fix it.

app/src/main/java/eu/kanade/tachiyomi/ui/catalogue/browse/BrowseCatalogueController.kt
```diff
@@ -232,11 +232,12 @@ open class BrowseCatalogueController(bundle: Bundle) :
             searchViewSubscription?.unsubscribe()
             searchViewSubscription = Observable.merge(writingObservable, submitObservable)
                     .map { it.queryText().toString() }
-                    .distinctUntilChanged()
```
arkon: Why isn't this needed anymore?

MCAxiaz: Never actually needed since searchWithQuery does a `presenter.query == newQuery` check before doing any work.

Normally the searchview will never emit the same time twice since it is a text change event after all. But the searchview does emit an empty string every time it's opened, so it would repeatedly emit empty string if you just repeatedly open and close the searchview without typing anything. But fixExpand invalidates the menu every time it's closed so we actually get a new subscription every time so that no longer applies.

Issue #2401, opened 2019-12-28
[Bug] Catalogue List View Collapses after Search


MCAxiaz:
### Device information
* Tachiyomi version: 0.8.4, latest dev
* Android version: 28

## Steps to reproduce
![device-2019-12-27-193844](https://user-images.githubusercontent.com/24946428/71538494-cd472a80-28e0-11ea-9492-738766d6056a.gif)

### Expected behavior
List view should not collapse

arkon:
Reproduced in stable 0.8.4.

arkon:
Does this happen with the change in #2396 too?

MCAxiaz:
Yea it's unrelated.

CarlosEsco:
verified pull request works

PR #2402, opened 2019-12-28
Issue #2401 Catalogue List View

MCAxiaz:
https://github.com/inorichi/tachiyomi/issues/2401

This should do the trick.

Recycler view should not be fixed size.


arkon:
Aren't there performance implications to this? I thought the fixed size setting was for its children.

PR #2403, opened 2019-12-28
MAL Tracking Improvements

MCAxiaz:
MAL API waiting room 💀

MAL has a known issue where it invalidates any login sessions if they are inactive for a couple of days.

Auto attempt a login refresh if MAL returns a 400 indicating the login is no longer valid.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeList.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeList.kt
@@ -112,11 +112,7 @@ class Myanimelist(private val context: Context, id: Int) : TrackService(id) {
                 .toCompletable()
     }
 
-    // Attempt to login again if cookies have been cleared but credentials are still filled
-    fun ensureLoggedIn() {
-        if (isAuthorized) return
-        if (!isLogged) throw Exception("MAL Login Credentials not found")
-
+    fun refreshLogin() {
         val username = getUsername()
         val password = getPassword()
         logout()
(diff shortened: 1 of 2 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeList.kt
```diff
@@ -131,6 +127,14 @@ class Myanimelist(private val context: Context, id: Int) : TrackService(id) {
         }
     }
 
+    // Attempt to login again if cookies have been cleared but credentials are still filled
+    fun ensureLoggedIn() {
+        if (isAuthorized) return
+        if (!isLogged) throw Exception("MAL Login Credentials not found")
```
arkon: This should be localized.

MCAxiaz: Yes but it doesn't make sense to make changes here. You'd end up having to inject context everywhere just for getting a string. 

And exceptions shouldn't be directly exposed to the user anyways (there's no guarantee all exceptions even have messages in the first place, and you can't control exceptions thrown by libraries). It really should be for logging and debugging purposes. 

If anything the app needs an overhaul to handle exceptions better and display sensible user-facing messages. (e.g. the good old 'Unknown Error' when logging into MAL).

Plus every single thrown exception in the app already uses hardcoded strings 🤷‍♂

arkon: Yeah, that's fair.

PR #2406, opened 2019-12-28
add width and height to listview for browseCatalogueController

CarlosEsco:
Fixes #2409  in a more performance minded way

@arkon  After you mentioned potential performance issues.  I realized the issue was for listview there it does not match parent for width and height like the grid view (catalogue_recycler_autofit) does


CarlosEsco:
whoops didnt update before opening.  Let me add the fixedSize back

PR #2409, opened 2019-12-30
Improvements to EPUB support.

Steveice10:
Fixes two problems I encountered trying to view local manga EPUBs in Tachiyomi:
* Image tags using relative paths were not resolved correctly.
* "META-INF/container.xml" couldn't be located in some EPUBs that used "\\" as a path separator.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/util/storage/EpubFile.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/util/storage/EpubFile.kt
@@ -93,25 +93,26 @@ class EpubFile(file: File) : Closeable {
      * Returns all the images contained in every page from the epub.
      */
     private fun getImagesFromPages(pages: List<String>, packageHref: String): List<String> {
+        val result = ArrayList<String>()
         val basePath = getParentDirectory(packageHref)
-        return pages.map { page ->
+        pages.forEach { page ->
             val entryPath = resolveZipPath(basePath, page)
             val entry = zip.getEntry(entryPath)
             val document = zip.getInputStream(entry).use { Jsoup.parse(it, null, "") }
             val imageBasePath = getParentDirectory(entryPath)
 
             val imgs = document.getElementsByTag("img")
-            if(imgs.isNotEmpty()) {
-                imgs.mapNotNull {
-                    resolveZipPath(imageBasePath, it.attr("src"))
-                }
-            } else {
-                val images = document.getElementsByTag("image")
-                images.mapNotNull {
-                    resolveZipPath(imageBasePath, it.attr("xlink:href"))
-                }
+            imgs.forEach {
+                result.add(resolveZipPath(imageBasePath, it.attr("src")))
             }
-        }.flatten()
+
+            val images = document.getElementsByTag("image")
+            images.forEach {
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/util/EpubFile.kt
```diff
@@ -43,27 +43,38 @@ class EpubFile(file: File) : Closeable {
      * Returns the path of all the images found in the epub file.
      */
     fun getImagesFromPages(): List<String> {
-        val allEntries = zip.entries().toList()
-        val ref = getPackageHref()
+        val pathSeparator = getPathSeparator()
```
arkon: Could potentially do this in an `init` block, along with the `val zip = ZipFile(file)` assignment? That way you can just have it as a field in the class and you don't need to pass it around.

app/src/main/java/eu/kanade/tachiyomi/util/EpubFile.kt
```diff
@@ -89,28 +100,56 @@ class EpubFile(file: File) : Closeable {
...
+            val imgs = document.getElementsByTag("img")
+            if(imgs.isNotEmpty()) {
+                imgs.mapNotNull {
+                    resolveZipPath(imageBasePath, it.attr("src"), pathSeparator)
+                }
+            } else {
+                val images = document.getElementsByTag("image")
+                images.mapNotNull {
+                    resolveZipPath(imageBasePath, it.attr("xlink:href"), pathSeparator)
+                }
```
arkon: What if there happens to be both? I assume that generally doesn't happen though, unless it's a really wonky EPUB.

Steveice10: I can change this to be able to handle both on the same page. In practice, I've only seen this used in calibre-generated title pages.

arkon:
Do you plan on updating this PR?

Steveice10:
Yes. Sorry for the delay, I just haven't gotten around to it yet.

Steveice10:
Sorry for taking so long to get back to this. I've rebased on top of the latest source and addressed any comments so far.

app/src/main/java/eu/kanade/tachiyomi/util/storage/EpubFile.kt
```diff
@@ -89,28 +92,69 @@ class EpubFile(file: File) : Closeable {
...
+
+            val imgs = document.getElementsByTag("img")
+            imgs.forEach {
+                result.add(resolveZipPath(imageBasePath, it.attr("src")))
+            }
+
+            val images = document.getElementsByTag("image")
+            images.forEach {
+                result.add(resolveZipPath(imageBasePath, it.attr("xlink:href")))
+            }
```
arkon, line 112: Wouldn't this result in weird behaviour if the page (for some reason) had `img` and `image` tags interchangeably?

Steveice10: Should be fixed now.

arkon:
Awesome, thanks for updating it!

Issue #1645, opened 2018-10-04
[Feature Request] Ability to search downloaded manga with tags


Amazingduf:
Is it possible to search the downloaded manga folder with tags ? Like trying Comedy and all the manga that I downloaded that have the Comedy tag appear

thu2468:
Would be nice to search/filter genre in library that you added already. Pretty sure it possible as it looks like the app already store genre. So you could make it so that it get all unique genre naming and then make a list out of that.

PR #2423, opened 2020-01-04
Local genre tag searching

Jays2Kings:
Using the search bar in My Library, you can search tags for manga (ie. "Romance") or exclude (ie. "-Comedy") You can also search multiple by seperating by commas (ie. "Romance, -Comedy")

Clicking the tag in manga info from the library also performs a local serach

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryItem.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryItem.kt
@@ -59,7 +59,22 @@ class LibraryItem(val manga: LibraryManga, private val libraryAsList: Preference
      */
     override fun filter(constraint: String): Boolean {
         return manga.title.contains(constraint, true) ||
-                (manga.author?.contains(constraint, true) ?: false)
+            (manga.author?.contains(constraint, true) ?: false) ||
+            if (constraint.contains(",")) {
+                val genres = manga.genre?.split(", ")
+                constraint.split(",").all { containsGenre(it.trim(), genres) }
+            }
+            else containsGenre(constraint, manga.genre?.split(", "))
+    }
+
+    private fun containsGenre(tag: String, genres: List<String>?): Boolean {
+        return if (tag.startsWith("-"))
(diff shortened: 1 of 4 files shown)


arkon:
Closes #1645, I guess?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryItem.kt
```diff
@@ -57,7 +57,22 @@ class LibraryItem(val manga: LibraryManga, private val libraryAsList: Preference
      */
     override fun filter(constraint: String): Boolean {
         return manga.title.contains(constraint, true) ||
-                (manga.author?.contains(constraint, true) ?: false)
+            (manga.author?.contains(constraint, true) ?: false) ||
```
arkon: Out of scope of this PR, but maybe we should be searching for artists too.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryItem.kt
```diff
@@ -57,7 +57,22 @@ class LibraryItem(val manga: LibraryManga, private val libraryAsList: Preference
...
-                (manga.author?.contains(constraint, true) ?: false)
+            (manga.author?.contains(constraint, true) ?: false) ||
+            if (constraint.contains(",")) {
+                val genres = manga.genre?.split(", ")
+                constraint.split(",").all { containsGenre(it.trim(), genres) }
+            }
+           else containsGenre(constraint, manga.genre?.split(", "))
+    }
+
+    private fun containsGenre(tag: String, genres: List<String>?): Boolean {
```
arkon: I do wonder if we should be doing partial matches (i.e. `contains`, like for title/author) instead of strict equality though.

Jays2Kings:
To me, contains makes sense for title and less for author and artist (but still does). But for tags I feel that it could false positives.

arkon:
I suppose that feels intuitive enough to me too.

Can you resolve the merge conflicts?

PR #2426, opened 2020-01-05
Improve Loading Speed When Skipping Pages in a Chapter

### LINKED ISSUE #2386: [Feature Request] Improve Page Loading When Skipping Pages in a Chapter

Currently if you decide to skip pages in the reader (to find a particular scene for example), you have to wait for all the previous pages that were queued to finish loading.

A workaround is to mark the chapter as unread, scroll to the desired page, then exit and reopen the reader to have it start loading from that page.

### What/Requirements

Page loading should cancel for any pages that the user has skipped.

Example:

If I go to a previous chapter and scroll to near the beginning, the reader should not load all the pages I skipped in descending page number. It should, after scrolling has stopped (perhaps with a 1 or 2 second debounce), cancel any pending loads and immediately start loading from the page I ended up at.

Same applies If I were to skip forward in a chapter.

MCAxiaz:
resolves #2386 

cancel queued loads when the page that requested the loads is destroyed

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/HttpPageLoader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/HttpPageLoader.kt
@@ -17,6 +17,7 @@ import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 import java.util.concurrent.PriorityBlockingQueue
 import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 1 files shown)


joseph619:
Is this for regular scroll or applies to seekbar as well? Currently the seekbar in reader is useless since it keeps loading previous pages forever

MCAxiaz:
Both.

joseph619:
Nice!

PR #2460, opened 2020-01-08
CloudflareInterceptor Update / Bypass

happywillow0:
Request for comments and review:

After staring at the code for who knows how long, I determined that the original check for the CF interceptor comes down to shouldOverrideUrlLoading 

Cloudflare 'I'm under attack' check utilizes a POST method to validate itself. As per the [developer reference:](https://developer.android.com/reference/android/webkit/WebViewClient.html#shouldOverrideUrlLoading(android.webkit.WebView,%20android.webkit.WebResourceRequest))
> Note: This method is not called for POST requests. 

The proposed change bypasses the challenge solution url check because webview isn't able to intercept a POST request. The 12 seconds should allow for the challenge to load and the post to go though via webview before the app proceeds and loads the manga catalog again. 

MangaFree is one I identified to take longer to load but still should be fine after hitting retry.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
@@ -54,7 +54,7 @@ class CloudflareInterceptor(private val context: Context) : Interceptor {
     }
 
     private fun isChallengeSolutionUrl(url: String): Boolean {
-        return "chk_jschl" in url
+        return true //"chk_jschl" in url
     }
 
     @SuppressLint("SetJavaScriptEnabled")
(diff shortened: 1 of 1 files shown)


arkon:
This seems... strange to me? It sounds weird that you can just wait for the CF page to go past. Also, what happens in cases where it's _not_ the "I'm under attack" page?

happywillow0:
> This seems... strange to me? It sounds weird that you can just wait for the CF page to go past. 

As far as I'm aware, we are using webview to resolve the CF check in the background. Webview loads the page and has javascript enabled to solve the check. 

The only issue right now is that we are unable to intercept the "solved" URL because it gets POSTed to the server and the response payload is main/requested page 

> Also, what happens in cases where it's _not_ the "I'm under attack" page?

The CF interceptor is only triggered by the http 503 error so if the "I'm under attack" page is disabled, everything is fully bypassed.

This is also why APC was presenting the same error because Wordfence was also reporting error 503.

PR #2472, opened 2020-01-10
remove google assistant search intent handling

MCAxiaz:
Pretty sure this only works if the app is published on the play store.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -34,7 +34,6 @@
             </intent-filter>
             <intent-filter>
                 <action android:name="android.intent.action.SEARCH" />
-                <action android:name="com.google.android.gms.actions.SEARCH_ACTION"/>
                 <category android:name="android.intent.category.DEFAULT"/>
             </intent-filter>
             <intent-filter>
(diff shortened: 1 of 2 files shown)


arkon:
Nope, it works fine. Pronunciation is weird though, but typing it definitely works.

MCAxiaz:
@arkon 🤷‍♂ 
device-2020-01-10-130416

manga woo is some tachiyomi clone on the playstore.

arkon:
Try Google Assistant, not Google Search.

MCAxiaz:
@arkon  🤷‍♂ 
still-nope

arkon:
Dunno what to tell you, it works on mine. :P

joseph619:
I guess it's just my phone then, closing

MCAxiaz:
-___- why is android such a pos. It doesn't work either on my old note 3 on 5.0.

It does work on my s9 on 8.1 but only the dev version. Stable doesn't work for some reason even though the code should be identical.

wopian:
> Stable doesn't work for some reason even though the code should be identical.

Yup, it was the last thing added to the 0.8.4 stable release 9 months ago and there hasn't been any code changes to the intent manifest or related code.

---

0.8.4 (installed on new profile) and r1609 works for me in Google Search (don't have assistant) on Android 9 and 10

AmineI:
Yeah tbh I've had this issue myself during testing when I was working on that feature's PR, where Tachiyomi wasn't recognized by Google app/assistant, but it was quite rare so I assumed the Google app didn't pick it up sometimes when installing it or something.

I think it's an issue with how or when the Google app registers that, but I don't know much more. During my testing reinstalling would usually do the trick, it was kind of random and rare so I couldn't really investigate. But it definitely works without releasing it to the play store.

PR #2495, opened 2020-01-12
Group notifications for Library updates

Jays2Kings:
Two down
Screenshot_20200112-124902
Screenshot_20200112-124914


app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
```diff
@@ -104,7 +120,7 @@ class NotificationReceiver : BroadcastReceiver() {
         val db = DatabaseHelper(context)
         val manga = db.getManga(mangaId).executeAsBlocking()
         val chapter = db.getChapter(chapterId).executeAsBlocking()
-
+        context.sendBroadcast(Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS))
```
arkon: What's this for?

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -88,6 +92,7 @@ class LibraryUpdateService(
             .setLargeIcon(notificationBitmap)
             .setOngoing(true)
             .setOnlyAlertOnce(true)
+            .setColor(ContextCompat.getColor(this, R.color.colorAccentLight))
```
arkon: I might just pull small parts like this into the repo separately so this PR stays a bit more specific to the library update notification functionality.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -437,39 +444,74 @@ class LibraryUpdateService(
...
+                setContentText(chaptersNames)
+                setStyle(NotificationCompat.BigTextStyle().bigText(chaptersNames))
+                priority = NotificationCompat.PRIORITY_HIGH
+                setGroup(Notifications.GROUP_NEW_CHAPTERS)
+                setContentIntent(
+                    NotificationReceiver.openChapterPendingActivity(
+                        this@LibraryUpdateService, manga, chapters.first()
+                    )
+                )
+                addAction(R.drawable.ic_glasses_black_24dp, getString(R.string.action_mark_as_read),
```
arkon: This is tangential, but I don't remember ever seeing icons. 

Less tangential, but existing actions all use rasterized icons for some reason, presumably for older versions of Android.

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
```diff
@@ -246,6 +293,32 @@ class NotificationReceiver : BroadcastReceiver() {
...
+            null) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                val groupKey = context.notificationManager.activeNotifications.find {
+                    it.id == notificationId
+                }?.groupKey
+                if (groupId != null && groupId != 0 && groupKey != null && groupKey.isNotEmpty()) {
+                    val notifications = context.notificationManager.activeNotifications.filter {
+                        it.groupKey == groupKey
+                    }
+                    if (notifications.size == 2) {
```
arkon: What exactly does this mean?

app/src/main/java/eu/kanade/tachiyomi/data/notification/Notifications.kt
```diff
@@ -23,15 +23,21 @@ object Notifications {
      * Notification channel and ids used by the library updater.
      */
     const val CHANNEL_LIBRARY = "library_channel"
-    const val ID_LIBRARY_PROGRESS = 101
-    const val ID_LIBRARY_RESULT = 102
+    const val ID_LIBRARY_PROGRESS = -101
```
arkon: Discussed with jay on Discord, but the intention behind making these negative is to avoid having them conflict with manga IDs.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
Jays2Kings: Iirc Android 8 or 9 plus doesn't have icons, and since these would only show on Android 7+ the icons would show there

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
Jays2Kings: I'm honestly trying to remember, it would closes and dialogs of a chapter was opened from them, but I don't think there's any dialogs that open chapters anymore. I'll look around and confirm

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
Jays2Kings: With group notifications there's always at least 2 notifications, even if there's a single manga update: the group summary notification (which looks like the notification now) and the single manga notification. 

If the single notification is dismissed by the system, ie by a user swipe or tapping on the notification, it will auto dismiss the group notification if there's no other single updates.

When programmatically dismissing this notification, the group notification is not dismissed, so say you mark the notification as read or open chapters. It will dismiss that notification and the old group notification will appear from nowhere. This logic is to dismiss that one as well if that notification and the group are the only ones left

app/src/main/java/eu/kanade/tachiyomi/data/notification/NotificationReceiver.kt
arkon: Did you ever remember why this was needed?

Jays2Kings: I believe it was used for when opening the reader activity was done using a pending broadcast and not a pending activity. Because it was a broadcast tapping on the notification didn't close the notification shade.

Since it now a pending activity, this is not needed. I'll try to fix these up and add the last few details tomorrow or over the week

Jays2Kings:
Closing due to renaming repo

PR #2497, opened 2020-01-13
Advance delete Selective Chapters

bairagyakushal4:
Delete chapters options from advance menu to mass delete all chapters and folder of selected mangas only. Improve version of "Clean up downloaded chapter" with option to select manga.


CarlosEsco:
dunno if you meant to open this request on main tachiyomi but closing so you can fix the commits.  Looks like you basically have all of jays commits included mistakenly

PR #2524, opened 2020-01-20
Do not suppress a type system error

pyos:
This code was sort of fine when it used raw Java types, but the Kotlin equivalent technically calls a method that takes a Nothing-typed argument with a value that is not of type Nothing. Whether that works depends on how lenient kotlinc is about inserting casts in bytecode.

The solution is to give the unknown type represented by a star an explicit name by capturing it in a type variable, then cast to that type instead of Nothing. This is guaranteed to be an unchecked, but valid, cast.


arkon:
Nice, thanks!

PR #2537, opened 2020-01-24
CloudflareInterceptor update

SnakeDoc83:
The idea here is that instead of checking the URL to see if Cloudflare has been successfully bypassed, instead we check if we have the "cf_clearance" cookie that's set when Cloudflare is bypassed.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/network/AndroidCookieJar.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/network/AndroidCookieJar.kt
@@ -31,27 +31,25 @@ class AndroidCookieJar : CookieJar {
         }
     }
 
-    fun remove(url: HttpUrl) {
+    fun remove(url: HttpUrl, cookieNames: List<String>? = null) {
         val urlString = url.toString()
         val cookies = manager.getCookie(urlString) ?: return
 
+        fun List<String>.filterNames(): List<String> {
+            return if (cookieNames != null) {
+                this.filter { it in cookieNames }
+            } else {
+                this
+            }
+        }
+
         cookies.split(";")
             .map { it.substringBefore("=") }
+            .filterNames()
             .onEach { manager.setCookie(urlString, "$it=;Max-Age=-1") }
     }
 
     fun removeAll() {
         manager.removeAllCookies {}
     }
-
-    fun removeCloudflare(url: HttpUrl) {
-        val urlString = url.toString()
-        val cookies = manager.getCookie(urlString) ?: return
-
-        cookies.split(";")
-                .filter { it.contains("__cfduid") || it.contains("cf_clearance") }
-                .map { it.substringBefore("=") }
-                .onEach { manager.setCookie(urlString, "$it=;Max-Age=-1") }
-    }
-
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/network/AndroidCookieJar.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/network/AndroidCookieJar.kt
@@ -31,7 +31,7 @@ class AndroidCookieJar : CookieJar {
         }
     }
 
-    fun remove(url: HttpUrl, cookieNames: List<String>? = null) {
+    fun remove(url: HttpUrl, cookieNames: List<String>? = null, maxAge: Int = -1) {
         val urlString = url.toString()
         val cookies = manager.getCookie(urlString) ?: return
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
```diff
@@ -135,16 +126,7 @@ class CloudflareInterceptor(private val context: Context) : Interceptor {
...
-        val solution = solutionUrl ?: throw Exception("Challenge not found")
-
-        return Request.Builder().get()
-            .url(solution)
-            .headers(request.headers)
-            .addHeader("Referer", origRequestUrl)
-            .addHeader("Accept", "text/html,application/xhtml+xml,application/xml")
-            .addHeader("Accept-Language", "en")
-            .build()
+        return
```
arkon: Don't need this.

app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
```diff
@@ -22,6 +23,10 @@ class CloudflareInterceptor(private val context: Context) : Interceptor {
 
     private val handler = Handler(Looper.getMainLooper())
 
+    private val networkHelper: NetworkHelper by injectLazy()
+
+    private var bypassSucceeded: Boolean? = null
```
arkon: Not sure if there might be issues with having this at the class level. I think it'd be possible for CF to kick in again at some point later with the same instance, but maybe that's only on higher levels of protection?

app/src/main/java/eu/kanade/tachiyomi/network/AndroidCookieJar.kt
```diff
@@ -44,4 +44,14 @@ class AndroidCookieJar : CookieJar {
         manager.removeAllCookies {}
     }
 
+    fun removeCloudflare(url: HttpUrl) {
```
arkon: I'd prefer if you modify the existing `remove` function to take in an optional list of cookie names. This class ideally shouldn't be concerned about Cloudflare specifically.

app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
SnakeDoc83: Ah, yeah.  With the changes I've made I'm returning a Boolean now.

app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
SnakeDoc83: Changed just in case.

app/src/main/java/eu/kanade/tachiyomi/network/AndroidCookieJar.kt
SnakeDoc83: Done.  One thing I'm not sure about with my implementation, would there be an issue if `.onEach` tries to act on an empty list (which could happen if none of the cookies in the manager have names that are in `cookieNames`?

SnakeDoc83:
Ugh, need to fix something.

SnakeDoc83:
Think those are going to be the last changes until you look at this again.

app/src/main/java/eu/kanade/tachiyomi/network/AndroidCookieJar.kt
arkon: It's fine if it's an empty list, it just won't do anything basically.

app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
```diff
@@ -35,14 +39,21 @@ class CloudflareInterceptor(private val context: Context) : Interceptor {
...
 
         // Check if Cloudflare anti-bot is on
         if (response.code == 503 && response.header("Server") in serverCheck) {
             try {
                 response.close()
-                val solutionRequest = resolveWithWebView(chain.request())
-                return chain.proceed(solutionRequest)
+                networkHelper.cookieManager.remove(originalRequest.url, listOf("__cfduid", "cf_clearance"), 0)
+                val oldCookie = networkHelper.cookieManager.get(originalRequest.url)
+                        .firstOrNull { it.name == "cf_clearance" }
```
arkon: Do we not care about the `__cfduid` cookie?

SnakeDoc83: Not really,  When we hit a website with an active anti-bot `_ cfduid` gets set on the initial request/challenge page, then `cf_clearance` gets set once we're past it.

PR #2538, opened 2020-01-24
Manga Info Page UI Update

### LINKED ISSUE #1699: [Bug] Description hidden when there are too many genre tags, scroll is also missing
**App version:** r1454

**Issue/Request:** 
1. The description text is hidden because there are too many tags. Even the title **Description** itself is overlapped.
2. The genre tags area is not scrollable. In the image below, there are still missing genres that cannot be shown/scrolled to.

image

The above problems occur only in portrait mode. How about moving the TagGroup view into the NestedScrollView containing the manga summary, so that we can achieve something like this?

image

MCAxiaz:
Fixes #1699 

Also relevant #1194. 

**Current**:
- Some titles are still too long to fit.
- Tiny font title looks bad imo.
- Autosize text sucks and never behaves consistently (case in point: note the extra space below the title that disappears after refresh). Also technically it's meant for fixed-sized containers, not `wrap_content` so really shouldn't have used it in the first place.

ezgif com-resize

**Changes**:
- Move title above the cover image to have it span the full screen width; remove autosize and max 2 lines.
- Allow the entire page to scroll and get rid of the description nested scroll.
- Add edge fade to the scroll views (pretty subtle but it comes for free so why not).

ezgif com-resize(2)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
@@ -223,7 +223,7 @@ class MangaInfoController : NucleusController<MangaInfoPresenter>(),
                     .centerCrop()
                     .into(manga_cover)
 
-            if (backdrop != null) {
+            backdrop?.post {
                 GlideApp.with(view.context)
                         .load(manga)
                         .diskCacheStrategy(DiskCacheStrategy.RESOURCE)
--- a/app/src/main/res/layout-land/manga_info_controller.xml
+++ b/app/src/main/res/layout-land/manga_info_controller.xml
@@ -9,11 +9,8 @@
     android:layout_height="match_parent">
 
     <androidx.constraintlayout.widget.ConstraintLayout
-        xmlns:app="http://schemas.android.com/apk/res-auto"
-        xmlns:tools="http://schemas.android.com/tools"
         android:layout_width="match_parent"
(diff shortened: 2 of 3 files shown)


CarlosEsco:
How bad does this one look?

https://mangadex.org/manga/41726/

arkon:
Yeah, not that great with super long titles.

Screenshot_20200124-114551

I kind of liked the cover being in the top left. Maybe could have it toggle between truncated/full text on tap or something?

CarlosEsco:
Another option is change the text for the title to the narrow font from the library.  Though that custom text view doesnt like a straight replacement of the normal text view

MCAxiaz:
> I have feeling its gonna look ugly. Now that the title is match parent width couldn't you go back to autosize text and do a limit of say 3 lines?

Still need to constrain the height to some fixed value instead of `wrap_content` which doesn't seem feasible in case.

@arkon that shouldn't happen. If it does it's another bug. How often are you able to repro that issue? It shows fine for me.
Screenshot_1579889394

MCAxiaz:
Ah ok, If the manga isn't saved in the database then during the network request the view has a chance to load using the default values (the title only has one line), and then the image gets loaded using that height instead of waiting for the height to adjust to the new height after updating the title. Should be simple, I'll fix it later today.

MCAxiaz:
Using that PTSansTextView.
Screenshot_1579927149

arkon:
Going to be making some larger changes to this view.

PR #2563, opened 2020-02-03
Library View: Add latest chapter sorting and revert last updated sorting changes

### LINKED ISSUE #2539: [Feature Request] Revert change to "last updated" library sorting and add "latest chapter date" sort option
"Last updated" more closely reflects fetch date as ee8a5318 had it, and "latest chapter date" describes exactly what the current behavior is as of a22c79c5. This issue is for making those both available as distinct sort options.

FlaminSarge:
Latest chapter is as it sounds
Last update is now any changes to the chapter list (addition, removal, rename, etc.; this is what the behavior was prior to the commits referenced in #2539)

Fixes #2539

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
@@ -117,7 +117,7 @@ class LibraryNavigationView @JvmOverloads constructor(context: Context, attrs: A
 
         private val source = Item.MultiSort(R.string.manga_info_source_label, this)
 
-        private val latestChapter = Item.MultiSort(R.string.manga_info_last_chapter_label, this)
+        private val latestChapter = Item.MultiSort(R.string.action_sort_latest_chapter, this)
 
         override val items = listOf(alphabetically, lastRead, lastUpdated, unread, total, source, latestChapter)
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryNavigationView.kt
```diff
@@ -117,7 +117,9 @@ class LibraryNavigationView @JvmOverloads constructor(context: Context, attrs: A
 
         private val source = Item.MultiSort(R.string.manga_info_source_label, this)
 
-        override val items = listOf(alphabetically, lastRead, lastUpdated, unread, total, source)
+        private val latestChapter = Item.MultiSort(R.string.manga_info_last_chapter_label, this)
+
+        override val items = listOf(alphabetically, lastRead, lastUpdated, unread, total, source, latestChapter)
```
FlaminSarge: May need to change this string. Currently reuses 'Last chapter' string from manga info.

May also need to move latestChapter around in the sorting type list.

CarlosEsco: Not your fault cause it was already like this but i feel the order should be alphabetical

arkon: I'm not really sure it matters if it'll be different in other languages anyway.

arkon:
Did you ever decide on better names?

FlaminSarge:
I'll do that today and update.

FlaminSarge:
Debating "Latest chapter" vs "Latest chapter date", but none of the others use the word 'date' so... consistency.

(If we did use the word 'date', the others would be e.g. "Last update date", "Last read date")

arkon:
LGTM for now. Naming is a whole other headache that I agree we can come back to later. :)

PR #2595, opened 2020-02-16
Fix fullscreen reader for notch device

onlymash:
- Notch device support fullscreen 
- Make fullscreen switching smoother

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.ui.reader
 
 import android.annotation.SuppressLint
+import android.annotation.TargetApi
 import android.app.ProgressDialog
 import android.content.Context
 import android.content.Intent
@@ -120,9 +121,6 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
             else -> R.style.Theme_Reader
         })
         super.onCreate(savedState)
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-            window.attributes.layoutInDisplayCutoutMode = WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES
-        }
         setContentView(R.layout.reader_activity)
 
         if (presenter.needsInit()) {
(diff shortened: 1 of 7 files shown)


onlymash:
- Before
VID_20200216_222002_02

- After
VID_20200216_222002_03

wopian:
Won't this cause text to be cut off for those that have "Crop Borders" enabled?

arkon:
This could replace #2121 if it's updated.

I'd prefer for this to be configurable.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -646,6 +653,14 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
             }
         }
 
+        @TargetApi(Build.VERSION_CODES.P)
+        private fun setCutoutShort(enabled: Boolean) {
```
arkon: Is there a way to force a re-layout after this? It seems that toggling the option doesn't make it take effect until I show/hide the toolbars again.

app/src/main/res/values/strings.xml
```diff
@@ -184,6 +184,7 @@
 
       <!-- Reader section -->
     <string name="pref_fullscreen">Fullscreen</string>
+    <string name="pref_cutout_short">Content renders into the cutout area</string>
```
arkon: Maybe something shorter like "Show content in cutout area"?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
```diff
@@ -74,6 +74,13 @@ class SettingsReaderController : SettingsController() {
             titleRes = R.string.pref_fullscreen
             defaultValue = true
         }
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
```
arkon: It'd be nice if this only showed up when there actually is display cutout(s), but I'm not sure if it's actually possible to check that.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
onlymash: recreate activity(the worst practice)

arkon: Hmm... I guess it's fine for now then. We can revisit it if people complain.

Issue #2487, opened 2020-01-11
[Bug] Page Scrolling bar is off


shogunreaper:
### Device information
* Tachiyomi version: r1624 is the latest one i've seen it on but i'm pretty sure its been there since the new reader was made.
* Android version: 5 I think? I'm using a kindle fire 10 hd 2017) so its the custom amazon version.

## Steps to reproduce
1. Open manga
2. tap on screen to bring the page scroller up

### Expected behavior
This is what it looked like in older versions before the new reader.
https://i.imgur.com/ew8w6zZ.jpg

### Actual behavior
this is what it looks like now
https://i.imgur.com/MDqoNOj.jpg

arkon:
I'm pretty sure there's an open issue about this but I can't find it right now.

Have you changed your scaling settings at all?

shogunreaper:
no i havent i didnt even know there was scaling settings

arkon:
I mean on the device (not the app), but I assume you haven't

shogunreaper:
Oh no i haven't but like i said its only on the newer versions so i'd have to assume it's something wrong with the app itself.

PR #2601, opened 2020-02-18
Show page number only when toolbar is hidden

onlymash:
The toolbar or navigation bar is translucent, so the page number should be hidden


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -37,12 +37,10 @@ import eu.kanade.tachiyomi.util.lang.plusAssign
...
 import eu.kanade.tachiyomi.util.system.GLUtil
 import eu.kanade.tachiyomi.util.system.toast
-import eu.kanade.tachiyomi.util.view.gone
-import eu.kanade.tachiyomi.util.view.visible
+import eu.kanade.tachiyomi.util.view.*
 import eu.kanade.tachiyomi.widget.SimpleAnimationListener
 import eu.kanade.tachiyomi.widget.SimpleSeekBarListener
 import kotlinx.android.synthetic.main.reader_activity.*
-import me.zhanghai.android.systemuihelper.SystemUiHelper
 import nucleus.factory.RequiresPresenter
```
arkon: I believe you can remove a dependency now.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -37,12 +37,10 @@ import eu.kanade.tachiyomi.util.lang.plusAssign
 import eu.kanade.tachiyomi.util.storage.getUriCompat
 import eu.kanade.tachiyomi.util.system.GLUtil
 import eu.kanade.tachiyomi.util.system.toast
-import eu.kanade.tachiyomi.util.view.gone
-import eu.kanade.tachiyomi.util.view.visible
+import eu.kanade.tachiyomi.util.view.*
 import eu.kanade.tachiyomi.widget.SimpleAnimationListener
 import eu.kanade.tachiyomi.widget.SimpleSeekBarListener
 import kotlinx.android.synthetic.main.reader_activity.*
-import me.zhanghai.android.systemuihelper.SystemUiHelper
```
arkon: I believe you can remove a dependency now.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -570,9 +579,6 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
             subscriptions += preferences.trueColor().asObservable()
                     .subscribe { setTrueColor(it) }
 
-            subscriptions += preferences.fullscreen().asObservable()
```
arkon: Shouldn't we still have something like this to handle it when it gets toggled in the reader?

onlymash: setMenuVisibility will be called

arkon:
Tested in API 21, and it looks like this after toggling fullscreen off:

Screen Shot 2020-02-18 at 17 29 22

It's fine after the activity gets recreated though (i.e. exiting and entering again or rotating).

app/src/main/res/layout/reader_activity.xml
```diff
@@ -84,9 +83,7 @@
             <eu.kanade.tachiyomi.ui.reader.ReaderSeekBar
                 android:id="@+id/page_seekbar"
                 android:layout_width="0dp"
-                android:layout_height="wrap_content"
-                android:maxHeight="?attr/actionBarSize"
-                android:minHeight="?attr/actionBarSize"
```
arkon: Can you double check that this doesn't cause a regression for #2487?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -278,7 +283,12 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
     private fun setMenuVisibility(visible: Boolean, animate: Boolean = true) {
         menuVisible = visible
         if (visible) {
-            systemUi?.show()
+            if (preferences.fullscreen().getOrDefault()) {
+                window.showBar()
+            } else {
+                reader_menu.setPadding(0, 0, 0, 0)
+                window.defaultBar()
+            }
```
arkon: Might be nice to extract this into a separate function.

PR #2603, opened 2020-02-18
Support follow system theme[Android 10]

onlymash:

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -56,12 +56,7 @@ class PreferencesHelper(val context: Context) {
 
     fun themeMode() = prefs.getString(Keys.themeMode, Values.THEME_MODE_LIGHT)
 
-    val themeDark: Int
-        get() = when (prefs.getString(Keys.themeDark, Values.THEME_DARK_DEFAULT)) {
-            Values.THEME_DARK_DEFAULT -> R.style.Theme_Tachiyomi_Dark
-            Values.THEME_DARK_AMOLED -> R.style.Theme_Tachiyomi_Amoled
-            else -> R.style.Theme_Tachiyomi_DarkBlue
-        }
+    fun themeDark() = prefs.getString(Keys.themeDark, Values.THEME_DARK_DEFAULT)
 
     fun rotation() = rxPrefs.getInteger(Keys.rotation, 1)
 
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -15,6 +15,7 @@ import java.text.DateFormat
 import java.text.SimpleDateFormat
 import java.util.Locale
 import eu.kanade.tachiyomi.data.preference.PreferenceKeys as Keys
+import eu.kanade.tachiyomi.data.preference.PreferenceValues as Values
 
 fun <T> Preference<T>.getOrDefault(): T = get() ?: defaultValue()!!
 
@@ -53,12 +54,12 @@ class PreferencesHelper(val context: Context) {
 
     fun clear() = prefs.edit().clear().apply()
 
-    fun themeMode() = prefs.getInt(Keys.themeMode, 1)
+    fun themeMode() = prefs.getString(Keys.themeMode, Values.THEME_MODE_LIGHT)
 
     val themeDark: Int
-        get() = when (prefs.getInt(Keys.themeDark, 1)) {
-            1 -> R.style.Theme_Tachiyomi_Dark
-            2 -> R.style.Theme_Tachiyomi_Amoled
+        get() = when (prefs.getString(Keys.themeDark, Values.THEME_DARK_DEFAULT)) {
+            Values.THEME_DARK_DEFAULT -> R.style.Theme_Tachiyomi_Dark
+            Values.THEME_DARK_AMOLED -> R.style.Theme_Tachiyomi_Amoled
             else -> R.style.Theme_Tachiyomi_DarkBlue
         }
 
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -53,7 +53,14 @@ class PreferencesHelper(val context: Context) {
 
     fun clear() = prefs.edit().clear().apply()
 
-    fun theme() = prefs.getInt(Keys.theme, 1)
+    fun themeMode() = prefs.getInt(Keys.themeMode, 1)
+
+    val themeDark: Int
+        get() = when (prefs.getInt(Keys.themeDark, 1)) {
+            1 -> R.style.Theme_Tachiyomi_Dark
+            2 -> R.style.Theme_Tachiyomi_Amoled
+            else -> R.style.Theme_Tachiyomi_DarkBlue
+        }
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
```diff
@@ -62,6 +64,24 @@ class SettingsGeneralController : SettingsController() {
                 true
             }
         }
+        intListPreference {
```
arkon: Would be nice if this only appears when relevant.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
```diff
@@ -49,11 +49,13 @@ class SettingsGeneralController : SettingsController() {
...
-            titleRes = R.string.pref_theme
-            entriesRes = arrayOf(R.string.light_theme, R.string.dark_theme,
-                    R.string.amoled_theme, R.string.darkblue_theme)
-            entryValues = arrayOf("1", "2", "3", "4")
+            key = Keys.themeMode
+            titleRes = R.string.pref_theme_mode
+            entriesRes = arrayOf(
+                    R.string.theme_light,
+                    R.string.theme_dark,
+                    R.string.theme_system)
```
arkon: Might be worth hiding this for older Android versions where it isn't relevant?

onlymash: Some custom ROMs have supported the system dark theme before Android 10

onlymash: E.g. LineageOS 15 / 16

arkon: Yeah for sure. I use it in Pie, but it's definitely not relevant for versions before Oreo.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
```diff
@@ -49,11 +50,20 @@ class SettingsGeneralController : SettingsController() {
...
-                    R.string.amoled_theme, R.string.darkblue_theme)
-            entryValues = arrayOf("1", "2", "3", "4")
+            key = Keys.themeMode
+            titleRes = R.string.pref_theme_mode
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                entriesRes = arrayOf(
+                        R.string.theme_light,
+                        R.string.theme_dark,
+                        R.string.theme_system)
+                entryValues = arrayOf("1", "2", "3")
```
arkon: Since we're changing the keys anyway, can we have better value names? I'm not a fan of using ints since they're not very meaningful.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
```diff
@@ -48,20 +50,55 @@ class SettingsGeneralController : SettingsController() {
...
+            }
+            defaultValue = Values.THEME_MODE_LIGHT
             summary = "%s"
 
             onChange {
                 activity?.recreate()
                 true
             }
         }
+        listPreference {
```
arkon: Can you hide this preference if the above one isn't set to dark or following the system theme?

Example: https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadController.kt#L92-L93

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
```diff
@@ -48,20 +50,55 @@ class SettingsGeneralController : SettingsController() {
...
+                    R.string.theme_dark_blue)
+            entryValues = arrayOf(
+                    Values.THEME_DARK_DEFAULT,
+                    Values.THEME_DARK_AMOLED,
+                    Values.THEME_DARK_BLUE)
+            defaultValue = Values.THEME_DARK_DEFAULT
+            summary = "%s"
+
+            onChange {
+                if (preferences.themeMode() != Values.THEME_MODE_LIGHT) {
```
arkon: Why specifically the light theme?

app/src/main/res/values/strings.xml
```diff
@@ -118,11 +118,14 @@
...
-    <string name="dark_theme">Dark</string>
-    <string name="amoled_theme">AMOLED dark</string>
-    <string name="darkblue_theme">Dark blue</string>
+    <string name="pref_theme_mode">App theme</string>
+    <string name="theme_light">Light</string>
+    <string name="theme_dark">Dark</string>
+    <string name="theme_system">Follow system</string>
+    <string name="pref_theme_dark">Dark theme</string>
+    <string name="theme_dark_default">Default</string>
+    <string name="theme_dark_amoled">AMOLED</string>
```
arkon: "AMOLED black", maybe?

app/src/main/java/eu/kanade/tachiyomi/ui/base/activity/BaseActivity.kt
```diff
@@ -17,11 +21,17 @@ abstract class BaseActivity : AppCompatActivity() {
...
     override fun onCreate(savedInstanceState: Bundle?) {
-        setTheme(when (preferences.theme()) {
-            2 -> R.style.Theme_Tachiyomi_Dark
-            3 -> R.style.Theme_Tachiyomi_Amoled
-            4 -> R.style.Theme_Tachiyomi_DarkBlue
-            else -> R.style.Theme_Tachiyomi
+        setTheme(when (preferences.themeMode()) {
+            Values.THEME_MODE_LIGHT -> R.style.Theme_Tachiyomi
+            Values.THEME_MODE_DARK -> preferences.themeDark
+            else -> {
```
arkon: I'd prefer for this case to be explicit and the fallback to remain as the default light theme like before.

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
```diff
@@ -53,7 +54,14 @@ class PreferencesHelper(val context: Context) {
...
 
-    fun theme() = prefs.getInt(Keys.theme, 1)
+    fun themeMode() = prefs.getString(Keys.themeMode, Values.THEME_MODE_LIGHT)
+
+    val themeDark: Int
+        get() = when (prefs.getString(Keys.themeDark, Values.THEME_DARK_DEFAULT)) {
+            Values.THEME_DARK_DEFAULT -> R.style.Theme_Tachiyomi_Dark
+            Values.THEME_DARK_AMOLED -> R.style.Theme_Tachiyomi_Amoled
+            else -> R.style.Theme_Tachiyomi_DarkBlue
+        }
```
arkon: Might make more sense to put the `when` logic in the BaseActivity along with the other theme-picking logic.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
onlymash: Light Theme does not need to recreate activity for dark theme changes

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
onlymash: This is not only associated with the follow system

arkon: What do you mean? It still respects what's chosen here when it switches to dark mode.

PR #2608, opened 2020-02-19
Change default theme to System Default

Soitora:
_Hopefully this was the only string that dictated the default setting._

This is the recommended way of handling application theme defaults according to the Android Developers page about Dark themes.

> When running on Android 10 (API level 29) and higher, the recommended options are different, to allow the user to override the system default:
> 
> **Light**
> **Dark**
> **System default** (the recommended default option)
- https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#changing_themes_in-app

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -54,7 +54,7 @@ class PreferencesHelper(val context: Context) {
 
     fun clear() = prefs.edit().clear().apply()
 
-    fun themeMode() = rxPrefs.getString(Keys.themeMode, Values.THEME_MODE_LIGHT)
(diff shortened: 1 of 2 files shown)


arkon:
You need to update it in the preferences helpers too. Also need to double check that it'll fall back to the light theme in cases where the system theme isn't actually supported.

Soitora:
Are you referring to this?
```kt
fun themeMode() = rxPrefs.getString(Keys.themeMode, Values.THEME_MODE_LIGHT)
```
Also, looking at the PR for the system follow support, it looks like it will based on the code (if I'm not looking at the wrong thing)
https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/base/activity/BaseActivity.kt#L43

Other than that, I can't actually check it since I don't know how to compile an app in Android Studio, and my only device accessible is Android 10.

arkon:
> Are you referring to this?

Yes.

Issue #2180, opened 2019-08-14
[Feature Request] Option to divide pages in Webtoon mode


TheUnlocked:
I like to be able to scroll part-way through a page when reading (like mentioned in #778, which was rejected for technical reasons). Webtoon mode supports that, but when you use Webtoon mode, every page is directly against one another rather than separated. This of course makes sense for Webtoons which are supposed to exist on an infinite canvas, but it works less well for page-based manga. To make the Webtoon viewing mode work better for page-based manga, I propose an option to add a small amount of vertical padding between each page in Webtoon mode.

Here's a comparison image between unnamed-other-manga-reading-app (left) and current Tachiyomi (right) to explain what I mean (don't mind the difference in scanlation):

![image](https://user-images.githubusercontent.com/10186337/63063645-4dc93980-beb2-11e9-8318-e6d4e82bae35.png)

hXtreme:
Is there any particular reason for why any other view doesn't work for you when reading page based manga?

TheUnlocked:
As I said, I like to be able to vertically scroll part-way through a page as I'm reading (makes me need to move my eyes around less), and vertical mode only supports viewing a full page at a time. This isn't the kind of thing which would be a dealbreaker for me, but it would be a nice feature to have.

metroid112:
Webtoon mode should be left as is, since that works perfect for webtoons which need no padding between vertical "pages".

I think the solution would either be adding a new type of viewer "webtoon with padding" or make the padding an option by itself.

PR #2618, opened 2020-02-23
Added Webtoon with Padding viewer

TheUnlocked:
Adds another viewer type for webtoons with padding between each page, following discussion in #2180.

image

### DIFF
--- a/app/src/main/res/layout/reader_settings_sheet.xml
+++ b/app/src/main/res/layout/reader_settings_sheet.xml
@@ -250,11 +250,11 @@
         app:layout_constraintTop_toBottomOf="@id/webtoon_prefs" />
 
     <androidx.appcompat.widget.SwitchCompat
-        android:id="@+id/pad_pages_webtoon"
+        android:id="@+id/pad_pages_vert_webtoon"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_marginTop="16dp"
-        android:text="@string/pref_webtoon_padding"
+        android:text="@string/pref_webtoon_padding_vert"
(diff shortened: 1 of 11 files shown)


joseph619:
How is this any different from Vertical?

TheUnlocked:
Vertical scrolls whole pages at a time.

arkon:
I'd prefer for it to be an option for the Webtoon pager rather than an entirely separate pager.

It could work alongside the margins from #2349, I suppose.

TheUnlocked:
image
image

app/src/main/res/layout/reader_settings_sheet.xml
```diff
@@ -249,6 +249,15 @@
         android:textColor="?android:attr/textColorSecondary"
         app:layout_constraintTop_toBottomOf="@id/webtoon_prefs" />
 
+    <androidx.appcompat.widget.SwitchCompat
+        android:id="@+id/pad_pages_webtoon"
```
arkon: I believe you need to add this to the `webtoon_prefs_group` group too, otherwise it'll show up when it's a different pager too.

app/src/main/res/values/strings.xml
```diff
@@ -207,6 +207,7 @@
     <string name="pref_show_page_number">Show page number</string>
     <string name="pref_true_color">32-bit color</string>
     <string name="pref_crop_borders">Crop borders</string>
+    <string name="pref_webtoon_padding">Add padding between pages</string>
```
arkon: "Vertical padding between pages"?

To avoid confusion when horizontal padding is added too.

PR #2633, opened 2020-02-26
add ktlint

CarlosEsco:
Set it up to required lint before a build, and lint requires a format.
It formats files then lints.  Only thing format can't fix is imports so you manually have to fix.  Any lint errors fails the build.  So nobody can commit bad code.

I also added an intellij code style based off their recommendations.

Note  if/when you merge this you should run a make on your computer and fix the imports as it will fail the build.  Then merge in.  It basically is gonna format 90% of files in the repo.


CarlosEsco:
@Jays2Kings  after this merges in you will probably want to pull into md2 and run yourself.

PR #2642, opened 2020-02-29
Fix #1770

### LINKED ISSUE #1770: Can not delete chapters of local manga
**App version:** 0.8.0
Android 8.1

Deleting local chapters doesn't work from within the app. You have to go to the local folder in file manager and delete from there.

udf:
Adds a special case for `LocalSource` inside `findSourceDir` so it uses the correct path.

See [my comment on the issue](https://github.com/inorichi/tachiyomi/issues/1770#issuecomment-592994791) for more information.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
@@ -8,6 +8,7 @@ import eu.kanade.tachiyomi.data.database.models.Chapter
 import eu.kanade.tachiyomi.data.database.models.Manga
 import eu.kanade.tachiyomi.data.preference.PreferencesHelper
 import eu.kanade.tachiyomi.data.preference.getOrDefault
+import eu.kanade.tachiyomi.source.LocalSource
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.util.storage.DiskUtil
 import uy.kohesive.injekt.injectLazy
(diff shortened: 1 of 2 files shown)


arkon:
I'm going to hold off on merging this until I add a snackbar with the ability to undo a deletion.

joseph619:
Doesn't J2k have it already? You could pull it from that.

arkon:
> Doesn't J2k have it already? You could pull it from that.

It doesn't.

udf:
Hold up, my commit has a pretty big bug. It ends up deleting the entire local manga directory after deleting the selected chapters.

This is because we [check if the directory is empty](https://github.com/inorichi/tachiyomi/blob/b104bec49d5046a47f745f1aab1a7e3a22c2f7a6/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt#L172) by calling [`DownloadCache.getDownloadCount`](https://github.com/inorichi/tachiyomi/blob/b104bec49d5046a47f745f1aab1a7e3a22c2f7a6/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt#L96), which returns 0 even if it can't find the source directory (which it can't because it looks in the downloads directory). So, the code thinks the directory is empty and removes it.

I'm not sure how to go about correcting this, the easy fix would be to make `getDownloadCount` return -1 if it is unable to count the files in the directory. But I'm not sure if that would break things elsewhere.

Perhaps a better fix would be to not ask the `DownloadCache` for the count, and to actually count the number of files directly in `deleteChapters`?

arkon:
Closing since we never got around to addressing the fatal flaw described above. I'll keep the issue open though.

PR #2672, opened 2020-03-06
Add Start/Finish date support for MAL

### LINKED ISSUE #315: MAL start date
When binding manga, if it has 0 chapters read and 0 volumes read (or maybe if it wasn't in any list) set start date to today.

### LINKED ISSUE #1358: [Feature Request] Add option to edit MyAnimeList start and finish date from the Tracking view
**App version:** 0.7.2

**Issue/Request:** It would be nice if you could edit your personal start and finish date of a manga directly from Tachiyomi, just like you can edit your status and score.

**Other details:** Screenshot of the MyAnimeList edit panel: https://puu.sh/zZLTm/a68c15a79c.png

### LINKED ISSUE #2670: [Feature Request] Add Start and End date on MAL support
Right now in order to set the start/end date of any anime/manga you need to open MAL on a browser and update it manually. It would be awesome to be able to edit the advanced details in the app. 

### What/Requirements
This can be achieved by either adding the 2 fields in the app details or by adding a "advanced settings" button. (I'd prefer the former)

ITALIA195:
Closes #2670 
Closes #315 
Closes #1358 

PS: First time working with Kotlin and Android applications so I won't guarantee anything.


ITALIA195:
This is blocked by unknown MAL API [here](https://github.com/ITALIA195/tachiyomi/blob/5d673ebbbcc73e358b218ed60ea55e94612ca55a/app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListApi.kt#L314-L322)

I need to know if there is a start/finish date api and if so what it is. Tried to guess it but it doesn't seem to be neither `start_date_day` (`start_date_month`, `start_date_year`) nor `start_date` with "yyyy-MM-dd".

arkon:
Did you try observing the network call even doing it from your desktop browser?

bbtora:
I see AniList file being changed too, doesn't their API do this automatically?

ITALIA195:
I tried replicating what tachi does on a [Python script](https://gist.github.com/ITALIA195/07088a1de07052c90c7a96066a37bdac) but I can't seem to update it. (I'm getting HTTP 400 on the update even though the login works (the page content has my data on it))

Regarding AniList being changed I've added a property that has to be true for the Start date and Finish date fields to appear on the tracking tab. (This approach may change later)

ITALIA195:
That said replicating what tachi does only makes testing easier. I still don't know how to update start and finish date through the `edit.json` api. The export uses the "yyyy-MM-dd" format and the edit page uses a selection box for the year, month and day.

I'll try using `start_date` with epoch as the parameter and I guess I'll try again the other approches just to be sure.

ITALIA195:
Okay small update:
I did not realize that the `edit.json` api was actually used by the browser (woopsie). The edit.json is called when we edit the manga from the manga page. The problem is that normally you can't change your `start_date` from that page. You need to open the edit page (https://myanimelist.net/ownlist/manga/ID/edit) and there you can update the date.

Problem is that that page requires way more information than the edit.json one. It seems that it does not work if you don't supply one of those parameters:
- `add_manga[sns_post_type]`
- `add_manga[is_asked_to_discuss]`
- `add_manga[status]`

And even bigger problem is that it changes every option that's not supplied to the default value. Most of the things that are asked for are not present in the `Track` class so there are only 2 approches I could think of:
1. Download the edit page (GET), get the informations, edit them, update the manga (POST to the same page).
2. Try using `edit.json` but without knowledge of the api (most likely you can't update date through there but we can try).
3. Wait for MAL to update their API and use that.

app/src/main/java/eu/kanade/tachiyomi/data/database/DbOpenCallback.kt
```diff
@@ -20,7 +20,7 @@ class DbOpenCallback : SupportSQLiteOpenHelper.Callback(DATABASE_VERSION) {
         /**
          * Version of the database.
          */
-        const val DATABASE_VERSION = 8
+        const val DATABASE_VERSION = 9
```
ITALIA195: I believe it's neeeded to add the start and finish reading dates to the database. I'm not sure what's the best way to add those fields without having to reset every tachi user's database.
For now I just bumped this number by one, please tell me if this is a problem.

ITALIA195:
I don't think that this is compatible with previous tachi database versions (had to clear the app storage to make it work) and I'd need to advice on how the database is structured so that I can add those fields without users' data loss.
Other than that this is ready for review

ITALIA195:
Ready for review.

ITALIA195:
Can someone look at this? It might break the more updates the repo undergoes

@arkon @

app/src/main/java/eu/kanade/tachiyomi/data/database/mappers/TrackTypeMapping.kt
```diff
@@ -54,6 +60,14 @@ class TrackPutResolver : DefaultPutResolver<Track>() {
         put(COL_STATUS, obj.status)
         put(COL_TRACKING_URL, obj.tracking_url)
         put(COL_SCORE, obj.score)
+        put(COL_START_DATE, obj.started_reading_date?.toDateString() ?: "")
+        put(COL_FINISH_DATE, obj.finished_reading_date?.toDateString() ?: "")
```
arkon, line 64: I'd prefer these to follow the same convention as other dates in the DB (i.e. epoch longs).

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
```diff
@@ -24,12 +25,18 @@ interface Track : Serializable {
 
     var status: Int
 
+    var started_reading_date: Calendar?
+
+    var finished_reading_date: Calendar?
```
arkon, line 30: Likewise, these should just be `Long`s.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/TrackImpl.kt
```diff
@@ -22,6 +24,10 @@ class TrackImpl : Track {
 
     override var status: Int = 0
 
+    override var started_reading_date: Calendar? = null
+
+    override var finished_reading_date: Calendar? = null
```
arkon, line 29: `Long`s here too.

app/src/main/java/eu/kanade/tachiyomi/data/track/TrackService.kt
```diff
@@ -22,6 +22,8 @@ abstract class TrackService(val id: Int) {
     // Name of the manga sync service to display
     abstract val name: String
 
+    abstract val supports_reading_dates: Boolean
```
arkon, line 25: Could just default it to `false`, i.e. `open val supportsReadingDates: Boolean = false`

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListApi.kt
```diff
@@ -76,14 +80,72 @@ class MyAnimeListApi(private val client: OkHttpClient, interceptor: MyAnimeListI
 
     fun updateLibManga(track: Track): Observable<Track> {
         return Observable.defer {
-            authClient.newCall(POST(url = updateUrl(), body = mangaPostPayload(track)))
+            authClient.newCall(GET(url = editPageUrl(track.media_id)))
```
arkon, line 83: There's a lot going on here. Can you refactor it (split it into smaller functions, add comments, etc.) to explain what's going on?

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListApi.kt
```diff
@@ -269,6 +331,66 @@ class MyAnimeListApi(private val client: OkHttpClient, interceptor: MyAnimeListI
...
+
+        private fun Element.getEditFinishDate(): Calendar? {
+            val month = select("#add_manga_finish_date_month > option[selected]").`val`().toIntOrNull()
+            val day = select("#add_manga_finish_date_day > option[selected]").`val`().toIntOrNull()
+            val year = select("#add_manga_finish_date_year > option[selected]").`val`().toIntOrNull()
+            if (year == null || month == null || day == null)
+                return null
+
+            return GregorianCalendar(year, month - 1, day)
+        }
```
arkon, line 380: This is a bit repetitive.

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListApi.kt
```diff
@@ -269,6 +331,66 @@ class MyAnimeListApi(private val client: OkHttpClient, interceptor: MyAnimeListI
...
+
+        private fun Element.searchDateXml(field: String): Calendar? {
+            val text = selectText(field, "0000-00-00")!!
+            if (text == "0000-00-00")
+                return null
```
arkon, line 385: Can you add a comment here? Not sure why this is a case.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/SetTrackReadingDatesDialog.kt
```diff
@@ -0,0 +1,119 @@
...
+import java.util.GregorianCalendar
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+class SetTrackReadingDatesDialog<T> : DialogController
+        where T : Controller, T : SetTrackReadingDatesDialog.Listener {
+
+    private val item: TrackItem
+
+    private val setting: Int
```
arkon, line 24: Why an integer?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/SetTrackReadingDatesDialog.kt
```diff
@@ -0,0 +1,119 @@
...
+                    else
+                        listener.setFinishDate(item, null)
+                    dismissDialog()
+                }
+                .onPositive { dialog, _ ->
+                    val view = dialog.customView
+                    if (view != null) {
+                        val dayPicker: NumberPicker = view.findViewById(R.id.day_picker)
+                        val monthPicker: NumberPicker = view.findViewById(R.id.month_picker)
+                        val yearPicker: NumberPicker = view.findViewById(R.id.year_picker)
```
arkon, line 67: Use the synthetic view imports.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/SetTrackReadingDatesDialog.kt
```diff
@@ -0,0 +1,119 @@
...
+                }
+                .show()
+
+        val view = dialog.customView
+        if (view != null) {
+            val today = GregorianCalendar.getInstance()
+
+            val dayPicker: NumberPicker = view.findViewById(R.id.day_picker)
+            val monthPicker: NumberPicker = view.findViewById(R.id.month_picker)
+            val yearPicker: NumberPicker = view.findViewById(R.id.year_picker)
```
arkon, line 94: Use the synthetic view imports.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackHolder.kt
```diff
@@ -39,6 +50,15 @@ class TrackHolder(view: View, adapter: TrackAdapter) : BaseViewHolder(view) {
                     if (track.total_chapters > 0) track.total_chapters else "-"
             track_status.text = item.service.getStatus(track.status)
             track_score.text = if (track.score == 0f) "-" else item.service.displayScore(track)
+
+            divider_2.visibleIf { item.service.supports_reading_dates }
+            date_support_container.visibleIf { item.service.supports_reading_dates }
+            if (item.service.supports_reading_dates) {
+                val dateFormat = DateFormat.getDateInstance(DateFormat.SHORT)
```
arkon, line 57: Should use the date format set by the user.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/SetTrackReadingDatesDialog.kt
ITALIA195: I can't get synthetic view imports to work. Looking at other track-related dialogs it doesn't seem like they are used either.
https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/SetTrackScoreDialog.kt#L56

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListApi.kt
ITALIA195: I did but I'm not familiar with Observable and I'm unsure whether this will work as intended:
https://github.com/ITALIA195/tachiyomi/blob/540435622222eb5bebb2e4afa3b47e4d74236b2c/app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListApi.kt#L111-L132
(the update works but not sure if what I do with requests and observable is right)

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/SetTrackReadingDatesDialog.kt
arkon: That's fine then.

arkon:
There seems to be an issue with removing the dates.

1. Set start date (and/or end date)
2. Remove it
3. Refresh

Note that the date reappears.

ITALIA195:
Fixed in 89abff2842a9a8ab6b2de6dc7b4c0485ea132ece

ITALIA195:
MAL supports partial dates (i.e. day null, month null and year 2020) but to support that in the application we would need to save each individual field in the database and this may cause problems with implementing other services' date functionality.
It's unlikely that someone uses partial dates so I don't think this will cause problems.

ITALIA195:
Ready to merge.

With the update of MaterialDialogs in 932a47a8a703d140c5676d2a0e25b9afaefa5703 I can now replace the datepicker with the [MaterialDialog's one](https://github.com/afollestad/material-dialogs/blob/master/documentation/DATETIME.md#Date) but will probably open another PR for that.

PR #2677, opened 2020-03-08
Fix Cloudflare Interceptor when User-Agent is Empty

TFSThiagoBR98:
Avoid User-Agent mismatch between WebView and OkHttp when User-Agent is Empty, see https://developer.android.com/reference/android/webkit/WebSettings#setUserAgentString(java.lang.String)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
@@ -55,7 +55,19 @@ class CloudflareInterceptor(private val context: Context) : Interceptor {
             val oldCookie = networkHelper.cookieManager.get(originalRequest.url)
                     .firstOrNull { it.name == "cf_clearance" }
             resolveWithWebView(originalRequest, oldCookie)
-            return chain.proceed(originalRequest)
+
+            // Avoid use empty User-Agent
+            return if (originalRequest.header("User-Agent").isNullOrEmpty()) {
+                val newRequest = originalRequest
+                        .newBuilder()
+                        .removeHeader("User-Agent")
+                        .addHeader("User-Agent",
+                                DEFAULT_USERAGENT)
+                        .build()
+                chain.proceed(newRequest)
+            } else {
+                chain.proceed(originalRequest)
+            }
         } catch (e: Exception) {
             // Because OkHttp's enqueue only handles IOExceptions, wrap the exception so that
             // we don't crash the entire app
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
```diff
@@ -55,7 +55,19 @@ class CloudflareInterceptor(private val context: Context) : Interceptor {
...
             resolveWithWebView(originalRequest, oldCookie)
-            return chain.proceed(originalRequest)
+
+            // Avoid use empty User-Agent
+            return if (originalRequest.header("User-Agent").isNullOrEmpty()) {
+                val newRequest = originalRequest
+                        .newBuilder()
+                        .removeHeader("User-Agent")
+                        .addHeader("User-Agent",
+                                "Mozilla/5.0 (Windows NT 6.3; WOW64)")
```
arkon: Any reason for this particular user agent string?

TFSThiagoBR98: None, I used the one I found in the file https://github.com/inorichi/tachiyomi/blob/4358f9fd2a2d40d2c8dcb321411b1aeb8b407e52/app/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt#L77
I haven't tested with other values, but I think they can also work

arkon: Ah right, makes sense. Can you make that a constant and just reference it?

TFSThiagoBR98: Ok, done

app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt
```diff
@@ -77,20 +89,23 @@ class CloudflareInterceptor(private val context: Context) : Interceptor {
...
             webview.settings.userAgentString = request.header("User-Agent")
+                    ?: DEFAULT_USERAGENT
 
             webview.webViewClient = object : WebViewClientCompat() {
                 override fun onPageFinished(view: WebView, url: String) {
                     fun isCloudFlareBypassed(): Boolean {
                         return networkHelper.cookieManager.get(origRequestUrl.toHttpUrl())
                                 .firstOrNull { it.name == "cf_clearance" }
-                                .let { it != null && it != oldCookie }
+                                .let { it != null && (it != oldCookie || withUserAgent) }
```
arkon: I ended up removing this part since I can't see why falling back to the default user agent should impact the check here.

Let me know if I misunderstood what this was for.

TFSThiagoBR98: This avoid Old Cookie verification, because the 503 block is caused by the absence of User-Agent String, the Cookie still valid, but not for that User-Agent.

arkon: Wouldn't that only be the case for sessions that were created prior to this change? There should always be a UA string now, unless you mean the UA string is changing.

TFSThiagoBR98: > There should always be a UA string now

Only for that request, a new one can come without the UA String

It is also possible to separate the code that adds the UA String in a new Interceptor that runs before the CloudflareInterceptor, so it will run always with an UA String and avoid to bypass Cloudflare in every request.

https://github.com/inorichi/tachiyomi/blob/07e2bdac81f109887cd14c662249747dbbf8599a/app/src/main/java/eu/kanade/tachiyomi/network/CloudflareInterceptor.kt#L61-L70

https://github.com/inorichi/tachiyomi/blob/07e2bdac81f109887cd14c662249747dbbf8599a/app/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt#L21-L23

arkon: That sounds reasonable to me, assuming it works fine. Could you open another PR? :)

TFSThiagoBR98: Ok, done #2685

Issue #1653, opened 2018-10-10
Remove metadata section from in between chapters.


LalitSinghRana:
With update 8.0, when one goes to next chapter there comes a section which tells about previous chapter and next chapter.

It isn't really needed there and disturb the reading experience.

inorichi:
It might seem not needed, but it has more uses besides showing you chapter titles. It shows a loading progress if the next chapter isn't loaded yet or a retry button if it fails.

Previously if the next chapter couldn't load you'd reach the end of the chapter and didn't have any feedback of what was going on. Furthermore, it's also used to know when to load the previous chapter.

Sparh4wk:
so can you please make it optional?

inorichi:
No :disappointed:

Tempest042:
Love your work but I would greatly appreciate if this feature was made optional.

somedood0:
I would like to add that it is extremely annoying when reading single-page manga or 4koma. It breaks the flow from page to page.

LinkSlayer64:
I would also like this made optional. I had no problems with the previous setup.

k0rak0ra:
I'd also vote for removing the page (since it cant be made optional). Previous setup was fine, this feels like a really annoying and unneeded option.

LinkSlayer64:
I was just thinking for a minute and an idea came to mind - would it be possible to disable/run a faster check/something the in-between page for series where the the next chapter is registered as "downloaded" - from my own issues, it breaks the flow with my downloaded chapters, whereas it's seems to be a feature more geared towards reading on the fly, because you want to make sure the next chapter begins to download.

ooluyid2:
I back the option to disable it.

Acoha:
I also want to this page to be optional, atleast for downloaded manga. I understand why it’s there: to show download progress for the next chapter but it is completely unnecessary for locally stored manga. It is especially annoying on low pages/daily manga. 

I hope you change your mind on this matter.

Th3N3rdyGam3r:
I also agree that it should become optional for both downloaded manga and non-dowbloaded manga since it is rather annoying when reading 1-2 page manga or downloaded manga.

rathus:
Came to post this as well, mostly reading 4-koma and other 1-page comics, and this screen is utterly annoying.  1 Page > Black Screen > 1 Page > Black Screen > 1 Page > Black Screen.   

[...]

PR #2682, opened 2020-03-09
Added config to hide transition page when not needed

morcefaster:
As per issue https://github.com/inorichi/tachiyomi/issues/1653

The transition page will not show if the "force" setting is off and the next/previous chapter has loaded. At a normal reading pace, this should be consistently the case due to last page kicking off preload.

Transition/loading page will show regardless of the setting if the page hasn't loaded yet or failed to load.

### DIFF
--- a/app/src/main/res/layout/reader_settings_sheet.xml
+++ b/app/src/main/res/layout/reader_settings_sheet.xml
@@ -152,11 +152,20 @@
         android:textColor="?android:attr/textColorSecondary"
         app:layout_constraintTop_toBottomOf="@id/keepscreen" />
 
+    <androidx.appcompat.widget.SwitchCompat
+        android:id="@+id/always_show_chapter_transition"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="16dp"
+        android:text="@string/pref_always_show_chapter_transition"
+        android:textColor="?android:attr/textColorSecondary"
+        app:layout_constraintTop_toBottomOf="@id/long_tap" />
+
     <androidx.legacy.widget.Space
         android:id="@+id/end_general_preferences"
         android:layout_width="0dp"
         android:layout_height="0dp"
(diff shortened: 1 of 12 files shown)


morcefaster:
@CarlosEsco The setting is there (Settings->Reader) via SettingsReaderController.kt

app/src/main/res/layout/reader_settings_sheet.xml
```diff
@@ -258,22 +268,31 @@
...
+        app:constraint_referenced_ids="pager_prefs,scale_type_text,scale_type,zoom_start_text,zoom_start,crop_borders,page_transitions,always_show_chapter_transition"
         tools:visibility="visible" />
 
     <androidx.constraintlayout.widget.Group
         android:id="@+id/webtoon_prefs_group"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:visibility="gone"
-        app:constraint_referenced_ids="webtoon_prefs,crop_borders_webtoon,pad_pages_vert_webtoon" />
+        app:constraint_referenced_ids="webtoon_prefs,crop_borders_webtoon,pad_pages_vert_webtoon,always_show_chapter_transition_webtoon" />
```
arkon: I'm not sure it makes sense to duplicate the option in the 2 groups.

morcefaster: I figured the setting belongs in Pager category, so I had to duplicate it so it appears in both Webtoon and Pager modes, akin to other settings like "Crop borders" which are also duplicated.

On that note, should this setting be global or per-reader? (i.e. no transition page for default pager, but always show for webtoon scroller)

CarlosEsco: yea it should be global not per reader

arkon: I'm not really sure if crop borders should be duplicated either tbh

morcefaster: I'll move it to the General group then and get rid of duplication.

Issue #476, opened 2016-09-22
Configurable zoom on double-tap


jotumn:
Right now, the app zooms the page to a fixed size on double-tap. This probably works fine on tablets but on smartphones the zoom is too large so it's not very useful. I'd like it if you could configure the zoom percentage in the settings.

TheMostCuriousThing:
I agree, but I think the easiest solution is simpler than a configurable percentage: a toggle between two different `Scale type`s (say, `Primary scale type` and `Secondary scale type`).

An example use case for smartphones in portrait would be:
- `Primary scale type`: `Fit screen`
- `Secondary scale type`: `Fit height`

...which would allow for simple, "zoom-free" reading on smartphones with a 16:9 or greater aspect ratio.

edenpc:
I second this.

Matth7878:
Same here.
I think issue #228 is identical to this one.

Battler624:
I also wish this is implemented.

arkon:
Duplicate of #228

PR #2707, opened 2020-03-15
Added support for tap-centered zoom on double tap in Pager viewer

morcefaster:
Zoom style for double tap on Pager viewer has been hardcoded to ZOOM_FOCUS_FIXED, working around image center.

A more intuitive type of zoom is ZOOM_FOCUS_CENTER, which zooms around double tap location.

This PR adds a setting to configure centering of the double-tap zoom. Can also serve as baseline for other zoom double-tap zoom configurations (i.e. #476)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
@@ -135,6 +135,8 @@ object PreferenceKeys {
 
     const val alwaysShowChapterTransition = "always_show_chapter_transition"
 
+    const val doubleTapZoomStyle = "double_tap_zoom_style"
+
     fun trackUsername(syncId: Int) = "pref_mangasync_username_$syncId"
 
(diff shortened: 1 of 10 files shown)


arkon:
I wonder if it's worth just doing `ZOOM_FOCUS_CENTER` always, since that _is_ more intuitive and is how the webtoon reader behaves.

@CarlosEsco @Jays2Kings Any thoughts?

arkon:
@CarlosEsco and @Jays2Kings both agree that `ZOOM_FOCUS_CENTER` should be the default or only option. I'm in favour of only doing that, rather than adding the complexity of having it configurable.

Soitora:
I agree that `ZOOM_FOCUS_CENTER` seems more intuitive and should be left as the default, I don't see any reasoning behind keeping the `ZOOM_FOCUS_FIXED` even as an option.

morcefaster:
Created new PR with ZOOM_FOCUS_CENTER as the only and default option.
(https://github.com/inorichi/tachiyomi/pull/2713)

Issue #2715, opened 2020-03-16
Downloaded manga bug.


Battler624:
I've been having this weird bug, I have about 500 chapters of a manga downloaded but every time I go to a new chapter it takes a long time to load (even tho it's already downloaded) but if I quit the viewer (not the whole app) it immediately loads.

This bug happens frequently but not always, I've been trying to replicate it but so far I've read 5 chapters while writing this without it happening again. So weird.

saintbepis:
Happened to me too, it's not big but it's a bug

morcefaster:
I've had it on manga with as few as 20 chapters too. I'm planning to try to debug it sometime soon since it happens relatively frequently.

I'm using android 8, @Battler624 @saintbepis could you tell me what android version you're on?

saintbepis:
@morcefaster I'm on android 10 using the latest dev build of the app

Battler624:
@morcefaster Android 10, note 9 device exynos.

Battler624:
Happened again, about 15 chapters of reading.

http://imgur.com/gallery/eSpyA1A

morcefaster:
Yeah, it's same bug we're talking about. I'm now trying to reproduce it in an emulator.

PR #2718, opened 2020-03-16
Fix for perpetually loading chapters

morcefaster:
Related issue: https://github.com/inorichi/tachiyomi/issues/2715

From time to time, a downloaded chapter's pages don't show up, instead
showing an infinite progress spinner with no error message or timeout. 

After reproducing said issue, the reason behind it is that `chapter.pageLoader` is `null` while the `state` is `Loaded`. PagerPageHolder simply has a stub for that case:
`val loader = page.chapter.pageLoader ?: return`

Normally this shouldn't ever happen since `state` gets set to `Wait` immediately `pageLoader` gets recycled. It might possibly be happening if pages finish loading after chapter gets recycled, for some reason, which would set the `state` to `Loaded` again (https://github.com/inorichi/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/ChapterLoader.kt#L47-L59)

This fix simply extends chapter ready check with loader present.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/ChapterLoader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/ChapterLoader.kt
@@ -26,7 +26,7 @@ class ChapterLoader(
      * completes if the chapter is already loaded.
      */
     fun loadChapter(chapter: ReaderChapter): Completable {
-        if (chapter.state is ReaderChapter.State.Loaded) {
+        if (chapterIsReady(chapter)) {
             return Completable.complete()
         }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/ChapterLoader.kt
```diff
@@ -65,7 +65,7 @@ class ChapterLoader(
      * Checks [chapter] to be loaded based on present pages and loader in addition to state.
      */
     private fun chapterIsReady(chapter: ReaderChapter): Boolean {
-        return chapter.state is ReaderChapter.State.Loaded && chapter.pageLoader != null && chapter.pages?.count() != 0
+        return chapter.state is ReaderChapter.State.Loaded && chapter.pageLoader != null
```
CarlosEsco: why page count check removed?

morcefaster: I put it there at first but upon consideration I found it excessive since unlike the loader, there is no race condition on setting the `Loaded` state+pages. Also I believe the download manager already has an exception for empty page list.

Issue #2728, opened 2020-03-17
[Bug] Chapter Transition sometimes shows up even if "Always show chapter transition" is turned off


X9VoiD:
### Device information
* Tachiyomi version: 0.8.5-1920
* Android version: 9.0

## Steps to reproduce
1. No clear steps to reproduce other than it is happening more frequently with a slow network and manhwas/webtoon. Check "other details" below.

### Expected behavior
"Always show chapter transition" is turned off so there shouldn't be any chapter transitions for a seamless reading experience.

### Actual behavior
But sometimes the chapter transition shows up briefly, then disappears after a while (or maybe disappears after the last page loads, check "other details"), offsetting the scroll of the current manhua/webtoon.

### Other details
Just a guess but the chapter transition shows up mistakenly when the last page of the current chapter is not loaded yet and the user already reached the end of the chapter.

CarlosEsco:
If your connection is slow and it can't load the next chapters page within the last 3 pages of the current chapter. It is going to show.

arkon:
We should probably word it better, but this is expected behaviour.

morcefaster:
The opposite of "Always show page transition" is "Not always show page transition" (i.e. sometimes show). It will still appear when the next chapter is not loaded and that's expected, but there _are_ some issues that can make it appear and disappear when unneeded too, such as #2727.

I'm now investigating that issue and will also see if preloading can be made better.

CarlosEsco:
@morcefaster  it already preloads.  If you have a slow connection and read fast there's not much to do.    Just pre download chapters is best solution

morcefaster:
> @morcefaster it already preloads. If you have a slow connection and read fast there's not much to do. Just pre download chapters is best solution

I know it preloads, hence I said if it can be made "better". I believe there might be some bugs around it, maybe with observers, that cause wrong chapters/pages to preload or load/preload out of order.

I see aforementioned issues on pre-downloaded chapters (I always download them ahead of time).

morcefaster:
Related issue: #2727

Both these issues are caused by a race condition in chapter (pre-)loading. 

While the transition page will still show up if the chapter **really** isn't loaded yet, it most commonly shows up with this bug right now.

morcefaster:
This can be closed. Wrongful appearances fixed with #2731, appearing when page is still loading is by design.

Issue #2727, opened 2020-03-17
[Bug] Chapters skipping


morcefaster:
While #2718 fixed perpetual loading on chapters, it appears that from time to time a chapter gets skipped entirely now (after a very short progress bar).

This might point to what the origin of the race condition was previously. Creating this issue to track the investigation.

EDIT1: The chapter loads, but skips to the last page. Seek bar at the bottom shows last page of previous chapter. It seems that the wrong pages are loaded into the chapter.

morcefaster:
Got'em. I was right about the race condition.

```2020-03-18 00:57:31.830 Loading pages for 8.1
2020-03-18 00:57:31.836 Recycling chapter Vol.3 Chapter 8.1 (8.1)
2020-03-18 00:57:31.853 Loaded chapter 8.1 with 10 pages
```
It seems that for some reason chapter got recycled after pages started loading, which leads to chapter having state=Loaded with no page loader (leading to permanent loading issue, #2715)

The quickfix which made the chapter reload again if the loader is not present loads new Page objects, confusing the hell out of @CarlosEsco's DirectionalViewPager.dataSetChanged() which issues a repopulation and sets current item to what is usually the last page, leading to "skipping" the chapter.

Now going to investigate the reason behind the weird mid-task ref drop.

morcefaster:
Fixed with #2731

PR #2731, opened 2020-03-18
Faulty preload fix

morcefaster:
Fixes issues #2715, #2727, #2728 

### The issue and its reproduction

1. Disable "Always show chapter transition" in settings
2. Download a regular manga, with >3 chapters and >3 pages per chapter.
3. Open a middle chapter in manga, let's call it chapter 5. Casually scroll to the next chapter.
	Current chapter is now set to 6, prev is 5, next is 7.
4. Scroll one page back onto the last page.
	What happens here is the following:
	- We're at last page, so next chapter's (7) preload is requested. That chapter hasn't been loaded yet, so the preload begins.
	- We're at different chapter than previously, so chapters get shifted. Current is now 5, previous 4, next 6.
	- At the same time, since chapter 7 is not referenced anymore, it gets recycled.
	- Preload of chapter 7 is finished, and its status gets set to Loaded.

5. Scroll forward to chapter 7.
	- If fix #2717 is included, you will see a loading page that gets turned into pages of chapter 8 (since chapter 7 got reloaded with new page objects)
	- If fix #2717 is not included, you will get a never-ending loading animation.

We fix that by making sure preload is *really* needed on page change.

---
Also:

- Webtoon viewer used `adapter.items[pages.length+1]` to find next chapter for preload. Since we load additional 2 pages from previous chapter + the chapter transition into items, this would mostly return a ReaderPage object and made preload not work correctly. Also didn't work with disabled ChapterTransitions.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
@@ -140,38 +140,26 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
 
     private fun checkAllowPreload(page: ReaderPage?): Boolean {
         // Page is transition page - preload allowed
-        if (page == null)
-            return true
+        page == null ?: return true
 
         // Initial opening - preload allowed
-        if (currentPage == null) {
-            return true
+        currentPage == null ?: return true
+
+        // Allow preload for
+        // 1. Going to next chapter from chapter transition
+        // 2. Going between pages of same chapter
+        // 3. Next chapter page
+        when (page?.chapter) {
+            (currentPage as? ChapterTransition.Next)?.to -> return true
+            (currentPage as? ReaderPage)?.chapter -> return true
+            adapter.nextTransition?.to -> return true
         }
 
-        // Going to previous chapter page from a chapter transition - no preload
-        if (page.chapter == (currentPage as? ChapterTransition.Prev)?.to) {
-            return false
-        }
-
-        // Going to next chapter from chapter transition - allow preload
-        if (page.chapter == (currentPage as? ChapterTransition.Next)?.to) {
-            return true
-        }
-
-        // Going between pages of same chapter - can preload
-        if (page.chapter == (currentPage as? ReaderPage)?.chapter) {
-            return true
-        }
-
-        // Next chapter page - allow preload, necessary for 1-3 page chapters.
-        if (page.chapter == adapter.nextTransition?.to) {
-            return true
-        }
-
-        // Other cases, no preload is necessary.
+        // Other cases, no preload is allowed.
         // 1. Going back to chapter from Next transition
         // 2. Going back to chapter from From transition
         // 3. Going back to a loaded page of previous chapter (no transition)
+        // 4. Going to previous chapter page from a chapter transition
         return false
     }
 
@@ -185,7 +173,7 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
         activity.onPageSelected(page)
 
         // Preload next chapter once we're within the last 3 pages of the current chapter
-        val inPreloadRange = (pages.size - page.number < 3)
+        val inPreloadRange = pages.size - page.number < 3
         if (inPreloadRange && allowPreload) {
             Timber.d("Request preload next chapter because we're at page ${page.number} of ${pages.size}")
             adapter.nextTransition?.to?.let {
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
@@ -129,26 +129,64 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
     private fun onPageChange(position: Int) {
         val page = adapter.items.getOrNull(position)
         if (page != null && currentPage != page) {
+            val allowPreload = checkAllowPreload(page as? ReaderPage)
             currentPage = page
             when (page) {
-                is ReaderPage -> onReaderPageSelected(page)
+                is ReaderPage -> onReaderPageSelected(page, allowPreload)
                 is ChapterTransition -> onTransitionSelected(page)
             }
         }
+    }
+
+    private fun checkAllowPreload(page: ReaderPage?): Boolean {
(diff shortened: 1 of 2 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -129,26 +129,64 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
...
-                is ReaderPage -> onReaderPageSelected(page)
+                is ReaderPage -> onReaderPageSelected(page, allowPreload)
                 is ChapterTransition -> onTransitionSelected(page)
             }
         }
     }
 
+    private fun checkAllowPreload(page: ReaderPage?): Boolean {
+        // Page is transition page - preload allowed
+        if (page == null)
```
CarlosEsco: page == null ?: return true

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -129,26 +129,64 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
...
+            return true
+
+        // Initial opening - preload allowed
+        if (currentPage == null) {
```
CarlosEsco: currentPage == null ?: return true

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -129,26 +129,64 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
...
+            return true
+        }
+
+        // Going to previous chapter page from a chapter transition - no preload
+        if (page.chapter == (currentPage as? ChapterTransition.Prev)?.to) {
```
CarlosEsco: when block for the rest of this?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -129,26 +129,64 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
...
-    private fun onReaderPageSelected(page: ReaderPage) {
+    private fun onReaderPageSelected(page: ReaderPage, allowPreload: Boolean) {
         val pages = page.chapter.pages!! // Won't be null because it's the loaded chapter
         Timber.d("onReaderPageSelected: ${page.number}/${pages.size}")
         activity.onPageSelected(page)
 
         // Preload next chapter once we're within the last 3 pages of the current chapter
-        val inPreloadRange = pages.size - page.number < 3
-        if (inPreloadRange) {
+        val inPreloadRange = (pages.size - page.number < 3)
```
CarlosEsco: why the parens?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonViewer.kt
```diff
@@ -122,6 +123,40 @@ class WebtoonViewer(val activity: ReaderActivity) : BaseViewer {
...
+            return true
+        }
+
+        // Other cases, no preload is necessary.
+        // 1. Going back to chapter from Next transition
+        // 2. Going back to chapter from From transition
+        // 3. Going back to a loaded page of previous chapter (no transition)
+        return false
+    }
+
```
CarlosEsco: so i see this slightly different then other stuff since its not really ripe for pulling it out into a common function i would at least do what you can to the methods like comments above

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
morcefaster: Doing that breaks smart cast of page to non-nullable.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -129,26 +129,52 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
...
+        page == null ?: return true
+
+        // Initial opening - preload allowed
+        currentPage == null ?: return true
+
+        // Allow preload for
+        // 1. Going to next chapter from chapter transition
+        // 2. Going between pages of same chapter
+        // 3. Next chapter page
+        when (page?.chapter) {
```
CarlosEsco: page.chapter since page isnt null at this point

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -129,26 +129,52 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
...
+            (currentPage as? ChapterTransition.Next)?.to -> return true
+            (currentPage as? ReaderPage)?.chapter -> return true
+            adapter.nextTransition?.to -> return true
```
CarlosEsco: else -> return false

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
morcefaster: Either we use ?: for returning null at the start or this - ?:return true breaks smart cast and page.chapter would throw an exception.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
morcefaster: Fixed

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
CarlosEsco: page!!.chapter

since we know its not null at this point.  or it would have returned earlier

morcefaster: Done

PR #2788, opened 2020-03-30
Add auto migrate to latest chapter

owosenpie:
**Why/user benifet/user problem**
Currently you can update to the first source witch is based off scource name alphabetically, and migrate to most chapters which always selects for manga with multiple of the same chapter, adding migrate by latest chapter will allow people to find the manga with the most chapters that was most recently updated not accounting for repeats.
**What/requirements**
When automigrating use the last chapter and updated date info to find the manga with the highest last chapter that was most recently updated.


Jays2Kings:
Bruh... this isn't how pull requests work

CarlosEsco:
Are you trying to merge all of j2k into tachiyomi? Lol

Issue #2107, opened 2019-03-24
Can't install/download extensions


gabriel-dehan:
Hello, I just installed this app for the first time and I just can't install any extension, when I try it just stays in a pending state for a very long time before asking me to retry.

Issue seems to be similar to : https://github.com/inorichi/tachiyomi/issues/1848
But I have no VPN, I have no issue downloading from the Google Play Store, and I have allowed the app to download from unknown sources.

My phone is a Xiaomi MI A1

I have tested the following app versions: 0.8.0, 0.8.1, r1474

Thanks a lot.

hXtreme:
For Xiaomi you need to individually authorize apps to install apps.

Best,
Harsh

hXtreme:
Does it read Allowed under Tachiyomi in

Settings > System & Device > Additional settings > Privacy > Special App Access > Install unknown apps > Tachiyomi

bsd1787:
Hi! I have found the solution. In my case I cannot update or download too and I tried using a different connection and it worked perfectly.
I guess Internet Service Protocol had blocked the address in my router because when I connected to my data I had no problem updating or downloading.

StevenInacio:
Hi! I'm currently running the dev builds, was a bit behind the updates unil a few days ago where I updated to r2024.
Now I'm facing the same problem.
I've got an Asus ZenFone 3 Z017D, battery optimisations are off, even autostart is enabled. I can't install or update any extensions through the app. Updating the app itself was no problem when I updated to 2028 today.

StevenInacio:
Hi! I'm currently running the dev builds, was a bit behind the updates unil a few days ago where I updated to r2024.
Now I'm facing the same problem.
I've got an Asus ZenFone 3 Z017D, battery optimisations are off, even autostart is enabled. I can't install or update any extensions through the app. Updating the app itself was no problem when I updated to 2028 today.

MCAxiaz:
@StevenInacio Are you getting a "unable to parse package" error or something else?
Also can you try to revert to an [earlier version such as r2011](https://tachiyomi.kanade.eu/tachiyomi-r2011.apk). Are you able to install extensions on that earlier version?

StevenInacio:
Hi @MCAxiaz, unfortunately I can't install anything on that version either.
I'm not getting any error message, it just says "pending" whenever I try to install something. If I go back a screen and back into the extensions it reads "update"/"install" again.

[...]

PR #2859, opened 2020-04-10
Fixes Issue #2107

MCAxiaz:
Fixes https://github.com/inorichi/tachiyomi/issues/2107

Honestly I still have no idea what the exact issue is. My best guess is it's the same issue as the one described in the comment block, except it also affects certain devices post N. 
`Due to a bug in Android versions prior to N, the installer can't open files that do not contain the extension in the path, even if you specify the correct MIME.`

Unfortunately since the issue is device dependent, I only have a sample size of one person that confirmed this change works (Android 7.0, vivo 1713)

Also tested on pre and post N emulators without noticing any issues.


arkon:
Nice, thanks. I'm fine with changing it since it doesn't seem to be any _worse_ than before.

PR #2875, opened 2020-04-14
Using the "android-gif-drawable" library to improve the GIF's FPS

DumbCodeGenerator:
I have a problem with GIFs(they are way too laggy, I don't know if this is a phone problem). I switched from the Glide's GifDrawable to the "[android-gif-drawable](https://github.com/koral--/android-gif-drawable)" library and this fixed the GIF problem for me. 

I guess it's not the best solution, but at least the GIFs are playing smooth now

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -204,6 +204,9 @@ dependencies {
     kapt "com.github.bumptech.glide:compiler:$glide_version"
     implementation 'jp.wasabeef:glide-transformations:4.1.0'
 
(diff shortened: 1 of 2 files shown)


arkon:
I actually need to see if SSIV can just show GIFs normally, maybe with some tweaks to the underlying image library, which already has libgif.

arkon:
It'd probably be better to just switch to Coil like J2K/Neko.

PR #2887, opened 2020-04-17
toggle reader menu if transition by tapping are disabled

raslmenea:


arkon:
Thanks for the contribution!

PR #2906, opened 2020-04-19
Update packagingOptions to exclude some META-INF

TacoTheDank:
META-INF really isn't needed, so let's Thanos snap it.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -89,11 +89,14 @@ android {
     }
 
     packagingOptions {
-        exclude 'META-INF/DEPENDENCIES'
         exclude 'LICENSE.txt'
+        exclude 'META-INF/DEPENDENCIES'
         exclude 'META-INF/LICENSE'
         exclude 'META-INF/LICENSE.txt'
         exclude 'META-INF/NOTICE'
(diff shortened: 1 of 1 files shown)


Flat:
Does that mean we only delete half of it?

TacoTheDank:
@Flat uhm... hrm... should've thought about it more

arkon:
I'm a bit hesitant on this change. I think _some_ of it is useful?

TacoTheDank:
@arkon If you're hesitant, I can change it to just exclude guaranteed pointless files.

Although, I've never experienced any sort of problem from an app that had nothing in META-INF.

Here's some related links:
- https://stackoverflow.com/questions/39305775/what-are-the-purposes-of-files-in-meta-inf-folder-of-an-apk-file/39305776
- https://discuss.kotlinlang.org/t/disable-meta-inf-generation-in-gradle-android-project/3830/6

arkon:
It's just my gut feeling. If there's proof against it, I'm fine with excluding all of it.

TacoTheDank:
Don't really need this tbh

PR #2908, opened 2020-04-20
Bug/2894 covers not updating

### LINKED ISSUE #2894: [Bug] Covers not updating like they used to
### Device information
* Tachiyomi version: r2028 and r2064
* Android version: 9
* Device: Samsung Galaxy Tab A

## Steps to reproduce
1. Pull down on the manga info page to refresh cover, but it doesn't.

### Expected behavior
If the source has a new cover, it shows up within a few seconds.

### Actual behavior
The new cover doesn't show up until after restarting Tachi.

### Other details
Someone brought this up in discord yesterday.  They were trying to get a new cover for their manga but they couldn't get it to update by using the pull-refresh method or by clearing their metadata.  They stated that it would only refresh the cover if they removed the manga from their library and then re-added it.  

In trying to reproduce their issue I found that my covers also wouldn't update unless I removed and re-added it or until after I had restarted Tachi.  The initial bug reporter was also on a recent version of dev.  I tested with MangaDex and MangaSee covers and their selectors look fine; so I don't think it's an extension issue.

MCAxiaz:
Fixes #2894 

Problem was in two places. First, `MangaInfoController` only requested reload if the image is null, so refreshing is not possible since the image would've already been set.

Secondly, Glide uses mem and disk caches. There's already code like the `MangaSignature` class that handles the disk cache. But the mem caches use the `equals()/hashCode()` methods of the model, in this case `Manga`, which only compares the manga's url. So even when the thumbnail url has been updated, glide would not reload the image until it is evicted from the cache (which happens after restarting the app among other things).

The model loader now takes in a wrapper `MangaThumbnail` which also checks the thumbnail url for equality.

Also made some `url.startsWith("http")` checks case insensitive.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaThumbnail.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaThumbnail.kt
@@ -1,0 +1,28 @@
+package eu.kanade.tachiyomi.data.glide
+
+import eu.kanade.tachiyomi.data.database.models.Manga
+
+class MangaThumbnail(val manga: Manga) {
+
+    val url = manga.thumbnail_url
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as MangaThumbnail
+
+        if (manga != other.manga) return false
+        if (url != other.url) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = manga.hashCode()
+        result = 31 * result + (url?.hashCode() ?: 0)
+        return result
+    }
+}
+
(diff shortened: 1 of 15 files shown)


CarlosEsco:
This doesn't address 1 issue though. Where the covers url doesn't change. 

It would be 10x easier to just clear glide cache and force an update whenever a thumbnail url is not null or empty

MCAxiaz:
I didn't really want to open that can of worms with same url, though I guess there is something I want to try.

It's not that simple since the covers are backed by local files handled by `CoverCache`, so changes need to be done there regardless of what Glide does. Besides, Glide doesn't support clearing the cache for specific files to begin with unless you want to nuke the entire app-wide cache.

arkon:
I'm fine with that being a separate change.

MCAxiaz:
Sure I can open a separate PR for that.

PR #2932, opened 2020-04-23
Update nucleus because we can

TacoTheDank:
La-dee-dah...♪ realizing nucleus can be updated without any changes needing to be made... don't mind me... ♪


TacoTheDank:
@arkon Quick question: do you plan on fully replacing RxJava with Coroutines?

arkon:
That's a pipe dream considering the app heavily relies on it. It'd be nice though.

PR #2962, opened 2020-04-26
Pull basic manga/chapter metadata from EPUB files.

Steveice10:
Follow-up to my previous pull request for some additional EPUB enhancements.

Pulls the following information from the first chapter for local manga:
* Author
* Description

Pulls the following information for each chapter:
* Title
* Scanlator (Set to Publisher if found, otherwise Creator)
* Publication Date

Also contains code to fuzzy-match manga titles to remove them from chapter names, for cases in which a manga's title within chapter metadata contains characters that can't be represented in the manga directory name.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
@@ -112,12 +112,20 @@ class LocalSource(private val context: Context) : CatalogueSource {
                     }
                 }
 
-                // Copy the cover from the first chapter found.
-                if (thumbnail_url == null) {
-                    val chapters = fetchChapterList(this).toBlocking().first()
-                    if (chapters.isNotEmpty()) {
+                val chapters = fetchChapterList(this).toBlocking().first()
+                if (chapters.isNotEmpty()) {
+                    val chapter = chapters.last()
+                    val format = getFormat(chapter)
+                    if (format is Format.Epub) {
+                        EpubFile(format.file).use { epub ->
+                            epub.fillMangaMetadata(this)
+                        }
+                    }
+
+                    // Copy the cover from the first chapter found.
+                    if (thumbnail_url == null) {
                         try {
-                            val dest = updateCover(chapters.last(), this)
+                            val dest = updateCover(chapter, this)
                             thumbnail_url = dest?.absolutePath
                         } catch (e: Exception) {
                             Timber.e(e)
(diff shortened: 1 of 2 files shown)


app/src/main/java/eu/kanade/tachiyomi/util/storage/EpubFile.kt
```diff
@@ -44,6 +48,48 @@ class EpubFile(file: File) : Closeable {
...
+        }
+
+        if (publisher != null) {
+            chapter.scanlator = publisher.text()
+        } else if (creator != null) {
+            chapter.scanlator = creator.text()
+        }
+
+        if (date != null) {
+            val parsedDate = ISO8601Utils.parse(date.text(), ParsePosition(0))
```
arkon, line 88: Probably don't want to rely on internal Gson stuff. Can refer to other places in the app that use `SimpleDateFormat`.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -180,6 +196,40 @@ class LocalSource(private val context: Context) : CatalogueSource {
...
+                }
+
+                if (invalidMangaChar) {
+                    mangaTitleIndex++
+                }
+            } else {
+                chapterNameIndex++
+                mangaTitleIndex++
+            }
+        }
```
arkon, line 228: This seems pretty overkill tbh...

app/src/main/java/eu/kanade/tachiyomi/util/storage/EpubFile.kt
Steveice10: SimpleDateFormat is apparently not ISO 8601 compliant, but I'll try to work out a better way of doing this.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
Steveice10: The issue here is that a manga's actual title may contain characters like colons that aren't supported in directory names, and as a result the previous code wouldn't be able to strip the manga title out of chapter titles. The only solution I could think of was to put together code that ignores characters that aren't alphanumeric or whitepsace when trying to remove the manga title.

app/src/main/java/eu/kanade/tachiyomi/util/storage/EpubFile.kt
```diff
@@ -44,6 +49,58 @@ class EpubFile(file: File) : Closeable {
...
+        }
+
+        if (publisher != null) {
+            chapter.scanlator = publisher.text()
+        } else if (creator != null) {
+            chapter.scanlator = creator.text()
+        }
+
+        if (date != null) {
+            val dateFormat = SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssXXX", Locale.getDefault())
```
arkon, line 92: We can't use `X` since only Android SDK 24+ supports that.

Steveice10: Replaced with `ZZZZZ`.

Issue #2959, opened 2020-04-25
[Bug] Rendering issues while using dark theme


AmplifyPaulo:
### Device information
* Tachiyomi version: dev r2112
* Android version: 8.1.0
* Device: Meizu 16th

## Steps to reproduce
1. Go to the General Settings and change the theme to dark theme

### Expected behavior
It was expected that both the app and the phone would become dark with white text and icons

### Actual behavior
While the app does become dark with with text and icons the phone header and footer don't, and instead they get a dark background with dark icons and text.

### Other details
While this looks like a bug from the phone itself this issue doesn't happen in the stable release of tachiyomi v0.8.5 so certainly something must have changed. So I decided to check more revisions for the dev version and found out that r1905 and prior render's the dark theme correctly while r1908 and further does not. I don't have the intermediate version between 1905 and 1908 so I'm not exactly sure which revision introduced the issue

r2112
![S00425-16252792](https://user-images.githubusercontent.com/33487678/80284420-4f702000-8716-11ea-808d-d2a0a0638ae4.jpg)

stable 0.8.5
![S00425-16350405](https://user-images.githubusercontent.com/33487678/80284427-51d27a00-8716-11ea-8399-e86d9f344b37.jpg)

r1905
![S00425-17054301](https://user-images.githubusercontent.com/33487678/80284551-28661e00-8717-11ea-87db-25a8de68ef8c.jpg)

arkon:
Diff between r1905 and r1908 for future reference: https://github.com/inorichi/tachiyomi/compare/a6c1de1cb2249956643b119d9d819d3fc8223c7f..e41edc1fb754a577dc2c367ba0e6dd8135d53d60

Moravuscz:
![Screenshot_20200426-082608](https://user-images.githubusercontent.com/5168699/80299812-a10cbf00-8797-11ea-8474-15c5e5fadfd7.png)
I have this on the stable 0.8.5 release, but only the navbar... Haven't tested Dev versions.
Phone: Sony Xperia XZ Premium
Android: 9 (stock Sony ROM, no root or anything)

arkon:
Seems like a ROM issue.

Moravuscz:
Probably, but 0.9.x doesn't do this anymore...
![Screenshot_20200507-060705](https://user-images.githubusercontent.com/5168699/81253492-02028580-9029-11ea-96a9-a005a8172546.png)

Oh well, don't worry about it...

PR #2993, opened 2020-04-30
fix copy paste mistake in PreferencesHelper.kt

0x766F6964:
I noticed this when reading the diff comparison in issue #2959. I doubt this solves the issue but it was almost certainly a mistake.


arkon:
Whoops good catch. Thanks!

PR #3101, opened 2020-05-07
Manga cover updates

MCAxiaz:
Some changes to how covers (including custom covers work). No functional changes besides one extra piece of UI for removing custom covers.

### General Cover Loading Changes

- Glide caching is now done using a new `metadata_date` field in a `manga` object. This is persisted to db, and is only updated when certain actions occur:
   - swipe refresh in manga info
   - refresh library metadata in advanced settings
   - editing a cover
   - removing a manga from library
- Fix issue where old cover file is left in cache if the `thumbnail_url` changes.

### Custom Covers

- Improved responsiveness. Edited covers should now immediately be loaded into the app.

- Custom covers used to just overwrite the normal cover file. This meant custom covers could not be added to manga without covers. Custom covers are now stored independently of the normal cover. This does not apply to local manga, which will continue to use a single cover file.

<details>
<summary>UI Change</summary>

When edit cover is selected for a non-local manga that already has a custom cover, a dialog will appear to provide an option to delete the custom cover. In all other cases the dialog will not be shown and the file picker will be opened as before.

Screenshot_1588824756

</details>


app/src/main/java/eu/kanade/tachiyomi/data/cache/CoverCache.kt
```diff
@@ -17,51 +18,89 @@ import java.io.InputStream
...
+    fun deleteFromCache(manga: Manga, deleteCustomCover: Boolean = false): Int {
+        var deleted = 0
 
-        destFile.outputStream().use { inputStream.copyTo(it) }
+        getCoverFile(manga)?.let {
+            if (it.exists() && it.delete()) ++deleted
+        }
+
+        if (deleteCustomCover) {
+            if (deleteCustomCover(manga)) ++ deleted
```
arkon, line 84: Nitpick: remove space after `++`

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
```diff
@@ -16,6 +16,8 @@ interface Manga : SManga {
 
     var chapter_flags: Int
 
+    var metadata_date: Long
```
arkon, line 19: The name isn't super clear what it is. Seems like it's the time that the cover was last updated?

app/src/main/java/eu/kanade/tachiyomi/data/glide/LibraryMangaCustomCoverFetcher.kt
```diff
@@ -0,0 +1,25 @@
...
+
+open class LibraryMangaCustomCoverFetcher(
+    private val manga: Manga,
+    private val coverCache: CoverCache
+) : FileFetcher() {
+
+    override fun loadData(priority: Priority, callback: DataFetcher.DataCallback<in InputStream>) {
+        getCustomCoverFile()?.let {
+            loadFromFile(it, callback)
+        } ?: callback.onLoadFailed(Exception("Custom cover file not found"))
```
arkon, line 19: So this would throw if it's been set but it couldn't actually find the file on disk?

app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaThumbnailModelLoader.kt
```diff
@@ -48,12 +48,6 @@ class MangaThumbnailModelLoader : ModelLoader<MangaThumbnail, InputStream> {
      */
     private val defaultClient = Injekt.get<NetworkHelper>().client
 
-    /**
-     * LRU cache whose key is the thumbnail url of the manga, and the value contains the request url
-     * and the file where it should be stored in case the manga is a favorite.
-     */
-    private val lruCache = LruCache<GlideUrl, File>(100)
```
arkon, line 55: Did you remove it because you don't directly handle files outside of the caches anymore?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -314,23 +317,21 @@ class LibraryPresenter(
     fun removeMangaFromLibrary(mangas: List<Manga>, deleteChapters: Boolean) {
         // Create a set of the list
         val mangaToDelete = mangas.distinctBy { it.id }
-        mangaToDelete.forEach { it.favorite = false }
-
-        Observable.fromCallable { db.insertMangas(mangaToDelete).executeAsBlocking() }
-            .onErrorResumeNext { Observable.empty() }
-            .subscribeOn(Schedulers.io())
-            .subscribe()
```
arkon, line 322: Hrm I'm not entirely sure why this was separate before. To ensure we remove them from the DB even if the download deletion fails?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -357,21 +358,42 @@ class LibraryPresenter(
...
+     * @param data uri of the cover resource.
      */
-    @Throws(IOException::class)
-    fun editCoverWithStream(inputStream: InputStream, manga: Manga): Boolean {
-        if (manga.source == LocalSource.ID) {
-            LocalSource.updateCover(context, manga, inputStream)
-            return true
-        }
+    fun editCover(manga: Manga, context: Context, data: Uri) {
+        Observable
```
arkon, line 366: This does have to be an observable? We should try to use coroutines when possible for new code.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -357,21 +358,42 @@ class LibraryPresenter(
...
+            )
+    }
 
-        if (manga.thumbnail_url != null && manga.favorite) {
-            coverCache.copyToCache(manga.thumbnail_url!!, inputStream)
-            return true
-        }
-        return false
+    fun deleteCustomCover(manga: Manga) {
+        Observable
```
arkon, line 387: Same here re: observable vs coroutine.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoController.kt
```diff
@@ -113,7 +111,7 @@ class MangaInfoController(private val fromSource: Boolean = false) :
 
         // Set SwipeRefresh to refresh manga data.
         binding.swipeRefresh.refreshes()
-            .onEach { fetchMangaFromSource() }
+            .onEach { fetchMangaFromSource(true) }
```
arkon, line 114: Can you explicitly write what the argument is too (i.e. `manualFetch = true`)?

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryHolder.kt
```diff
@@ -63,12 +63,10 @@ class HistoryHolder(
 
         // Set cover
         GlideApp.with(itemView.context).clear(cover)
-        if (!manga.thumbnail_url.isNullOrEmpty()) {
```
arkon, line 66: Wait, so what exactly happens if it _is_ null or empty? Some sources might not provide thumbnails.

MCAxiaz: It just doesn't load anything. [The loader already handles null urls](https://github.com/inorichi/tachiyomi/pull/3101/files#diff-801ccb870281f6050f0bbadb37c02307L91). This was basically a micro optimization since I guess it is technically faster if you just skip the image loading entirely.

app/src/main/java/eu/kanade/tachiyomi/data/glide/MangaThumbnailModelLoader.kt
MCAxiaz: The loader is used when the file is not in memory cache, and the fetcher is used when the file is not in disk cache. Any work you do in the loader is wasted if the image is in the disk cache already. Files are also only used be the fetcher.

So rather than doing this caching where the file might not even be used it's better to just defer file handling to the fetcher.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
MCAxiaz: Hm yea I named it before I even thought of custom covers so it was only intended to be changed on swipe refresh or refresh library metadata, hence the name. I'm thinking maybe changing it to `cover_last_modified` or something.

app/src/main/java/eu/kanade/tachiyomi/data/glide/LibraryMangaCustomCoverFetcher.kt
MCAxiaz: A custom cover is only considered "set" if it can find the file, since there's no explicit variable tracking whether or not a custom cover has been added. So it throws if it hasn't been set.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
MCAxiaz: Not too sure either, might've just been so the two could occur without waiting for each other. Though I would just slap a `try-catch` in the `forEach` to be safe. Or I guess if you really care about rx conventions, turn the `fromCallable` into a proper stream.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
MCAxiaz: Is there some kind of coroutine api for accessing the controller from the presenter?

arkon: Ah my bad, I missed that. No there isn't.

app/src/main/java/eu/kanade/tachiyomi/data/cache/CoverCache.kt
```diff
@@ -17,51 +18,89 @@ import java.io.InputStream
  */
 class CoverCache(private val context: Context) {
 
+    companion object {
+        private const val COVERS_DIR = "covers"
+        private const val CUSTOM_COVERS_DIR = "custom covers"
```
arkon, line 23: Just a thought, maybe it could be `covers/` and `covers/custom/`?

MCAxiaz:
I added the "refresh library metadata" option back in as "refresh covers", since updating covers on every library update is a bit too much. Instead, the regular library update will only refresh covers if the url has changed.

arkon:
This is great, thanks!

PR #3117, opened 2020-05-08
Download queue - clean

vanessabluellie:
Hello, I wanted to add a improvement for the next glorious update for the best manga app on android! 

App version: tachiyomij2k-1.0.2

On the recent tabs, when you pull up the downloads, they have all the chapters for all the manga you've queued. Rather than having just all the chapters one by one for every manga, when you add something to the queue, could you make it so that the manga title appears first. And then when you click on the manga, the chapters for their respective manga qill be there. It could be in 'list' form!

I have a screenshot of what the downloads tab could look like rather than all separated chapters of different manga! Thanks!


CarlosEsco:
Don't open a pull request merging j2k into stable

PR #3118, opened 2020-05-08
String Formatting Fixes

MCAxiaz:
~~Fix hindi formatting strings~~ Strings changed on weblate
Prefer `context.getString()` for formatting since android studio is able to detect formatting errors.

Also I'm using this PR to plug an issue with the title of `TrackLogoutDialog`, since `R.string.logout_title` is a format string but is no longer being passed any args.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
@@ -94,8 +94,9 @@ internal class DownloadNotifier(private val context: Context) {
                 )
             }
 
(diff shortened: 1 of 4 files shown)


arkon:
Can you fix the strings on Weblate instead?

MCAxiaz:
Done.

Do you want me to [change this line as well](https://github.com/inorichi/tachiyomi/blob/ce6848b3c0c48b7321e8b21278867e49ab4c6992/app/src/main/java/eu/kanade/tachiyomi/widget/preference/TrackLogoutDialog.kt#L22)? I assume it's actually supposed to be calling `getString()`, though commit history shows you actually removed it so I'm not too sure what you intended it to be.

arkon:
Yeah I didn't even realize it was broken.

PR #3129, opened 2020-05-09
Add integrated dummy restore

jobobby04:
Adds restore support for read chapters using dummy extensions.
Dummy extensions will have to be updated to return -99F when fetching chapter list, SnakeDoc83 has done this and helped me test my code, and telling me any errors that occurred.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/ChapterTypeAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/ChapterTypeAdapter.kt
@@ -14,6 +14,7 @@ object ChapterTypeAdapter {
     private const val READ = "r"
     private const val BOOKMARK = "b"
     private const val LAST_READ = "l"
+    private const val NUMBER = "n"
 
     fun build(): TypeAdapter<ChapterImpl> {
         return typeAdapter {
@@ -22,6 +23,8 @@ object ChapterTypeAdapter {
                     beginObject()
                     name(URL)
                     value(it.url)
+                    name(NUMBER)
+                    value(it.chapter_number.toDouble())
                     if (it.read) {
                         name(READ)
                         value(1)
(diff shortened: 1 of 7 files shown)


arkon:
I would much rather it be an option when doing the restore and not actually relying on an external dummy source.

CarlosEsco:
Arkon how would you know which manga should use dummy code and which should use normal?

arkon:
If a source is missing => stub it

jobobby04:
I think I can try something, but I am not confidant on it being able to do what you want

jobobby04:
So I made manga and chapters not have to have a source or dummy to restore.

I also added chapter numbers to the backup so the chapters can migrate if someone restores to a fork with automatic source migration, but this causes other forks to not restore because they get stuck on the chapter number. 

I can revert this change, but it makes chapter restore almost useless unless they are able to get the original source working, as automatic source migration works with chapter numbers. 

Also I am not sure if you wanted the broken source migration the way I did it, personally I am not too advanced with Kotlin yet, so anything else is likely out of my doable range

app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/ChapterTypeAdapter.kt
```diff
@@ -22,6 +23,8 @@ object ChapterTypeAdapter {
                     beginObject()
                     name(URL)
                     value(it.url)
+                    name(NUMBER)
+                    value(it.chapter_number.toDouble())
```
arkon, line 27: Is the motivation for this just so you can retain the reading history? I'm not a fan of changing these types and breaking backwards compatibility.

jobobby04: That's pretty much it, it would allow people to migrate their read chapters, its not full proof, but it definitely helps.

Though the reason it gets stuck is because there is no else statement in the when statement at the bottom of the file, so we could add a else statement to cover for future changes if someone decides to add another variable to the backup

arkon:
Closing this in favour of eventually adding the better backup/restore system.

PR #2847, opened 2020-04-09
Fix zh-rCN plural strings

Lyaiya:

Issue #2714, opened 2020-03-16
Weblate plural translation


Lyaiya:
I translate the language in Weblate, but I found a error at plural translation.
After I translated Chinese plural string in Weblate, it should be <item quantity="other">, not <item quantity="one"> in string.xml (zh-rCN).
Because of the Language Plural Rules #zh, plural only have "other" category and no "one" category.
May be caused by Weblate? Sorry my poor English.

Language Plural Rules #zh: https://unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html#zh

arkon:
Seems related to https://github.com/WeblateOrg/weblate/issues/520

I'm not sure if we can fix that from the Weblate UI or not. I could fix them manually in the source files in this repo. @j2ghz Any thoughts?

PR #3142, opened 2020-05-10
Fix zh-rCN plural strings manually

Lyaiya:
Look at [Weblate/#3837](https://github.com/WeblateOrg/weblate/issues/3837). The context of `<item quantity="other">` forced move to `<item quantity="one">` in `app/src/main/res/values-zh-rCN/strings.xml` after fix the bug.

So I need to fix zh-rCN plural strings manually. And when tachiyomi add new plurals, I should do the same work.

Related links:

- #2847
- #2714


arkon:
There were some merge conflicts, so I did it manually: https://github.com/inorichi/tachiyomi/commit/1da169319d95c96df51a81e19df16c4ede682133

Hopefully Weblate fixes this.

Issue #3139, opened 2020-05-10
[Bug] Covers set as blank after checking for updates


Matth7878:
### Device information
* Tachiyomi version: Preview r2289
* Android version: 10
* Device: OP-7

## Steps to reproduce
Update Library

### Expected behavior
Library updates and if cover failed to be retrieved or is blank it should not be updated.

### Actual behavior
If you update the library and it fails to retrieve cover you end up with manga without one.

### Other details
I think it's because covers are now updated when checking sources but if there is no cover from the source previous cover should be kept.

Edit : edited to respect template for submitting bugs

arkon:
Addressed by #3101

PR #3150, opened 2020-05-10
Cover Update Followup to Address #3139

MCAxiaz:
https://github.com/inorichi/tachiyomi/issues/3139

My PR actually potentially causes this kind of issue, since even though the thumbnail url won't get set to null, the local cover file will still be deleted. And it's possible the current url is no longer valid. Added some additional logic to prevent that from happening.

Not sure what happened in the original issue though, since my PR hadn't even been merged yet. My guess it was just a networking error with fetching the new url.


arkon:
Do you mind removing the concurrency parts from this PR? I'm refactoring that stuff in my local changes too.

PR #3175, opened 2020-05-14
Some more forgotten translations 1

j2ghz:
can't merge automatically, so if your translation is here and you really want it let me know


arkon:

PR #3177, opened 2020-05-14
Some more forgotten translations 2

j2ghz:


yshalsager:
Arabic seems to be okay.

asereze:
Sardinian is fine as well.

alessandrojean:
The Brazilian Portuguese translation in the three PRs is outdated and doesn't need to be saved.

arkon:

PR #3178, opened 2020-05-14
Some more forgotten translations 3

j2ghz:


arkon:

Issue #1962, opened 2019-04-11
Sort by release date of chapter


NetherKing1357:
**App version:**
r1489
**Android version:**
8.0.0
**Request:**
Please add an option to sort issues/chapters by release date of chapter. This option (if added) would preferably go in the 'Sorting mode' tab in the hamburger menu of a page of the manga. 

**Benefits:**
Omakes, specials and annuals will be sorted correctly to improve ease of reading

Titus30:
Does readcomiconline extension works

7sevenseven:
I agree with you.

arkon:
"Latest chapter" sort was added some time ago.

bbtora:
They are talking about chapter list sorting. not library sorting.

emlautarom1:
Please, take a look at #3179

PR #3179, opened 2020-05-15
Added sorting chapters by upload date

emlautarom1:
Solves #1962. Spanish `strings.xml` contains the proper translation for the new feature.

---

A few things as a first-time contributor:

- Handling masks is a real pain when it comes to adding new features. I guess this is done for performance reasons or for easier serialization, but it's actually necessary?
https://github.com/inorichi/tachiyomi/blob/8a4c4c346ab8f23fbe3e12d091ad291cbb0dcc7c/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt#L58-L85

- The sorting selection function for chapters redefines each function if the order is ascending/descending, requiring code duplication:
https://github.com/inorichi/tachiyomi/blob/cac11f8be4b67f23670bceb9f9a5326cf271bd80/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersPresenter.kt#L197-L210

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
@@ -78,7 +78,8 @@ interface Manga : SManga {
 
         const val SORTING_SOURCE = 0x00000000
         const val SORTING_NUMBER = 0x00000100
-        const val SORTING_MASK = 0x00000100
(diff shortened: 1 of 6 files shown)


MCAxiaz:
There area other places that rely on the sorting as well, such as the reader.

emlautarom1:
Yes, my bad, I didn't check properly all places. I'm making the changes right now.

emlautarom1:
So,

PR #3180, opened 2020-05-15
Sort chapters by upload date

emlautarom1:
Solves #1962. Spanish `strings.xml` contains the proper translation for the new feature.

---

A few things as a first-time contributor:

- Handling masks is a real pain when it comes to adding new features. I guess this is done for performance reasons or for easier serialization, but it's actually necessary?
https://github.com/inorichi/tachiyomi/blob/8a4c4c346ab8f23fbe3e12d091ad291cbb0dcc7c/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt#L58-L85

- The use of masks makes it harder to find every place that uses the sorting method, so I missed a few in my first PR.

- The sorting selection function for chapters redefines each function if the order is ascending/descending, requiring code duplication:
https://github.com/inorichi/tachiyomi/blob/cac11f8be4b67f23670bceb9f9a5326cf271bd80/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersPresenter.kt#L197-L210

- The sorting functionality is spread across multiple files and repeated, for example:
https://github.com/inorichi/tachiyomi/blob/9e830f1c55a9514462290868864b84194aef7b15/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt#L98-L102
https://github.com/inorichi/tachiyomi/blob/9e830f1c55a9514462290868864b84194aef7b15/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersPresenter.kt#L197-L208

Just a thought, but having each sorting method as a sealed class and placing all sorting logic inside that class would have made this 10x easier. There is a [ChapterLoadStrategy](app/src/main/java/eu/kanade/tachiyomi/ui/reader/ChapterLoadStrategy.kt) but it does not handle all the necessary logic for sorting

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
@@ -98,6 +98,7 @@ class HistoryPresenter : BasePresenter<HistoryController>() {
         val sortFunction: (Chapter, Chapter) -> Int = when (manga.sorting) {
             Manga.SORTING_SOURCE -> { c1, c2 -> c2.source_order.compareTo(c1.source_order) }
             Manga.SORTING_NUMBER -> { c1, c2 -> c1.chapter_number.compareTo(c2.chapter_number) }
+            Manga.SORTING_UPLOAD_DATE -> { c1, c2 -> c1.date_upload.compareTo(c2.date_upload) }
             else -> throw NotImplementedError("Unknown sorting method")
         }
 
@@ -116,6 +117,12 @@ class HistoryPresenter : BasePresenter<HistoryController>() {
                         it.chapter_number > chapterNumber &&
                             it.chapter_number <= chapterNumber + 1
                     }
+            }
+            Manga.SORTING_UPLOAD_DATE -> {
+                val dateUpload = chapter.date_upload
+                ((currChapterIndex + 1) until chapters.size)
+                    .map { chapters[it] }
+                    .firstOrNull { it.date_upload > dateUpload }
             }
             else -> throw NotImplementedError("Unknown sorting method")
         }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
@@ -119,10 +119,8 @@ class HistoryPresenter : BasePresenter<HistoryController>() {
                     }
             }
             Manga.SORTING_UPLOAD_DATE -> {
-                val dateUpload = chapter.date_upload
-                ((currChapterIndex + 1) until chapters.size)
-                    .map { chapters[it] }
-                    .firstOrNull { it.date_upload > dateUpload }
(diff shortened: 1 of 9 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
```diff
@@ -117,6 +118,12 @@ class HistoryPresenter : BasePresenter<HistoryController>() {
                             it.chapter_number <= chapterNumber + 1
                     }
             }
+            Manga.SORTING_UPLOAD_DATE -> {
+                val dateUpload = chapter.date_upload
+                ((currChapterIndex + 1) until chapters.size)
+                    .map { chapters[it] }
+                    .firstOrNull { it.date_upload > dateUpload }
```
arkon, line 125: Should this be `>=`?

app/src/main/res/values/strings.xml
```diff
@@ -473,6 +473,7 @@
     <string name="sorting_mode">Sorting mode</string>
     <string name="sort_by_source">By source</string>
     <string name="sort_by_number">By chapter number</string>
+    <string name="sort_by_upload_date">By chapter upload date</string>
```
arkon, line 476: "By upload date" is fine.

arkon:
Thanks for the contribution!

---

Addressing the points from the PR overview:

> Handling masks is a real pain when it comes to adding new features. I guess this is done for performance reasons or for easier serialization, but it's actually necessary?

I don't have context around it, but I believe that is the reason. It's allows for easier backups and migrations with a smaller space footprint. I don't expect it to change for 0.x since it'd be a hassle to do with compatibility across versions/forks, but maybe we can reconsider it for the future rearchitected version.

> The sorting selection function for chapters redefines each function if the order is ascending/descending, requiring code duplication:

> The sorting functionality is spread across multiple files and repeated, for example:

Yup, we could probably clean that up somehow. You can open up another PR for that refactoring if you want to take a stab at it.

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
emlautarom1: I wasn't very sure about this one:
- 'date_upload' is a `Long`, not a proper `Date`, so I went with a strict `>` for "safety". I'll change it for a `>=` but I'm not sure if this could break something later on.
- This tries to find the first chapter with a "bigger" upload date but starting from the current chapter, relying on the current sorting mode (if I'm not mistaken). I don't know if this could end up creating some bugs later on.

arkon: I suggested it since the date values are typically on the day itself, rather than super granular (i.e. to a specific second), and there tends to be chapters released on the same day.

I can only think of it resulting in a bug if the implementation between this and the `ChaptersPresenter` differs. Ideally we should refactor it so the logic is common between the two, like you mentioned.

arkon:
This is great, thanks for the contribution!

PR #3193, opened 2020-05-17
download new chapters changes

### LINKED ISSUE #2325: Issue with "download new chapters" feature!
**App version:** 0.8.4

**Android version:** 8.1.0

**Issue/Request:**
Recently began having issues auto downloading new chapters. I've had to manually download my manga chapters for the last few weeks. I really enjoyed the download new chapters feature and it's kind of disappointing that it isn't working.

**Steps to reproduce (if applicable)**

1. Go into settings, then select downloads and enable "download new chapters".
 2. Select a category or categories to include in the download.
 3. Wait for new chapters to appear once library updates, then it should download new chapters... But it's doesn't!

**Other details:**
I use multiple extensions: webtoons, mangarock, mangasee, mangahere. And none of them are auto downloading new chapters.

This is the only issue I've ever experienced with Tachiyomi. I love this app and the people behind it. It's the best manga reader I've ever used! 

**Edit:**
Since opening this issue I discovered that using multiple category tags on a manga prevents autodownloads from working. It's hard to explain but the image below should help. I have 3 categories: reading, manga, webtoons. Reading is set to download new chapters but the manga and webtoons inside Reading use multiple tags. When I removed all but one category tag auto downloads began to work for those manga. It's also possible that what I experienced was just a coincidence but who knows.

Screenshot_20191114-001216

### LINKED ISSUE #3168: [Feature Request] Allow "Download New Chapters" when updating from chapter list
Currently the "download new chapters" setting only works for library updates. If I swipe refresh on the chapters list of a manga, the new chapters are not downloaded.

### What/Requirements
Swipe refreshing on chapters list of a library manga should also respect the download settings.

MCAxiaz:
Closes #3168 
Fixes #2325

Also some other related changes

- Fix bug with library updates where new chapters do not download if the manga is in multiple categories and download categories are specified.

- Allow auto initial chapters fetch on `ChaptersController` when not navigating from browse source.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -23,6 +23,7 @@ import eu.kanade.tachiyomi.source.SourceManager
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.util.chapter.syncChaptersWithSource
 import eu.kanade.tachiyomi.util.prepUpdateCover
+import eu.kanade.tachiyomi.util.shouldDownloadNewChapters
 import eu.kanade.tachiyomi.util.storage.getUriCompat
 import eu.kanade.tachiyomi.util.system.acquireWakeLock
 import eu.kanade.tachiyomi.util.system.isServiceRunning
@@ -256,10 +257,6 @@ class LibraryUpdateService(
         val newUpdates = ArrayList<Pair<LibraryManga, Array<Chapter>>>()
         // List containing failed updates
         val failedUpdates = ArrayList<Pair<Manga, String?>>()
-        // List containing categories that get included in downloads.
-        val categoriesToDownload = preferences.downloadNewCategories().get().map(String::toInt)
-        // Boolean to determine if user wants to automatically download new chapters.
-        val downloadNew = preferences.downloadNew().get()
         // Boolean to determine if DownloadManager has downloads
         var hasDownloads = false
 
@@ -278,11 +275,7 @@ class LibraryUpdateService(
                     // Filter out mangas without new chapters (or failed).
                     .filter { pair -> pair.first.isNotEmpty() }
                     .doOnNext {
-                        if (downloadNew && (
-                            categoriesToDownload.isEmpty() ||
-                                manga.category in categoriesToDownload
-                            )
-                        ) {
+                        if (manga.shouldDownloadNewChapters(db, preferences)) {
                             downloadChapters(manga, it.first)
                             hasDownloads = true
                         }
@@ -309,7 +302,7 @@ class LibraryUpdateService(
 
                 if (newUpdates.isNotEmpty()) {
                     notifier.showUpdateNotifications(newUpdates)
-                    if (downloadNew && hasDownloads) {
+                    if (hasDownloads) {
                         DownloadService.start(this)
                     }
                 }
@@ -326,14 +319,9 @@ class LibraryUpdateService(
     }
 
     private fun downloadChapters(manga: Manga, chapters: List<Chapter>) {
-        // we need to get the chapters from the db so we have chapter ids
-        val mangaChapters = db.getChapters(manga).executeAsBlocking()
-        val dbChapters = chapters.map {
-            mangaChapters.find { mangaChapter -> mangaChapter.url == it.url }!!
-        }
         // We don't want to start downloading while the library is updating, because websites
         // may don't like it and they could ban the user.
-        downloadManager.downloadChapters(manga, dbChapters, false)
+        downloadManager.downloadChapters(manga, chapters, false)
     }
 
     /**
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -278,11 +275,7 @@ class LibraryUpdateService(
                     // Filter out mangas without new chapters (or failed).
                     .filter { pair -> pair.first.isNotEmpty() }
                     .doOnNext {
-                        if (downloadNew && (
-                            categoriesToDownload.isEmpty() ||
-                                manga.category in categoriesToDownload
```
MCAxiaz, line 283: The list of manga to update is created here with a [distinctBy call on the manga id](https://github.com/inorichi/tachiyomi/blob/5f315957d825b4ea40677504de153cb0d64e6660/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt#L232), since a separate `LibraryManga` object is created for each category of a manga. So the final list only contains the `LibraryManga` object corresponding to the first category that a manga belongs to.

So if a manga belongs to category 1 and 2, and I set categories to download to 2, the chapters will not download since it thinks the manga is only in category 1.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -326,14 +319,9 @@ class LibraryUpdateService(
     }
 
     private fun downloadChapters(manga: Manga, chapters: List<Chapter>) {
-        // we need to get the chapters from the db so we have chapter ids
-        val mangaChapters = db.getChapters(manga).executeAsBlocking()
-        val dbChapters = chapters.map {
-            mangaChapters.find { mangaChapter -> mangaChapter.url == it.url }!!
-        }
```
MCAxiaz, line 333: No longer necessary since [this commit](https://github.com/inorichi/tachiyomi/commit/cc7735e284d6084cca2ecc1eff61085ff9f8ff2f#diff-7aee8c5fe9ec96f6fd30d85e193ea207R123)

app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChaptersController.kt
```diff
@@ -280,15 +276,16 @@ class ChaptersController :
     }
 
     private fun initialFetchChapters() {
-        // Only fetch if this view is from the catalog and it hasn't requested previously
-        if ((parentController as MangaController).fromSource && !presenter.hasRequested) {
```
MCAxiaz, line 284: I'm not sure if there was a particular reason why this had to be restricted to fromSource. I think even if it is from library, doing a single auto fetch when the chapters list is empty is reasonable. 

There was also an edge case where if you just added a manga to library by long tapping from browse source without ever going into manga details, then going back to library and opening the manga there, the chapters list would be blank and require a manual refresh. So I think just doing an auto fetch is much nicer behavior.

arkon: I think it made sense if it was a standalone function that was called from multiple places and you wanted to avoid unnecessary calls, but it was really only ever called from this one place for this one scenario.

It might actually be more straightforward to just move this logic directly into `onNextChapters`. As it is now, having the call to `initialFetchChapters()` seems misleading unless you actually read this part.

app/src/main/java/eu/kanade/tachiyomi/util/MangaExtensions.kt
```diff
@@ -47,3 +48,17 @@ fun Manga.updateCoverLastModified(db: DatabaseHelper) {
...
+
+fun Manga.shouldDownloadNewChapters(db: DatabaseHelper, prefs: PreferencesHelper): Boolean {
+    // Boolean to determine if user wants to automatically download new chapters.
+    val downloadNew = prefs.downloadNew().get()
+
+    val categoriesToDownload = prefs.downloadNewCategories().get().map(String::toInt)
+
+    val categoriesForManga by lazy {
+        db.getCategoriesForManga(this).executeAsBlocking().mapNotNull { it.id }
+    }
```
arkon, line 60: There isn't much point to lazily instantiating this if it's going to be run as part of this function anyway.

MCAxiaz: It's only accessed as part of the last condition check, so on larger libraries it saves a a lot of db accesses if the evaluation gets short circuited.

arkon: Ah, I see. In that case, I'd rather there be explicit short circuiting in the function, e.g. after checking the `downloadNew` preference.

MCAxiaz: done

PR #3197, opened 2020-05-17
Add a simple way to detect the default user language

jobobby04:
This adds a simple way to detect the users language and add it to the default enabled languages on first launch

Locale.getDefault().language returns values such as en, he, ja, and other simple language codes, so I believe they can be matched to the extension ones(though its likely not 100%)


arkon:
`Locale.getDefault().language` is already in the default for `enabledLanguages()` though.

jobobby04:
I thought of this when I saw someone say in support they needed to enable their language for a extension to work, it can be closed if it's actually already there

arkon:
Yeah, I don't think this would help if the existing default value doesn't already work. Not sure why it wouldn't work though, unless they had set a value already before the default locale was added as part of the preference's default value.

Issue #1338, opened 2018-04-04
Rematch episodes (like migrate from source to the same source)


444nonymous:
Sometimes mangas are re-uploaded in a manga site with a new chapter naming. Tachichomi does not detect chapters as already downloaded. The old ones are unlisted and new ones are re-downloaded.

Another problem even if you manually set old chapters as downloaded is that if you zip the whole manga, chapters will be out of order, because different name schemes are used so sorting is never done right by offline readers or Tahiyomi's local mangas.

Also a new default chapter naming would be very useful for storage. Long chapter names could be stored in the database as usual, but directory folders (or zips when available) could always be just "Chapter or MANGA + Number" (the fixed word CHAPTER or MANGA name + the variable number).

LakotaGray:
In regards to the initial point about new chapter naming and re-uploaded chapters, if the url link is not identical, Tachi sees it as different, which means its "new." From my understanding, this has to do with the way Tachi stores the info and images and is part of the core way it operates. 

I'm not sure what exactly you mean by the second part of your post, but if you take the images from downloaded and zip them it would need to be added to local. The naming schemes in local manga can be altered however you want, so make sure its in an order that makes sense to you.

444nonymous:
It's not the URL. I manually renamed the folders and Tachiyomi detects now chapters as downloaded.

Old naming: Manga name + Chapter (The word chapter) + Number
Was reuploaded in server as: Manga name + Number

By removing the word "Chapter" it was fixed. Tachiyomi could use a default storage naming to avoid this problem.

inorichi:
> Tachiyomi could use a default storage naming to avoid this problem.

What do you mean?

CarlosEsco:
@inorichi basically wants us to ignore chapter names and use our own naming convention. I don't think it would work well

444nonymous:
@inorichi

But only for storage (folder names). Chapter names can be displayed as usual in Tachiyomi. In fact there is already an option to "show title" or "show chapter number". 

It would be just using "show chapter number" info for creating the folders instead of the current "show title"

inorichi:
What if the chapter numbers are not recognized?

444nonymous:
What about using both? Tachiyomi can create an standard pattern that includes everything.

The problem is created when using server chapter names. Tachiyomi could standardize the names with parameters. So all folder names could start with Chapter + "number" + "text".

arkon:
Closed by #3216

PR #3216, opened 2020-05-21
Rename downloaded chapters

ChrisJAllan:
This renames downloaded chapters when the chapter names change on the source, because it is very annoying to have all of my downloads disappear as soon as volume numbers or chapter titles are added.

Addresses #1338, so long as the chapter's url stays the same.


arkon:
Nice idea!

Issue #1822, opened 2019-01-28
[Feature requiest] Consistent zoom


Maurogch:
App version: 0.8.0

Request: Add option to keep zoom level when going to next page/s

Other details: If I (pinch) zoom, I want the next pages to have the same zoom applied to them.

bahari15:
Bumping this one because I too, want this feature.

CarlosEsco:
@bahari15  don't bump things all it does is email people.  then we delete it cause its annoying.  It doesnt move it up

CreativeC0der:
Is there a possibility for this feature to come in the next update??cause its in high demand

EyangKodok:
Please make it a thing!! \(^0^)/

JennieGucci:
Please make this a thing... It's really hard to read Manga with 2 pages merge with it. :(

panyann:
> Please make this a thing... It's really hard to read Manga with 2 pages merge with it. :(

@JennieGucci Use this Tachiyomi fork, it has this feature and **tons** of other https://github.com/Jays2Kings/tachiyomiJ2K/releases

It has also an option to keep the margins always.

JennieGucci:
ohh really... thanks

phanirithvij:
If you'd like people to not spam here why not have a issues repo for people to comment on? Marking every comment as spam looks like you don't want people to say anything here.

Flat:
There is a reason there is a discord. GitHub issues are not for commenting or chatting.

amcsi:
+1

In fact, maybe it didn't even need to be an option, and could just be the behavior.
I could imagine that when you swiped left to vote the next manga chapter, you could get taken to the top right with the same zoom level, and going back would take you to the bottom left.

nikhilCad:
I think that this issue is considered not priority by devs, but I still think you should give it more priority. It's the third most "Liked" issue on this repository, and will be a huge QoL improvement.

LeitHunt:
Note: 
I searched if these request were already done. 
They were already requested but I wanna present these requests in new perspective :)
[Thanks for such a flawless app & user experience - From a beginner in android development :)]

I have 2 request:
-Pre-set zoom level
-Auto-scroll in vertical mode

This may also help in reduction of battery drain...in amoled screens 
How?:
-Pre-set zoom level:
I have set background full black & always read manga in landscape mode with 50% zoom as by double taping on every page.
So the result is only half screen is covered by a manga page leaving remaining half screen full black.

[...]

PR #3269, opened 2020-06-02
Continuous  Zoom Level

JennieGucci:
Please add a Continuous Zoom Level per page option in a manga... Sometimes, there are manga series, especially the old ones that has 2 pages per page cause it was scanned using it from the original Book. Continuous Zoom Level will continue zooming the page with the right exact size you zoom in per page and chapter, it will only stop Continuous Zoom if you turned off the option. Thanks!


arkon:
#1822

PR #3319, opened 2020-06-13
Added category-wise display setting

### LINKED ISSUE #3288: Allow different library categories to have different display options
>  Allow different library categories to have different display options

As it stands, All the categories in the Library can be set to only one kind of display.

so how about an option to have an independent display option for each category? (that can be checked or unchecked depending on preference)

This would allow users to make some categories show up as grid view while others as list view.
Having multiple display options for different categories would enable making reading lists within the app easier. Coupled with adding a 'date added' filter option would make tachiyomi more usable as a reading list+reader app combo.

Here's an image to visualize it better: https://i.imgur.com/DXDB8CS.jpg

Rogavactive:
closes #3288 
category wise options stored in preferences.
added switch in settings to turn setting on or off.
added helper method to store preference enum map as preferences.
list and grid options both now are `AutofitRecyclerView` -s:
 unable to access category information at `onCreate` of `LibraryCategoryView` , so instantiate both as `AutofitRecyclerView` grids and then manipulate with `spanCount`. Tested and works fine.


app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
```diff
@@ -141,6 +141,8 @@ object PreferenceKeys {
 
     const val defaultCategory = "default_category"
 
+    const val categorisedDisplaySettings = "categorised_display_settings"
```
arkon, line 144: Very nitpick: can we use "categorized" instead? The codebase is already using Americanized spelling.

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryController.kt
```diff
@@ -36,6 +39,11 @@ class CategoryController :
     CategoryRenameDialog.Listener,
     UndoHelper.OnActionListener {
 
+    /**
+     * Preferences helper necessary to apply changes.
+     */
```
arkon, line 44: Can remove this, it's pretty redundant. Also consider using `injectLazy()`

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryController.kt
```diff
@@ -179,6 +187,10 @@ class CategoryController :
 
         when (item.itemId) {
             R.id.action_delete -> {
+                adapter.selectedPositions.forEach {
+                    adapter.getItem(it)?.category?.id?.let { it1 -> preferences.getCategoryDisplayPreference(it1).delete() }
```
arkon, line 191: Use more meaningful names than `it` or `it1`

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
```diff
@@ -247,7 +247,7 @@ class LibraryController(
     }
 
     fun showSettingsSheet() {
-        settingsSheet?.show()
+        adapter?.categories?.get(binding.libraryPager.currentItem)?.id?.let { settingsSheet?.show(it) }
```
arkon, line 250: Nitpick: put the `let` body on its one line rather than having it all in one line for readability

app/src/main/res/values/strings.xml
```diff
@@ -195,6 +195,7 @@
     <string name="pref_category_library_categories">Categories</string>
     <string name="default_category">Default category</string>
     <string name="default_category_summary">Always ask</string>
+    <string name="categorised_display_settings">Display settings by Category</string>
```
arkon, line 198: "Per-category display settings"?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsLibraryController.kt
```diff
@@ -107,6 +108,19 @@ class SettingsLibraryController : SettingsController() {
...
+
+            switchPreference {
+                key = Keys.categorisedDisplaySettings
+                titleRes = R.string.categorised_display_settings
+                defaultValue = false
+
+                onChange {
+                    // Set default value to global settings. This will ensure,
+                    // that default library settings will not mess anything up.
+                    preferences.libraryDisplayMode().set(PreferenceValues.DisplayMode.COMPACT_GRID)
```
arkon, line 120: I'm not sure I understand what's going on here.

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
Rogavactive: Sure. But I think it should be at least categorizedDisplay. There could be more we can "categorize".

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryController.kt
Rogavactive: Sure. Forgot about lazy injection.

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryController.kt
Rogavactive: Done. Even did a comment explaining what's going on here.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
Rogavactive: Not one-liner anymore. Also used named argument instead it.

app/src/main/res/values/strings.xml
Rogavactive: Changed it.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsLibraryController.kt
Rogavactive: As I remember I added it early on, but I found cause of the problem now and fixed it, so I don't need these lines and removed. If library was on grid mode Per-category display mode was acting strangely so I was reseting that preference here.

arkon:
As a heads up, I don't plan on adding this until after the next release hence the open state.

That said, as I was poking around some other stuff, I noticed that `Category` actually has a `flags` field that could be used for storing this sort of stuff too, similar to how we deal with chapter flags. Did you explore that when implementing this?

arkon:
Superseded by #5232

PR #3320, opened 2020-06-13
Chapter nth to the last deletion logic fixed

### LINKED ISSUE #3252: [Bug] Delete nth to the last doesn't work sometimes
### Device information
* Tachiyomi version: Stable 0.9.2
* Android version: 5.1
* Device: Huawei TIT-U02

## Steps to reproduce (way 1, real usecase)
1. Enable remove fifth chapter to the last.
2. Enable download new chapters.
3. Update library.
4. While it's updating, open first found manga with new chapter. You'll see that the chapter is queued.
5. Open the chapter and read it.

## Steps to reproduce (way 2, easier to debug)
1. Enable remove second chapter to the last.
2. Add new manga to the library.
3. Add to download queue first and second chapters of the new manga.
4. When the first one is downloaded, pause the queue (second chapter remains in the queue).
5. Read those two chapters.

### Expected behavior
nth chapter to the last should be removed.

### Actual behavior
nth chapter to the last isn't removed.

Rogavactive:
closes #3252 
It seems like when attempting to delete previous nth to last chapter it is supposed to check if that chapter was actually downloaded, but instead was checking current chapter.
fix: now retrieves chapter to delete before attempting to enqueue and checks if it has to delete before attempting.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
@@ -361,13 +361,29 @@ class ReaderPresenter(
         if (selectedChapter.pages?.lastIndex == page.index) {
             selectedChapter.chapter.read = true
             updateTrackChapterRead(selectedChapter)
-            enqueueDeleteReadChapters(selectedChapter)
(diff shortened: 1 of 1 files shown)


arkon:
Nice catch!

PR #3360, opened 2020-06-21
Add option to reverse tapping

### LINKED ISSUE #3266: [Feature Request] Invert Tap Navigation Option
The only option I miss in Tachiyomi is to invert the tapping direction.
If I'm holding my tablet with the right hand I wish a tap on the right side of the screen to go forward (to the left) in a manga.

GHOSTBEAR:
This adds the option to reverse tapping.
By default, the setting is off

Demo with it on:
ezgif com-video-to-gif
And with it off it like normal
ezgif com-video-to-gif (1)

Closes #3266


GHOSTBEAR:
With Webtoon also inverted I think the option should be an enum instead of a boolean. This to let the user choose if they want it inverted when tapping horizontally, vertically, or both

arkon:
I feel like the option should be in the reader sheet too.

GHOSTBEAR:
The option is now also in the reader sheet

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
```diff
@@ -190,6 +192,24 @@ class SettingsReaderController : SettingsController() {
...
+                )
+                entryValues = arrayOf(
+                    TappingInvertMode.NONE.name,
+                    TappingInvertMode.HORIZONTAL.name,
+                    TappingInvertMode.VERTICAL.name,
+                    TappingInvertMode.BOTH.name
+                )
+                defaultValue = TappingInvertMode.NONE.name
+                summary = "%s"
+            }.apply { dependency = Keys.readWithTapping }
```
arkon, line 212: Rather than the dependency, it'd be nice to just hide it completely.

e.g. https://github.com/inorichi/tachiyomi/blob/dev/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadController.kt#L106-L107

catriuspham:
@arkon is this available in v0.9.2? If not, when will it be?

jobobby04:
There is no ETA for the next release, but you can download the latest preview from tachiyomi.org as it contains the feature

Issue #1561, opened 2018-08-03
Order of downloaded pages wrong if more than 1000 pages in single chapter


Zumochi:
Version 0.7.4.

1. find manga/webtoon/whatever that has more than 1000 pages in a single "chapter" with simple numerical filenames
2. download said chapter
3. go to page 100
4. see sudden skip to different page when going to page 101, it's actually going to 1000
5. at page 111 it goes back to 101
6. page 112 goes to 1010

[Example (SFW)](https://e-hentai.org/g/1151347/725a7d2a39/):
[01 - 100-101 web](https://user-images.githubusercontent.com/5372678/43667304-aeb7ef24-9777-11e8-9686-f28814838226.jpg)
[01 - 100-1000 tachiyomi](https://user-images.githubusercontent.com/5372678/43667306-aed3224e-9777-11e8-9d08-7ee00463eddc.jpg)

[02 - 101-102 web](https://user-images.githubusercontent.com/5372678/43667321-ba895e8c-9777-11e8-8e5f-919bb1d6f456.jpg)
[02 - 111-112 tachiyomi](https://user-images.githubusercontent.com/5372678/43667322-baa8abca-9777-11e8-83c0-8a7c47d6ae48.jpg)

Relevant code seems to be [here](https://github.com/inorichi/tachiyomi/blob/34a40b01314fe9f7ba886724395af4f7a04b559d/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt#L121-L124) and I think it can be fixed by doing something [like this](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.comparisons/natural-order.html), but I'm no Java, let alone Kotlin programmer (or even a full-fledged programmer to begin with 😅)

hXtreme:
I'll try to take a look at it in the next week or so if I don't get bogged down by exams!

uap4:
[Google Translate] I can see that chapters are downloading fine, but they are named in the format "%03d". Because of this, their real names are first 001-999, and then 1000-9999. I think that the reader uses the simple sorting, not the downloader.
A different sorting should be used in the reader.
Or it has already been suggested to expand to five digits here - https://github.com/inorichi/tachiyomi/pull/3361.
I am using TachiyomiAZ.

uap4:
[...]

PR #3361, opened 2020-06-21
Dirty "Fix" to issue #1561

jamesma0508:
"Fixes" the issue of the file name %03d in the download.kt that causes sorting problems if the chapter has more than 1k pics. I changed it to %05d since it's quite rare for pics to have more than 100k in 1 chapter (and is also a dirty quick fix :P).

_New to git and first-time pull request here, if I am doing anything wrong please correct me!_

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
@@ -330,7 +330,7 @@ class Downloader(
             return Observable.just(page)
         }
 
-        val filename = String.format("%03d", page.number)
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -330,7 +330,7 @@ class Downloader(
             return Observable.just(page)
         }
 
-        val filename = String.format("%03d", page.number)
+        val filename = String.format("%05d", page.number)
```
arkon, line 333: This will cause the following logic of checking for existing files to fail, so you'd need to check for both the old and new filenames.

PR #3372, opened 2020-06-23
Make for IOS

Yronk:
Please..I think lot of people need it to..for ipad resolution..this is the best manga app ive seen so far


Soitora:
What the f, don't try to pull an entire other fork into the main app? 🤔

PR #3400, opened 2020-06-29
Fix pause notification

raslmenea:
this fixes a bug in downloader when sorting or changing the download item position (drag and drop) shows a pause notification when download completes, and the caused of it beacuase when we sort download items we paused the downloader but did not set DownloadNotifier pause to false.


CarlosEsco:
Fix your commit history then reopen a new pull request.

PR #3489, opened 2020-07-19
fix: Download on WiFi regardless of metered status

### LINKED ISSUE #3395: [Bug] Fails to download over Wifi if 'Only download over Wifi' is enabled
### Device information
* Tachiyomi version: 0.9.2
* Android version: 9
* Device: Lenovo M10 tablet plus

## Steps to reproduce
1. Connect to your Wifi network.
2. Turn on 'Only download over Wifi' in the settings.
3. Attempt to download any manga.

### Expected behavior
It downloads over the Wifi connection.

### Actual behavior
A non-dismissable notification appears: 'No Wifi connection available'. The item gets stuck at 'Queued' and never downloads. Video showing the download working with the option off and failing with the option on: https://m.imgur.com/a/4TSYduW

###Workarounds
Turn off 'Only download over Wifi', force stop Tachiyomi and try again. It will download successfully.

### Other details
Same error occurs on my old Samsung tablet (Android 4.4) but not on my phone (Android 6). This makes me believe that this error only occurs on devices that can only connect to the internet via Wifi and nothing else.

Flat:
fixes #3395

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadService.kt
@@ -4,6 +4,7 @@ import android.app.Notification
 import android.app.Service
 import android.content.Context
 import android.content.Intent
+import android.net.ConnectivityManager
 import android.net.NetworkInfo.State.CONNECTED
 import android.net.NetworkInfo.State.DISCONNECTED
 import android.os.Build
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadService.kt
```diff
@@ -143,7 +144,7 @@ class DownloadService : Service() {
     private fun onNetworkStateChanged(connectivity: Connectivity) {
         when (connectivity.state) {
             CONNECTED -> {
-                if (preferences.downloadOnlyOverWifi() && connectivityManager.isActiveNetworkMetered) {
+                if (preferences.downloadOnlyOverWifi() && connectivityManager.activeNetworkInfo.type == ConnectivityManager.TYPE_MOBILE) {
```
arkon, line 147: This is deprecated, but I guess it's fine for now....

Regardless, shouldn't we be explicitely checking if it's `!wifi`?

Flat: It's deprecated in Java but not the Android SDK somehow... But yeah you're right that's probably the more logical way to do it.

Flat:
The non-depreciated way to do this requires API 24, I think this is fine for now. If we decide to bump target API I will look at re-writing the network monitoring for the download service.

Issue #3521, opened 2020-07-26
[Feature Request] Add manga-wised rotation mode settings (similar to viewer)


bboyz269:
### Why/User Benefit/User Problem
I'm reading a mix of mangas and manhwas in vertical continuous viewer. Force portrait is good for reading manhwa or webtoon but it's impossible to read manga as the texts are so smalls, especially on low resolution phones.  
So I have to frequently change the rotation mode (portrait - landscape) for each manga / manhwa. And it's quite a problem for heavy reader like me.

### What/Requirements
Exactly like the viewer settings.  

A system-wised default rotation mode settings.  
![Screenshot_1595725457](https://user-images.githubusercontent.com/4453811/88469178-a0abcd80-cf18-11ea-9dc5-a89dad54ae44.png)

And a manga-wised rotation mode settings which is override the default settings.  
![Screenshot_1595727177](https://user-images.githubusercontent.com/4453811/88469333-d05bd500-cf1a-11ea-90a7-5f48bd235264.png)
![Screenshot_1595727234](https://user-images.githubusercontent.com/4453811/88469336-d18d0200-cf1a-11ea-9f31-97934fc4df32.png)

### Pull request
I took myself and created a working example codes based on the Viewer settings. Would you take a look?

https://github.com/inorichi/tachiyomi/pull/3522

arkon:
> I took myself and created a working example codes based on the Viewer settings. Would you take a look?

Sure, submit a PR.

bboyz269:
> > I took myself and created a working example codes based on the Viewer settings. Would you take a look?
> 
> Sure, submit a PR.

@arkon I updated the issue with the link to pull request or [here](https://github.com/inorichi/tachiyomi/pull/3522).

PR #3522, opened 2020-07-26
[Feature Request] Add manga-wised rotation mode settings (similar to viewer)

bboyz269:
[[Feature Request] Add manga-wised rotation mode settings (similar to viewer)](https://github.com/inorichi/tachiyomi/issues/3521)

- Most of the code was cloned from "Viewer" settings.
- Minor change to how to screen rotation is set
- Modify translations based on existing ones and google translate


app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/MangaTypeAdapter.kt
```diff
@@ -28,6 +29,7 @@ object MangaTypeAdapter {
                 manga.title = nextString()
                 manga.source = nextLong()
                 manga.viewer = nextInt()
+                manga.rotationType = nextInt()
```
arkon, line 32: Won't this be problematic with old backups?

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
```diff
@@ -16,6 +16,8 @@ interface Manga : SManga {
 
     var viewer: Int
```
arkon, line 17: Maybe we could use this `viewer` field as flags similar to what's done for `chapter_flags`? It could store the viewer config, rotation, and other per-manga options. I think that'd allow us to make it backwards compatible without any database differences and the same size footprint.

Thoughts?

app/src/main/res/values-ar/strings.xml
```diff
@@ -140,7 +140,7 @@
     <string name="zoom_start_left">اليسار</string>
     <string name="zoom_start_right">اليمين</string>
     <string name="zoom_start_center">المركز</string>
-    <string name="pref_rotation_type">الدوران</string>
+    <string name="pref_rotation_type">الدوران الافتراضي</string>
```
arkon, line 143: Please only include English string changes. Translations are handled externally via Weblate.

app/src/main/java/eu/kanade/tachiyomi/data/backup/serializer/MangaTypeAdapter.kt
bboyz269: There will be problems with old backups. It could be fixed by move the `rotationType` the the last token of json and put it in `try catch` for compatibility.  
However, I think I'll just use the follow the current of approach of using `chapter_flags`.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
bboyz269: As I want the changes to be as minimum as possible, I will follow existing approach.  

Though, I have no clue what these number or how to create ones for the new flag.  
I try to put some random ones but value set to `chapter_flags` and value return from rotationType#get is always 0. Could you shed some light on this?

```kotlin
    var rotationType: Int
        get() = chapter_flags and ROTATION_MASK
        set(rotation) = setFlags(rotation, ROTATION_MASK)

```
```kotlin
        const val ROTATION_DEFAULT = 0x00000180
        const val ROTATION_FREE = 0x00000190
        const val ROTATION_LOCK = 0x00000200
        const val ROTATION_FORCE_PORTRAIT = 0x00000210
        const val ROTATION_FORCE_LANDSCAPE = 0x00000220
        const val ROTATION_MASK = 0x00001000
```

app/src/main/res/values-ar/strings.xml
bboyz269: Will be removed.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
bboyz269: Nevermind, I figured it out thanks to [this savior](http://bitwisecmd.com/)

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
```diff
@@ -57,6 +57,10 @@ interface Manga : SManga {
         get() = chapter_flags and SORTING_MASK
         set(sort) = setFlags(sort, SORTING_MASK)
 
+    var rotationType: Int
+        get() = chapter_flags and ROTATION_MASK
```
arkon, line 61: I think we should repurpose/rename `viewer` for these options instead.

In my mind:
- `chapter_flags` = settings in the chapters list view
- `viewer` (or whatever it'll be named, maybe `viewer_flags`) = things in the reader

Of course, that'd mean that we'll need to have the masks for the existing `viewer` values.

bboyz269: To sum up your idea:
- `chapter_flags` stay as it is
- new `viewer_flags` contains flags for `readingMode` (renamed from `viewer`) and `rotationType`.

I'll try to make adjustments following above.

bboyz269: Added changes but I couldn't do a test run yet since I'm at work.  
Could you help review the changes first?

app/src/main/java/eu/kanade/tachiyomi/data/database/tables/MangaTable.kt
```diff
@@ -86,4 +94,23 @@ object MangaTable {
...
+     */
+    val updateViewerValues: String
+        get() = "UPDATE $TABLE SET $COL_VIEWER = CASE $COL_VIEWER " +
+            "WHEN 1 THEN ${MangaImpl().apply { readingMode = Manga.READING_L2R }.viewer_flags} " +
+            "WHEN 2 THEN ${MangaImpl().apply { readingMode = Manga.READING_R2L }.viewer_flags} " +
+            "WHEN 3 THEN ${MangaImpl().apply { readingMode = Manga.READING_VERTICAL }.viewer_flags} " +
+            "WHEN 4 THEN ${MangaImpl().apply { readingMode = Manga.READING_WEBTOON }.viewer_flags} " +
+            "WHEN 5 THEN ${MangaImpl().apply { readingMode = Manga.READING_CONT_VERTICAL }.viewer_flags} " +
+            "ELSE ${MangaImpl().apply { readingMode = Manga.READING_DEFAULT }.viewer_flags} " +
+            "END"
```
arkon, line 109: This doesn't seem necessary if the values are still the same.

app/src/main/java/eu/kanade/tachiyomi/data/database/tables/MangaTable.kt
```diff
@@ -86,4 +94,23 @@ object MangaTable {
...
+
+    /**
+     * Rename columna name `viewer` to `viewer_flags`.
```
arkon, line 112: I'm fine with keeping it with the old name for the sake of reducing database schema changes.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
```diff
@@ -32,30 +32,42 @@ interface Manga : SManga {
...
         chapter_flags = chapter_flags and mask.inv() or (flag and mask)
     }
 
+    private fun setViewerFlags(flag: Int, mask: Int) {
+        viewer_flags = viewer_flags and mask.inv() or (flag and mask)
+    }
+
     // Used to display the chapter's title one way or another
     var displayMode: Int
         get() = chapter_flags and DISPLAY_MASK
```
arkon, line 45: May be good to rename these to be prefixed with `CHAPTER_` and the new ones with `VIEWER_`

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderSettingsSheet.kt
```diff
@@ -151,4 +157,80 @@ class ReaderSettingsSheet(private val activity: ReaderActivity) : BottomSheetDia
         }
         setSelection(intValues.indexOf(pref.get()), false)
     }
+
+    /**
```
arkon, line 161: You could probably eliminate these mappings by using enum classes for the constants instead.

app/src/main/java/eu/kanade/tachiyomi/data/database/tables/MangaTable.kt
bboyz269: YOu're right. It's not needed.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderSettingsSheet.kt
bboyz269: Since we have to maintain a mapping between position and value, we can only tidy up code a bit, not completely eliminate them.  
What I'm thinking of is
```kotlin
package eu.kanade.tachiyomi.ui.reader

import java.lang.IllegalArgumentException

enum class ReadingMode(val position: Int, val value: Int) {
    DEFAULT(0, 0x00000000),
    LEFT_TO_RIGHT(1, 0x00000001),
    RIGHT_TO_LEFT(2, 0x00000002),
    VERTICAL(3, 0x00000003),
    WEBTOON(4, 0x00000004),
    CONTINOUS_VERTICAL(5, 0x00000005);

    companion object {
        const val MASK = 0x00000007

        fun ofPosition(position: Int) = ReadingMode.values().find { it.position == position }
                ?: throw IllegalArgumentException()

        fun ofValue(value: Int) = ReadingMode.values().find { it.value == value }
                ?: throw IllegalArgumentException()
    }
}
```

What's your idea?

app/src/main/java/eu/kanade/tachiyomi/data/database/tables/MangaTable.kt
bboyz269: It's strongly recommended to name the column to reflect its value. But I'll respect your comment and keep it as it was.

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
bboyz269: Are we talking about renaming the constants? Ex: `CHAPTER_DISPLAY_MASK`, `CHAPTER_DISPLAY_NAME`, `CHAPTER_DISPLAY_NUMBER`, ... ect ?

arkon: Yup, just to make it a bit clearer.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderSettingsSheet.kt
arkon: Yeah, something more or less like that.

arkon:
Superseded by #4841

PR #3540, opened 2020-07-30
Add option to disable extension update notifications

### LINKED ISSUE #3525: [Feature Request] Option to disable extension update notifications per extension.
Would really like an option to be able to disable extension update notifications for individual extensions.
---

I mainly use two extensions for 95% of what I read. There is one however that I only use for a few manga (Madara) that is constantly updating. Almost daily and sometime multiple times per day. I feel like there is almost always a notification that there is an update for this extension and it's quite annoying as it is always at the top of my notifications list. As it is the only way to avoid this is to turn off notifications for all extensions. Doing so causes me to not realize when there is an update for the ones I use regularly which unlike the Madra extension often stop working when an update is missed. I assume this is due do most of these updates seeming to just be new sources being added to the Madra extension.

### What/Requirements
All that would be required is a toggle in a menu somewhere for the update notifications for each installed extension.

midyHamdoun:
Closes #3525 (kinda).
If my first time contributing to an open source project so I expect my code to be a mess.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
@@ -157,6 +157,8 @@ object PreferenceKeys {
 
     const val searchPinnedSourcesOnly = "search_pinned_sources_only"
 
+    const val disableExtUpdateNotification = "disable_ext_update_notification"
+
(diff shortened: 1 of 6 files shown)


arkon:
You can just disable notification channels via Android settings.

arkon:
To clarify, if the issue is actually to be addressed, it should be per-extension.

Issue #3033, opened 2020-05-02
[Feature Request] Remove Package name in X-Requested-With


armangido:
### Why/User Benefit/User Problem
Removing the package name in X-Requested-With
Because few sites don't give further access if its been found on headers
### What/Requirements
intercepting or removing the header

arkon:
Are you talking about in the WebView? Regular requests don't include that.

armangido:
it does include when solving the cloudflare challenge

PR #3617, opened 2020-08-06
Update WebViewActivity.kt

armangido:
This code added is for some extension that blocks tachiyomi, by tricking it that it was sent by a android browser, nothing major changes.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/webview/WebViewActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/webview/WebViewActivity.kt
@@ -100,7 +100,9 @@ class WebViewActivity : BaseActivity<WebviewActivityBinding>() {
 
             binding.webview.webViewClient = object : WebViewClientCompat() {
                 override fun shouldOverrideUrlCompat(view: WebView, url: String): Boolean {
-                    view.loadUrl(url)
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/webview/WebViewActivity.kt
```diff
@@ -100,7 +100,9 @@ class WebViewActivity : BaseActivity<WebviewActivityBinding>() {
 
             binding.webview.webViewClient = object : WebViewClientCompat() {
                 override fun shouldOverrideUrlCompat(view: WebView, url: String): Boolean {
-                    view.loadUrl(url)
+                    val android_browser: MutableMap<String, String> = HashMap()
+                    android_browser["X-Requested-With"] = "com.android.browser"
+                    view.loadUrl(url,android_browser)
```
arkon, line 105: Better yet, we should probably be adding this to the existing `headers` map and using that here.

We could also do it in `CloudflareInterceptor` to take care of #3033.

armangido: does CloudflareInterceptor use webview ?

arkon: Yes.

PR #3649, opened 2020-08-11
Optimize images

TacoTheDank:
Done with https://ezgif.com/optipng.


AbdullahM0hamed:
OptimizationMan™

CarlosEsco:
@arkon  is it time to retire the old book icon?  no reason to have old book, android 7 round icon, and the adapative icon for newer android devices

arkon:
Why? It's still used by some launchers and older versions of Android.

CarlosEsco:
sure but you can just make the default icon the round one.

TacoTheDank:
Correct me if I'm wrong, but Android versions under Nougat can't access round icons, right? In that case, we need to keep it for versions below Nougat.

CarlosEsco:
you would just do this

```
        android:icon="@mipmap/ic_launcher"
        android:roundIcon="@mipmap/ic_launcher_round"
```

to

```
        android:icon="@mipmap/ic_launcher_round"
```

arkon:
I'd rather not shoehorn the rounded icon to versions where unique icon shapes are still the norm.

TacoTheDank:
Sure

PR #3650, opened 2020-08-11
Yet another gradle and library update PR

TacoTheDank:
### 1st commit
- Adds a gradle wrapper validation action (see more [here](https://github.com/gradle/wrapper-validation-action))
- Updates gradle wrapper to 6.5.1 using git bash

### 2nd commit
- Remove a few unneeded things
- Make Kotlin use the BuildPluginsVersion
- Rename the KTLINT plugin to differentiate between the two
- Update aboutlibraries-plugin 8.2.0 -> 8.3.0 ([changelog](https://github.com/mikepenz/AboutLibraries/releases))
- Update `org.jlleitschuh.gradle.ktlint` 9.2.1 -> 9.3.0 ([changelog](https://github.com/JLLeitschuh/ktlint-gradle/releases))
- Update ktlint 0.36.0 -> 0.37.2 ([changelog](https://github.com/pinterest/ktlint/releases))
- Update AssertJ test library 3.12.2 -> 3.16.1 ([changelog](https://assertj.github.io/doc/#assertj-core-release-notes))


app/build.gradle
```diff
@@ -260,21 +260,20 @@ dependencies {
     implementation "io.github.reactivecircus.flowbinding:flowbinding-viewpager:$flowbinding_version"
 
     // Licenses
-    final aboutlibraries_version = '8.3.0'
-    implementation "com.mikepenz:aboutlibraries-core:$aboutlibraries_version"
-    implementation "com.mikepenz:aboutlibraries:$aboutlibraries_version"
+    // NOTE: REMEMBER TO UPDATE GRADLE PLUGIN
+    implementation 'com.mikepenz:aboutlibraries:8.3.0'
```
CarlosEsco, line 264: Why did you remove core?

app/build.gradle
```diff
@@ -260,21 +260,20 @@ dependencies {
...
 
     // Tests
     testImplementation 'junit:junit:4.13'
-    testImplementation 'org.assertj:assertj-core:3.12.2'
+    testImplementation 'org.assertj:assertj-core:3.16.1'
     testImplementation 'org.mockito:mockito-core:1.10.19'
 
     final robolectric_version = '3.1.4'
```
CarlosEsco, line 271: This could probably be moved into the versions  since there are 3 uses of it

app/build.gradle
TacoTheDank: The one class that's used isn't located in core in the first place (for whatever reason), so only the all-encompassing one is needed.

app/build.gradle
TacoTheDank: What do you mean "moved into the versions"? Isn't that where it already is?

CarlosEsco: never mind i didn't realize arkon had not moved any other versions to https://github.com/inorichi/tachiyomi/blob/dev/buildSrc/src/main/kotlin/Dependencies.kt

Issue #3659, opened 2020-08-12
Change tracking guide wording


vengiare:
![Screenshot_20200812-042412~01](https://user-images.githubusercontent.com/63973749/90011302-f253a680-dcd3-11ea-9618-80a4aff1e168.png)

Small nitpick: tracking tab no longer exist. Don't really know how to word it tho, "tracking menu" maybe.

shashank-p:
To use Tracker you need to login before from Setting -> Tracking tab once you logged in Tracking button is visible in Manga Info Page beside library button

scb261:
@shashank-p screenshot is from Settings -> Tracking. The issue is that tracking (and chapter list) were moved from tabs to manga info screen and there is no tabs now, but screenshot tells users that tracking still in the tab

PR #3663, opened 2020-08-13
Update and clarify message on tracking services

specterflare:
Relates to #3659. If this is not the proper way to update the app strings then please let me know.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -337,7 +337,7 @@
       <!-- Tracking section -->
     <string name="pref_auto_update_manga_sync">Update chapter progress after reading</string>
     <string name="services">Services</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -337,7 +337,7 @@
       <!-- Tracking section -->
     <string name="pref_auto_update_manga_sync">Update chapter progress after reading</string>
     <string name="services">Services</string>
-    <string name="tracking_info">One-way sync to update the chapter progress in tracking services. Set up tracking for individual manga entries from their tracking tabs.</string>
+    <string name="tracking_info">This sets up a one-way sync with your tracking service(s) that allows Tachiyomi to update the chapter progress and other details of your tracked manga. Set up tracking for individual manga entries from their tracking chip.</string>
```
arkon, line 340: This is too verbose. Can you keep the first sentence as it was originally?

specterflare: Ok

Issue #3520, opened 2020-07-26
[Feature Request] Searchable Settings


mpm11011:
### Why/User Benefit/User Problem
Tachiyomi has a large number of settings. It would be nice to search for a setting by name and load it's respective controller directly to that setting. That would help users find a setting that they know a keyword for, but not where it lives, which would improve speed and ease of use.

### What/Requirements
Similarly to Windows 10 settings or Android settings, you search for a setting by keyword, get a list of results, click a result, and it loads the correct page and autoscrolls right to that setting.

### Roadblocks
I can include a search icon on the SettingsMainController; clicking the icon opens up a search bar, just like LibrarySearchController or GlobalSearchController. 

[Mockups](https://imgur.com/a/Ivw62pP)

However, I see two problems from here. First, the settings names are stored in the strings.xml, so searching for a string, with a string, is not all that helpful on its own. You need to tie the string to an actual Preference object, which leads to problem number 2: The SettingsController subclasses load their Preference objects in a JSON-like collection, but the Preferences have no information as to which controller they belong to. You need to go from Preference objects in your search results to the correct SettingsController subclass to load. 

I am completely new to Kotlin but I would be happy to contribute to the effort.

arkon:
> The SettingsController subclasses load their Preference objects in a JSON-like collection,

The "JSON"-ness is just syntactic sugar. They're still regular old `Preference` objects.

> but the Preferences have no information as to which controller they belong to.

This is indeed an issue.

I know there's library like [SearchPreference](https://github.com/ByteHamster/SearchPreference), but those are based on XML preferences, but Tachiyomi defines them programmatically. My initial thought would be to do some codegen on build to create mappings of preference title strings -> the controller class and the search functionality would do a lookup based on that.

mpm11011:
I had considered using reflection on the controllers to do that lookup , but I'm not sure that would work since the preferences are set up in the setupPreferenceScreen function, they're not properties or members of the class itself.

lmj0011:
>I can include a search icon on the SettingsMainController; clicking the icon opens up a search bar, just like LibrarySearchController or GlobalSearchController.

@mpm11011 
If you want to go ahead and start a PR, I can look into what arkon is suggesting. I too think this is a good thing to have implemented

>My initial thought would be to do some codegen on build to create mappings of preference title strings -> the controller class and the search functionality would do a lookup based on that.

mpm11011:
[...]

PR #3683, opened 2020-08-16
#3520 Searchable Settings

mpm11011:
Per issue #3520, Added UI elements for a search button in the settings menu, along with some classes and design files that should be useful in setting up a search results page.

One of the greater hurdles to implementing this feature is that the various settings menus are populated at runtime, programmatically. This means that when a search result is clicked, we need to a) use reflection to find the correct menu, or b) rely on a compile-time-generated dictionary to map the clicked search result to the correct menu. The latter method could also help greatly with search itself.

My knowledge of Kotlin and Android design patterns is very limited, so please keep an eye out for any mistakes I may have made. Thanks!


app/build.gradle
```diff
@@ -201,6 +201,8 @@ dependencies {
 
     // Preferences
     implementation 'com.github.tfcporciuncula:flow-preferences:1.3.0'
+    implementation 'com.github.ByteHamster:SearchPreference:v1.0.3'
+    implementation "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
```
jobobby04, line 205:
```suggestion
    implementation "org.jetbrains.kotlin:kotlin-reflect:$BuildPluginsVersion.KOTLIN"
```

Also maybe put it where the build.gradle contains the kotlin stdlib

lmj0011:
> This means that when a search result is clicked, we need to a) use reflection to find the correct menu, or b) rely on a compile-time-generated dictionary to map the clicked search result to the correct menu.

I plan to tinker with a runtime approach first since a build time one seems it would be a bit more challenging.

app/build.gradle
lmj0011: I'll fix this.

lmj0011:
@mpm11011 Here's my PR to your branch https://github.com/mpm11011/tachiyomi/pull/1
Once merged, I'll close this PR https://github.com/inorichi/tachiyomi/pull/3695

demo: https://streamable.com/zuqg4p

lmj0011:
Some other notes:

- The Search could be added to the top of the "More" Screen, for quicker access. 
- Scrolling to/Highlighting the preference after controller navigation would be a nice touch 
- Doesn't seem like SettingsSearchPresenter is needed

@mpm11011 I'll leave it to you to make those additions if you choose to.

mpm11011:
@arkon My plan is to merge in @lmj0011's changes locally, test out how it works in an emulator, and if it all works well, merge in his PR. Then I will see if I can take care of point #3 and/or #1 in his previous message. Point #2 I wouldn't know where to start on how to implement that.

Also I should figure out this build check failure we're getting here, unless that has already been fixed.

mpm11011:
Resolved the conflict in build.gradle. Set up a query hint for searching the settings.

About adding the search to the top of the More screen: @lmj0011 are we picturing the same spyglass icon in the top right corner? I don't know if there's enough visual context for the user to know that that icon would only search the settings and not, say, the Categories, About, or Help pages. Maybe we could have a menu item right above the Settings item with a proper label, which is far less ambiguous?

lmj0011:
>About adding the search to the top of the More screen: @lmj0011 are we picturing the same spyglass icon in the top right corner?

That what I was thinking since the More screen is just a page of Preferences just like the Settings screen is.


> Maybe we could have a menu item right above the Settings item with a proper label, which is far less ambiguous?

If you mean putting the search icon there instead, then that seems fine to me

mpm11011:
Oh, I did not notice those. So indeed, we ARE searching the About page as well. Then yeah, let's add the icon.

mpm11011:
I removed the MoreController from the list of controllers in the Helper class. Reason being is that we call setupPreferenceScreen on each controller in that list in initPreferenceSearchResultCollection, but we also need to call initPreferenceSearchResultCollection in MoreController.setupPreferenceScreen... the result was a never ending loop. Removing the MoreController is the simplest way of solving that - there's nothing to search for in that controller itself that can't be searched in other controllers.

lmj0011:
@mpm11011 

> I removed the MoreController from the list of controllers in the Helper class. Reason being is that we call setupPreferenceScreen on each controller in that list in initPreferenceSearchResultCollection, but we also need to call initPreferenceSearchResultCollection in MoreController.setupPreferenceScreen... the result was a never ending loop.

https://github.com/mpm11011/tachiyomi/pull/2 should fix that

> Removing the MoreController from the list is the simplest way of solving that - there's nothing to search for in that controller itself that can't be found in other controllers.

true, it's not needed.


As a note, this feature adds around 1 second to the app's cold start time. This would be [the reason](https://github.com/mpm11011/tachiyomi/blob/437a34b5dce46fb7a1bb773a3a0af1fc66d48009/app/src/main/java/eu/kanade/tachiyomi/ui/base/activity/BaseActivity.kt#L82), but [that method](https://github.com/mpm11011/tachiyomi/blob/437a34b5dce46fb7a1bb773a3a0af1fc66d48009/app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt#L53-L68) can't be called off the UI thread unfortunately. Open to suggestions.

lmj0011:
>Scrolling to/Highlighting the preference after controller navigation would be a nice touch

currently looking into this. 

<hr>

- [ ] all Preferences will need a `key` set or else the highlighting effect will have no effect on it.

arkon:
> About adding the search to the top of the More screen

You could also do something similar to the Android settings screen: https://icdn4.digitaltrends.com/image/digitaltrends/now-playing-settings-1440x2960.jpg

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchItem.kt
```diff
@@ -0,0 +1,69 @@
...
+ * @param results the search results.
+ */
+class SettingsSearchItem(val settingsSearchResult: SettingsSearchHelper.SettingsSearchResult, val results: List<SettingsSearchItem>?) :
+    AbstractFlexibleItem<SettingsSearchHolder>() {
+
+    /**
+     * Set view.
+     *
+     * @return id of view
+     */
```
arkon, line 23: Remove this.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchItem.kt
```diff
@@ -0,0 +1,69 @@
...
+     *
+     * @return holder of view.
+     */
+    override fun createViewHolder(view: View, adapter: FlexibleAdapter<IFlexible<RecyclerView.ViewHolder>>): SettingsSearchHolder {
+        return SettingsSearchHolder(view, adapter as SettingsSearchAdapter)
+    }
+
+    /**
+     * Bind item to view.
+     */
```
arkon, line 39: Remove this.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchItem.kt
```diff
@@ -0,0 +1,69 @@
...
+        payloads: List<Any?>?
+    ) {
+        holder.bind(this)
+    }
+
+    /**
+     * Used to check if two items are equal.
+     *
+     * @return items are equal?
+     */
```
arkon, line 53: Remove this.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchItem.kt
```diff
@@ -0,0 +1,69 @@
...
+            return settingsSearchResult == settingsSearchResult
+        }
+        return false
+    }
+
+    /**
+     * Return hash code of item.
+     *
+     * @return hashcode
+     */
```
arkon, line 65: Remove this.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
```diff
@@ -0,0 +1,69 @@
...
+ * @param sourceManager manages the different sources.
+ * @param db manages the database calls.
+ * @param preferences manages the preference calls.
+ */
+open class SettingsSearchPresenter(
+    val initialQuery: String? = "",
+    val initialExtensionFilter: String? = null,
+    val sourceManager: SourceManager = Injekt.get(),
+    val db: DatabaseHelper = Injekt.get(),
+    val preferences: PreferencesHelper = Injekt.get()
```
arkon, line 28: You could move these down to where `extensionManager` is defined too.

I realize the project isn't very consistent with this. Having it in the constructor would be convenient for unit tests... if we had any.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Fetches the different sources by user settings.
+     */
+    private var fetchSourcesSubscription: Subscription? = null
+
+    /**
+     * Subscription for fetching images of manga.
+     */
+    private var fetchImageSubscription: Subscription? = null
```
arkon, line 45: These aren't used, are they?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    private val extensionManager by injectLazy<ExtensionManager>()
+
+    private var extensionFilter: String? = null
```
arkon, line 49: Are these used?

app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
```diff
@@ -38,7 +48,8 @@ class MoreController :
     private var isDownloading: Boolean = false
     private var downloadQueueSize: Int = 0
 
-    override fun setupPreferenceScreen(screen: PreferenceScreen) = with(screen) {
+    override fun setupPreferenceScreen(screen: PreferenceScreen) = screen.apply {
+        SettingsSearchHelper.initPreferenceSearchResultCollection(context, preferenceManager)
```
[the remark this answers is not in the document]
lmj0011: that method causes a bit of UI lag so I put in the activity so it would be called only once. 

see https://github.com/inorichi/tachiyomi/pull/3683#issuecomment-681624477

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
```diff
@@ -0,0 +1,104 @@
...
+        SettingsTrackingController::class,
+        ExtensionFilterController::class,
+        SourceFilterController::class,
+        AboutController::class
+    )
+
+    /**
+     * Must be called to populate `prefSearchResultList`
+     */
+    fun initPreferenceSearchResultCollection(context: Context, preferenceManager: PreferenceManager) {
```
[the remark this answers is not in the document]
lmj0011: Yea, I can give loading it with via LiveData a try

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
```diff
@@ -0,0 +1,202 @@
...
+     */
+    fun setItems(searchResult: List<SettingsSearchItem>) {
+        adapter?.updateDataSet(searchResult)
+    }
+
+    /**
+     * Opens a catalogue with the given search.
+     */
+    override fun onTitleClick(ctrl: SettingsController) {
+        router.replaceTopController(ctrl.withFadeTransaction())
```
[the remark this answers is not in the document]
lmj0011: I chose replace to avoid going back to `SettingsSearchController` because it didn't seem the natural thing to go back to after hitting the back button.

lmj0011:
@mpm11011 you may want to merge https://github.com/mpm11011/tachiyomi/pull/2 and https://github.com/mpm11011/tachiyomi/pull/3 before working on those review changes

mpm11011:
Will merge in those 2 PR's and then tackle @arkon's suggestions.

app/build.gradle
```diff
@@ -273,7 +273,11 @@ dependencies {
     testImplementation "org.robolectric:shadows-multidex:$robolectric_version"
     testImplementation "org.robolectric:shadows-play-services:$robolectric_version"
 
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$BuildPluginsVersion.KOTLIN"
```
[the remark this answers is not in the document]
mpm11011: Removed that line. Does the same need to be done for "org.jetbrains.kotlin:kotlin-reflect" right below it?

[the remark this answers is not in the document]
arkon: AFAIK no.

app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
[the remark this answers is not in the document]
arkon: Ideally it'd happen in the IO thread, not the main one.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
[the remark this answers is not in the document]
arkon: I feel like it'd be more natural to go back to what you were actually at before, which is the search results.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
arkon: Don't pull in LiveData, try to just do it with existing constructs (e.g. coroutines, flows).

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchAdapter.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Restores the view state of the given holder.
+     *
+     * @param holder The holder to restore.
+     */
+    private fun restoreHolderState(holder: RecyclerView.ViewHolder) {
+        val key = "holder_${holder.bindingAdapterPosition}"
+        val holderState = bundle.getSparseParcelableArray<Parcelable>(key)
+        if (holderState != null) {
```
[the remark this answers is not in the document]
mpm11011: Sorry, I need some clarification on this one. You want the "if" statement replaced by this "let" right? If so, what should we do about the ```holderState``` variable in this line: ```holder.itemView.restoreHierarchyState(holderState)```?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
```diff
@@ -0,0 +1,202 @@
...
+import eu.kanade.tachiyomi.R
+import eu.kanade.tachiyomi.databinding.SettingsSearchControllerBinding
+import eu.kanade.tachiyomi.ui.base.controller.NucleusController
+import eu.kanade.tachiyomi.ui.base.controller.withFadeTransaction
+import eu.kanade.tachiyomi.ui.setting.SettingsController
+
+/**
+ * This controller shows and manages the different search result in settings search.
+ * This controller should only handle UI actions, IO actions should be done by [SettingsSearchPresenter]
+ * [SettingsSearchAdapter.WhatListener] called when preference is clicked in settings search
```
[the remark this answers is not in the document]
mpm11011: Yeah, wasn't sure until after re-reading lmj's code.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
```diff
@@ -0,0 +1,202 @@
...
+ */
+open class SettingsSearchController(
```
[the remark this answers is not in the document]
mpm11011: No, there's probably nothing that needs to inherit from this class. And if there is, then we go back and make it open again.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchAdapter.kt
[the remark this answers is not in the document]
arkon: It's just the more Kotlin-ish way of doing it. The default variable in the lambda would be named `it`, but you can rename it.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
```diff
@@ -0,0 +1,202 @@
...
+//
+//        adapter.allBoundViewHolders.forEach { holder ->
+//            val item = adapter.getItem(holder.bindingAdapterPosition)
+//            if (item != null && pref.key == item.pref.key) {
+//                return holder as SettingsSearchHolder
+//            }
+//        }
+//
+//        return null
+//    }
```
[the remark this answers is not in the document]
mpm11011: @lmj00110 commented that out - I assume because it's not needed anymore.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
```diff
@@ -0,0 +1,104 @@
...
+import kotlin.reflect.full.createInstance
+
+object SettingsSearchHelper {
+    var prefSearchResultList: MutableList<SettingsSearchResult> = mutableListOf()
+        private set
+
+    /**
+     * All subclasses of `SettingsController` should be listed here, in order to have their preferences searchable.
+     */
+    private val settingControllersList: List<KClass<out SettingsController>> = listOf(
```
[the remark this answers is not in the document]
mpm11011: I'll remove the AboutController from that list then. If we're limiting to what is in the Settings menu, do we want to remove the search ability from the More page?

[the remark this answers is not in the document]
arkon: Yes.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    /**
+     * Extracts the data needed from a `Preference` to create a `SettingsSearchResult`, and then adds it to `prefSearchResultList`
+     */
+    private fun getSettingSearchResult(ctrl: SettingsController, pref: Preference, breadcrumbs: String = "") {
+        when (pref) {
+            is PreferenceGroup -> {
+                val breadcrumbsStr = breadcrumbs + " > ${pref.title}"
```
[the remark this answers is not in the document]
mpm11011: I assume RTL means "Right to Left"? How do these locales usually display those kinds of paths? I assume it would be like this, but just want to confirm:

LTR: ```root node > middle > middle > leaf ```
RTL: ```leaf node < middle < middle < root```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
```diff
@@ -0,0 +1,104 @@
...
+    )
+
+    /**
+     * Must be called to populate `prefSearchResultList`
+     */
+    fun initPreferenceSearchResultCollection(context: Context, preferenceManager: PreferenceManager) {
+        prefSearchResultList.clear()
+
+        settingControllersList.forEach { kClass ->
+            val ctrl = kClass.createInstance()
```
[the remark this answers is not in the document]
mpm11011: Can you give an example? I can go test that out and see what happens.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
arkon: Yup.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHolder.kt
```diff
@@ -0,0 +1,37 @@
+package eu.kanade.tachiyomi.ui.setting.settingssearch
+
+import android.view.View
+import eu.kanade.tachiyomi.ui.base.holder.BaseFlexibleViewHolder
+import kotlin.reflect.full.createInstance
+import kotlinx.android.synthetic.main.settings_search_controller_card.*
```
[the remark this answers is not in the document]
mpm11011: Fixed. Weirdly enough, Android Studio likes to auto-correct to ```.*``` if I let it handle them.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
mpm11011: Moved em. Should they be private like ```extensionManager```? 

> unit tests... if we had any.

Long term roadmap project?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
mpm11011: Doesn't seem so. Removed em. Also removed the onDestory, which is now redundant.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
arkon: The automatic backup settings are hidden if backup frequency is disabled.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
mpm11011: Doesn't seem so. Will try removing them and see if anything breaks.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
mpm11011: Done.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
mpm11011: Will work on a helper function to write that string correctly depending on the current locale. Do you happen to know what variable name that determines the locale (and which way it reads)?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
mpm11011: So no matter what value you set for Backup Frequency, the Backup Location and Maximum Backups options do show up in the search results. If you click on them while Backup Frequency is set to Manual (which hides them) then you end up in the Backup settings page, the options are still hidden properly, and nothing is highlighted. At least the app doesn't crash outright.

Generally as a user, I prefer seeing disabled items rather than hidden items. If they're disabled, at least I know they exist and can find the right way to interact with them. If they're hidden, I may never notice them. This is just my personal preference, but I design GUI's like this because I think it's more beneficial to the user.

So here are some options for dealing with this. I'm open to more suggestions as well.

1. Accept this behavior as is.
2. Figure out which settings would be hidden in their respective controllers, and likewise remove them from the settings search list... 
   a) when the settings search list is constructed initially, before any user query.
   b) after the user has entered a query and the list is filtered.
3. Change the behavior of dependent settings so that they are disabled instead of hidden.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
mpm11011: Nothing bad seems to have happened.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchAdapter.kt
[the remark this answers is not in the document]
mpm11011: Renamed ```holderState``` to ```it``` and that resolved the missing reference. Not sure if this ends up breaking anything, but I haven't noticed any problems in testing yet.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
arkon: https://stackoverflow.com/a/47119946/4421500

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
arkon: I'm sort of okay with "hidden" ones (i.e. ones that are shown based on other ones) being surfaced. What about ones that are hidden based on OS level requirements, like the 32-bit color reader option?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
arkon: Not really worth it at this point tbh

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
mpm11011: I was just about to do that until I noticed this: 
```app\src\main\java\eu\kanade\tachiyomi\util\system\ContextExtensions.kt```
```Resources.getSystem().isLTR```

Your thoughts?

File: app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt (hunk 19 above)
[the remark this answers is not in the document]
Comment (mpm11011) [reply app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt:55]: I'm emulating with a Pixel 3a, and that option is available under Reader settings, and it shows up in the search results. Let me set up an environment that doesn't have this option available.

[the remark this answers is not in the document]
Comment (mpm11011) [reply app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt:55]: So if ```Build.VERSION.SDK_INT >= Build.VERSION_CODES.O``` evaluates to false (SDK isn't Oreo or greater) then the 32-bit color option is never added to the ```SettingsReaderController```, which means it is never found when creating the initial search list. So it doesn't show up in the controller, and it doesn't show up when you search for it.

File: app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt [hunk 21]
```diff
@@ -0,0 +1,202 @@
...
+            }
+        } else {
+            SettingsSearchHelper.prefSearchResultList
+                .filter {
+                    val inTitle = it.title.contains(query, true)
+                    val inSummary = it.summary.contains(query, true)
+                    val inBreadcrumb = it.breadcrumb.contains(query, true)
+
+                    return@filter inTitle || inSummary || inBreadcrumb
+                }
```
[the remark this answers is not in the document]
mpm11011: This is probably a question for @lmj0011.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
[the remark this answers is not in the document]
mpm11011: Tried out ```pushController``` instead of ```replaceTopController```: you do go back to the search results, but your query is lost. If we do this, we should at least retain the query.

app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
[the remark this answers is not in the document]
lmj0011: https://github.com/inorichi/tachiyomi/pull/3683#issuecomment-681624477

As a note, this feature adds around 1 second to the app's cold start time. This would be the reason, but **that method can't be called off the UI thread unfortunately. Open to suggestions.**

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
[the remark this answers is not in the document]
lmj0011: yes, it can be put in `SettingsSearchHelper`

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
lmj0011: not familiar with Flows, so will need to look it up later, but aren't they similar to LiveData?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
lmj0011: looks like they will be in the rewrite though https://github.com/tachiyomiorg/android-app/issues/2 :eyes:

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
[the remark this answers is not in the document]
mpm11011: ```if (query.isNullOrBlank()) {```, do we want to display all the results? I think it would be cleaner to display nothing, but I'm open to discussion.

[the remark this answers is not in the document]
lmj0011: I say to display nothing. It nay actually make the wait for search results to be more natural in case I can't find a way to load them ahead of time off the main thread. 

Android system settings has this behavior also where they don't display anything until you start typing, Even shows a progress bar too.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
arkon: That's probably fine too.

app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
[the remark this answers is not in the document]
arkon: > but that method can't be called off the UI thread unfortunately

Why?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
arkon: We don't use LiveData in this app, and I'd rather not introduce it for this.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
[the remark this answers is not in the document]
mpm11011: Done. No promise on the progress bar though. :)

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
mpm11011: I'm assuming the thumbs up means "good" and we can close this line of discussion. Let me know if we need to revisit.

mpm11011:
Pushed out a bunch of updates. Conversations not marked as resolved are still on the table code-wise. Picking out which thread to use, Flows, etc. are little over my head at the moment.

lmj0011:
>Picking out which thread to use, Flows, etc. are little over my head at the moment.

I'll be looking into this later so don't stress too much about it you can't figure it out

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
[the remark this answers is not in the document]
lmj0011: I was asking if they're similar not that I'm going to use it.

app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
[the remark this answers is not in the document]
lmj0011: idk, That's the question I'm trying to figure out. All I know currently is when I put it in the IO thread search results don't get generated. Going to take some more digging to find out why is that the case.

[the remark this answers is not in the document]
lmj0011: This is what happens when you call SettingsSearchHelper.initPreferenceSearchResultCollection() in the IO thread.

It seems `kClass.createInstance()` needs access to the Main thread.

```
2020-08-30 21:00:13.383 9396-9434/eu.kanade.tachiyomi.debug E/ACRA: ACRA caught a InvocationTargetException for eu.kanade.tachiyomi.debug
    java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343)
        at kotlin.reflect.jvm.internal.calls.CallerImpl$Constructor.call(CallerImpl.kt:41)
        at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:108)
        at kotlin.reflect.jvm.internal.KCallableImpl.callDefaultMethod$kotlin_reflection(KCallableImpl.kt:159)
        at kotlin.reflect.jvm.internal.KCallableImpl.callBy(KCallableImpl.kt:112)
        at kotlin.reflect.full.KClasses.createInstance(KClasses.kt:280)
        at eu.kanade.tachiyomi.ui.setting.settingssearch.SettingsSearchHelper.initPreferenceSearchResultCollection(SettingsSearchHelper.kt:55)
        at eu.kanade.tachiyomi.ui.base.activity.BaseActivity$onCreate$1.invokeSuspend(BaseActivity.kt:82)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
     Caused by: java.lang.RuntimeException: Can't create handler inside thread Thread[DefaultDispatcher-worker-2,5,main] that has not called Looper.prepare()
        at android.os.Handler.<init>(Handler.java:207)
        at android.os.Handler.<init>(Handler.java:119)
        at androidx.preference.PreferenceController$1.<init>(PreferenceController.java:137)
        at androidx.preference.PreferenceController.<init>(PreferenceController.java:137)
        at eu.kanade.tachiyomi.ui.setting.SettingsController.<init>(SettingsController.kt:31)
        at eu.kanade.tachiyomi.ui.setting.SettingsAdvancedController.<init>(SettingsAdvancedController.kt:35)
        at java.lang.reflect.Constructor.newInstance0(Native Method) 
        at java.lang.reflect.Constructor.newInstance(Constructor.java:343) 
        at kotlin.reflect.jvm.internal.calls.CallerImpl$Constructor.call(CallerImpl.kt:41) 
        at kotlin.reflect.jvm.internal.KCallableImpl.call(KCallableImpl.kt:108) 
        at kotlin.reflect.jvm.internal.KCallableImpl.callDefaultMethod$kotlin_reflection(KCallableImpl.kt:159) 
        at kotlin.reflect.jvm.internal.KCallableImpl.callBy(KCallableImpl.kt:112) 
        at kotlin.reflect.full.KClasses.createInstance(KClasses.kt:280) 
        at eu.kanade.tachiyomi.ui.setting.settingssearch.SettingsSearchHelper.initPreferenceSearchResultCollection(SettingsSearchHelper.kt:55) 
        at eu.kanade.tachiyomi.ui.base.activity.BaseActivity$onCreate$1.invokeSuspend(BaseActivity.kt:82) 
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33) 
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56) 
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571) 
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738) 
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678) 
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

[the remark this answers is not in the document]
lmj0011: trying to fit Flow into this was a bit more complicated than it's worth, imo. I just moved the call into a more appropriate place.
https://github.com/mpm11011/tachiyomi/pull/4

lmj0011:
>Pushed out a bunch of updates. Conversations not marked as resolved are still on the table code-wise. Picking out which thread to use, Flows, etc. are little over my head at the moment.

I played with Flow a little bit before before giving up. I just moved the call to `SettingsSearchHelper.initPreferenceSearchResultCollection` into the `SettingsSearchController`, so now it doesn't add on to app's cold start time.

https://github.com/mpm11011/tachiyomi/pull/4

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
```diff
@@ -165,6 +165,48 @@ object PreferenceKeys {
 
     const val enableDoh = "enable_doh"
 
+    const val disableBatteryOptimization = "pref_disable_battery_optimization"
```
arkon, line 168: I'd actually rather these keys just be inlined if they're not actually reused elsewhere.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsController.kt
```diff
@@ -50,14 +73,25 @@ abstract class SettingsController : PreferenceController() {
...
         val tv = TypedValue()
         activity!!.theme.resolveAttribute(R.attr.preferenceTheme, tv, true)
         return ContextThemeWrapper(activity, tv.resourceId)
     }
 
+    private fun animatePreferenceHighlight(view: View) {
+        val duration = 500L
+        val repeat = 2
+
+        val colorAnimation = ValueAnimator.ofObject(ArgbEvaluator(), Color.TRANSPARENT, Color.WHITE)
```
arkon, line 88: You might actually want to get the `rippleColor` attribute rather than white to better match themes.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsController.kt
```diff
@@ -50,14 +73,25 @@ abstract class SettingsController : PreferenceController() {
...
+        colorAnimation.duration = duration
+        colorAnimation.repeatCount = repeat
+        colorAnimation.addUpdateListener { animator -> view.setBackgroundColor(animator.animatedValue as Int) }
+        colorAnimation.reverse()
```
arkon, line 92: You could `.apply{ }` this stuff on the declaration rather than assigning it to a variable and setting them afterwards.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
```diff
@@ -82,4 +91,31 @@ class SettingsMainController : SettingsController() {
...
+        // Create query listener which opens the global search view.
+        searchView.queryTextEvents()
+            .filterIsInstance<QueryTextEvent.QueryChanged>()
+            .onEach {
+                performSettingsSearch(it.queryText.toString())
+            }
+            .launchIn(scope)
+    }
+
+    private fun performSettingsSearch(query: String) {
```
arkon, line 116: Wait, what exactly happens with this `query`?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
```diff
@@ -82,4 +91,31 @@ class SettingsMainController : SettingsController() {
...
+        val searchItem = menu.findItem(R.id.action_search)
+        val searchView = searchItem.actionView as SearchView
+        searchView.maxWidth = Int.MAX_VALUE
+
+        // Change hint to show global search.
+        searchView.queryHint = applicationContext?.getString(R.string.action_search_settings)
+
+        // Create query listener which opens the global search view.
+        searchView.queryTextEvents()
+            .filterIsInstance<QueryTextEvent.QueryChanged>()
```
arkon, line 109: We might want to put some restrictions here like requiring a minimum of characters before doing a search, otherwise you could just search something like "a" and get a ton of probably useless results?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
```diff
@@ -82,4 +91,31 @@ class SettingsMainController : SettingsController() {
...
+            .onEach {
+                performSettingsSearch(it.queryText.toString())
+            }
+            .launchIn(scope)
+    }
+
+    private fun performSettingsSearch(query: String) {
+        router.pushController(
+            SettingsSearchController().withFadeTransaction()
+        )
```
arkon, line 119: I haven't actually tried pulling the code and using it, but how's this actually work? It's just repeatedly pushing new controllers as you're typing?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
```diff
@@ -0,0 +1,167 @@
...
+                setItems(getResultSet())
+                return true
+            }
+
+            override fun onMenuItemActionCollapse(item: MenuItem?): Boolean {
+                return true
+            }
+        })
+
+        searchView.setOnQueryTextListener(object : SearchView.OnQueryTextListener {
```
arkon, line 93: Any reason not to use the flow bindings here?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
```diff
@@ -0,0 +1,167 @@
...
+     * returns a list of `SettingsSearchItem` to be shown as search results
+     */
+    fun getResultSet(query: String? = null): List<SettingsSearchItem> {
+        val list = mutableListOf<SettingsSearchItem>()
+
+        if (!query.isNullOrBlank()) {
+            SettingsSearchHelper.getFilteredResults(query)
+                .forEach {
+                    list.add(SettingsSearchItem(it, null))
+                }
```
arkon, line 146:
```
return SettingsSearchHelper.getFilteredResults(query)
    .map { SettingsSearchItem(it, null) }
```

?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Query from the view.
+     */
+    var query = ""
+        private set
+
+    val sourceManager: SourceManager = Injekt.get()
+    val db: DatabaseHelper = Injekt.get()
+    val preferences: PreferencesHelper = Injekt.get()
```
arkon, line 25: Are these used somewhere?

app/src/main/res/layout/settings_search_controller_card.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
arkon, line 2: Redundant wrapper?

app/src/main/res/layout/settings_search_controller_card.xml
```diff
@@ -0,0 +1,52 @@
...
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:ellipsize="end"
+            android:maxLines="1"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@+id/search_result_pref_summary"
+            tools:text="Location" />
+
+
```
arkon, line 49: Nitpick: remove extra empty line

app/src/main/res/values/dimens.xml
```diff
@@ -23,4 +23,8 @@
     <dimen name="text_small_body">14sp</dimen>
 
     <dimen name="bottom_sheet_width">0dp</dimen>
+
+    <dimen name="search_result_wrapper_margin_start">50dp</dimen>
+    <dimen name="search_result_wrapper_margin_top">20dp</dimen>
+    <dimen name="search_result_wrapper_margin_bottom">20dp</dimen>
```
arkon, line 29: I'd actually rather these be inlined if they're not reused. They should ideally also be multiples of 4, e.g. 48dp and 16dp (where 8dp/16dp is commonly used).

arkon (commented):
Generally looks pretty close to being ready to be merged.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
lmj0011: iirc, that's function is not used.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsController.kt
mpm11011: Replacing ```Color.WHITE``` with ```R.attr.rippleColor```.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsController.kt
mpm11011: The function would look like this instead?

```
    private fun animatePreferenceHighlight(view: View) {
        ValueAnimator
                .ofObject(ArgbEvaluator(), Color.TRANSPARENT, R.attr.rippleColor)
                .apply {
                    duration = 500L
                    repeatCount = 2
                    addUpdateListener { animator -> view.setBackgroundColor(animator.animatedValue as Int) }
                    reverse()
                }
    }
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
mpm11011: Nothing it seems. Shortening that section up by moving the ```pushController``` call into ```onCreateOptionsMenu``` and removing the ```performSettingsSearch``` function.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
mpm11011: I can add a 3 character minimum, unless you think another number would be more appropriate?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
mpm11011: Cleaned this section up. Does ```it.queryText``` need to be ```it.queryText.toString()``` before asking for the length?

```
        // Create query listener which opens the global search view.
        searchView.queryTextEvents()
            .filterIsInstance<QueryTextEvent.QueryChanged>()
            .onEach {
                if (it.queryText.length >= 3) {
                    router.pushController(
                            SettingsSearchController().withFadeTransaction()
                    )
                }
            }
            .launchIn(scope)
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
mpm11011: Is that more "Kotlinic"? I assume the results of both operations are the same.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchPresenter.kt
mpm11011: Preferences is used for ```initPreferenceSearchResultCollection```, but we can remove the other two.

app/src/main/res/layout/settings_search_controller_card.xml
mpm11011: ```global_search_controller_card.xml``` has the same LinearLayout wrapper, but ```global_search_controller_card_item.xml``` does not. Since this is also a "card" and not a "card_item" I'm inclined to keep it the way it is, but maybe it's redundant since we are not displaying nearly as much info as the global search. Will try removing it.

app/src/main/res/values/dimens.xml
mpm11011: 20 is not a multiple of 4 eh? Just kidding, changing them from ```50/20/20``` to ```48/16/16```.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
lmj0011: could you link an example?

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
mpm11011: Inlined every string between ```pref_disable_battery_optimization``` and ```pref_about_licenses```.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsController.kt
```diff
@@ -50,14 +73,25 @@ abstract class SettingsController : PreferenceController() {
...
 
     private fun getThemedContext(): Context {
         val tv = TypedValue()
         activity!!.theme.resolveAttribute(R.attr.preferenceTheme, tv, true)
         return ContextThemeWrapper(activity, tv.resourceId)
     }
 
+    private fun animatePreferenceHighlight(view: View) {
+        ValueAnimator
+            .ofObject(ArgbEvaluator(), Color.TRANSPARENT, R.attr.rippleColor)
```
arkon, line 86: This seems to be getting the default attribute rather than the themes.

Try:

```
view.context.getResourceColor(R.attr.rippleColor)
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
arkon: I'm not sure, what do you think? 3 might actually be an issue for other languages too, so we could just leave it unless it poses an issue.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
```diff
@@ -82,4 +91,27 @@ class SettingsMainController : SettingsController() {
...
+        // Initialize search option.
+        val searchItem = menu.findItem(R.id.action_search)
+        val searchView = searchItem.actionView as SearchView
+        searchView.maxWidth = Int.MAX_VALUE
+
+        // Change hint to show global search.
+        searchView.queryHint = applicationContext?.getString(R.string.action_search_settings)
+
+        // Create query listener which opens the global search view.
+        searchView.queryTextEvents()
```
arkon, line 108: It still confuses me what this is really doing. Is it just because the search view is technically the same between the controllers?

Maybe the code'd be more intuitive if `action_search` just directly goes to `SettingsSearchController` with the SearchView expanded/focused?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        searchItem.setOnActionExpandListener(object : MenuItem.OnActionExpandListener {
+            override fun onMenuItemActionExpand(item: MenuItem?): Boolean {
+                searchView.onActionViewExpanded() // Required to show the query in the view
+                setItems(getResultSet())
+                return true
+            }
+
+            override fun onMenuItemActionCollapse(item: MenuItem?): Boolean {
+                return true
```
arkon, line 89: Should this be doing something similar as the other function? I'm not sure if this is what's responsible for the behaviour where:

1. Tap search icon
2. Search for something, e.g. "read"
3. Tap the up button (i.e. the arrow at the top left)
4. Note that the search is cleared and we're just looking at an empty view

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
```diff
@@ -0,0 +1,135 @@
...
+            val inBreadcrumb = it.breadcrumb.contains(query, true)
+
+            return@filter inTitle || inSummary || inBreadcrumb
+        }
+    }
+
+    /**
+     * Extracts the data needed from a `Preference` to create a `SettingsSearchResult`, and then adds it to `prefSearchResultList`
+     */
+    private fun getSettingSearchResult(ctrl: SettingsController, pref: Preference, breadcrumbs: String = "") {
```
arkon, line 81: You could maybe do this separately afterwards if you want, but it might be nice if the matched text is bolded or something. Not necessary though.

app/src/main/java/eu/kanade/tachiyomi/ui/more/AboutController.kt
```diff
@@ -91,7 +96,10 @@ class AboutController : SettingsController() {
         }
 
         preferenceCategory {
+            titleRes = R.string.about_resources
```
arkon, line 99: I kind of preferred it without a title. Is this required?

lmj0011: `About` is not included in the search results, so no.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
mpm11011: That's a good point. I do know that when typing in Japanese on GBoard hiragana can automatically convert to kanji, so as you type your query length may actually become shorter. I'm sure there are similar considerations in other languages as well.

e.g.　
1. Enter "kanji" in the keyboard: ```query = "かんじ"```; ```query.length == 3```
2. Hit tab or use the autocomplete: ```query = "漢字"```; ```query.length == 2```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
mpm11011: I've noticed this behavior as well. I can try my hand at a fix, but it will be trial and error for me.

app/src/main/java/eu/kanade/tachiyomi/ui/more/AboutController.kt
mpm11011: Removed that line.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsController.kt
mpm11011: The flashing is now a light gray on most themes, rather than a harsh blue.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
mpm11011: Seems to be fixed with this. Any concern about the validity of suddenly popping the current controller (should be SettingsSearchController)?

```
            override fun onMenuItemActionCollapse(item: MenuItem?): Boolean {
                searchView.onActionViewCollapsed() // Required to show the query in the view
                router.popCurrentController()
                return true
            }
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchHelper.kt
mpm11011: Adding a comment to this effect on that function. Might require some passing of the query text to do the matching, which I don't think the Helper class has right now.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
mpm11011: I'm not sure I understand it myself. I agree that when you read it, it looks like you're pushing a new ```SettingsSearchController``` onto the stack every time the query text changes, but that doesn't happen. Instead the controller stack always contains these elements before the push:

```MoreController -> SettingsMainController```

and these elements after the push:

```MoreController -> SettingsMainController -> SettingsSearchController```

And no matter how much I type or delete in the query, the push only seems to occur when you hit the search icon. I've confirmed this with a breakpoint on the push call, and a watch variable on ```router.backstack.```

Maybe @lmj0011 can explain this better, but my understanding is that this structure of callbacks is simply what is necessary to set up the search, and it does not result in a runaway collection of new controllers.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
mpm11011: As it turns out, this wasn't an appropriate solution, as when you click the search icon in GBoard after entering a query, you get sent back to the main Settings page instead of being shown your search results.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsMainController.kt
lmj0011: resolved https://github.com/mpm11011/tachiyomi/pull/5

app/src/main/java/eu/kanade/tachiyomi/ui/setting/settingssearch/SettingsSearchController.kt
lmj0011: resolved https://github.com/mpm11011/tachiyomi/pull/5

arkon:
Sigh, forgot to make sure the setting was for squashing the PR first yet again.

Thanks for making this feature though!

mpm11011:
Thank you @lmj0011 and @arkon for your help!

PR #3695, opened 2020-08-19
Searchable Settings

### LINKED ISSUE #3520: [Feature Request] Searchable Settings

lmj0011:
closes #3520
see: https://github.com/inorichi/tachiyomi/pull/3683#issuecomment-676473467

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -202,7 +202,6 @@ dependencies {
     // Preferences
     implementation 'com.github.tfcporciuncula:flow-preferences:1.3.0'
     implementation 'com.github.ByteHamster:SearchPreference:v1.0.3'
-    implementation "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
 
     // Model View Presenter
     final nucleus_version = '3.0.0'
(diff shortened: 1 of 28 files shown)


arkon:
Closing in favour of #3683

PR #3743, opened 2020-09-01
Change sources sort to case-insensitive

scb261:
Extensions are sorted by something like `eu.kanade.tachiyomi.extension.all.mangadex` so it's already case-insensitive, but sources sorted by it's normal name like `MangaDex` so it's case-sensitive. I just added `toLowerCase()` in few places to make sort consistent for sources and extensions. In examples you can see that `Mangabat` was below `MangaTX` but now it's above `MangaBob`.

<details>
  <summary>How it was before</summary>

before

</details>

<details>
  <summary>How it is in this PR</summary>
  
after

</details>

I did two changes in `GlobalSearchPresenter.kt`, I suppose that one is used when global search is initialized and second one to display results?


arkon:
Thanks!

Issue #3392, opened 2020-06-27
[Feature Request] Always show bottom navigation


pervertkirito:
### User Problem
it irritates me when this happen.

[![video](https://i.imgur.com/CuYEMGf.jpg)](https://streamable.com/9z1hzn)

Bottom navigation won't show up if your library not big enough. Check video ☝️☝️

### Requirements
Add an option for **always show bottom navigation** on settings.
( I actually like the old bottom navigation style :P )

lmj0011:
>Bottom navigation won't show up if your library not big enough.

This seems to only happens when you have the Library display set to show 3 items per row in portrait mode and have exactly 2 rows in a particular category. Changing the "items per row" count to any other number then it works as expected.

scb261:
Tachiyomi version: Stable 0.9.2

Same problem when a category is empty (for example, filtered by unread when all mangas in the category are read). 

~~Also I just don't like when it disappears at all since it takes few seconds for me to recall direction of scroll to made it appear (in the end I just try one way and if it doesn't work, then another one), maybe I just not used to it yet. Not sure how many users share my opinion, but the option _always show bottom navigation_ would be nice.~~ And while writing this comment I understood my problem - **if you scroll up only a little bit, upper bar is shown while the menu isn't and that's what made me confused**. It's better to make them appear together.

PR #3758, opened 2020-09-06
Add option to hide bottom bar when scrolling

scb261:
#3392, but I'm not sure if it really closes it. Some users including myself just don't like when bottom navbar is hidden, but in #3392 mentioned problem that the navbar won't show up in particular usecase, and the problem is still there if the option is set to default.


app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
```diff
@@ -334,6 +341,7 @@ class MainActivity : BaseActivity<MainActivityBinding>() {
         if (to is RootController) {
             // Always show bottom nav again when returning to a RootController
             showBottomNav(visible = true, collapse = from !is RootController)
+            setBottomBarBehaviorOnScroll()
```
scb261, line 344: I'm not sure about this one. Right now it is called even when it isn't needed, but adding 
```
                onChange {
                    activity?.recreate()
                    true
                }
```
recreates whole activity and I'm not sure that it's right decision here.

arkon: Maybe just call it on the pref flow on each change?

scb261: I'm pretty new to Android development and Kotlin. How can I get current instance of `MainActivity` inside of `SettingsGeneralController`?

jobobby04: SettingsController should have the activity variable internally(as all controllers do)

arkon: I ended up making the change myself after merging this.

PR #3762, opened 2020-09-06
WebtoonAdapter don't like small images

### LINKED ISSUE #3749: [Bug] Certain chapters get stuck when going backwards with continuous scroll
---

## Device information
* Tachiyomi version: 0.10.4
* Android version: 10
* Device: OnePlus 6 64gb

## Steps to reproduce
1. Change reader setting to continuous scroll
2. In this manga: Please Go Home, Akutsu-san! (Mangakalot or mangadex, haven't checked others but happened with both)
3. Chapters which have both the images I attached at the end of them, from 8 to 15 more or less, reproduced with other forks as well
4. Enter a chapter 10 for example, scroll up to return to chapter 9 and then it gets stuck

### Expected behavior
This should happen: return to previous chapters normally

### Actual behavior
This happened instead: heck stuck on last page or two of previous chapter

## Other details
Additional details and attachments.
Happens with chapters that have both these images, with the violet evergarden being the last one.
Please Go Home, Akutsu-san! - Vol 1 Chapter 12 - 7
Please Go Home, Akutsu-san! - Vol 1 Chapter 12 - 6

Link to video of it happening: https://drive.google.com/file/d/1uvlBwAtRERpaklvijxIx674KKGO-5BWQ/view?usp=drivesdk

ghostbear:
closes #3749 

### Where does this happen
This happens in `Reading mode` -> `Webtoon` and `Continous vertical`

### Why does it happen
From observation, it seems to happen when there are ~~two~~ small pages in sequence to one each other

### How was this fixed
~~This was fixed by making `WebtoonAdapter` take one more page in both directions.
Why both directions because my assumption is that if the next chapter also has two small images it will also cause this bug~~
This was fixed by adding a `SwipeRefreshLayout` which causes a forced chapter change

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
@@ -531,9 +531,9 @@ class ReaderActivity : BaseRxActivity<ReaderActivityBinding, ReaderPresenter>() 
     /**
      * Called every time images are to small to trigger a chapter change
      **/
-    fun forceChapterChange(currentChapter: ReaderChapter?, any: Any) {
-        Toast.makeText(this, "Warning! The images seams to be small, are you using the right reading mode? Reader will not operate properly", Toast.LENGTH_LONG).show()
-        presenter.forceChapterChange(currentChapter, any)
+    fun forceChapterChange(page: ReaderPage) {
+        Toast.makeText(this, getString(R.string.reader_manual_chapter_change_warning), Toast.LENGTH_LONG).show()
+        presenter.forceChapterChange(page)
     }
 
     /**
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
@@ -20,6 +20,7 @@ import android.view.WindowManager
 import android.view.animation.Animation
 import android.view.animation.AnimationUtils
 import android.widget.SeekBar
+import android.widget.Toast
 import androidx.core.view.ViewCompat
 import androidx.core.view.isVisible
 import androidx.core.view.setPadding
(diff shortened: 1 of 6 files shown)


scb261:
Now it's stuck if there is 3 small pages at the end of previous chapter, but I guess I just zoomed out too much (still could happen without zooming out if source serves small images). 

Also it loads images in next chapter just fine (I created local manga with really small images and it handles it right when scrolling to next chapter even when there are ~20 pages that fits into screen at the same time), I guess the app handles next and previous chapters differently.

ghostbear:
I couldn't find where the trigger for when the previous chapter becomes the current chapter.
But I found that by making the chapter transition match parent height it always triggers this change

ghostbear:
I have wrapped the `RecyclerView` in a ``SwipeRefreshLayout` so if a user hits the top of the `RecyclerView` and continues to try scrolling to the previous chapter it will trigger a chapter change. It will also notify the user that images are small and the experience will not be as good. Also when the forced chapter change happens it sometimes jumps to a page in the chapter and not the last page

ghostbear:
I will close this due to the poor way of fixing it.

**For the one who wants to look into the problem**
The problem occurs when there are small pages in sequence to one another.

If I understand the code correctly this is where the problem occurs
```kotlin
recycler.addOnScrollListener(
  object : RecyclerView.OnScrollListener() {
    override fun onScrolled(recyclerView: RecyclerView, dx: Int, dy: Int) {
      val position = layoutManager.findLastEndVisibleItemPosition() // <-- This line to be precise
      val item = adapter.items.getOrNull(position)
      val allowPreload = checkAllowPreload(item as? ReaderPage)
      if (item != null && currentPage != item) {
        currentPage = item
        when (item) { // <-- Results in this only doing onTransitionSelected
          is ReaderPage -> onPageSelected(item, allowPreload) // <-- This changes the chapter
          is ChapterTransition -> onTransitionSelected(item)
        }
      }
      ...
    }
  }
)
```

So when there are two small images and a chapter transition in sequence scrolling backward the item that `layoutManager.findLastEndVisibleItemPosition()` returns is the chapter transition.

A way of fixing this would be to take another image and hope it is enough the trigger the chapter change, in the WebtoonAdapter. But if the chapter only contains images with a small height it will still get stuck.

A way to force a chapter change is with a SwipeRefreshLayout this is janky and is not good for the reading experience

PR #3796, opened 2020-09-13
Add ability to choose what to remove for selected mangas

### LINKED ISSUE #3334: [Feature Request] Add a "delete downloaded chapters only" when selecting multiple titles in the library section
I want to delete some downloaded content but want to keep the manga itself in my library. I find it really tedious to individually open a title to delete the downloaded chapters.

scb261:
Closes #3334

| before | after |
| --- | --- |
| before | after |


app/src/main/res/values/strings.xml
```diff
@@ -443,8 +443,10 @@
     <string name="library_search_hint">Title or author…</string>
     <string name="updating_category">Updating category</string>
     <string name="local_source_badge">Local</string>
-    <string name="confirm_delete_manga">Are you sure you want to remove selected manga?</string>
-    <string name="also_delete_chapters">Also delete downloaded chapters</string>
+    <string-array name="delete_selected_mangas">
+        <item>Manga from library</item>
+        <item>Downloaded chapters</item>
```
arkon, line 448: Put regular strings in this file and have a string array in `arrays.xml`.

joseph619:
What's the benefit of this? Now we have to give one more extra click than which is worse.

scb261:
Benefit is that user can clean up space easily without entering each manga or browsing file manager where manga doesn't have categories and no covers are displayed, it was asked in the issue and several times on discord. I can make first option enabled by default or remember what user selected before, what is better?

CarlosEsco:
I feel like the remove downloads only should just be in an option inside the manga.  This completely changes the long press action now.  Instead of removing manga it can be used for a download cleaner

joseph619:
I agree with Carlos. 

@scb261 Make the first option enabled by default at least, as that's the primary and original function for the delete button.

arkon:
Yeah, I'd go with making the first option defaulted to checked.

I'm not against changing the action to be able to be used as a download cleaner.

app/src/main/res/values/strings.xml
scb261: Done

scb261:
I changed to radio buttons to have one click less when user wants to remove only downloaded chapters. If you prefer checkboxes, I'll revert it.

<details>
  <summary>How it looks now</summary>
  
Screenshot_2020-09-14-23-35-34
  
</details>

arkon:
I personally preferred the checkboxes.

scb261:
Reverted

Issue #3772, opened 2020-09-09
[Feature Request] Option to turn off History.


saieeshb:
## Why/User Benefit/User Problem
This would be useful for users who don't want their history to be tracked. Prevents repeatedly deleting manga in 

## What/Requirements

The option to turn just tracking of history off.
  By this, I mean the manga you have recently read would not show up in the history tab, or show no history at all. This does not mean resetting the progress in the manga itself.

Thank you!

PR #3855, opened 2020-09-27
[Feature Request] Option to turn off History - #3772

Platiplus:
Implementing feature to turn on/off reading history of chapters

I'm not really sure where to put it (on which category of preferences), but I'm happy to help and fix any issues with this pull request.

Resolves issue #3772

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
@@ -14,7 +14,16 @@ import eu.kanade.tachiyomi.ui.category.CategoryController
 import eu.kanade.tachiyomi.ui.download.DownloadController
 import eu.kanade.tachiyomi.ui.setting.SettingsController
 import eu.kanade.tachiyomi.ui.setting.SettingsMainController
-import eu.kanade.tachiyomi.util.preference.*
+import eu.kanade.tachiyomi.util.preference.add
+import eu.kanade.tachiyomi.util.preference.iconRes
+import eu.kanade.tachiyomi.util.preference.iconTint
+import eu.kanade.tachiyomi.util.preference.onClick
+import eu.kanade.tachiyomi.util.preference.preference
+import eu.kanade.tachiyomi.util.preference.preferenceCategory
+import eu.kanade.tachiyomi.util.preference.summaryRes
+import eu.kanade.tachiyomi.util.preference.switchPreference
+import eu.kanade.tachiyomi.util.preference.titleRes
+import eu.kanade.tachiyomi.util.preference.defaultValue
 import eu.kanade.tachiyomi.util.system.getResourceColor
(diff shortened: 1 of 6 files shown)


joseph619:
Just a suggestion, but perhaps this setting could be put in More and be enabled with a toggle called Incognito mode? Then a description would say that it disables recording history in History tab

Platiplus:
> Just a suggestion, but perhaps this setting could be put in **More** and be enabled with a toggle called **Incognito mode**? Then a description would say that it disables recording history in History tab

Got it done! You can check it now on the current commit.

jobobby04:
How about putting it in More instead? Like downloaded only is
eu/kanade/tachiyomi/ui/more/MoreController.kt

Platiplus:
> How about putting it in More instead? Like downloaded only is
> eu/kanade/tachiyomi/ui/more/MoreController.kt

Done!

app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
```diff
@@ -14,15 +14,7 @@ import eu.kanade.tachiyomi.ui.category.CategoryController
...
-import eu.kanade.tachiyomi.util.preference.add
-import eu.kanade.tachiyomi.util.preference.iconRes
-import eu.kanade.tachiyomi.util.preference.iconTint
-import eu.kanade.tachiyomi.util.preference.onClick
-import eu.kanade.tachiyomi.util.preference.preference
-import eu.kanade.tachiyomi.util.preference.preferenceCategory
-import eu.kanade.tachiyomi.util.preference.summaryRes
-import eu.kanade.tachiyomi.util.preference.switchPreference
-import eu.kanade.tachiyomi.util.preference.titleRes
+import eu.kanade.tachiyomi.util.preference.*
```
jobobby04, line 17: Probably a unintended change, Tachi's linter doesn't allow wildcard imports

app/src/main/res/values/strings.xml
```diff
@@ -318,6 +318,9 @@
     <string name="webtoon_side_padding_15">15%</string>
     <string name="webtoon_side_padding_20">20%</string>
     <string name="webtoon_side_padding_25">25%</string>
+    <string name="more_category_header">More</string>
```
jobobby04, line 321: Probably also remove this

app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
Platiplus: Yup, totally unintended. It was the auto import. Fixed.

app/src/main/res/values/strings.xml
Platiplus: Forgot to do that! Fixed!

arkon:
Thanks for the contribution!

PR #3858, opened 2020-09-28
Darken the description colors

Soitora:
Darkens the description color on Manga Info page as well as the Updates view.
It closely matches the color of "DOWNLOADED", although that one was not a regular android attribute so couldn't figure out how to make `getResourceColor` to work for those.

`0.62f` value wasn't gotten organically from other colors, I tried finding something similiar but I thought that (`100`-`38`=`62`), `100` being unread, `38` being read, `62` now being sub-unread.

Feel free to grill me on this one but I felt kannaKMS when looking at full white both texts... vampySmug

| Original | Edited |
| -------- | ------ |
| Original manga info | Edited manga info |
| Original update list | Edited update list |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChapterHolder.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/chapter/ChapterHolder.kt
@@ -32,13 +32,19 @@ class ChapterHolder(
         }
 
         // Set correct text color
-        val chapterColor = when {
+        val chapterTitleColor = when {
             chapter.read -> adapter.readColor
             chapter.bookmark -> adapter.bookmarkedColor
             else -> adapter.unreadColor
         }
-        chapter_title.setTextColor(chapterColor)
-        chapter_description.setTextColor(chapterColor)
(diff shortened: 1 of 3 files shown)


arkon:
I'm a bit iffy on this. It makes sense for the chapters list since the second line is additional info, but the second line seems like important info in Updates.

Soitora:
I get that it's important in Updates but keeping them the same color but having different meanings and sizes feels odd

Soitora:
Even though I disagree here I've tweaked the PR to not include the changes for the Updates view.

vengiare:
It looks so weird having it dark

image
Also these colors are too close for comfort

vengiare:
Screenshot_20201004-201259
I haven't checked but does the secondary color use the same color as this

Soitora:
It does not use the same color, _could_ be tweaked to use the same one.
The current reading progress one could in general be overhauled but I think the contrast between the two colors are fine

vengiare:
I think it should use the secondary color from extensions list then. I believe it's slightly lighter than the one here and also of course for consistency.

Soitora:
I switched it now, still looks pleasing to the eye 👍

PR #3862, opened 2020-09-29
Fix color filter sheet content alignment

### LINKED ISSUE #3860: Color filter sheet content alignment
Preview

ghostbear:
Fixes #3860 
Before:
| Blueprint | Preview |
| --- | --- |
| image | image |

After:
| Blueprint | Preview |
| --- | --- |
| image | image |


arkon:
Ideally should use a [Barrier](https://developer.android.com/reference/androidx/constraintlayout/widget/Barrier) instead.

ghostbear:
Changed it to use a barrier instead
| Blueprint | Preview |
| --- | --- |
| image | image |

PR #3869, opened 2020-10-01
Viewer navigation

### LINKED ISSUE #2361: [Feature Request] Left screen border to also proceed to next page
Currently right edge goes to the next page and left edge goes to the previous page. This makes it impossible to hold a tablet with the left hand on the left side and still read.

### What/Requirements
Both left and right edge taps should go to the next page, and a bottom tap or a slide can be used to change page back. This is fine because going to the next page is many times more common than going to a previous page.

I would do this myself if someone can point out where in the code the edge reader behaviors are.

### LINKED ISSUE #3678: [Feature Request] Tap left or right side of screen to go to next page
Currently you're only able to have it so that tapping the right side of the screen will flip to next page while tapping left side will flip to previous (or inverted) but you can't have it so that both sides of the screen will go next. I'd prefer it that way (while I'd swipe to go previous). 

---

More user customization.

## What/Requirements
Should be in the options.

ghostbear:
Closes #2152 
Closes #2361, and closes #3678 with Edge navigation

**What differs from the other PR**
- Default navigation behaves more like current reader navigation
- Adds the ability to invert viewer navigation
- Adds additional Edge navigation

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceKeys.kt
@@ -65,6 +65,10 @@ object PreferenceKeys {
 
     const val readWithVolumeKeysInverted = "reader_volume_keys_inverted"
 
+    const val navigationModePager = "reader_navigation_mode_pager"
+
+    const val navigationModeWebtoon = "reader_navigation_mode_webtoon"
+
     const val webtoonSidePadding = "webtoon_side_padding"
 
     const val portraitColumns = "pref_library_columns_portrait_key"
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceValues.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferenceValues.kt
@@ -35,7 +35,10 @@ object PreferenceValues {
         NONE,
         HORIZONTAL,
         VERTICAL,
-        BOTH
+        BOTH;
+
+        fun invertHorizontal() = this == HORIZONTAL || this == BOTH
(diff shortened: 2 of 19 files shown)


Charuru:
Does this fix the problem here? https://github.com/tachiyomiorg/tachiyomi/issues/2361 Or not?

ghostbear:
@Charuru this doesn't specifically fix your problem.

But if I understand your issue correctly L Navigation should partially fix your problem
```
L Navigation tap zones
+---+---+---+
| N | N | N |   P: Previous
+---+---+---+
| N | M | P |   M: Menu
+---+---+---+
| P | P | P |   N: Next
+---+---+---+
```

Tap zone can be inverted by the axis also
```
L Navigation tap zones (horizontal invert)
+---+---+---+
| N | N | N |   P: Previous
+---+---+---+
| P | M | N |   M: Menu
+---+---+---+
| P | P | P |   N: Next
+---+---+---+
```

---

But if I understand your issue correctly you would want it more like this
```
Tap zones
+---+---+---+---+
| N | M | M | N |   P: Previous
+---+---+---+---+
| N | M | M | N |   M: Menu
+---+---+---+---+
| N | P | P | N |   N: Next
+---+---+---+---+
```

Then both left and right edge taps should go to the next page, and bottom-center tap is used to change the page back. If this is more of the expected behavior you want I could implement it

Charuru:
Yes your last diagram is exactly how I would like it, it would be great if that were an option. Thanks a lot.

arkon:
This is great stuff, thanks for expanding on this, @ghostbear !

hXtreme:
Wow! So happy to see this finally implemented.

You da boss @ghostbear !

PR #3870, opened 2020-10-01
Remove from queue after read

### LINKED ISSUE #2077: Remove from queue after read
r1512

So if you have a chapter in queue (not downloaded), and you go ahead and read that chapter without it downloaded.
With the option to delete chapter after read, it doesn't clear the queued chapter.
Long select the chapter and press delete does remove it from queue.
So my guess is the two delete action isn't running the same function.

Thanks.

ghostbear:
Closes #2077

**What it does**
If the setting is enabled it will remove the download from the queue for the chapter when the chapter being marked as read

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadController.kt
@@ -128,12 +128,6 @@ class SettingsDownloadController : SettingsController() {
                     }
                     .launchIn(scope)
             }
-            switchPreference {
-                key = Keys.deleteFromQueue
-                titleRes = R.string.pref_delete_from_queue
-                summary = resources?.getString(R.string.pref_delete_from_queue_summary)
-                defaultValue = false
-            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsDownloadController.kt
```diff
@@ -92,6 +92,11 @@ class SettingsDownloadController : SettingsController() {
                 titleRes = R.string.pref_remove_bookmarked_chapters
                 defaultValue = false
             }
+            switchPreference {
+                key = Keys.deleteFromQueue
+                titleRes = R.string.pref_delete_from_queue
+                defaultValue = false
+            }
```
arkon, line 99: Should probably be in the "Download new chapters" section.

ghostbear:
image
"From queue after marked as read" didn't make sense under Download new chapter.
So I changed it to "Delete from queue after marked as read" which became to long.
Which resulted in splitting it to "Delete from queue" with a summary of "After being marked as read"

arkon:
Thinking about it, does it even really need to be a preference? I think it'd be fine if it just removes it from the queue when marked as read. If the user still wants it to be downloaded, they can trigger the download again.

ghostbear:
Yeah, I really don't need to be a preference either. It was more if someone wanted to keep the download and let it be removed by "After reading". But sure I could remove the preference and when someone complains about it add it back

PR #3871, opened 2020-10-01
Ability to delete reading history

### LINKED ISSUE #1799: [Feature Request] Option to bulk clear recently read
**App version:**
0.8.0
**Issue/Request:**
An option to bulk clear recently read manga in the Recently read section

**Other details:**

ghostbear:
Closes #1799

**What it does**
It adds the ability to delete all reading history with one button

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -712,5 +712,9 @@
     <string name="tapping_inverted_horizontal">Horizontal</string>
     <string name="tapping_inverted_vertical">Vertical</string>
     <string name="tapping_inverted_both">Both</string>
+    <string name="clear_history_completed">History deleted</string>
+    <string name="clear_history_confirmation">Are you sure? All history will be lost</string>
+    <string name="pref_clear_history">Clear history</string>
+    <string name="pref_clear_history_summary">Delete history for all manga</string>
 
 </resources>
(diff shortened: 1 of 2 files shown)


joseph619:
Could you also add an undo option for this? If some user does this by mistake/curiosity, they would lose all history if they didn't have auto backup enabled.

ghostbear:
> Could you also add an undo option for this? If some user does this by mistake/curiosity, they would lose all history if they didn't have auto backup enabled.

There is already a warning which warns users that the history data will be deleted before they confirm the deletion. And from how the restore backup works they could use it to restore history

app/src/main/res/values/strings.xml
```diff
@@ -712,5 +712,9 @@
     <string name="tapping_inverted_horizontal">Horizontal</string>
     <string name="tapping_inverted_vertical">Vertical</string>
     <string name="tapping_inverted_both">Both</string>
+    <string name="clear_history_completed">History deleted</string>
```
arkon, line 715: Can you group these with the other Advanced section strings, e.g. `battery_optimization_setting_activity_not_found`?

app/src/main/res/values/strings.xml
```diff
@@ -712,5 +712,9 @@
...
+    <string name="clear_history_confirmation">Are you sure? All history will be lost</string>
```
arkon, line 716: Nitpick: add a period at the end.

app/src/main/res/values/strings.xml
```diff
@@ -712,5 +712,9 @@
...
+    <string name="pref_clear_history">Clear history</string>
+    <string name="pref_clear_history_summary">Delete history for all manga</string>
```
arkon, line 718: "reading history"?

arkon (changes requested):

PR #3884, opened 2020-10-03
Add manga count to the library header

### LINKED ISSUE #2937: [Feature Request] Show Count/Number of Favourites (Manga/Webtoon)
It helps to keep count of favourite manga which might be helpful while migrating etc

### What/Requirements
In My Library (Total Count) and for each Category (category count) we can show count in Badges (or Text)

TheUnlocked:
Resolves #2937

Adds a number indicating the number of manga in each category/in the library:

image

image

I don't have images for every situation because that would be a lot of images, but I have tested this with basically every combination of viewing categories as tabs and having/not having the default category and having everything in one category, etc., and it seems to work for everything.

As far as the code goes, I don't know how much of this code is actually idiomatic to Kotlin and Tachiyomi, but I tried to comment it. :P


joseph619:
Can you make it a checkbox under Display in Library filter sheet? I wouldn't want to see the manga by default count unless I selected everything.

TheUnlocked:
Done. 
image

joseph619:
Thanks!

joseph619:
Is it possible to make the number of manga count under Library header be displayed as a secondary text? The parenthesis looks bad and out of place.

For example, see this: https://github.com/inorichi/tachiyomi/pull/3972

It could say 252 manga/favorites under Library header on the toolbar

arkon:
That'd look worse IMO.

vengiare:
image

I think the problem with these is it looks more like an unread indicator than a manga count indicator

scb261:
Maybe like this? Filters could affect this as well in this case

TheUnlocked:
> Maybe like this? Filters could affect this as well in this case
> 

I like that idea, though I'm fairly busy right now so I probably won't have time to work on it any time soon. If someone else wants to experiment with that layout but is worried about duplicated work, just go ahead.

PR #3885, opened 2020-10-03
Update issue_closer.yml

sainAk:
Limit the issue_closer workflow to run only when the issue is opened, edited or reopened


arkon:
Good idea, thanks!

Could also do it for https://github.com/inorichi/tachiyomi-extensions/blob/master/.github/workflows/issue_closer.yml

PR #3891, opened 2020-10-04
Workflow improvements

sainAk:
Added some steps in build_check for faster builds in github runner


sainAk:
btw if we use NDK version 21.3.6528147 we could reduce the build time further by a minute

sainAk:
Should I add a ci skip condition check in the workflow?

arkon:
> Should I add a ci skip condition check in the workflow?

Could add that in a separate PR.

sainAk:
ok merge this pr but what about using NDK v21.3.6528147

arkon:
I'm actually not sure what dictates the NDK version being used. I assume the current is just some default. Do you know?

sainAk:
the ndk version can be defined in the app:build.gradle just like build tools but if it is not specified it will use the default version set by [agp](https://developer.android.com/studio/projects/install-ndk#default-ndk-per-agp) in theis case v21.0.6113669 but v21.3.6528147 is the current stable version thats why this version is preloaded in the runner images

arkon:
Cool. Feel free to submit a PR to set that first, then you can remove the caching from this PR.

sainAk:
ok

sainAk:
one last question is travisCI still needed?

checksum.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
arkon, line 1: Could you put this into the `.github` directory as well? You could maybe put this and `ci-gradle.properties` together into a subdirectory.

sainAk: sure

arkon:
> one last question is travisCI still needed?

For now, yes. I do intend on replacing the `master` branch tag builds with GitHub Actions too though. I also have [another PR](https://github.com/inorichi/tachiyomi/pull/3873) open to eventually remove the preview build stuff from Travis, but I'm letting it run for at least another week so that there's a better chance for people running old builds to pick up a newer build that checks for preview builds from GitHub instead of the old server.

sainAk:
the PR is ready to be merged I'll create a seperate pr for ci skip later

.github/runner-files/checksum.sh
```diff
@@ -0,0 +1,23 @@
...
+FILES=()
+while read -r -d ''; do
+  FILES+=("$REPLY")
+done < <(find . -type f \( -name "build.gradle*" -o -name "Dependencies.kt" -o -name "gradle-wrapper.properties" \) -print0)
+
+# Loop through files and append MD5 to result file
+for FILE in ${FILES[@]}; do
+  echo $(checksum_file $FILE) >> $RESULT_FILE
+done
+# Now sort the file so that it is
```
arkon, line 22: Did you forget to finish this sentence?

.github/workflows/build_check.yml
```diff
@@ -16,8 +16,21 @@ jobs:
...
           java-version: 1.8
-      - name: Install NDK
-        run: sudo ${ANDROID_HOME}/tools/bin/sdkmanager --install "ndk;21.0.6113669" &> /dev/null
+
+      - name: Copy CI gradle.properties
+        run: mkdir -p ~/.gradle ; cp .github/runner-files/ci-gradle.properties ~/.gradle/gradle.properties
+
+      - name: Generate Gradle cache key
+        run: ./.github/runner-files/checksum.sh checksum.txt
+      - name: Cache gradle dependencies
```
arkon, line 25: Just nitpicking, but can you capitalize "Gradle" here?

arkon (commented):
Super minor comments, otherwise LGTM.

.github/workflows/build_check.yml
sainAk: ok

sainAk:
https://docs.github.com/en/free-pro-team@latest/actions/guides/storing-workflow-data-as-artifacts
the generated apk can be stores as artifacts if you want

.github/runner-files/checksum.sh
sainAk: i used this script from here https://github.com/chrisbanes/tivi/blob/main/checksum.sh
i think it should be "Now sort the file so that it is in same order"

arkon:
That's fine, I'm not really interested in surfacing APKs for every build.

We do weekly Preview builds via [this repo](https://github.com/tachiyomiorg/android-app-preview) so that I can dogfood changes for a bit and to avoid spamming people with constant updates.

I'll use [TachiyomiSY's build script](https://github.com/jobobby04/TachiyomiSY/blob/master/.github/workflows/TachiyomiSY-Release-Builder.yml) as a reference for the release builds.

PR #3902, opened 2020-10-05
skip ci in commit message condition check

sainAk:


arkon:
Can the jobs have more descriptive names?

sainAk:
yeah what name do you want

sainAk:
@arkon about your commit e637f22
it will also skip builds of other branches on a fork repo like mine

arkon:
> it will also skip builds of other branches on a fork repo like mine

Yeah, my bad. It's an unintentional side effect of me not wanting it to run for other branches in this main repo, but I'm not sure if there's a good workaround for that.

Issue #3896, opened 2020-10-05
[Bug] <Recent Chapter History>


Toastypoptart:
## Device information
* Tachiyomi version: Preview r2763 (4cbe4997)
* Android version: 10
* Device: Samsung Galaxy S10 plus

## Steps to reproduce
1. Exits out of Manga I was reading
2. In History tab

### Expected behavior
Should say chapter 42 (most recent chapter) that I just read.

### Actual behavior
Says chapter 41 that I've read last time.

## Other details
On version r2763 of Tachiyomi my history tab doesn't update to the most recent chapter that I've read unless I click off the tab and click back on (example: clicks on browse and then clicks on history) or close the app and reopen it.

WildOrangutan:
I will try to solve this mystery! 😊

PR #3912, opened 2020-10-07
Some history fixes

WildOrangutan:
I've tried to address  #3896, but this project is still on the next level for me, so I'm giving up atm 😅. 

I think refreshing was missing, when history view is resumed, so I added that, but I'm not sure it should be on `onActivityResumed`. Even tough I've added refreshing, history now refreshes on resume, but it for some reason does not contain most recent history values. I did some experiments, like for example using `take(7)` (`HistoryPresenter.updateList()`), and most recent value eventually gets there.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
@@ -151,11 +151,7 @@ class ReaderPresenter(
     override fun onDestroy() {
         super.onDestroy()
         val currentChapters = viewerChaptersRelay.value
-        if (currentChapters != null) {
-            currentChapters.unref()
-            saveChapterProgress(currentChapters.currChapter)
-            saveChapterHistory(currentChapters.currChapter)
-        }
+        currentChapters?.unref()
     }
 
     /**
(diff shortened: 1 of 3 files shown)


scb261:
If this will make you interested again, here is commit that probably broke it https://github.com/inorichi/tachiyomi/commit/9d2adcd512c28872c9e958e9fcdcbccbd11b3b35

Also there is a fork that had infinite history before original repository and the same problem appeared after this commit https://github.com/jobobby04/TachiyomiSY/commit/9c76f1fd8fecd68d501b538aeb7312fba6338637

WildOrangutan:
I will check it out and see if that gives me any more ideas, thanks.

WildOrangutan:
My best guess, why this worked before 9d2adcd, is because `HistoryPresenter` permanently subscribed to all database changes. After it, only "load more" is permanently subscribed.

My fix with adding another `updateList()` didn't work, because `ReaderPresenter.onDestroy()` is called after `HistoryController.onResume()`.

With f0fa46e this should work now. I don't completely understand the framework you're using, but as far as I know, data should be saved to DB `onPause`, so I think moving saving to `ReaderPresenter.onSave` is correct decision, since it's called around that time.

arkon:
Thanks for trying, but this is superseded by #3921.

WildOrangutan:
No worries. Just note that fix in d90c44d795b3aa74f602 is still valid, even tough it's not related to #3896.

PR #3915, opened 2020-10-08
Add support with Android 5.0

lostb053:
My request is that support with lower Android versions should be added.
Well this might seem quite unnecessary but many ppl still use old phones who are interested in manga
I too am using one with Android 5.0 and the app wont support tho i like the app
Sorry if i am not providing enough info or placing this at won't place as i am new to this but this is my request


scb261:
1. If you want request something, you need to open an issue, not pull request.
2. Android 5 still is supported, I am personally using the app on Android 5.
3. If Android 5 will become unsupported (just like it was done to Android 4 in past), then it means that there were reasons for this and it won't be added back.

lostb053:
Ohh
Sorry for 2 things 
1. For pulling request while i should have opened issue
2. Well i was found either thing in wrong place
I was supposed to do this in tachiyomij2k's github acc. 
So i apologize fir the mistakes

joseph619:
Please don't do this anywhere. You will just annoy the repo maintainers. Support for older Android versions are dropped for a good reason, you should consider buying a new phone ideally.

PR #3928, opened 2020-10-11
Fix missing chapter warning when chapter number is not recognized

ghostbear:
mhrj pointed this out on Discord, [link to message](https://discordapp.com/channels/349436576037732353/566585518171881493/764861993009676288)
| Before | After |
|---|---|
| Screenshot_1602430970 | Screenshot_1602431650 |

I found that this chapter was recognized as 0. To fix this I added some RegEx to see if the chapter name contains any numbers.
If it doesn't contain any number it returns false. Then it uses isChapterRecognized to see if the number was recognized as a chapter number if it was not then it returns false. Else it continues with the function.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/MissingChapters.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/MissingChapters.kt
@@ -16,6 +16,7 @@ fun hasMissingChapters(higher: Chapter?, lower: Chapter?): Boolean {
 }
 
 fun hasMissingChapters(higherChapterNumber: Float, lowerChapterNumber: Float): Boolean {
+    if (higherChapterNumber == 0f || lowerChapterNumber == 0f) return false
     return floor(higherChapterNumber) - floor(lowerChapterNumber) - 1f > 0f
 }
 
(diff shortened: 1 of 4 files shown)


arkon:
I'm pretty sure it's _supposed_ to be -1 rather than 0 from `ChapterRecognition`.

ghostbear:
Well it is recognized as 0
image

jobobby04:
Its because of this https://github.com/inorichi/tachiyomi-extensions/blob/master/src/all/mangaplus/src/eu/kanade/tachiyomi/extension/all/mangaplus/MangaPlus.kt#L241

arkon:
Ah yeah, that'd explain it.

The automatic recognition stuff does get `-1`:

```
    @Test
    fun test_ex() {
        createManga("Monster #8")

        createChapter("ex - Illustration Compilation: Monster Views")
        ChapterRecognition.parseChapterNumber(chapter, manga)
        assertThat(chapter.chapter_number).isEqualTo(-1f)
    }
```

The extension should probably be fixed.

ghostbear:
Ah, but I think a guard case should be added for when a chapter is not recognized

scb261:
While it was fixed for some sources in https://github.com/inorichi/tachiyomi-extensions/pull/4596, the problem is still there for sources that don't assign `chapter_number` themselves. Examples are [*MangaSee*](https://github.com/inorichi/tachiyomi-extensions/blob/dad3bf88a28751573f538d86ec69d38c76530e11/src/en/mangasee/src/eu/kanade/tachiyomi/extension/en/mangasee/Mangasee.kt#L244-L252) and *MangaLife*, manga *Kingdom*, transition between chapter *Gathered at you* and *Chapter 0412*. Not sure if this PR fixes it, so I'll just leave this info here.

PR #4029, opened 2020-11-12
Add support for S Pen actions

RikudouSage:
Adds support for S Pen Air Actions for Galaxy Note 10 and newer devices. This basically allows you to use the pen to do some gestures and respond with actions - each gesture is converted to a key event that is sent to the app.

Only one activity can handle Air Actions in the app, Reader seemed the most useful to me.

All actions are configurable in the settings, the `<preference />` xml tag is only the default which can be changed by user.

  

If there are any other actions that you think would make sense in a reader, let me know, I can implement them.

The actions:

- **Next page** - default air action is single button press because it's probably the most used action
- **Previous page** - default air action is double button press
- **Next chapter** - default air action is swipe right
- **Previous chapter** - default air action is swipe left
- **Back** - default air action is counter clockwise circle

Available air actions (let me know if you think other default action makes sense):

- Single click
- Double click
- Swipe left
- Swipe right
- Swipe up
- Swipe down
- Circle Counter-clock-wise 
- Circle Clock-wise 

Also the app actions can be restricted to only button actions (single click, double click) or only motion actions, but I think there's no point in restricting these actions.

The official documentation is available here: https://developer.samsung.com/galaxy-spen-remote/air-actions.html

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -60,7 +60,14 @@
         </activity>
         <activity
             android:name=".ui.reader.ReaderActivity"
-            android:launchMode="singleTask" />
+            android:launchMode="singleTask">
+            <intent-filter>
+                <action android:name="com.samsung.android.support.REMOTE_ACTION" />
+            </intent-filter>
+
+            <meta-data android:name="com.samsung.android.support.REMOTE_ACTION"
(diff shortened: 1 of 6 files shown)


arkon:
Can't test this since I don't have a device with an S Pen, but looks neat. Thanks!

PR #4114, opened 2020-12-12
Enable manga blocker

rawfh:
Because so many sites do not have a filter option, this commit will help extension archive a manga blocker in popular, latest and search result. Extension can set manga's title to "REMOVEDMANGA" and indicates APP to drop it.

One possible scenario is that users edit their blocklist keywords in extension preference panel, and the extension can compare these keywords with manga's genre or title; If any matches, rename its title to "REMOVEDMANGA".

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/online/ParsedHttpSource.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/online/ParsedHttpSource.kt
@@ -24,7 +24,7 @@ abstract class ParsedHttpSource : HttpSource() {
 
         val mangas = document.select(popularMangaSelector()).map { element ->
             popularMangaFromElement(element)
-        }
+        }.filter { manga -> manga.title != "REMOVEDMANGA" }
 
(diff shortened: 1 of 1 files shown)


arkon:
The extensions shouldn't be checking the main app's preferences. With that in mind, that means an extension can already refer to its own preferences, and as such can do its own filtering based on its own blocklist, which means that this change should occur in the extension itself.

rawfh:
https://github.com/tachiyomiorg/tachiyomi/blob/d363d205c3e9a71de422fb81ef26cd12f42bed7a/app/src/main/java/eu/kanade/tachiyomi/source/online/ParsedHttpSource.kt#L22-L34
I forget to mention that  this commit is mainly for those extensions using `mangaSelector`.
According to the implementation of `popularMangaParse`, all manga items that match `popularMangaSelector` would be displayed in the result, so I can't find another way to archive a blocker only in extensions.

rawfh:
The mangaElement is  a`jsoup` element, which supports both CSS and **jquery-like** selector syntax. So it's unnecessary to change codes in main app.

PR #4121, opened 2020-12-13
Added the feature to exclude certain categories from getting the download removed

### LINKED ISSUE #1412: Per-category Delete After Read (Read/Marked as Read) setting
IDK if this has already been posted but anyway, please take the time to read this. And oh, I'm no programmer and I know it's hard to be one. I just wanna post this in case someone takes an interest in making a fix for em. I really like the app so far. (Ignore the manga rock app... It's just that I just found and installed tachiyomi yesterday.)

App Version: v7.3

**Part A**
**Problem 1**:  Tracking doesn't track the chapters that are only marked as read manually (but not actually read).

**Problem 2**: If I only "Mark as Read" a chapter, or a certain amount of them, I have to actually open one chapter so that those marked as read will be tracked.

**Problem 3**: If I marked a certain amount of chapters as read, for example, I marked 40 chapters as read, then I opened page 1 of chapter 33, only 33 chapters are tracked as read but not 40. 

**Here, take a look.** 
**1)** Here's a manga I'm reading. Clearly, nothing is marked as read. 

**2)** Here's the initial status of this manga that I'm reading. 

**3)** Now, I'm marking all 38 chapters as read.

**4)** Clearly, all 38 chapters are now "MARKED" as read.

**5)** Still, the status of this manga isn't updated. No read chapters has been tracked.

**6)** Now I'm opening **page one of chapter 37**.

**7)** Only 37 chapters has been marked as read. 

**8)** Now, I'm opening chapter 38

**9)** Now 38 chapters has been tracked.

**Part B**
Suggestion: I hope that the "Remove after read" be changed in such way that it will only affect one category that I made. 

For example, I have two categories. One for the Mangas and one for the Hentai Mangas. I only want the "Remove after read" to take effect on the mangas since I want to keep my hentai mangas saved even though I've already finished reading them. 

Idea for fix: Add a "Categories to include in Remove after read". Something like what's in the "Categories to include in download".

tobinstultiens:
Created it so that you can exclude certain categories from getting the downloaded data removed.
The only point I don't know how to improve the quality on is the reuse of [this](https://github.com/inorichi/tachiyomi/compare/dev...tobinstultiens:dev#diff-18f98d9e2efa09a66cce8c3a072a875e3044cd7cc12bcf9c33ff07a7e5f044a6R454)

Fixes #1412


arkon:
Thanks for the PR, but I've made it such that the categories list can have excluded items now too.

PR #4190, opened 2020-12-25
Reader snackbar dismissible 

ghostbear:
**What it does**
This makes the snackbar that is shown when opening the reader dismissible

**Why**
It is quite annoying waiting for it to disappear when you quickly want to jump to a specific page


arkon:
Maybe the reader should use a CoordinatorLayout instead so swiping the snack bar works as intended?

ghostbear:
It is now swipeable instead

PR #4196, opened 2020-12-27
Direct user back to global search when coming from source search results

clarencecastillo:
## Setup:
Browse or Library -> Global Search -> Source Search Results
1. Trigger global text search (from Browse, Library or anywhere else).
2. Press source name to view more results from that source.
3. Press back button.

## Expected:
Browse or Library <- Global Search <- Source Search Results
It's more natural to expect the back button in step 3 to direct user back to global search.

## Actual:
Browse or Library <- Global Search <- Source Page <- Source Search Results
Back button clears the search text, directing user to source's default page.
This inconveniences the user to press back twice just to go back to the global search results.


arkon:
Thanks!

PR #4201, opened 2020-12-27
Fix multi-select phantom anchor bug in manga chapters and library

### LINKED ISSUE #3769: [Bug] multi select not working after some steps
## Device information
* Tachiyomi version: 0.10.4
* Android version: 9
* Device: any device can be reproduced

## Steps to reproduce
1. open chapter menu
2. select one chapter by long press
3. select another chapter by single tap, and cancel this selection of this chapter
4. try to select multiple chapter from step 2 selected to here by long press

### Expected behavior
multiple chapter selected (behavior like select one chapter and long press another chapter to select select multiple chapter between them)

### Actual behavior
multi select not be triggered, only chapters in step 1 and 4 was being selected

## Other details
nope

### LINKED ISSUE #4189: [Bug] multi select also not work in library after single select and cancel
## Device information
* Tachiyomi version: 0.10.7 Stable
* Android version: 9
* Device: Across any devices

## Steps to reproduce
1. Enter your library
2. Long press select one manga and enter the active select mode
3. single select one manga and cancel it
4. Try to do multi select by long press another manga

### Expected behavior
Multi select will be done.

### Actual behavior
Unable to multi select, only work as doing single click select.

clarencecastillo:
Closes #3769

Closes #4189

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryCategoryView.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryCategoryView.kt
@@ -28,6 +28,7 @@ import reactivecircus.flowbinding.recyclerview.scrollStateChanges
 import reactivecircus.flowbinding.swiperefreshlayout.refreshes
 import rx.subscriptions.CompositeSubscription
 import uy.kohesive.injekt.injectLazy
+import java.util.ArrayDeque
 
 /**
  * Fragment containing the library manga for a certain category.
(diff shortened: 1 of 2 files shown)


arkon:
Wouldn't this in theory be an issue in other places where we also do multi-selection, like the main library grid/list? If so, maybe we should try to handle those cases in the same PR.

clarencecastillo:
Implemented the same fix to the Library page, which applies it to all the view adapters.

I noticed that Updates page also implements multi-select but it's not "sticky" and doesn't allow subset selection via long press. Didn't change that since it would be adding a new feature, and not a fix.

Closes #4189

arkon:
Probably need to find a way to clean all of this up some day, but thanks for the fix!

PR #4233, opened 2021-01-02
Navigation settings and split invert tapping for webtoon and pager

ghostbear:
- Adds viewer navigation settings to the actual settings page
- Split invert tapping for webtoon and pager (because one can have different navigation for webtoon and pager)
- Use `<include />` for instead of a constraint group because not being able to use `isVisible` with the latter one (also `reader_settings_sheet.xml` is already hard to work with)


arkon:
Thanks!

Issue #4237, opened 2021-01-03
[Bug] Tap navigation in reader is inverted without invert tap enabled


dafillio:
---

## Device information
* Tachiyomi version: r2959
* Android version: 10
* Device: Mi Mix 2

## Steps to reproduce
1. Set reader from left to right (or other views, I think they're all messed up).
2. Turn invert tap setting to none
3. Tap on right side of screen in reader

### Expected behavior
You should advance to the next page

### Actual behavior
It goes back a page

## Other details
I believe this affects all reader view settings, I've tried webtoon, left to right, and right to left personally. It seems like the default behaviour is now inverted. This might also affect other types of navigation but I don't use them personally. This behaviour is the opposite of what was default as of the previous build version

ghostbear:
The way tapping works has been changed and the default for Right to Left and Left to Right is
```
+---+---+---+
| N | M | P |   P: Previous
+---+---+---+
| N | M | P |   M: Menu
+---+---+---+
| N | M | P |   N: Next
+---+---+---+
```
I was thinking if the user doesn't like to change to the next page on the left side they can invert it to the right side

---

For what doesn't match the previous build
- Right to Left: Matches previous
- Left to Right: Inverted
- Webtoon: Inverted
- Vertical: Inverted
- Continous vertical: Inverted

For Webtoon, Vertical, and Continuous vertical I will fix them. 
For Right to Left and Left to Right, I'm unsure if adding separate default navigation for Left to Right will help due to them both having the same invert. Like I said before
> I was thinking if the user doesn't like to change to the next page on the left side they can invert it to the right side

dafillio:
Hmm I guess as long as the inverted option exists it's not a big deal but it seems a bit odd to have the default behaviour for left to right to proceed to the next page be on the left side. To me it would make more sense to have the default behaviour be that you would tap on the side in which you want to progress/turn towards. As long as this is expected behaviour that's fine it just seemed odd enough that I thought it was a bug and warranted a report. I strongly agree on reverting the vertical changes though, that ones seems to be a little more clear cut.

Matth7878:
Can't wait for this to be fixed. It's really annoying to not be able to tap right to go next on left right flow.
At least with webtoon mode there is the ability to invert tapping but for left to right and right to left the setting is shared. :-/

PR #4238, opened 2021-01-03
Fix navigation from feedback

ghostbear:
Based on feedback from #4237

This fixes navigation issues with L Navigation (Default for Webtoon, Vertical, and Continous Vertical) not matching previous build navigation. And (Right to Left)/(Left to Right) not using moveRight and -Left resulting in needing to invert navigation to match previous build

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
@@ -92,10 +92,18 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
 
             val pos = PointF(event.rawX / pager.width, event.rawY / pager.height)
             val navigator = config.navigator
-            when (navigator.getAction(pos)) {
-                ViewerNavigation.NavigationRegion.MENU -> activity.toggleMenu()
-                ViewerNavigation.NavigationRegion.NEXT -> moveToNext()
-                ViewerNavigation.NavigationRegion.PREV -> moveToPrevious()
+            if (this is VerticalPagerViewer) {
+                when (navigator.getAction(pos)) {
+                    ViewerNavigation.NavigationRegion.MENU -> activity.toggleMenu()
+                    ViewerNavigation.NavigationRegion.NEXT -> moveRight()
+                    ViewerNavigation.NavigationRegion.PREV -> moveLeft()
+                }
+            } else {
+                when (navigator.getAction(pos)) {
+                    ViewerNavigation.NavigationRegion.MENU -> activity.toggleMenu()
+                    ViewerNavigation.NavigationRegion.NEXT -> moveLeft()
+                    ViewerNavigation.NavigationRegion.PREV -> moveRight()
+                }
             }
         }
         pager.longTapListener = f@{
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
@@ -15,7 +15,7 @@ import eu.kanade.tachiyomi.ui.reader.model.ChapterTransition
 import eu.kanade.tachiyomi.ui.reader.model.ReaderPage
 import eu.kanade.tachiyomi.ui.reader.model.ViewerChapters
 import eu.kanade.tachiyomi.ui.reader.viewer.BaseViewer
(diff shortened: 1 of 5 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt
```diff
@@ -92,10 +92,18 @@ abstract class PagerViewer(val activity: ReaderActivity) : BaseViewer {
...
+                    ViewerNavigation.NavigationRegion.MENU -> activity.toggleMenu()
+                    ViewerNavigation.NavigationRegion.NEXT -> moveRight()
+                    ViewerNavigation.NavigationRegion.PREV -> moveLeft()
+                }
+            } else {
+                when (navigator.getAction(pos)) {
+                    ViewerNavigation.NavigationRegion.MENU -> activity.toggleMenu()
+                    ViewerNavigation.NavigationRegion.NEXT -> moveLeft()
+                    ViewerNavigation.NavigationRegion.PREV -> moveRight()
+                }
```
ghostbear, line 106: I'm not really sure this is the right approach because this will not only change `LNavigation` (which is inverted if this is not here) but also change the tapping for `KindlishNavigation` and `EdgeNavigation` works if using `VerticalPagerViewer`

This logic would be better for `PagerDefaultNavigation` to behave more like previous builds
```kt
if (navigator is PagerDefaultNavigation) {
  when (navigator.getAction(pos)) {
    ViewerNavigation.NavigationRegion.MENU -> activity.toggleMenu()
    ViewerNavigation.NavigationRegion.NEXT -> moveLeft()
    ViewerNavigation.NavigationRegion.PREV -> moveRight()
  }
  return@f
}
when (navigator.getAction(pos)) {
  ViewerNavigation.NavigationRegion.MENU -> activity.toggleMenu()
  ViewerNavigation.NavigationRegion.NEXT -> moveToNext()
  ViewerNavigation.NavigationRegion.PREV -> moveToPrevious()
}
```

Alternatively adding `RIGHT` and `LEFT` to the `NavigationRegion` enum would also solve this, then the code could be written like
```kt
when (navigator.getAction(pos)) {
  ViewerNavigation.NavigationRegion.MENU -> activity.toggleMenu()
  ViewerNavigation.NavigationRegion.NEXT -> moveToNext()
  ViewerNavigation.NavigationRegion.PREV -> moveToPrevious()
  ViewerNavigation.NavigationRegion.RIGHT -> moveRight()
  ViewerNavigation.NavigationRegion.LEFT -> moveLeft()
}
```
And `PagerDefaultNavigation` would use `RIGHT` and `LEFT` instead of `NEXT` and `PREV`. WebtoonViewer doesn't have moveRight and -Left methods here it would probably just ignore it or have `RIGHT` scroll down and `LEFT` scroll up or vise versa

arkon: We could potentially make all of the functions abstract and implement them in the WebtoonViewer too (where `moveLeft` would actually be the same as `scrollUp`, for example).

I'm okay with any of the approaches though.

ghostbear:
I have added additional enum values (`RIGHT` and `LEFT`) to NavigationRegion, so PagerViewer can move right or left instead of only next due to it being two-directional. And I don't see an exclusive ViewerNavigation for WebtoonViewer using these additional values but if it would happen it will scroll down on `RIGHT` and scroll up on `LEFT`

Charuru:
Please please please release. I really need the new nav, the preview doesn't keep your library.

arkon:
You should file an issue if the app isn't working properly. The preview version should retain your library between updates without any issues.

Charuru:
> You should file an issue if the app isn't working properly. The preview version should retain your library between updates without any issues.

I'm already using the stable release version, it doesn't keep your lib from that.

ghostbear:
Stable and preview doesn't share library

If you are using stable. You will need to backup your library in stable, and restore the backup in preview.

Charuru:
That's a pain, and i don't think the backup tool ports the already downloaded manga you got. That's why I'm just trying to rush him to release lol.

jobobby04:
Stable and preview use the same download location, so yes, downloads would be shared

Charuru:
> Stable and preview use the same download location, so yes, downloads would be shared

Thanks I appreciate the tip. It's not working for me right now but maybe that's because my device doesn't have internet access currently. I'll give it a try when I get back online thanks.

arkon:
This conversation isn't relevant to the PR. If you need support, please refer to the website or Discord.

PR #4310, opened 2021-01-18
Hide tracking when no tracker is logged in and change filter logic

### LINKED ISSUE #4307: Only show tracking filter if logged into a tracker

ghostbear:
closes #4307

Adds `isVisible` attribute to `MultiState` which is then used to make `MultiStateHolder` not visible

https://user-images.githubusercontent.com/6576096/104925531-4491a580-599f-11eb-9978-a52eb250774e.mp4

---

Also changed the logic for checking if a manga is being tracked

From
```kt
return db.getTracks().asRxObservable().map { tracks ->
  tracks.associate { track ->
    // Checks if any tracker is logged in for any manga being tracked
    val isLogged = tracks.any { trackManager.getService(it.sync_id)?.isLogged ?: false }
    // Manga id will always be true if one tracker is logged, a manga 
    // can use another tracker that is not logged in and still show up
    Pair(track.manga_id, isLogged) 
  }
}
```
| Result (Only MAL is logged in) |  |
|---|---|
| Screenshot_1610978317 | Screenshot_1610978319 |
| | ^Is being tracked by AniList which is not logged in |

To
```kt
return db.getTracks().asRxObservable().map { tracks ->
  tracks.groupBy { it.manga_id } // Groups tracks by manga id 
                .mapValues { tracksForMangaId ->
                    // Check if any of the trackers is logged in for the current manga id
                    tracksForMangaId.value.any {
                        trackManager.getService(it.sync_id)?.isLogged ?: false
                    }
                }
}
```
| Result (Only MAL is logged in) |  |
|---|---|
| Screenshot_1610978363 | Screenshot_1610978368 |
| ^Flying Witch is being tracked by only MAL, the rest is tracked by MAL and AniList | ^Is being tracked by MAL which is logged in and AniList which is not logged in |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
@@ -118,6 +118,7 @@ class LibraryPresenter(
         val filterUnread = preferences.filterUnread().get()
         val filterCompleted = preferences.filterCompleted().get()
         val tracking = preferences.filterTracking().get()
+        val isNotLogged = !trackManager.hasLoggedServices()
 
         val filterFnUnread: (LibraryItem) -> Boolean = unread@{ item ->
             if (filterUnread == State.IGNORE.value) return@unread true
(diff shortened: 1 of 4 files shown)


arkon:
Nice, thanks.

One other minor thing: can you change the string to "Tracked" to match the other filters?

ghostbear:
> One other minor thing: can you change the string to "Tracked" to match the other filters?

Changed

PR #4344, opened 2021-01-24
Extend track filter

ghostbear:
| No logged in tracker | One logged in tracker | Multiple logged in trackers |
|---|---|---|
| Screenshot_1611482424 | Screenshot_1611482412 | Screenshot_1611482391 |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
@@ -153,51 +153,40 @@ class LibraryPresenter(
         }
 
         val filterFnTracking: (LibraryItem) -> Boolean = tracking@{ item ->
-            if (loggedInTracks.values.size > 1) {
-                // Exclude > Include
-                if (isNotLogged || (loggedInTracks.all { it.value == State.IGNORE.value } && !loggedInTracks.any { it.value == State.EXCLUDE.value || it.value == State.INCLUDE.value })) return@tracking true
+            if (isNotLogged || (loggedInTracks.all { it.value == State.IGNORE.value } && !loggedInTracks.any { it.value == State.EXCLUDE.value || it.value == State.INCLUDE.value })) return@tracking true
 
-                val trackedManga = trackMap[item.manga.id ?: -1] ?: return@tracking false
+            val trackedManga = trackMap[item.manga.id ?: -1]
 
-                val exclude = loggedInTracks.filterValues { it == State.EXCLUDE.value }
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySettingsSheet.kt
```diff
@@ -64,20 +70,49 @@ class LibrarySettingsSheet(
...
+            private fun getServiceResId(service: TrackService): Int {
+                return when (service) {
+                    is Anilist -> R.string.anilist
+                    is MyAnimeList -> R.string.my_anime_list
+                    is Kitsu -> R.string.kitsu
+                    is Bangumi -> R.string.bangumi
+                    is Shikimori -> R.string.shikimori
+                    else -> R.string.unknown
+                }
+            }
```
arkon, line 114: I guess this is because refactoring the items to accept strings is a bit too much for this PR?

It's fine if so, but I can also clean this up later by making the `TrackService`s use these too after.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -149,11 +152,40 @@ class LibraryPresenter(
...
 
         val filterFnTracking: (LibraryItem) -> Boolean = tracking@{ item ->
-            if (isNotLogged || tracking == State.IGNORE.value) return@tracking true
+            if (isNotAnyLoggedIn || (loggedInServices.all { it.value == State.IGNORE.value } && !loggedInServices.any { it.value == State.EXCLUDE.value || it.value == State.INCLUDE.value })) return@tracking true
+
+            val trackedManga = trackMap[item.manga.id ?: -1]
+
+            var exclude: Collection<Boolean>
+            var include: Collection<Boolean>
+            val containsExclude = loggedInServices.filterValues { it == State.EXCLUDE.value }
```
arkon, line 161: This feels a little repetitive with the initial check in L155.

You could probably do something like:

```
if (isNotAnyLoggedIn)
  return true

val containsInclude = ...
val containsExclude = ...

// All ignored
if (!containsInclude.any() && !containsExclude.any())
  return true

val include = ...
val exclude = ...

...
```

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySettingsSheet.kt
ghostbear: Yeah it was either this or figuring out a way for items to accept strings which felt be a bit much

Issue #961, opened 2017-08-30
Handling of multiple categories, might remove categories


scriptmode:
When you select multiple series and then change categories, the categories which not all series have will be removed. It seems the cagegory-dialog only gets the list of categories which all series have shared. On save it writes those categories back to all selected series, overwriting the existing list, instead of just adding or removing changed categories.

Example:

    Manga_1 with [Categorie_A]
    Manga_2 with [Categorie_A, Categorie_B]

becomes after selection and saving:

    Manga_1 with [Categorie_A]
    Manga_2 with [Categorie_A]

Version: 0.6.1
Android 6.0

Jendrej:
That’s what’s supposed to happen. Tachiyomi likes to forcefully change status of things instead of taking the current state into account. So it is not a bug.

scriptmode:
Then it's a case of very bad UX. It's not the excepted result.

Jendrej:
For me it is, but I see why could you think otherwise. That is just how the
dev does things. It is the same with the whole MAL updating thing they
refused to call synchronisation and now I know why. If you change your
rating on MAL and not in Tachiyomi, when it updates the rating is
overwritten with what you had in the app, even if it was blank. Here I
agree that it is not what should be expected.

30 sie 2017 12:22 "scriptmode" <notifications@github.com> napisał(a):

Then it's a case of very bad UX. It's not the excepted result.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/inorichi/tachiyomi/issues/961#issuecomment-325949109>,
or mute the thread

j2ghz:
The problem here is what should happen. Do you have a better proposal?
Maybe it would be best to not allow changes to categories that not all selected manga have.
Related: #135

scriptmode:
The usual solution for this cases is to mark partly assigned elements in grey/half transparent way. When state is changed, you first remove the category for all elements, and behave normally for all future state-changes. If no state was changed, left the category untouched. It's called tripple-state. Not very hard to implement, but not sure how simple it's possible with android.

Anyway, this is clearly communicates what happens. The way Tachiyomi works now is a hidden behaviour. It's only obvious from the result, not from the interface itself. That's why it's bad.

@j2ghz
[...]

PR #4360, opened 2021-01-27
Fix UpdateCategories dialog to not remove old value

quangkieu:
Fix https://github.com/tachiyomiorg/tachiyomi/issues/961
Add preselected to updateCategoriesForMangas listener as reference
Add updateMangasToCategories that bulk update categories of mangas using
    old and new common categories

Signed-off-by: Quang Kieu <quangkieu1993@gmail.com>

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
@@ -613,7 +613,7 @@ open class BrowseSourceController(bundle: Bundle) :
      * @param mangas The list of manga to move to categories.
      * @param categories The list of categories where manga will be placed.
      */
-    override fun updateCategoriesForMangas(mangas: List<Manga>, categories: List<Category>) {
+    override fun updateCategoriesForMangas(mangas: List<Manga>, categories: List<Category>, preselected: List<Category>) {
         val manga = mangas.firstOrNull() ?: return
 
         presenter.changeMangaFavorite(manga)
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/ChangeMangaCategoriesDialog.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/ChangeMangaCategoriesDialog.kt
@@ -40,13 +40,14 @@ class ChangeMangaCategoriesDialog<T>(bundle: Bundle? = null) :
                 allowEmptySelection = true
             ) { _, selections, _ ->
                 val newCategories = selections.map { categories[it] }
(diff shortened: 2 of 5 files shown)


quangkieu:
@arkon I wonder who I need to ping for review?

arkon:
Nobody, I'll look at it eventually.

quangkieu:
Hopefully fixed forked branch too https://github.com/Jays2Kings/tachiyomiJ2K/issues/644

CarlosEsco:
Copying this from the J2K ticket
```

    Set one comic in both A, B catalogs
    Set another comic in both A, C catalogs
    Select both comics, the catalog menu would show A checked
    Now add catalog D
    Both comics now have tag A, D but none of them have catalog B or C
```
So while #961 is an issue, solving it this way just pushes the issue to the other extreme.

Glancing at the code, it keeps the uncommon categories on each manga, but what if i wanted to remove the uncommon ones also?.  Like if i had manga in A,B,C and some of them were also in D, and wanted to move them all the E this wouldnt work.

quangkieu:
@CarlosEsco 
User could search for category D
Then select all comic filtered
D should now be checked initially
Remove D and apply

quangkieu:
This is not ideal solution as this is based on dialog with only 2 states
But it would not completely remove old uncommon

arkon:
I'm having trouble wrapping my head around how this is _supposed_ to work given the change. Neither the steps that @CarlosEsco nor @quangkieu described are intuitive.

The dialog would need to be reworked with a custom view with a tri-state dialog that shows:
- [✓] Checked if all manga are in that category
- [-] Mixed state if some, but not all, manga are in that category
- [ ] Not checked if no manga are in that category

And clicking a checkbox would go to:
- [✓] All manga should be in this category
- [ ] No manga should be in this category

Where categories with the mixed state ([-]) would not be changed if not toggled.

quangkieu:
@arkon I fully understand that. I just not know how to set the [-]. If you could point to the document or code for that. I am familiar with Kotlin or Android much

arkon:
Unfortunately it won't be that straightforward in the code's current state, since it'd involve creating a new view to mimic the checkbox with an extra state.

We have something similar that we use for the filter sheets (it uses different drawables for the state) that you can refer to: https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/widget/ExtendedNavigationView.kt#L93

quangkieu:
@arkon Hmm, I would try to see if I could catch up on it. In the meantime, I wonder if we could push this one out so it would be least destruction to the Library. I would assume user would able to work around the quirk behavior.

arkon:
I have tri-state checkbox dialogs implemented for category exclusion settings, so I'll reuse that for addressing this at some point.

PR #4361, opened 2021-01-27
Unify history and update item

ghostbear:
It has annoyed me that history uses card background while updates don't. So this makes history and update item look more the same. 

| Updates | History | [Not in PR] History but with grey tinted ImageButtons |
|---|---|---|
| Screenshot_1611749515 | Screenshot_1611749512 | Screenshot_1611749488 |
| Title is now bold | Removed card background | I thought this looks weird, that why I didn't include it. But would match ImageButton tint on other screens |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryHolder.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryHolder.kt
@@ -68,9 +68,8 @@ class HistoryHolder(
             binding.mangaSubtitle.text = Date(history.last_read).toTimestampString()
         }
 
-        val requestOptions = RequestOptions().apply {
-            transform(CenterCrop(), RoundedCorners(8))
-        }
+        val radius = itemView.context.resources.getDimensionPixelSize(R.dimen.card_radius)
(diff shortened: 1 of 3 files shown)


CarlosEsco:
Thoughts on removing the cards for the dates?  I always thought it looked weird having the headers on cards.  Either no cards, or 1 card per a date is my though

ghostbear:
> Thoughts on removing the cards for the dates?

I'm okay with the date headers having a card, not the prettiest but they work.

Also currently it looks weird to remove the background from date headers due to them being sticky

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryHolder.kt
```diff
@@ -65,12 +68,16 @@ class HistoryHolder(
             binding.mangaSubtitle.text = Date(history.last_read).toTimestampString()
         }
 
+        val requestOptions = RequestOptions().apply {
+            transform(CenterCrop(), RoundedCorners(8))
```
arkon, line 72: Should use `R.dimen.card_radius` like the other places.

arkon:
Yeah, I'd rather the headings not change for now. I do like how the one at the top looks with the toolbar, but it'd look a bit odd if everything was grey.

PR #4375, opened 2021-02-02
Improving genre search started from the manga page of a popular manga

E3FxGaming:
Tapping on a genre while viewing a popular manga of a HttpSource/ParsedHttpSource now brings up better results,
if the extension developer implemented a genre filter into the extension filter list.

Until now tapping on a genre below the manga description of a popular manga simply started a search where the genre was put into the manga title field.
That's of course the incorrect way to search for a genre on most websites that offer manga genre.

Extension developers put effort into making the proper "search for genre" way accessible through filter - the changes made with this pull request bridge the gap between the filter opportunity and the simplicity of tapping on a genre of a manga to see similar manga with the same genre.

Works with genre filters implemented as groups of checkbox filters/tri-state filters, as well as with simple drop down filters.

Extension developers don't have to do any additional work beyond implementing the genre filter properly.
The genre in the filter must have the same name as the genre that show up below the manga description (upper/lower case ignored).

image

image

The new genre filter method has a built-in fallback to the old way tapping on a genre was handled:
If a genre name can't be found in the extension filters, it'll simply search for the genre as if it is a manga title.

I tested this with the extensions

- MangaPark (has a list of tri-state genre filter)
- Manga Mutiny (has a list of checkbox genre filter)
- Mangakakalot (has a drop down genre filter)


dionisioedu (approved):

arkon:
Had to fix some conflicts, so "merged" via 62955e73852ed1ca69a1168c855721076da9fb05

PR #4392, opened 2021-02-06
Add Right and Left for navigation mode

### LINKED ISSUE #4391: [Feature Request] Have another option for webtoon viewer navigation that is like the pager default navigation
The current default navigation for webtoon is the L navigation. I have often accidentally tapped the middle top and bottom row looking for the menu because I am used to the pager default navigation. Having another option for the webtoon navigation that is like the pager default would help fix this problem.

## What/Requirements
Have an extra option for the webtoon navigation that is like the pager default navigation, where the menu is in the full vertical centre row.

ghostbear:
and remove default navigation classes in favor of the navigation classes

closes #4391

### DIFF
--- a/app/src/main/res/values/arrays.xml
+++ b/app/src/main/res/values/arrays.xml
@@ -97,6 +97,7 @@
         <item>@string/l_nav</item>
         <item>@string/kindlish_nav</item>
         <item>@string/edge_nav</item>
+        <item>@string/right_and_left_nav</item>
     </string-array>
 
     <string-array name="webtoon_nav">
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/arrays.xml
```diff
@@ -97,12 +97,14 @@
         <item>@string/l_nav</item>
         <item>@string/kindlish_nav</item>
         <item>@string/edge_nav</item>
+        <item>@string/right_and_left_nav</item>
     </string-array>
 
     <string-array name="webtoon_nav">
```
arkon, line 103: Could just have one array since they're the same, couldn't we?

ghostbear: Yes we could but if we would like to add a specific navigation mode for a Viewer (Webtoon or Pager) it would mean we can't. I don't know how likely it would be to have a specific navigation mode for a Viewer, due to all navigation types is currently for both Viewers. So maybe one array would be better until they diverge from one another

PR #4470, opened 2021-02-13
Let users invert dual page split

### LINKED ISSUE #4468: [Feature Request] Implement An Optional Inverse Mode for Double Page Split.
Most manga are read from Right to Left, however the Left to Right **viewing** mode has been the default in Tachiyomi for ages, and I like it that way (because it's easy for right-handed usage, and I'm very used to it).

With the way the current Double Page Split feature works, if you're using the Left to Right mode, and if a page is split, the left side is shown first, rather than the right (it simply follows the order). This can be antithetical to conventional manga reading experience, no?

> ### Why not just use Right to Left viewing mode and use Inverse Tapping? 

Well, like I said, I'm too used to LTR **viewing** experience with an RTL **reading** experience. I did try doing this and I found myself not exactly being comfortable with it.

## What/Requirements

For those users who prefer a Left to Right **viewing** experience with a Right to Left **reading** experience (I'd suspect I'm not alone, as far as reading manga is concerned), why not make an option for the Double Page Split to show the right side first when using LTR? Basically, an inverse of the way it works currently.

Plus it would be only an option, so you wouldn't be forcing it on anyone who prefers the current method. I hope I've been able to convey my thoughts properly, and I hope this is considered seriously. Thanks for the great app, as usual.

ghostbear:
closes #4468 

This adds another option to let users invert dual page split

This would allow users to change the placement for Webtoon and Continous Vertical as it currently places them as the user was reading R2L

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderSettingsSheet.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderSettingsSheet.kt
@@ -6,6 +6,7 @@ import android.widget.Spinner
 import androidx.annotation.ArrayRes
 import androidx.core.view.isVisible
 import androidx.core.widget.NestedScrollView
+import androidx.lifecycle.lifecycleScope
 import com.tfcporciuncula.flow.Preference
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.data.preference.PreferencesHelper
@@ -15,7 +16,6 @@ import eu.kanade.tachiyomi.ui.reader.viewer.pager.PagerViewer
 import eu.kanade.tachiyomi.ui.reader.viewer.webtoon.WebtoonViewer
 import eu.kanade.tachiyomi.widget.IgnoreFirstSpinnerListener
 import eu.kanade.tachiyomi.widget.sheet.BaseBottomSheetDialog
-import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.flow.launchIn
 import uy.kohesive.injekt.injectLazy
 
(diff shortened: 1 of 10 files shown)


ImPudding-tech:
Small Suggestion:
WHEN Going to the Next Split Page. Change The Page Number to
1 - 1st Split
1.5 - 2nd Split

I dont Need to Open Another issue Because This Is Just A Small Request

ghostbear:
> Small Suggestion:
> WHEN Going to the Next Split Page. Change The Page Number to
> 1 - 1st Split
> 1.5 - 2nd Split
> 
> I dont Need to Open Another issue Because This Is Just A Small Request

LOL, you will need to open another issue. Because it is a separate issue from what this PR is trying to solve. And it can be a large request as much as you think it is small

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderSettingsSheet.kt
```diff
@@ -25,6 +28,8 @@ class ReaderSettingsSheet(private val activity: ReaderActivity) : BaseBottomShee
 
     private val binding = ReaderSettingsSheetBinding.inflate(activity.layoutInflater, null, false)
 
+    private val viewScope = MainScope()
```
arkon, line 31: Can just use `activity.lifecycleScope` instead.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerPageHolder.kt
```diff
@@ -264,24 +264,28 @@ class PagerPageHolder(
...
+            (viewer is R2LPagerViewer || viewer is VerticalPagerViewer) && page !is InsertPage -> ImageUtil.Side.RIGHT
+            else -> error("We should choose a side!")
+        }
 
-            inputStream = ImageUtil.splitInHalf(inputStream, side)
+        side = when {
+            viewer.config.dualPageInvert && side == ImageUtil.Side.RIGHT -> ImageUtil.Side.LEFT
+            viewer.config.dualPageInvert && side == ImageUtil.Side.LEFT -> ImageUtil.Side.RIGHT
+            else -> side
         }
```
arkon, line 282: I feel like it'd be slightly cleaner with just

```
if (viewer.config.dualPageInvert) {
  side = when (side) {
    // ...
  }
}
```

PR #4474, opened 2021-02-14
Add ability to exclude manga tagged with certain categories from update

yshui:
I find it useful to be able to tag mangas I don't want updated with a designated category, while still keep the mangas in categories related to their contents.

This adds an new option of a set of excluded categories. If any one of a manga's categories is in the exclusion list, the mange won't be updated.


Ocsidot:
Great propasl, I was thinking to request this feature 👍

dionisioedu (approved):

arkon:
Thanks for the PR, but I've made it such that the categories list can have excluded items now too.

PR #4489, opened 2021-02-15
Add Right and Left to reader settings

### LINKED ISSUE #4487: [Bug] Right and left reader navigation mode not selectable as default
## Device information
* Tachiyomi version: 0.10.9
* Android version: 10 (SDK 29)
* Device: Xiaomi Redmi Note 8 Pro

## Steps to reproduce
1. More → Settings → Reader
2. After checking Navigation → Tapping, Navigation Layout will be selectable for Paged and Webtoon.
3. Option Right and Left non-existant.

### Expected behavior
New Right and Left navigation mode selectable as default.

### Actual behavior
Option not there

## Other details
Right and Left navigation is selectable during reader mode. But then you'd have to change it for every single manga.

ghostbear:
closes #4487

Adds Right and Left to reader settings 

| Before | After |
|---|---|
| Screenshot_1613390377 | Screenshot_1613390227 |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
@@ -145,7 +145,7 @@ class SettingsReaderController : SettingsController() {
             intListPreference {
                 key = Keys.navigationModePager
                 titleRes = R.string.pref_viewer_nav
-                entries = resources?.getStringArray(R.array.webtoon_nav)?.also { values ->
+                entries = context.resources.getStringArray(R.array.pager_nav).also { values ->
                     entryValues = values.indices.map { index -> "$index" }.toTypedArray()
                 }
                 defaultValue = "0"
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
```diff
@@ -145,13 +145,9 @@ class SettingsReaderController : SettingsController() {
...
                 key = Keys.navigationModePager
                 titleRes = R.string.pref_viewer_nav
-                entriesRes = arrayOf(
-                    R.string.default_nav,
-                    R.string.l_nav,
-                    R.string.kindlish_nav,
-                    R.string.edge_nav
-                )
-                entryValues = arrayOf("0", "1", "2", "3")
+                entries = resources?.getStringArray(R.array.webtoon_nav)?.also { values ->
```
arkon, line 148: `pager_nav`?

ghostbear: Your are indeed right that this should be `pager_nav`

Issue #3627, opened 2020-08-06
[Bug] Search is cleared when app in background


scb261:
### Device information
* Tachiyomi version: 0.10.3
* Android version: 5.1
* Device: Huawei TIT-U02

## Steps to reproduce
1. Use global search or open some source.
2. Start to type search.
3. Switch to other app to see other part of manga's title.
4. Switch back to Tachiyomi.

### Expected behavior
Previously typed search stays.

### Actual behavior
Previously typed search disappears and you need to type it again.

### Other details
Right now you can save current search by tapping search button on keyboard before switching apps, so it searches with part that you already typed in. And you even could find desired manga with it. But the problem is mostly with Japanese titles since part that I can remember isn't enough to find the manga.

It also was happening on 0.9.2.

scb261:
Fixed by #4494

PR #4494, opened 2021-02-16
Add onPause\onResume persistence to searchView. Fixes issue #3627

AntoineGS:
Still learning the ropes with Android\Kotlin coming from Delphi so if something isn't Kotlin-y enough let me know :)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
@@ -121,7 +121,7 @@ abstract class BaseController<VB : ViewBinding>(bundle: Bundle? = null) :
      * [expandActionViewFromInteraction] should be set to true in [onOptionsItemSelected] when the expandable item is selected
      * This method should be called as part of [MenuItem.OnActionExpandListener.onMenuItemActionExpand]
      */
-    fun invalidateMenuOnExpand(): Boolean {
+    open fun invalidateMenuOnExpand(): Boolean {
         return if (expandActionViewFromInteraction) {
             activity?.invalidateOptionsMenu()
             false
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
@@ -1,0 +1,94 @@
+package eu.kanade.tachiyomi.ui.base.controller
+
+import android.app.Activity
+import android.os.Bundle
+import android.view.Menu
+import android.view.MenuInflater
+import androidx.appcompat.widget.SearchView
+import androidx.viewbinding.ViewBinding
+import eu.kanade.tachiyomi.R
+import nucleus.presenter.Presenter
+
+/**
+ * Implementation of the NucleusController that has a built-in ViewSearch
+ */
+abstract class SearchableNucleusController<VB : ViewBinding, P : Presenter<*>>
+(bundle: Bundle? = null) : NucleusController<VB, P>(bundle) {
+
+    /**
+     * Bool used to bypass the initial searchView being set to empty string after an onResume
+     */
+    protected var storeNonSubmittedQuery: Boolean = false
+
+    /**
+     * Store the query text that has not been submitted to reassign it after an onResume, UI-only
+     */
+    protected var nonSubmittedQuery: String = ""
+
+    protected fun commonCreateOptionsMenu(
+        menu: Menu,
+        inflater: MenuInflater,
+        menuId: Int,
+        searchItemId: Int,
+        useGlobalSearch: Boolean = false
+    ) {
+        // Inflate menu
+        inflater.inflate(menuId, menu)
+
+        // Initialize search option.
+        val searchItem = menu.findItem(searchItemId)
+        val searchView = searchItem.actionView as SearchView
+        searchView.maxWidth = Int.MAX_VALUE
+
+        if (nonSubmittedQuery.isNotBlank()) {
+            searchItem.expandActionView()
+            searchView.setQuery(nonSubmittedQuery, false)
+            storeNonSubmittedQuery = true // searchView.requestFocus() does not seem to work here
+        } else if (useGlobalSearch) {
+            // Change hint to show global search.
+            searchView.queryHint = applicationContext?.getString(R.string.action_global_search_hint)
+        }
+
+        initSearchHandler(searchView)
+    }
+
+    private fun initSearchHandler(searchView: SearchView) {
+        searchView.setOnQueryTextFocusChangeListener { _, hasFocus ->
+            storeNonSubmittedQuery = hasFocus
+        }
+
+        searchView.setOnQueryTextListener(object : SearchView.OnQueryTextListener {
+            // Save the query string whenever it changes to be able to store it for persistence
+            override fun onQueryTextChange(newText: String?): Boolean {
+                // Ignore events triggered when the search is not in focus
+                if (storeNonSubmittedQuery) {
+                    nonSubmittedQuery = newText ?: ""
+                }
+                // abstract function for implementation
+                onSearchViewQueryTextChange(newText)
+                return false
+            }
+
+            // Only perform search when the query is submitted
+            override fun onQueryTextSubmit(query: String?): Boolean {
+                // abstract function for implementation
+                onSearchViewQueryTextSubmit(query)
+                return true
+            }
+        })
+    }
+
+    override fun onActivityResumed(activity: Activity) {
+        super.onActivityResumed(activity)
+        // searchView.onQueryTextChange is triggered after this, and the query set to "", so we make
+        // sure not to save it (onActivityResumed --> onQueryTextChange
+        // --> OnQueryTextFocusChangeListener --> onCreateOptionsMenu)
+        storeNonSubmittedQuery = false
+    }
+
+    /**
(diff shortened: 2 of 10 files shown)


arkon (commented):
Considering a lot of the logic is actually duplicate, maybe we could put it in the parent class (`NucleusController`)? I think there's actually a few other places that could take advantage of  this too that aren't called out in the original issue.

AntoineGS:
Good point, I was making a conservative commit as my first contribution so I was hesitant to start refactoring a dozen files (I think there are 15 files with a very similar onCreateOptionsMenu that inherit from NucleusController).
I'll refactor and push another commit for this.

arkon:
Makes sense. If you want to keep the PR small, you could add the base logic and only apply it one controller, then do another PR (or more) to apply it to others.

AntoineGS:
Ok so this is small-ish commit.
I sub-classed BaseController so that I can work on the conversion without breaking the build or creating a massive PR.
Once everything is converted I can merge SearchableNucleusController into BaseController and make the necessary declaration changes.
SearchView has a lot of nitpicky behaviors, like initializing its query to "" after you have set a value and thus losing the set value.

OK so while writing this I realized I might be able to get around this whole loading\loaded issue by posting the query restore to the queue which will hopefully mean it will execute after the initialization to "".. If this works it should simplify the class.

AntoineGS:
Ok so it does not really remove the need to know when the SearchView is loaded, when resuming the activity I would still need to know or the nonCommitedQuery will be set to "".
Plus it makes the UI jitter as it gets set to "" and the to the actual search.
Between the two the committed implementation works better, let me know if you have any comments. 
It is unfortunately pretty complex for something that I had figured would be easy to implement, though making it centralized makes it easier to maintain.

AntoineGS:
Hey @arkon, once this is merged I will do the other controllers (in case you were waiting on more commits in this PR before merging as I am not planning to). 
Probably you are just busy with other stuff, if so no worries ;)

arkon:
I'll be sure to review this PR this weekend.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
```diff
@@ -121,7 +121,7 @@ abstract class BaseController<VB : ViewBinding>(bundle: Bundle? = null) :
      * [expandActionViewFromInteraction] should be set to true in [onOptionsItemSelected] when the expandable item is selected
      * This method should be called as part of [MenuItem.OnActionExpandListener.onMenuItemActionExpand]
      */
-    fun invalidateMenuOnExpand(): Boolean {
+    open fun invalidateMenuOnExpand(): Boolean {
```
arkon, line 124: It actually probably makes sense to move all of this search stuff from this class to your new `SearchableNucleusController` class, as it's not actually relevant for controllers without a `SearchView`.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    private fun acceptEmptyQuery(): Boolean {
+        return when (currentSearchViewState) {
+            SearchViewState.COLLAPSING, SearchViewState.FOCUSED -> true
+            else -> false
+        }
+    }
+
+    private fun setCurrentSearchViewState(newState: SearchViewState, fromState: SearchViewState? = null) {
```
arkon, line 155: Nitpick: let's use `to` and `from` (which we use elsewhere too, like in `syncActivityViewWithController`).

app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/SourceController.kt
```diff
@@ -3,13 +3,7 @@ package eu.kanade.tachiyomi.ui.browse.source
...
 import android.app.Dialog
 import android.os.Bundle
-import android.view.LayoutInflater
-import android.view.Menu
-import android.view.MenuInflater
-import android.view.MenuItem
-import android.view.View
-import android.view.ViewGroup
-import androidx.appcompat.widget.SearchView
+import android.view.*
```
arkon, line 6: You can turn this off from Preferences > Editor > Code Style > Kotlin > Imports

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    protected open fun onSearchMenuItemActionExpand(item: MenuItem?) {
+    }
+
+    protected open fun onSearchMenuItemActionCollapse(item: MenuItem?) {
+    }
+
+    /**
+     * During the conversion to SearchableNucleusController (after which I plan to merge its code
+     * into BaseController) this addresses an issue where the searchView.onTextFocus event is not
```
arkon, line 188: I sort of wonder if we _should_ merge them. I guess we could, but I feel like it's making `BaseController` a lot more complicated than it needs to be too.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Store the query text that has not been submitted to reassign it after an onResume, UI-only
+     */
+    protected var nonSubmittedQuery: String = ""
+
+    /**
+     * To be called by classes that extend this subclass in onCreateOptionsMenu
+     */
+    protected fun commonCreateOptionsMenu(
```
arkon, line 42: Nitpick: maybe just name this `createOptionsMenu`?

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
```diff
@@ -0,0 +1,200 @@
...
+        menu: Menu,
+        inflater: MenuInflater,
+        menuId: Int,
+        searchItemId: Int,
+        queryHint: String = "",
```
arkon, line 47: Let's actually make this a `@StringRes` `Int` to force uses of it to be localized strings.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
```diff
@@ -0,0 +1,200 @@
...
+            }
+
+            if (restoreCurrentQuery) {
+                // Restoring a query the user had submitted
+                if (query.isNotBlank()) {
+                    searchItem.expandActionView()
+                    searchView.setQuery(query, true)
+                    searchView.clearFocus()
+                    onSearchViewQueryTextChange(query)
+                    onSearchViewQueryTextSubmit(query)
```
arkon, line 101: Seems a bit odd to fire both.

arkon (commented):
Overall looks pretty good, just some minor comments.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
AntoineGS: Good point, I've taken note of it to remove it once all views with SearchView are moved onver to SearchableNucleusController.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
AntoineGS: Done

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
AntoineGS: I learned something new, thanks ;)

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
AntoineGS: The idea behind it is that some implementations (ie local searches) are triggered by the `onSearchViewQueryTextChange` event and do not need to implement the submit event.
Firing only the `onSearchViewQueryTextSubmit` event would require more implementations.
I do not mind doing it, let me know if you prefer it with only the submit being fired.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
AntoineGS: Done!

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
AntoineGS: Makes sense, there is no real need to merge them if there are view using the 'vanilla' `BaseController`. Plus it keeps all the complicated state logic in its own unit so it is more obvious that is it for the search capabilities only.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/SearchableNucleusController.kt
arkon: Probably fine for now. We can address it later if it seems problematic.

arkon:
Thanks again! We can follow up on this work in separate PRs.

PR #4557, opened 2021-02-28
Library update freq: add 4 & 8 hours

Riztard:


Bm12347:
My friend can you make it that we can add the time ourselves

Riztard:
> My friend can you make it that we can add the time ourselves

nah, i m too dumb for that, and i add 4 & 8 because i need it

PR #4641, opened 2021-03-13
Update URL for Local Manga guide

Soitora:
Updates to the new URL. Maybe create a Cloudflare redirect specifically for this one as well at least for a little while?

Alternatively I can create a page under that url on the website with a link to the updated url since Vuepress own Redirection tools doesn't seem to work as we want them to.


arkon:
A temporary page sounds fine to me, since it's easier to manage (and not forget about).

PR #4683, opened 2021-03-20
Add navigation layout overlay

ghostbear:
This adds an in-app overlay for navigation layouts 

The current behavior is that it will show when a user opens the reader for the first time or if a user has turned on `Show navigation overlay on start`. Then when a user starts tapping or swiping it will disaper. After that it will reaper of a user changes the navigation layout to another one

---

User opens the reader for the first time or users that have turned on `Show navigation overlay on start`

https://user-images.githubusercontent.com/6576096/111871984-2e976580-898d-11eb-94c2-1bfb94ee6640.mp4

---

After the user opens the reader for the first time

https://user-images.githubusercontent.com/6576096/111871986-30612900-898d-11eb-9c31-b8fc9d200d89.mp4

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderNavigationOverlayView.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderNavigationOverlayView.kt
@@ -1,0 +1,138 @@
+package eu.kanade.tachiyomi.ui.reader
+
+import android.content.Context
+import android.graphics.Canvas
+import android.graphics.Color
+import android.graphics.Paint
+import android.util.AttributeSet
+import android.view.MotionEvent
+import android.view.View
+import android.view.ViewPropertyAnimator
+import androidx.core.content.ContextCompat
+import androidx.core.view.isVisible
+import eu.kanade.tachiyomi.R
+import eu.kanade.tachiyomi.ui.reader.viewer.ViewerNavigation
+import kotlin.math.abs
+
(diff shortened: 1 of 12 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderNavigationOverlayView.kt
```diff
@@ -0,0 +1,140 @@
...
+            canvas?.drawText(region.type.name, x, y, textBorderPaint)
+            canvas?.drawText(region.type.name, x, y, textPaint)
+
+            canvas?.restore()
+        }
+    }
+
+    private fun paintForRegion(type: ViewerNavigation.NavigationRegion): Paint {
+        return Paint().apply {
+            when (type) {
```
arkon, line 98: Could maybe just have a map of region to ColorInt.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderNavigationOverlayView.kt
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    override fun performClick(): Boolean {
+        super.performClick()
+
+        if (viewPropertyAnimator == null && isVisible) {
+            viewPropertyAnimator = animate()
+                .alpha(0f)
+                .setDuration(1000L)
```
arkon, line 124: Could maybe extract this into a constant, in case we ever fiddle with it and forget to change the duration on the other end.

arkon (commented):
Looks pretty good overall. The actual colours could probably be tweaked but it's fine for now.

PR #4752, opened 2021-04-01
Replace reading mode snackbar with toast

T3sT3ro:
Currently snackbar displaying reading direction covers the navbar, so the user can't immediately see how many pages does a chapter have, nor can he scroll fast without dismissing the snackbar or waiting for it to disappear. This commit replaces a snackbar with centered toast

# snackbar:
image

# toast:
image


arkon:
I'm not really sure if this is the best solution or not, but I'll merge it and see what the feedback is. Thanks!

T3sT3ro:
I was also thinking about solution used in MangaRock when it was still alive - small icon of manga page appearing in the center, animated numbers with trail on panels and a hand swiping the page in proper direction. I'm not entirely convinced of that solution though.

PR #4765, opened 2021-04-02
Fix renaming download folder

BlueCat300:
This commit will remove the "_" from the beginning of chapter folders if the extension does not have chapter.scanlator


arkon:
This will break all existing downloads, so you'd need to handle that. That said, is there a strong benefit to this? I'm hesitant to check for more filename formats as it's just going to make things slower.

BlueCat300:
I just wanted to fix the bug of this condition. it is always true for extensions without chapter.scanlator string. If re-downloading the manga is very important for users, then it can be closed. I hope there is a better way to make this change.

PR #4773, opened 2021-04-04
add sort by date fetched in library

### LINKED ISSUE #4735: [Feature Request] <Allow users to sort library by fetch date>
## User Problem
Users that sort by "Latest Chapter" run into some issues when sources themselves are unreliable in how they list chapter dates. This leads to updating manga from certain sources appearing inconsistently in their library.

For example:

- Isekaiscan continually updates its manga as releasing [x seconds ago] for around 5 days after its actual release, meaning whenever manga is refreshed from this source its chapter date is continually modified to the current date, pushing them to the top of the library and keeping them there.
- Mangahub labels its manga a day later than it should, meaning any new updates do not rise to the top of the library and instead remain further down the chonological list than it should.
- Webtoon has a similar problem to mangahub, new updates are backdated by a day and do not rise to the top of the library.
- There have also been reports of manga sources labelling its date as "0", thereby always gathering at the bottom of the library, but i have not seen this first hand.

Problems like this are and will continue to be recurring when using a sort that relies on unreliable source dating. Since examples like these are a site issue, there is no fix on an individual level.

## What/Requirements
I propose a new library sort called "Recently Updated". This sort option will list manga in the library solely based on when a new chapter was fetched, disregarding source prescribed dates. When there is a new chapter, it's manga will rise to the top of the library. This will make for a much cleaner user experience, as newly updated manga will always be at the top of the library list. 

## User Benefits
- Completely avoids any problems relying on source-prescribed chapter dates currently cause/may cause in the future for users that want to view recently updated manga in their Library.
- Will allow users to gain a more accurate view of recent updates while still in the Library (which, unlike the Updates tab, allows for thumbnails, grid format, user-generated categories etc.).
- Sort by "Latest Chapter" will still be preserved for users that want to have Library listed in chonological order as their main priority (i.e. for those with manga with a wide range of time periods of release, and want their library to order based on that chronology)

Riztard:
Closes #4735

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySort.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySort.kt
@@ -8,8 +8,8 @@ object LibrarySort {
     const val UNREAD = 3
     const val TOTAL = 4
     const val LATEST_CHAPTER = 6
-    const val CHAPTER_FETCH_DATE = 7
-    const val DATE_ADDED = 8
+    const val CHAPTER_FETCH_DATE = 8
+    const val DATE_ADDED = 7
 
     @Deprecated("Removed in favor of searching by source")
     const val SOURCE = 5
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySort.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySort.kt
@@ -8,7 +8,8 @@ object LibrarySort {
     const val UNREAD = 3
     const val TOTAL = 4
     const val LATEST_CHAPTER = 6
-    const val DATE_ADDED = 7
+    const val CHAPTER_FETCH_DATE = 7
+    const val DATE_ADDED = 8
 
     @Deprecated("Removed in favor of searching by source")
     const val SOURCE = 5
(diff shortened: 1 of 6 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySort.kt
```diff
@@ -8,7 +8,8 @@ object LibrarySort {
     const val UNREAD = 3
     const val TOTAL = 4
     const val LATEST_CHAPTER = 6
-    const val DATE_ADDED = 7
+    const val CHAPTER_FETCH_DATE = 7
+    const val DATE_ADDED = 8
```
ghostbear, line 12: For anyone using `Date added` this will mean their sorting filter will be changed to `Date fetched`. And this may confuse/be unwanted for some when their sorting filter suddenly changes. You probably want to keep `Date added` as `7`. `Date fetched` can still be before `Date added` in the sorting options even though it is `8`
```suggestion
    const val CHAPTER_FETCH_DATE = 8
    const val DATE_ADDED = 7
```

Or is this meant to change the sorting filtering for anyone using `Date added` to `Date fetched`?

Riztard: ah right

>Or is this meant to change the sorting filtering for anyone using Date added to Date fetched?

no

PR #4791, opened 2021-04-07
add page split webtoon page (alpha)

Riztard:
split very long webtoon page, to make it less lag/smoother when scrolling

*based on dual page split
*need to split more than 2 if the image is very long or make it dynamic or at least just split into 3

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
@@ -6,6 +6,7 @@ import android.widget.LinearLayout
 import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.RecyclerView
 import eu.kanade.tachiyomi.ui.reader.model.ChapterTransition
+import eu.kanade.tachiyomi.ui.reader.model.InsertPage
 import eu.kanade.tachiyomi.ui.reader.model.ReaderChapter
 import eu.kanade.tachiyomi.ui.reader.model.ReaderPage
 import eu.kanade.tachiyomi.ui.reader.model.ViewerChapters
@@ -19,10 +20,31 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
     /**
      * List of currently set items.
      */
-    var items: List<Any> = emptyList()
+    var items: MutableList<Any> = mutableListOf()
         private set
 
     var currentChapter: ReaderChapter? = null
+    fun onPageSplitWebtoon(current: Any?, newPage: InsertPage, clazz: Class<out WebtoonViewer>) {
+        if (current !is ReaderPage) return
+
+        val currentIndex = items.indexOf(current)
+
+        val placeAtIndex = currentIndex + 1
+
+        // It will enter a endless cycle of insert pages
+        if (clazz.isAssignableFrom(WebtoonViewer::class.java) && items[placeAtIndex - 1] is InsertPage) {
+            return
+        }
+
+        // Same here it will enter a endless cycle of insert pages
+        if (items[placeAtIndex] is InsertPage) {
+            return
+        }
+
+        items.add(placeAtIndex, newPage)
+
+        notifyDataSetChanged()
+    }
 
     /**
      * Updates this adapter with the given [chapters]. It handles setting a few pages of the
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
@@ -24,21 +24,15 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
         private set
 
     var currentChapter: ReaderChapter? = null
-    fun onPageSplitWebtoon(current: Any?, newPage: InsertPage, clazz: Class<out WebtoonViewer>) {
-        if (current !is ReaderPage) return
+    fun onPageSplitWebtoon(current: ReaderPage?, newPage: InsertPage) {
+        if (current is InsertPage) return
 
         val currentIndex = items.indexOf(current)
 
         val placeAtIndex = currentIndex + 1
 
-        // It will enter a endless cycle of insert pages
-        if (clazz.isAssignableFrom(WebtoonViewer::class.java) && items[placeAtIndex - 1] is InsertPage) {
-            return
-        }
-
-        // Same here it will enter a endless cycle of insert pages
         if (items[placeAtIndex] is InsertPage) {
-            return
+            return // We return here to not insert more of the already split page
         }
 
         items.add(placeAtIndex, newPage)
@@ -209,6 +203,7 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
 
     fun cleanupPageSplit() {
         val insertPages = items.filterIsInstance(InsertPage::class.java)
+        if (insertPages.size == 0) return
         items.removeAll(insertPages)
         notifyDataSetChanged()
     }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
@@ -24,6 +24,7 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
         private set
 
     var currentChapter: ReaderChapter? = null
+
     fun onPageSplitWebtoon(current: ReaderPage?, newPage: InsertPage) {
         if (current is InsertPage) return
 
@@ -37,6 +38,13 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
 
         items.add(placeAtIndex, newPage)
 
+        notifyDataSetChanged()
+    }
+
+    fun cleanupPageSplit() {
+        val insertPages = items.filterIsInstance(InsertPage::class.java)
+        if (insertPages.size == 0) return
+        items.removeAll(insertPages)
         notifyDataSetChanged()
     }
 
@@ -199,12 +207,5 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
          * View holder type of a chapter transition view.
          */
         const val TRANSITION_VIEW = 1
-    }
-
-    fun cleanupPageSplit() {
-        val insertPages = items.filterIsInstance(InsertPage::class.java)
-        if (insertPages.size == 0) return
-        items.removeAll(insertPages)
-        notifyDataSetChanged()
     }
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
@@ -206,4 +206,10 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
          */
         const val TRANSITION_VIEW = 1
     }
+
+    fun cleanupPageSplit() {
+        val insertPages = items.filterIsInstance(InsertPage::class.java)
+        items.removeAll(insertPages)
+        notifyDataSetChanged()
+    }
 }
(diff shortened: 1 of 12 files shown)


ghostbear:
Just leaving an idea for splitting it into more than two.

Instead of using `InsertPage` which only supports right and left have a data class that holds an offset, and the parent. And a pre-defined height that the class would use as height for each split of the image. This class and height variable would work as a stencil to display the image. 

---

Let us say you have an image with a height of 4500 and the stencil height of 1000. This means the code should generate four stencil pages
```
*---*
|   | ReaderPage
|   |
*---* Y=1000
|   | StencilPage(offset=1000)
|   |
*---* Y=2000
|   | StencilPage(offset=2000)
|   |
*---* Y=3000
|   | StencilPage(offset=3000)
|   |
*---* Y=4000
|   | StencilPage(offset=4000)
*---* Y=4500
```
When it detects that it is a webtoon page `ReaderPage` will get the offset from 0 to the stencil height (1000), then each stencil page will use the offset along with the stencil height to render the reset of the webtoon page. The last page may or may not be 1000 in height so it would just render what's left in this case 500.

It would still work like dual page split inserting into the adapter but it would insert multiple items instead of one.

The stencil height could be the screen height or the screen height multiplied by 2 not necessarily 1000

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonViewer.kt
```diff
@@ -219,6 +220,10 @@ class WebtoonViewer(val activity: ReaderActivity, val isContinuous: Boolean = tr
         }
     }
 
+    fun onPageSplitWebtoon(currentPage: ReaderPage, newPage: InsertPage) {
+        adapter.onPageSplitWebtoon(currentPage, newPage, this::class.java)
+    }
+
```
ghostbear, line 226: You probably want to have a cleanup function as well else you will end up will duplicates
https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewerAdapter.kt#L155-L159

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonPageHolder.kt
```diff
@@ -315,6 +324,34 @@ class WebtoonPageHolder(
...
     }
 
+    private fun processSplitWebtoonPage(openStream: InputStream): InputStream {
+        var inputStream = openStream
+        val (isWebtoonPage, stream) = when (page) {
+            is InsertPage -> Pair(true, inputStream)
+            else -> ImageUtil.isWebtoonPage(inputStream)
+        }
+        inputStream = stream
+        if (isWebtoonPage) {
```
arkon, line 334: Short-circuit early, i.e.

```
if (!isWebtoonPage) {
  return ...
}

// Other logic
return ...
```

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonPageHolder.kt
```diff
@@ -315,6 +324,34 @@ class WebtoonPageHolder(
...
+            }
+
+            inputStream = ImageUtil.splitWebtoon(inputStream, upperSide)
+        }
+        return inputStream
+    }
+
+    private fun onPageSplitWebtoon() {
+        val newPage = page?.let { InsertPage(it) }
+        page?.let { newPage?.let { it1 -> viewer.onPageSplitWebtoon(it, it1) } }
```
arkon, line 352: Just use an if-statement to check if `page` is null, the `.let`s here aren't making it any more readable.

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonPageHolder.kt
```diff
@@ -315,6 +324,34 @@ class WebtoonPageHolder(
         addSubscription(readImageHeaderSubscription)
     }
 
+    private fun processSplitWebtoonPage(openStream: InputStream): InputStream {
+        var inputStream = openStream
```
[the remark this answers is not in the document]
ghostbear: It is needed because of reassignment later when the stream is read, else the image will just be blank. But it can be replaced with `InputStream.mark()` and `InputStream.reset()` in the functions in `ImageUtil`

Riztard:
ill just let ghostbear or others to continue this 🤣
i dont even know what did i write lol
my ape brain cant handle it :harold:

ghostbear:
A monke wrote it so another monke should be able to add to it

Riztard:
> A monke wrote it so another monke should be able to add to it

apes alone weak
apes together strong

ghostbear:
I think you should be able to go through the PR review
I will go through them and add suggestions to help you out a bit

ghostbear (commented):
I have made some suggestions to help you get through the PR review, it's based on arkon's comments. If you want me to explain more about why I did something in one of the suggestion just comment on them

ghostbear:
> A recent PR replaced this function with an extension function, so you could use that here too.

As for adding this, I think you should checkout 62ac664, rebase, and force push to this branch to get the extension functions

Just remember the things you changed in c64d44b

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
```diff
@@ -184,4 +206,10 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
          */
         const val TRANSITION_VIEW = 1
     }
+
+    fun cleanupPageSplit() {
+        val insertPages = items.filterIsInstance(InsertPage::class.java)
+        items.removeAll(insertPages)
+        notifyDataSetChanged()
+    }
```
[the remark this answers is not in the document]
Riztard: > arkon: Put this alongside the other split function

its not possible right? because pager and webtoon have different adapter? 🤔

app/src/main/java/eu/kanade/tachiyomi/util/system/ImageUtil.kt
```diff
@@ -87,6 +87,18 @@ object ImageUtil {
         return Pair(options.outWidth > options.outHeight, ByteArrayInputStream(imageBytes))
     }
 
+    /**
+     * Check whether the image is a webtoon image (width * 3 < height), return the result and original stream
```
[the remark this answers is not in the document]
Riztard: yea what if its just high quality images 🤔

Riztard:
hmm failed

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonViewer.kt
```diff
@@ -325,4 +332,13 @@ class WebtoonViewer(val activity: ReaderActivity, val isContinuous: Boolean = tr
             min(position + 3, adapter.itemCount - 1)
         )
     }
+
+    fun onPageSplitWebtoon(currentPage: ReaderPage, newPage: InsertPage) {
+        adapter.onPageSplitWebtoon(currentPage, newPage, this::class.java)
```
ghostbear, line 337: To many arguments
```suggestion
        adapter.onPageSplitWebtoon(currentPage, newPage)
```

arkon:
You don't really need to rebase, the core extension functions should be available to you anyway.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonAdapter.kt
```diff
@@ -184,4 +206,10 @@ class WebtoonAdapter(val viewer: WebtoonViewer) : RecyclerView.Adapter<RecyclerV
          */
         const val TRANSITION_VIEW = 1
     }
+
+    fun cleanupPageSplit() {
```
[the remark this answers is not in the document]
arkon: I mean grouping `onPageSplitWebtoon` and `cleanupPageSplit` together.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonPageHolder.kt
```diff
@@ -315,6 +324,38 @@ class WebtoonPageHolder(
...
+            is InsertPage -> Pair(true, inputStream)
+            else -> ImageUtil.isWebtoonPage(inputStream)
+        }
+        inputStream = stream
+        
+        if (!isWebtoonPage) {
+            return inputStream
+        }
+        
+        val upperSide = when {
```
arkon, line 339: Can you try making it so it's `when (page) { ... }`?

app/src/main/java/eu/kanade/tachiyomi/util/system/ImageUtil.kt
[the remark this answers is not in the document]
arkon: Super hi-res images will make it lag anyway.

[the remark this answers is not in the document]
Riztard: so if height more than what?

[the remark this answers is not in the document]
Riztard: make it like (height > width && height > x) ?
or dont compare to width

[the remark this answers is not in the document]
arkon: Kind of hard to say without any data, but I'm fine with @ghostbear's suggestion of 10,000px as a start. We could fiddle around with it more in the future. You just need to make sure that it's simple to change that predicate.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonPageHolder.kt
Riztard: that sounds simple, but im not famiiar with `when` LOL

arkon:
```
when (page) {
  is InsertPage -> ...
  !is InsertPage -> ...
}
```

arkon: Or maybe just a good old fashioned `if`/`else`.

Riztard: like this? 🤔
https://github.com/tachiyomiorg/tachiyomi/pull/4791/commits/01a621aa7ea2c4f6d41a6968105daa7fc3e1bf2e

arkon:
@Riztard Was this ready for another round of review, or are you still iterating on it?

Riztard:
maybe, i dont know what else to do 😂
ill let someone(you) clean it :tehe:

arkon:
@ghostbear I assume you were interested in continuing this work? How did you want to do that, do you want to PR into @Riztard's branch, or do you want to create your own, or do you want me to merge this into a different branch to work off of?

ghostbear:
I will probably just create my own PR superseding this and make Riztard a co-author on the commit containing most of his logic

ghostbear:
I have tested this and the stencil page idea on my old Sony Xperia M2 Aqua which is the lowest end device I have

Riztard's PR seems to work fine just that it's a bit janky jumping around when processing/adding the pages

For my stencil page idea, it throws OOM near-instantaneous this is mostly because it does everything on IO rather than the main thread. Even if I only do two pages like Riztard's PR does it still throws OOM

I found an alternative way to reduce lag which is to use `BitmapFactory` option `inSampleSize` to basically scale down the image, but at the cost of quality. And it doesn't throw OOM as soon as it's being processed. I believe this would be a better way to approach this rather than to introduce jankiness into the `WebtoonViewer`.
Maybe even keep some of the elements of this PR in this solution where it only scales down extremely tall images

FourTOne5:
> I have tested this and the stencil page idea on my old Sony Xperia M2 Aqua which is the lowest end device I have

Can you provide the code?

ghostbear:
> > I have tested this and the stencil page idea on my old Sony Xperia M2 Aqua which is the lowest end device I have
> 
> Can you provide the code?

It seems I have deleted the stash containing the code, so no I can't provide it.

But it is really similar to how it does dual page split except it determined how many pages it needed to split it to then loop over the original image to create the splits and inserted them into the recycler

FourTOne5:
> then loop over the original image to create the splits and inserted them into the recycler

Where exactly were you looping it? I tried to do by looping but it didn't went as planned

ghostbear:
> > then loop over the original image to create the splits and inserted them into the recycler
> 
> Where exactly were you looping it? I tried to do by looping but it didn't went as planned

Before calling `splitWebtoon` I changed the parameters to take in an int to represent which page it was creating and remove the side parameter. Also had a global constant to represent the height of the stencil image 

So something like this:
```kt
// Global constant somewhere
val STENCIL_HEIGHT = 10_000

// Loop
val imageHeight = ImageUtil.getHeight(inputStream)
val numberOfPages = imageHeight / STENCIL_HEIGHT
for (pageIndex in 0..numberOfPages) {
  val newImage = ImageUtil.splitWebtoon(inputStream, pageIndex)
  // Handle image
}
// First image gets returned to the current WebtoonPageHolder
// The rest of the images goes to the recycler 
```

FourTOne5:
Just wanted to clarify I'm working on this.

arkon:
Superseded by #5759

PR #4802, opened 2021-04-09
Draw edge-to-edge

ivaniskandar:
Insets are handled by [Insetter](https://chrisbanes.github.io/insetter/) so it should be consistent across APIs.

The navigation bar color will only be set transparent when it's positioned on the bottom (e.g., on landscape orientation with non-gesture navbar won't be set transparent).

[portrait](https://user-images.githubusercontent.com/12537387/114224914-e7a0fc80-999b-11eb-9987-8d6b36f24885.png)
[landscape gesture 1](https://user-images.githubusercontent.com/12537387/114224630-8547fc00-999b-11eb-88d6-e9114bf64efd.png)
[landscape gesture 2](https://user-images.githubusercontent.com/12537387/114224754-b0325000-999b-11eb-88f5-73b73ac36fc5.png)
[landscape 3-button 1](https://user-images.githubusercontent.com/12537387/114224820-c5a77a00-999b-11eb-841e-3ab1617a5a24.png)
[landscape 3-button 2](https://user-images.githubusercontent.com/12537387/114225267-54b49200-999c-11eb-82e5-6ce296cdf5bb.png)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
@@ -2,11 +2,16 @@ package eu.kanade.tachiyomi.ui.main
 
 import android.app.SearchManager
 import android.content.Intent
+import android.graphics.Color
+import android.os.Build
 import android.os.Bundle
 import android.view.View
 import android.view.ViewGroup
 import android.widget.Toast
 import androidx.coordinatorlayout.widget.CoordinatorLayout
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowCompat
+import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.isVisible
 import androidx.core.view.updateLayoutParams
 import androidx.lifecycle.lifecycleScope
@@ -18,6 +23,7 @@ import com.bluelinelabs.conductor.Router
 import com.bluelinelabs.conductor.RouterTransaction
 import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.behavior.HideBottomViewOnScrollBehavior
+import dev.chrisbanes.insetter.applyInsetter
 import eu.kanade.tachiyomi.BuildConfig
 import eu.kanade.tachiyomi.Migrations
 import eu.kanade.tachiyomi.R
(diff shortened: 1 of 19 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
```diff
@@ -85,6 +91,35 @@ class MainActivity : BaseViewBindingActivity<MainActivityBinding>() {
         setContentView(binding.root)
         setSupportActionBar(binding.toolbar)
 
+        // Needs windowLightNavigationBar to avoid navbar icon contrast issue
```
arkon, line 94: Could potentially set a translucent (e.g. 50% opacity) background for lower versions.

arkon (commented):
This is great! I didn't realize the insetter library got a nice DSL too, but I definitely struggled to get it to work properly before, so I appreciate this a lot.

app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
ivaniskandar: Ideally, with button navigation we should use translucent scrim like the default system behavior as stated [here](https://medium.com/androiddevelopers/gesture-navigation-going-edge-to-edge-812f62e4e83e), but the system-defined scrim is not fitting with the available color scheme (except light theme). And if we want to custom it, there's no official API to detect when gesture navigation with dynamic color adaptation is used.

I can add some hacky check I've used in other project, but it's proven to be broken on some devices.

For versions prior to Q, I'll try to use translucent scrims.

arkon: I was thinking just using the existing colours (white, blue, grey, black) but making them translucent, as the themes should already be handling the correct button colours.

We could leave this for a separate change too if you prefer (and if you're okay with actually making that change). This PR seems fine as-is.

ivaniskandar:
This applies to all versions prior to API 27
[API 21 default light with 54% md black scrim](https://user-images.githubusercontent.com/12537387/114257737-35455580-99ec-11eb-8195-9c5bea7897e0.png)
[API 21 light blue with 70% colorPrimary scrim](https://user-images.githubusercontent.com/12537387/114257648-a0daf300-99eb-11eb-9ae8-7418ebee4740.png)
[API 21 all dark themes with no scrim](https://user-images.githubusercontent.com/12537387/114257820-a4bb4500-99ec-11eb-9490-6439616238e9.png)

LGTM although I still prefer using translucent scrim on non-gesture navigation. This PR is good to go but if you want the translucent scrim I can add checks to apply it to all situations except when gesture nav is used.

arkon (approved):
LGTM. We could tweak it if there's any user feedback.

PR #4841, opened 2021-04-14
Add manga-wised rotation mode settings

ghostbear:
Supersedes #3522

The only difference is that I haven't moved `Default rotation type` to be beside `Default reading mode` in the settings

And the only thing missing is an icon for `DEFAULT` orientation

| Reader | Settings |
|---|---|
| Screenshot_1618389082 | Screenshot_1618389101 |

---

**Video**
Rotation default: Force Portrait
Reader default: Vertical 

https://user-images.githubusercontent.com/6576096/114695186-a0a86380-9d1b-11eb-93fd-37712ad62645.mp4

^ Orientation icon and force portrait not working when default has been fix

---

closes #3522


arkon:
What happens if you set a custom orientation and restore a backup into a version/fork that doesn't have it? I assume nothing happens, but it'd be best to make sure just to avoid crashes.

ghostbear:
```kt
viewer_flags = viewer_flags and 0x00000007.inv() or (0x00000002 and 0x00000007) // -> 2, we set a reader for the series
viewer_flags = viewer_flags and 0x00000038.inv() or (0x00000018 and 0x00000038) // -> 26, we set a orientation for the series
```
If the code is the same in the forks they should just use `R2LPagerViewer` if a custom orientation was set on `viewer_flags` when opening the reader

---

Tested with J2K it seems that opening the reader and reader settings will select `R2L` while the reader settings sheet will crash
image

ghostbear:
A solution would be that we use the mask when backing up manga so we only add the reading mode setting and not both

arkon:
We could potentially try to land a mitigation for that crash into J2K first (not necessarily add this feature, just avoid the crash), and let Neko pick that up.

SY should pick up the change from here quickly, and AZ can take it as needed. I don't particularly care about backwards compatibility either.

ghostbear:
J2K would only need...
```kt
var readingModeType: Int
        get() = viewer_flags and 0x00000007
        set(readingMode) { viewer_flags = viewer_flags and 0x00000007.inv() or (readingMode and 0x00000007) }
```
...to avoid a crash. This would then allow J2K to then add orientation per series if wanted. But the only thing really needed is the get field to avoid crashes, it could still set to viewer_flags.

Or were you thinking of changing something related to how the spinner works to avoid the `IndexOutOfBoundsException`?

arkon:
We can sync up with Jay on Discord to see what he's comfortable with, but the mask probably makes sense.

CarlosEsco:
what happened to not breaking backups :sob:

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
```diff
@@ -83,8 +85,10 @@ class SettingsReaderController : SettingsController() {
                     R.string.rotation_force_portrait,
                     R.string.rotation_force_landscape,
                 )
-                entryValues = arrayOf("1", "2", "3", "4", "5")
-                defaultValue = "1"
+                entryValues = OrientationType.values().drop(1)
+                    .filterNot { value -> value == OrientationType.LOCKED_LANDSCAPE } // Lock has two types in this enum while only being one option
```
arkon, line 89: This isn't true anymore.

PR #4845, opened 2021-04-14
Add navigation bar scrim

ivaniskandar:
This checks for a internal flag `config_navBarNeedsScrim` added since API 29. It exists in pretty much all of my test devices (Pixel 3, Xperia 5II and Galaxy A50).

If the flag explicitly set to false, navbar color will be set to transparent, else it will use 70% of navigationBarColor.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
@@ -24,7 +24,6 @@ import com.bluelinelabs.conductor.Router
 import com.bluelinelabs.conductor.RouterTransaction
 import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.behavior.HideBottomViewOnScrollBehavior
-import dev.chrisbanes.insetter.Insetter
 import dev.chrisbanes.insetter.applyInsetter
 import eu.kanade.tachiyomi.BuildConfig
 import eu.kanade.tachiyomi.Migrations
@@ -103,26 +102,16 @@ class MainActivity : BaseViewBindingActivity<MainActivityBinding>() {
                 margin(top = true)
             }
         }
-        binding.rootFab.applyInsetter {
-            type(navigationBars = true) {
-                margin()
-            }
-        }
         binding.bottomNav.applyInsetter {
             type(navigationBars = true) {
                 padding()
             }
         }
-        Insetter.builder()
-            .consume(Insetter.CONSUME_ALL)
-            .setOnApplyInsetsListener { view, insets, _ ->
-                val systemInsets = insets.getInsets(WindowInsetsCompat.Type.systemBars())
-                view.isVisible = systemInsets.bottom > 0
-                view.updateLayoutParams<ViewGroup.LayoutParams> {
-                    height = systemInsets.bottom
-                }
+        binding.rootFab.applyInsetter {
+            type(navigationBars = true) {
+                margin()
             }
-            .applyToView(binding.navigationScrim)
+        }
 
         // Make sure navigation bar is on bottom when making it transparent
         ViewCompat.setOnApplyWindowInsetsListener(binding.root) { _, insets ->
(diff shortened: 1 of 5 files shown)


app/src/main/java/eu/kanade/tachiyomi/util/system/InternalResourceHelper.kt
```diff
@@ -0,0 +1,25 @@
...
+     * @return 0 if not available
+     */
+    private fun getResourceId(resName: String, type: String): Int {
+        return Resources.getSystem().getIdentifier(resName, type, "android")
+    }
+
+    fun getBoolean(context: Context, resName: String, defaultValue: Boolean): Boolean {
+        val id = getResourceId(resName, "bool")
+        return if (id != 0) {
+            context.createPackageContext("android", 0).resources.getBoolean(id)
```
ivaniskandar, line 20: Before anyone asks why I didn't use Resources.getSystem() it's because overlaid resources won't get picked up with it.

app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
```diff
@@ -103,33 +104,27 @@ class MainActivity : BaseViewBindingActivity<MainActivityBinding>() {
...
 
-        // Make sure navigation bar is on bottom when making it transparent
+        // Make sure navigation bar is on bottom before we modify it
         ViewCompat.setOnApplyWindowInsetsListener(binding.root) { _, insets ->
             if (insets.getInsets(WindowInsetsCompat.Type.navigationBars()).bottom > 0) {
-                // Keep scrim on light theme if windowLightNavigationBar is not available
-                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O_MR1 || isDarkMode) {
-                    window.navigationBarColor = Color.TRANSPARENT
+                window.navigationBarColor = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q &&
+                    !InternalResourceHelper.getBoolean(this, "config_navBarNeedsScrim", true)
```
arkon, line 122: I can't say I'm a fan of using internal flags, considering how wacky some ROMs can be (especially those Chinese ones...).

arkon:
Another thing that may be of note is that [this change](https://github.com/tachiyomiorg/tachiyomi/commit/15bed1ac4c9140e7c0da214fdf6cd721cd7e4ff1) to fix the weird toolbar snapping behaviour also results in the content getting pushed down by the height of the status bar. I do have a local change to compensate for that, but it's not particularly elegant.

app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
ivaniskandar: Well I'm not either but if the system doesn't have the flag then it will just return the default value, so those ROMs will just get the scrim all the time..

ivaniskandar:
> Another thing that may be of note is that [this change](https://github.com/tachiyomiorg/tachiyomi/commit/15bed1ac4c9140e7c0da214fdf6cd721cd7e4ff1) to fix the weird toolbar snapping behaviour also results in the content getting pushed down by the height of the status bar. I do have a local change to compensate for that, but it's not particularly elegant.

I don't have the issue mentioned in this commit, what was the issue on the first place? Is it only happening with oppo devices?

ivaniskandar:
Looking at the provided video, I'd like to think that it's the Toolbar snapping behavior that's broken when padding is applied.

There's also this attribute set for the parent AppBarLayout [here](https://github.com/tachiyomiorg/tachiyomi/blame/master/app/src/main/res/layout/main_activity.xml#L14). It doesn't do anything on my device but who knows what the other ROMs do..

ivaniskandar:
So I can kind of replicate the snap issue without https://github.com/tachiyomiorg/tachiyomi/commit/15bed1ac4c9140e7c0da214fdf6cd721cd7e4ff1, and looks like a similar issue is reported [here](https://github.com/material-components/material-components-android/issues/2001).

I'd say remove the snap scroll flag and call it a day...

https://user-images.githubusercontent.com/12537387/114825370-55ac5000-9df0-11eb-8142-872980148195.mp4

app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
arkon: That's fair.

PR #4896, opened 2021-04-18
Support parsing volume numbers as chapters for volume-only titles

tombriden:
For cases where manga is a volume, the previous behaviour would result
in the "volume" tag and number being removed and all files end up as
"chapter -1".

This change adds an additional check for the single occurrence of
numbers before removing the unwanted tags, so files named, eg
`attackontitan_vol1.cbz` or epubs with metadata `Attack on Titan
Vol 1`  can still have the "chapter" number parsed for sorting
in the library, while continuing to support cases where the single
number match would happen after the volume tags are removed

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterRecognition.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterRecognition.kt
@@ -54,6 +54,19 @@ object ChapterRecognition {
             it.forEach { occurrence -> name = name.replace(occurrence.value, occurrence.value.trim()) }
         }
 
+        // Start by checking for one number occurrence before any name manipulation
+        // to support volume-only files
+        val occurrences: MutableList<MatchResult> = arrayListOf()
+        occurrence.findAll(name).let {
+            it.forEach { occurrence -> occurrences.add(occurrence) }
+        }
+
+        if (occurrences.size == 1) {
+            if (updateChapter(occurrences[0], chapter)) {
+                return
+            }
+        }
+
         // Remove unwanted tags.
         unwanted.findAll(name).let {
             it.forEach { occurrence -> name = name.replace(occurrence.value, "") }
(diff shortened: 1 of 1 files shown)


arkon:
What's the benefit of having the volume number as the "chapter" number? I can see people potentially being confused when trackers aren't updated correctly, for example.

tombriden:
> What's the benefit of having the volume number as the "chapter" number? I can see people potentially being confused when trackers aren't updated correctly, for example.

I use tachiyomi for local manga from places like humblebundle that tend to distribute as volumes. They also have no consistency in image resolutions for the available formats so I end up with a mixture of cbz and epub for the same manga. Source sorting doesn't work as cbz uses filename and epub uses metadata and the chapter number is always -1 so when reading, the next volume often isn't right. This way allows the volumes to be sortable even though its not technically the chapter number. 
I hadn't considered the tracker side of things though so can see that being a problem. I was going for minimum change but guess I should look into parsing the volume separately and displaying/sorting one or the other or both depending on what's available?

arkon:
> guess I should look into parsing the volume separately and displaying/sorting one or the other or both depending on what's available?

That'd make sense, although I'm not quite sure how feasible it is without any model changes (which is problematic for extensions and whatever vague level of backwards compatibility we strive to achieve with older versions).

paulo27ms:
> What's the benefit of having the volume number as the "chapter" number? I can see people potentially being confused when trackers aren't updated correctly, for example.

Trackers already won't update properly anyway and Tachiyomi doesn't overwrite tracking chapter data if the chapter read is smaller so it's unlikely this would be confusing at all and probably would go unnoticed entirely (because people will have to update the chapter count manually and the chapter number will be a lot higher than the volume count).

tombriden:
> > What's the benefit of having the volume number as the "chapter" number? I can see people potentially being confused when trackers aren't updated correctly, for example.
> 
> Trackers already won't update properly anyway and Tachiyomi doesn't overwrite tracking chapter data if the chapter read is smaller so it's unlikely this would be confusing at all and probably would go unnoticed entirely (because people will have to update the chapter count manually and the chapter number will be a lot higher than the volume count).

I also noticed some of my volumes metadata don't even contain the word "volume" (eg `Attack on Titan 27`) so were being parsed as chapters already, which results in the same problem if tracking is enabled

arkon:
Closing this since it doesn't seem like it'd be very reliable. Feel free to open another PR (@ anyone reading this) if you figure out a way to do so, but also with some unit tests.

PR #4903, opened 2021-04-19
Status bar icon colors in webview activity

Fmaldonado6:
Added a new theme for the Webview activity just to change the icon color in the status bar

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -84,7 +84,7 @@
         </activity>
         <activity
             android:name=".ui.security.BiometricUnlockActivity"
-            android:theme="@style/Theme.Splash" />
+            android:theme="@style/Theme.Base" />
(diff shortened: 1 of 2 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -87,6 +87,7 @@
             android:theme="@style/Theme.Splash" />
         <activity
             android:name=".ui.webview.WebViewActivity"
+            android:theme="@style/Theme.Webview"
```
ivaniskandar, line 90: I think it's better to change the app theme to Theme.Base instead. I forgot to do that on my last PR.

Fmaldonado6: Oh yeah! that works aswell and it's better than adding another theme, I just pushed a new commit with the changes, thanks!

arkon:
Thanks!

PR #4914, opened 2021-04-20
MainActivity: Show bottom nav when the tab page is changed

### LINKED ISSUE #4895: [Bug] Bottom Navbar stays disabled when switching categories
## Device information
* Tachiyomi version: Stable 0.10.10
* Android version: 10
* Device: OnePlus 6

## Steps to reproduce
1. Create two categories - One with multiple items (big) allowing for vertical scrolling and another with few items (small) where scrolling is not available.
2. Go to the big category and scroll down. This hides the bottom navbar.
3. Switch to the small category.

### Expected behavior
The bottom navbar should become visible as there is no scrolling functionality in the small category.

### Actual behavior
The bottom navbar remains hidden until you switch back to the big category and scroll up.

## Other details
[Video](https://imgur.com/a/787Wsxw) demonstrating this.

ivaniskandar:
Fixes #4895

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
@@ -24,7 +24,6 @@ import com.bluelinelabs.conductor.Router
 import com.bluelinelabs.conductor.RouterTransaction
 import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.behavior.HideBottomViewOnScrollBehavior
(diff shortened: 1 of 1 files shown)


arkon:
I'm a bit iffy on this experience, actually. Maybe we could add a listener to show it if the `AppBarLayout` is fully expanded?

ivaniskandar:
I think it's possible even if it's not pretty. I'll check in a bit.

ivaniskandar:
nvm it looks pretty much the same... it's just this listener is called more often.

PR #4926, opened 2021-04-21
Fixed russian locale

BlueCat300:
Fixes the following bug:

Every time a last chapter ends with 1 it will show "1 chapter" instead of the actual value.


CarlosEsco:
All translations are done through weblate. 
https://hosted.weblate.org/projects/tachiyomi/

PR #4956, opened 2021-04-24
Cleanup dual page split

ghostbear:
This is a cleanup of the dual page split logic and also changes so it processes the stream in the same place like J2K.

I have not noticed a performance benefit with this when using dual page split it is basically the same, but I have noticed a performance benefit if we do something like my stencil page idea in #4791 it will be a lot smoother if it processes the stream on IO rather than the main thread

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/util/system/ImageUtil.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/util/system/ImageUtil.kt
@@ -79,13 +79,17 @@ object ImageUtil {
     /**
      * Check whether the image is a double image (width > height), return the result and original stream
      */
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/tachiyomi/util/system/ImageUtil.kt
```diff
@@ -79,13 +79,17 @@ object ImageUtil {
     /**
      * Check whether the image is a double image (width > height), return the result and original stream
```
arkon, line 80: Should update this JavaDoc too.

arkon (commented):
LGTM, just one minor comment.

PR #4959, opened 2021-04-25
Fix download error icon color tint

ivaniskandar:
Sorry I just realized my last PR breaks the error icon tint, so here's the fix..


app/src/main/res/layout/chapter_download_view.xml
```diff
@@ -7,16 +7,6 @@
     android:padding="8dp"
     android:background="?selectableItemBackgroundBorderless">
 
-    <ImageView
-        android:id="@+id/download_icon_border"
-        android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:padding="2dp"
-        android:scaleType="fitXY"
-        app:srcCompat="@drawable/border_circle"
```
arkon, line 16: Can delete this drawable too. Clever to just set the progress indicator to 100. :)

app/src/main/res/drawable/ic_check_circle_24dp.xml
```diff
@@ -2,7 +2,8 @@
     android:width="24dp"
     android:height="24dp"
     android:viewportWidth="24"
-    android:viewportHeight="24">
+    android:viewportHeight="24"
+    android:tint="?android:textColorPrimary">
```
arkon, line 6: Nitpick: `?android:attr/` for consistency.

arkon:
I just remembered that we actually have [an extension function](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/util/view/ImageViewExtensions.kt#L14) for setting a drawable with a specified tint. Probably better to just use that rather than making the icons have specific tint colours.

PR #4976, opened 2021-04-28
Show notification to disable Incognito Mode when it's enabled

ivaniskandar:


app/src/main/java/eu/kanade/tachiyomi/App.kt
```diff
@@ -65,6 +76,33 @@ open class App : Application(), LifecycleObserver, ImageLoaderFactory {
 
         // Reset Incognito Mode on relaunch
         preferences.incognitoMode().set(false)
+
+        // Show notification to disable Incognito Mode when it's enabled
+        baseContext.registerReceiver(disableIncognitoReceiver, IntentFilter(ACTION_DISABLE_INCOGNITO_MODE))
```
arkon, line 81: Could maybe keep this local instead (i.e. with `registerLocalReceiver`)?

app/src/main/res/values/strings.xml
```diff
@@ -771,9 +771,12 @@
     <string name="channel_new_chapters">Chapter updates</string>
     <string name="channel_ext_updates">Extension updates</string>
     <string name="channel_crash_logs">Crash logs</string>
+    <string name="channel_incognito_mode">Incognito</string>
```
arkon, line 774: Could just reuse the existing "Incognito mode" string. Should ideally keep the amount of strings minimal to avoid translation churn.

app/src/main/res/values/strings.xml
```diff
@@ -771,9 +771,12 @@
...
 
     <!-- S Pen actions -->
     <string name="spen_previous_page">Previous page</string>
     <string name="spen_next_page">Next page</string>
 
+    <string name="notification_incognito_text">Disable incognito mode</string>
```
arkon, line 780: Can you group this with the other Incognito mode strings?

app/src/main/java/eu/kanade/tachiyomi/App.kt
ivaniskandar: Uhhhhh I don't think so, since it's used to receive the Intent sent by SystemUI [set here](https://github.com/tachiyomiorg/tachiyomi/pull/4976/files#diff-650565d0685bd1f618386dbce217b777c842a8acc6f2e5ff8ea73e63f1a7531fR98).

I can make the receiver registered only when incognito is enabled if you don't like it registered all the time.

ivaniskandar: Okay it's registered on-demand now

arkon: Eeer right. Evidently I shouldn't be reviewing code right after I wake up.

PR #4996, opened 2021-05-01
Automatic background color for PagerViewer

### LINKED ISSUE #584: Automatic background colour
At the moment, the only options for background colour are black or white. This is fine (after all, white is the right answer almost all of the time), but there are some apps (CBLoader, for example) that have a very basic "automatic" background setting, and I think Tachiyomi could benefit from implementing the same, to prevent white borders on the occasional black page.

The easiest version is to sample a single pixel (usually (0,0), the top-leftmost) of each page, and set the background colour to that, since most manga pages have generous blank margins and so the (0,0) pixel will be the right colour nearly every time.

A slightly more involved version is sampling eight or so pixels from various points around the edges (mostly top and bottom) and checking if more of them are black or white. I don't think that's really necessary, but if you felt like it, it's an idea!

ghostbear:
closes #584

This adds J2K implementation of automatic background color, which was the more preferable implementation https://github.com/tachiyomiorg/tachiyomi/issues/584#issuecomment-827070474

The implementation differs a little bit like it doesn't have J2K feature `Smart (by theme)`. Changed many of the `getPixel` function calls to be in variables instead. Also tweaked the code a little bit but it works the same way.

WebtoonViewer will see that `Automatic` is selected but it will use `Black`

**WARNING: THE VIDEO CONTAINS SOME NSFW CONTENT**

https://user-images.githubusercontent.com/6576096/116786783-d7fb6c00-aaa0-11eb-9d8b-9821799fc550.mp4

^Left is Tachiyomi and right is J2K, using the same implementation

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerConfig.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerConfig.kt
@@ -23,6 +23,9 @@ class PagerConfig(
     preferences: PreferencesHelper = Injekt.get()
 ) : ViewerConfig(preferences, scope) {
 
+    var automaticBackground = false
+        private set
+
     var dualPageSplitChangedListener: ((Boolean) -> Unit)? = null
 
     var imageScaleType = 1
@@ -35,6 +38,9 @@ class PagerConfig(
         private set
 
     init {
+        preferences.readerTheme()
+            .register({ automaticBackground = it == 3 }, { imagePropertyChangedListener?.invoke() })
+
         preferences.imageScaleType()
             .register({ imageScaleType = it }, { imagePropertyChangedListener?.invoke() })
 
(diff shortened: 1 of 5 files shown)


Cerxi:
I'm a little confused, in #584 users voted in favour of the implementation without split top/bottom, but this appears to have it? I admit I'm not exactly sure what "J2K" is.

ghostbear:
> I admit I'm not exactly sure what "J2K" is.

J2K is an alternative version of Tachiyomi that has automatic background and some other additional features

---


> in #584 users voted in favour of the implementation without split top/bottom, but this appears to have it?

Yes, the J2K implementation (which was the favored one) does have the split background color but it more on pages that transition from a white to a black background or vice versa. I admit that I didn't choose the greatest chapter to demonstrate with, in #584 else it would have been shown in that video too. But there are cases where a background split makes sense

For example:
| Auto | Manual | Manual |
|---|---|---|
| Screenshot_1620199141 | Screenshot_1620199154 | Screenshot_1620199526 | 
| This is a great example where it makes sense because the background of the page transitions from white to black which makes the whole reader more immersive | But if we only have one color it's a bit weird looking. Here it looks weird at the bottom  | And here it looks weird at the top |

There are people who like this, you may not be one of them. But there are places where it makes sense to have a split background color

---

But the J2K implementation is not perfect 

For example:
| Auto | Manual |
|---|---|
| Screenshot_1620199085 | Screenshot_1620199098 |
| Here it chooses to have a black bottom because it more black than white along the bottom  | But we can clearly see that it's a white page |


But comparing J2K implementation to the other implementation it will do these split background color much less

PR #5033, opened 2021-05-06
[IDEA] change default to daily backup

OncePunchedMan:
This prevents some frustration over lost manga libraries.

It's not THE solution but the first step imo


arkon:
We'd need to ensure we actually have storage permissions first.

This could be changed maybe after we deal with the underlying storage changes, since we'd likely have to prompt the user to grant permissions to a single folder for that anyway (and we can enable the backups at that point if not set).

Riztard:
make it weekly by default? for someone who doesnt even know, doesnt really need daily backup, weekly is enough imo

PR #5037, opened 2021-05-07
Added update titles function

simakover:
Added update titles fuction, logic is same as update cover. 
Im have an issue a long ago with my manga list, there are many mangas have wrong titles. Im dont know deserve this changes pull or not, but im made it) 
Also this is my first try in andriod dev\forking, so any advices welcome

### DIFF
--- a/app/src/main/res/values-ru/strings.xml
+++ b/app/src/main/res/values-ru/strings.xml
@@ -529,7 +529,6 @@
     </plurals>
     <string name="battery_optimization_setting_activity_not_found">Не удалось открыть настройки устройства</string>
     <string name="pref_refresh_library_covers">Обновить обложки библиотечной манги</string>
-    <string name="pref_refresh_library_titles">Обновить заголовки библиотечной манги</string>
     <plurals name="restore_completed_message">
         <item quantity="one">Выполнено за %1$s с %2$s ошибкой</item>
         <item quantity="few">Выполнено за %1$s с %2$s ошибками</item>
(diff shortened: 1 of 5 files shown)


paulo27ms:
Is updating titles generally a good idea? Does this break download folders?

app/src/main/res/values-ru/strings.xml
```diff
@@ -529,6 +529,7 @@
     </plurals>
     <string name="battery_optimization_setting_activity_not_found">Не удалось открыть настройки устройства</string>
     <string name="pref_refresh_library_covers">Обновить обложки библиотечной манги</string>
+    <string name="pref_refresh_library_titles">Обновить заголовки библиотечной манги</string>
```
Riztard, line 532: i believe for other lang strings, you can only change it through weblate
or it will break something

Riztard:
why not just use refresh metadata when update library, did it not work? 🤔

simakover:
> why not just use refresh metadata when update library, did it not work? 🤔

not working for titles

app/src/main/res/values-ru/strings.xml
simakover: revert changes in commit 009dc4539c4f6f5f1eb927f04b67146b9b2ee486

simakover:
Looks like after changing app/src/main/java/eu/kanade/tachiyomi/source/model/SManga.kt with adding title it start updating via metadata, but sometimes im have error: "lateinit property title has not been initialized"

PR #5049, opened 2021-05-10
Add Komga as an unattended track service

gotson:
This PR is an attempt to add tracking support between Tachiyomi and Komga (server).

It works quite differently from regular tracking services. The Komga tracker can only be used for mangas that are using the Komga source. It will use the unique IDs of the manga to match it on the server, so no search dialog is necessary. A manga will be either a match, or not.

It also does not need to login to the tracking service. The TrackService will use the same OkHttp client than the Komga extension, and as such will be able to access the session cookie for a particular domain. This has the benefit of having a single Komga tracker than can track multiple instances of the Komga extension, because it relies on the URL of the manga to find the correct server and session cookie.

~~It is still a work in progress, i would like to add:~~
- ~~support for read lists. Read Lists can be added as manga, so they should be trackable also. At the moment the code doesn't handle read lists at all, and would not find any match for a read list.~~
- ~~auto-bind when adding a Komga source manga to the Tachiyomi library. Given there is no need for user interaction, auto-bind should be achievable.~~
- ~~2-way sync. This was proposed in a [PR](https://github.com/tachiyomiorg/tachiyomi/pull/3705).~~

Feedbacks on this WIP PR is appreciated, as I am not very familiar with Android development!

UPDATE:
- it requires Komga `0. 95.2+`
- ir requires the Komga Tachiyomi extension `1.2.24+`


gotson:
I have added auto-bind when adding to library. This will attempt to match the manga with any UnattendedTrackService, and can be disabled by a new preference in the Tracking settings.

Screenshot 2021-05-10 at 15 12 25

gotson:
I have added 2-way sync for read progress. Chapters are marked as read in Tachiyomi upon tracker refresh. This will only apply to unattended tracking services though.



https://user-images.githubusercontent.com/2139133/117624060-12eb5700-b1a7-11eb-9e90-b42ee7081f02.mp4

ghostbear:
> I have added 2-way sync for read progress. Chapters are marked as read in Tachiyomi upon tracker refresh. This will only apply to unattended tracking services though.

Okay nice 👍

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
```diff
@@ -749,6 +758,14 @@ class MangaPresenter(
         }
     }
 
+    private fun syncChaptersRead(lastChapterRead: Int) {
+        val sortedChapters = chapters.sortedBy { it.chapter_number }
+        sortedChapters.forEachIndexed { index, chapter ->
+            chapter.read = index < lastChapterRead
```
goldbattle, line 764: If you read chapters out of order, then this will mark ones read as unread?

goldbattle:
This PR looks very promising and opens a way for 2way since for services which provide tracking in a very seamless manor. I am a big fan. I was wondering about your comment:
```
The TrackService will use the same OkHttp client than the Komga extension, and as such will be able to access the session cookie for a particular domain
```

Can you point to me where this is the case?
The client build is from the tracking service [here](https://github.com/tachiyomiorg/tachiyomi/pull/5049/files#diff-81a344ab9e41c8464f4c53a122c348b8f623748da273880c8b447daf9c532c0bR27-R30), and I only see the interaction directly with the extension [here](https://github.com/tachiyomiorg/tachiyomi/pull/5049/files#diff-3abcd51424f6288497b857220699c9b184308e580b01d2ced558cf681105256eR88-R104).

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
goldbattle: True, and that is understandable. A thought would be to add a list of "read" chapters ids to the `Track` model which `UnattendedTrackService`s only use to sync per-chapter read status. For example in the [getTrackSearch()](https://github.com/tachiyomiorg/tachiyomi/pull/5049/files#diff-b7b82e522af71b0f5f674816ae2fe443db46cc18decfb2b3b37878dd851b5071R21-R62) this custom field could be set to an array of chapter ids which have been read. Going the other way from Tachi to Komga api the [updateProgress()](https://github.com/tachiyomiorg/tachiyomi/pull/5049/files#diff-b7b82e522af71b0f5f674816ae2fe443db46cc18decfb2b3b37878dd851b5071R64-R75) to be changed to just setting a "read / unread" boolean per chapter (unless you also track it as a single chapter number).

gotson: I just ran into that very case myself. For series with specials, i name the files so they end up last in the directory, so they don't mess up the automatic numbering of normal issues when adding new ones. This breaks the current tracking, a workaround is to keep those specials as unread for now.

I will work on a new branch with your suggestion, so we can line up another PR for later 👍🏻

gotson:
I have changed a bit the logic on server side for the computation of the last read book, now it will return the index of the last read book in a continuous fashion. So if a series has 10 books, and you read 1-3 and 6, it will still return 3. That way the read books out of order will not mess up the count.

The server will only mark books as read when receiving tracker update from Tachiyomi. I will not mark unread books as unread.

How 2-way sync works:
- retrieve the remote track with the `lastRead`
- mark local chapters as read if they were unread for chapters < `lastRead`
- update the remote tracker with a new track, composed of read chapters

So if you had books 1-2 marked as read on server, and 1-4 on Tachiyomi:
- retrieve remote track with `lastRead` = 2
- mark 1-2 as read
- update remote tracker with lastRead = 4

app/src/main/java/eu/kanade/tachiyomi/data/track/komga/KomgaApi.kt
```diff
@@ -0,0 +1,80 @@
...
+                }
+            } catch (e: Exception) {
+                Timber.w(e, "Could not get item: $url")
+                null
+            }
+        }
+
+    suspend fun updateProgress(track: Track): Track {
+        val progress = ReadProgressUpdateDto(track.last_chapter_read)
+        val payload = Json.encodeToString(ReadProgressUpdateDto.serializer(), progress)
```
arkon, line 60: There should be an extension function that lets you drop the explicit serializer.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourcePresenter.kt
```diff
@@ -260,6 +266,27 @@ open class BrowseSourcePresenter(
             manga.removeCovers(coverCache)
         } else {
             ChapterSettingsHelper.applySettingDefaults(manga)
+
+            if (prefs.autoAddTrack()) {
```
arkon, line 270: Nitpick: extract this out into a function

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackHolder.kt
```diff
@@ -49,6 +49,12 @@ class TrackHolder(private val binding: TrackItemBinding, adapter: TrackAdapter)
                 if (track.total_chapters > 0) track.total_chapters else "-"
             binding.trackStatus.text = item.service.getStatus(track.status)
             binding.trackScore.text = if (track.score == 0f) "-" else item.service.displayScore(track)
+            if (item.service.getScoreList().isEmpty()) {
+                with(binding.trackScore) {
+                    text = "Not supported"
```
arkon, line 54: Should be localized.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSheet.kt
```diff
@@ -70,7 +78,33 @@ class TrackSheet(
...
-        TrackSearchDialog(controller, item.service).showDialog(controller.router, TAG_SEARCH_CONTROLLER)
+
+        if (item.service is UnattendedTrackService) {
+            if (item.track != null) {
+                controller.presenter.unregisterTracking(item.service)
+                return
+            }
+
+            if (!item.service.accept(sourceManager.getOrStub(manga.source))) {
+                controller.presenter.view?.applicationContext?.toast("Source is not supported")
```
arkon, line 89: Should be localized.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSheet.kt
```diff
@@ -70,7 +78,33 @@ class TrackSheet(
...
+            }
+
+            launchIO {
+                try {
+                    val track = item.service.match(manga)
+                    if (track != null) {
+                        controller.presenter.registerTracking(track, item.service)
+                        return@launchIO
+                    }
+                    withUIContext { controller.presenter.view?.applicationContext?.toast("No match found") }
```
arkon, line 100: Should be localized.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSheet.kt
```diff
@@ -70,7 +78,33 @@ class TrackSheet(
...
+                } catch (e: Exception) {
+                    withUIContext { controller.presenter.view?.applicationContext?.toast("No match found") }
```
arkon, line 102: Should be localized.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSheet.kt
```diff
@@ -95,7 +129,7 @@ class TrackSheet(
 
     override fun onScoreClick(position: Int) {
         val item = adapter.getItem(position) ?: return
-        if (item.track == null) return
+        if (item.track == null || item.service.getScoreList().isEmpty()) return
```
arkon, line 132: What exactly is the expectation here as the user? I just can't set the score?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsTrackingController.kt
```diff
@@ -76,9 +86,14 @@ class SettingsTrackingController :
             {
                 onClick {
                     if (service.isLogged) {
-                        val dialog = TrackLogoutDialog(service)
-                        dialog.targetController = this@SettingsTrackingController
-                        dialog.showDialog(router)
+                        if (service is NoLoginTrackService) {
```
arkon, line 89: Is this actually necessary? I don't see a problem with confirming even if the login is basically a no-op.

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterTrackSync.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    // update remote
+    remoteTrack.last_chapter_read = localLastRead
+
+    launchIO {
+        try {
+            service.update(remoteTrack)
+            db.insertTrack(remoteTrack).executeAsBlocking()
+        } catch (e: Throwable) {
+        }
```
arkon, line 34: Probably want to at least log something?

arkon (changes requested):
- Is there a simpler icon for the tracker? It looks out of place.
- Do you expect the API on Komga's side to change in the future? You'll be tied tightly with app updates, which likely won't move as quickly as you need it to.

gotson:
> * Is there a simpler icon for the tracker? It looks out of place.

I took the one from the extension. The original Komga logo is round, you can see it at https://komga.org/. I also have the [original SVG logo](https://github.com/gotson/komga/blob/master/res/komga_text_as_path.svg), so I could probably do something with it. I'm not sure however what's so out of place when looking at the others, they don't seem to have a common trait. Please let me know !

> * Do you expect the API on Komga's side to change in the future? You'll be tied tightly with app updates, which likely won't move as quickly as you need it to.

The element (series or readlist) API is solely used to get the _title_, _summary_ and _publishing status_, but i think only the title is used in the track sheet, the summary and publishing status are used in the interactive dialog to choose the right manga to track, but that is not used here. Those APIs have been quite stable since their introduction.

There is a newly added API endpoint solely for tachiyomi tracking (`/read-progress/tachiyomi`), so the impact would be quite limited.

---

I will have a look at all your comments on the code now.

app/src/main/java/eu/kanade/tachiyomi/data/track/komga/Komga.kt
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    override fun getCompletionStatus(): Int = COMPLETED
+
+    override fun getScoreList(): List<String> = emptyList()
+
+    override fun displayScore(track: Track): String = ""
+
+    override suspend fun add(track: Track): Track {
+        TODO("Not yet implemented: add")
```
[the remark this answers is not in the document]
gotson: the `add` function is defined in `TrackService` but never called in any common code. It is only called from inside each tracker implementation.

Screenshot 2021-05-20 at 09 22 45

I would suggest to remove it from `TrackService`.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSheet.kt
gotson: If the score list is empty, this will crash the app. I noticed this while coding, it probably went unnoticed since all other trackers have score.

This is because of https://github.com/tachiyomiorg/tachiyomi/blob/535abcbb8b2197faef77a6b0ba56aa4309067d4c/app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/SetTrackScoreDialog.kt#L66

There may be a better way to avoid that crash directly in `SetTrackScoreDialog` though, but given it's a pop-up dialog, we should not open the dialog at all if there's no score.

In addition to the above change, which is really to prevent a crash in case we incorrectly change other parts of the code, I added [this](https://github.com/gotson/tachiyomi/blob/622e68270b2ce94673e30d910ce65f5b206747ce/app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackHolder.kt#L55) to disable the score button entirely.

If you want to keep only the disabled one, i can remove that additional check.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsTrackingController.kt
gotson: I thought it would make the experience better. There's i no point asking to logout if you didn't ask for login. That makes the preference a simple toggle at each tap.

If you think it clutters the code i can remove it.

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterTrackSync.kt
gotson: Good catch, i added some logs for the execption.

gotson:
I have rebased, fixing some merge conflicts, and addressed the majority of your comments @arkon

gotson:
I also changed the logo for a webp version generated from the original (rounded) logo.

app/src/main/java/eu/kanade/tachiyomi/data/track/komga/Komga.kt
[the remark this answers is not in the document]
arkon: Yeah, the tracker interface isn't that great in general.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSheet.kt
arkon: It may be better to just make it not render the score selection at all if there's trackers that don't support it. The previous assumption was that all trackers support that.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsTrackingController.kt
arkon: I'm kind of split on it, but I'm fine with keeping it.

arkon (approved):

arkon:
Can followup in separate PRs on the comments if necessary.

PR #5103, opened 2021-05-18
Fix unofficial extensions being labeled as "Obsolete" instead of "Unofficial"

datreesezcup:
This PR addresses a small, purely visual issue in the extension browser that due to the ordering of the statements in a "when" clause in ExtensionHolder.kt, an unofficial extension's label will read "OBSOLETE", when it should read "UNOFFICIAL"

Before:
before-fix

After:
after-fix


arkon:
Thanks!

PR #5108, opened 2021-05-18
confirmation mark as read/unread in library

### LINKED ISSUE #5100: [Feature Request] confirm mark as read/unread in library
people may accidentally press wrong button like mark as read/unread

## What/Requirements
add confirmation to continue for mass mark as read/unread or undo button

image

Riztard:
add dialog confirmation for mark as read/unread in library
lol i manage to add it myself 
Closes #5100

<table>
<tr>
	<td>
	<td>
</table>


joseph619:
Seems annoying to do this for individual chapters

CrepeTF:
If you've already committed to go into action select mode wouldn't that in itself be a confirmation? And on top of that you need to actually select read/unread

CrepeTF:
It might make more sense to have an undo snackbar instead

arkon:
Agreed, undo would be less annoying.

PR #5109, opened 2021-05-19
Ripple overhaul

Soitora:
Tweaks and tries to unify the ripples.

It's not 100% perfect but, I did the best I could.
Feel free to 👎🏻 if I spent 20 hours to learn the best spaghetti recipe.
Please comment if I can improve some bits, I really am not familiar with Kotlin.

Some areas (tab ripples) I noticed it isn't 100% the exact same HEX, but it still follows the same variable now (as all things should).
I think these areas has different opacity set for the ripples but I did not find these, but it's not that noticable.

I added a different ripple variable for the BottomNavigation, I think it looks much better now with a little hint of color, instead of just different shades of somewhere between Black/White.

**EDIT** Could potentially use somewhat colored ripples/selection highlights for other things as well.

| New | Old |
| ------------- | ------------- |
| new-dark  | old-dark  |
| new-light | old-light |

https://user-images.githubusercontent.com/10836780/118740275-babdf000-b84b-11eb-961f-8dac748b16b5.mp4


Soitora:
Reverted the two layout changes I had, because the ripple was circular without it.
Did not revert the pin button layout changes.

| Changed pin layout | Old behavior |
| --------------------- | --------------- |
| changed-behavior | old-behavior |

arkon:
I'm going to be honest with you, reviewing this is a pain so I'm just going to merge it and you can handle fixing bugs if they get reported.

Soitora:
Fully understandable, it was a huge pain making too.

Let's just pray no one has issues 🌚

PR #5133, opened 2021-05-22
Remove material-design-dimens

ivaniskandar:
Also added paddings to some buttons to make its touch target larger

### DIFF
--- a/app/src/main/res/layout/navigation_view_radio.xml
+++ b/app/src/main/res/layout/navigation_view_radio.xml
@@ -16,7 +16,7 @@
         android:clickable="false"
         android:gravity="center_vertical|start"
         android:maxLines="1"
-        android:paddingStart="@dimen/material_component_lists_icon_left_padding"
+        android:paddingHorizontal="16dp"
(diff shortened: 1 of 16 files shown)


app/src/main/res/layout/navigation_view_radio.xml
```diff
@@ -16,7 +16,7 @@
         android:clickable="false"
         android:gravity="center_vertical|start"
         android:maxLines="1"
-        android:paddingStart="@dimen/material_component_lists_icon_left_padding"
+        android:paddingHorizontal="16dp"
```
arkon, line 19: Isn't this SDK 26+?

ivaniskandar: It should be backwards compatible as long as aapt2 is used. Check out [aapt2 changelog](https://android.googlesource.com/platform/frameworks/base/+/master/tools/aapt2/readme.md).

arkon: Ah nice, I never heard about that.

PR #5134, opened 2021-05-22
Reinstate Proguard/R8

ivaniskandar:
These are month old stats but it should be similar.
According to APK analyzer
Before: 18044 classes with 113387 methods in 3 dex files
After: 12604 classes with 68010 methods in 2 dex files
APK download size is reduced by 2.8 MB

And I believe this should also be enabled on preview builds, but it uses debug build type so we can either make a new build type for previews with R8 enabled or a separate debug build type with R8 disabled.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -55,6 +55,15 @@ android {
         named("debug") {
             versionNameSuffix = "-${getCommitCount()}"
             applicationIdSuffix = ".debug"
+
+            isShrinkResources = true
+            isMinifyEnabled = true
+            proguardFiles("proguard-android-optimize.txt", "proguard-rules.pro")
+        }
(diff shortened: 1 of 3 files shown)


arkon:
> And I believe this should also be enabled on preview builds, but it uses debug build type so we can either make a new build type for previews with R8 enabled or a separate debug build type with R8 disabled.

Yeah, how the dev builds are set up is something that inorichi regrets doing IIRC, and something that we're kind of locked into if we want to keep updates working properly.

I'm fine with adding another build type (`dev`) or something for local development. I'd also be fine with just having it always on if it doesn't slow the builds down _too_ much. I do find it useful for testing sometimes.

ivaniskandar:
`debug` builds is now minified and added new `debugFull` build type that's not minified for faster build when needed.

Issue #5135, opened 2021-05-22
[Bug] Library button dissapear


Riztard:
## Device information
* Tachiyomi version: r3368
* Android version: 10
* Device: redmi 8

## Steps to reproduce
1. search something in library
2. open manga
3. back to library
4. remove search input

### Expected behavior
button on top show back

### Actual behavior
button replaced by unusable three dot 

## Other details
didnt happen to stable

<table>
<tr>
	<td>Before
	<td>After
<tr>
	<td><img src="https://cdn.discordapp.com/attachments/825346400953827349/845612144254124043/IMG_20210522_174038.jpg">
	<td><img src="https://cdn.discordapp.com/attachments/825346400953827349/845612285681598474/IMG_20210522_174111.jpg">
</table>

AntoineGS:
Good catch ;)

PR #5152, opened 2021-05-23
Prevent view from being removed if a to is found (#5135)

AntoineGS:
Introduced in c68e7c8da78309a0b091c11379a8da5c6222572d


arkon:
Thanks!

ghostbear:
This PR seems to have introduced some weird transition artifacts

https://user-images.githubusercontent.com/6576096/119270193-d2a2c480-bbfb-11eb-9712-27fc79d2ae55.mp4

On some views the background is transparent which makes it so one can see the library when transitioning. This doesn't seem to be present on 98461f9bcab66c5a78cabf35d758a4166a7b282e which was the commit before this

AntoineGS:
I just created a new PR (#5189), it seems removing the view from the container causes this issue when the text view is expanded.
As a workaround we can animate the from view from 0f to 0f and it seems to work fine.
I haven't found a cleaner way to do this.

PR #5164, opened 2021-05-24
Refactor TrackService.kt to remove unused `add` function

gotson:
removed the add function from TrackService.kt as it's not used except within the tracker implementations


app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
```diff
@@ -130,7 +130,7 @@ class Anilist(private val context: Context, id: Int) : TrackService(id) {
         }
     }
 
-    override suspend fun add(track: Track): Track {
+    suspend fun add(track: Track): Track {
```
arkon, line 133: Can you make these private?

arkon (changes requested):

arkon (approved):

PR #5166, opened 2021-05-24
Background tracker update during Library update

gotson:
This adds a new preference in the Library screen to enable trackers update during Library update. It defaults to off. Should probably be merged after #5163 to avoid issues with AniList.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -21,6 +21,7 @@ import eu.kanade.tachiyomi.data.library.LibraryUpdateService.Companion.start
 import eu.kanade.tachiyomi.data.notification.Notifications
 import eu.kanade.tachiyomi.data.preference.PreferencesHelper
 import eu.kanade.tachiyomi.data.track.TrackManager
+import eu.kanade.tachiyomi.data.track.TrackService
 import eu.kanade.tachiyomi.data.track.UnattendedTrackService
 import eu.kanade.tachiyomi.source.SourceManager
 import eu.kanade.tachiyomi.source.model.SManga
@@ -210,14 +211,10 @@ class LibraryUpdateService(
             stopSelf(startId)
         }
         updateJob = ioScope.launch(handler) {
-            when {
-                target == Target.CHAPTERS && preferences.autoUpdateTrackers() -> {
-                    updateChapterList()
-                    updateTrackings()
-                }
-                target == Target.CHAPTERS -> updateChapterList()
-                target == Target.COVERS -> updateCovers()
-                target == Target.TRACKING -> updateTrackings()
+            when (target) {
+                Target.CHAPTERS -> updateChapterList()
+                Target.COVERS -> updateCovers()
+                Target.TRACKING -> updateTrackings()
             }
         }
         updateJob?.invokeOnCompletion { stopSelf(startId) }
@@ -277,6 +274,7 @@ class LibraryUpdateService(
         val newUpdates = mutableListOf<Pair<LibraryManga, Array<Chapter>>>()
         val failedUpdates = mutableListOf<Pair<Manga, String?>>()
         var hasDownloads = false
+        val loggedServices by lazy { trackManager.services.filter { it.isLogged } }
 
         mangaToUpdate.forEach { manga ->
             if (updateJob?.isActive != true) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -210,10 +210,14 @@ class LibraryUpdateService(
...
         updateJob = ioScope.launch(handler) {
-            when (target) {
-                Target.CHAPTERS -> updateChapterList()
-                Target.COVERS -> updateCovers()
-                Target.TRACKING -> updateTrackings()
+            when {
+                target == Target.CHAPTERS && preferences.autoUpdateTrackers() -> launch {
+                    updateChapterList()
+                    updateTrackings()
+                }.join()
```
jobobby04, line 217:
```suggestion
                target == Target.CHAPTERS && preferences.autoUpdateTrackers() -> {
                    updateChapterList()
                    updateTrackings()
                }
```

jobobby04: No need to launch and join it, you can just use a block

gotson: Thanks!

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -210,10 +210,14 @@ class LibraryUpdateService(
...
         updateJob = ioScope.launch(handler) {
-            when (target) {
-                Target.CHAPTERS -> updateChapterList()
-                Target.COVERS -> updateCovers()
-                Target.TRACKING -> updateTrackings()
+            when {
+                target == Target.CHAPTERS && preferences.autoUpdateTrackers() -> {
+                    updateChapterList()
+                    updateTrackings()
+                }
```
arkon, line 217: With this implementation, it'd update all the manga chapters/metadata and then all the tracking in a sequential fashion. I wonder if it'd be better to update them together (i.e. sets of [chapters/metadata/tracking] in sequence)? That would at the very least make it not seem like it's updating twice.

gotson: Indeed. I can try to refactor the existing code to pull the tracking update per manga in a private function, and reuse that conditionally inside `updateChapterList` to also perform tracking update if the preference is `true`.

gotson: The latest change should address that.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -273,6 +274,7 @@ class LibraryUpdateService(
         val newUpdates = mutableListOf<Pair<LibraryManga, Array<Chapter>>>()
         val failedUpdates = mutableListOf<Pair<Manga, String?>>()
         var hasDownloads = false
+        val loggedServices by lazy { trackManager.services.filter { it.isLogged } }
```
arkon, line 277: Should be able to remove the instance at L407 since you have this now.

gotson: They are different ones:
- L277 is for `updateChapters` and is `lazy` because it can not be used if you didn't tick the preference.
- L407 is for `updateTrackings` and is not lazy given we need it every time.

I thought of pulling it at the class level, but it wouldn't be dynamic anymore, so logging in/out of trackers would require an app restart to update the `loggedServices`. With the current way we get the logged services zero or once per job.

arkon: Whoops, I didn't expand the context and thought this was the class level. I'm pretty sure it'd get reinstantiated for different service invocations, but I guess this is fine too.

arkon (approved):

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
gotson: You are right, i just tested with the debugger, the `LibraryUpdateService` is instanciated every time. I can pull that to class level in another PR if you want.

arkon: It's fine for now. The service needs some unrelated cleanup/changes anyway, so maybe it'll get done at that point...

PR #5174, opened 2021-05-24
About screen visual adjustments

ivaniskandar:
[Preview](https://user-images.githubusercontent.com/12537387/119382361-1f65c880-bcec-11eb-987f-a219ede88c93.png)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/more/AboutController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/more/AboutController.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.ui.more
 
 import android.app.Dialog
+import android.os.Build
 import android.os.Bundle
 import androidx.core.os.bundleOf
 import androidx.preference.PreferenceScreen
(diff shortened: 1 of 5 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/more/AboutController.kt
```diff
@@ -15,91 +16,59 @@ import eu.kanade.tachiyomi.ui.base.controller.DialogController
...
-                    "Stable ${BuildConfig.VERSION_NAME} (${getFormattedBuildTime()})"
-                }
+                key = "pref_about_check_for_updates"
+                titleRes = R.string.check_for_updates
 
-                onClick {
-                    activity?.let {
-                        val deviceInfo = CrashLogUtil(it).getDebugInfo()
-                        it.copyToClipboard("Debug information", deviceInfo)
-                    }
```
arkon, line 65: Can you keep this behaviour of copying the debug info when tapping the version (or the header)?

Soitora: Maybe add a text button above What's New that says "Copy debug info" or something?

CrepeTF:
I'm not really a fan of having the logo header not only in the initial More page but also the About page.

Also as for the design of the About page, I'd much rather see it be kept the same, but just compacting the socials to buttons as you've done.

Especially keeping the Check for updates button too.

vengiare:
Keep Version button even with version number on the header. Header version would be (Stable, 0.10.12 / Preview, r3417) and Version button would have the extra details.

Soitora:
> I'm not really a fan of having the logo header not only in the initial More page but also the About page.

Not having the logo in there severely makes it an empty piece of shell, and is also quite necessary for the version print

> Especially keeping the Check for updates button too.

That button isn't gone, it's just not displayed in local environment

app/src/main/java/eu/kanade/tachiyomi/ui/more/AboutController.kt
ivaniskandar: > Can you keep this behaviour of copying the debug info when tapping the version (or the header)?

Ok will do

vengiare:
mock

mockup for my suggestion

T3sT3ro:
I'll add my two cents:
- I would rearrange "More" page to the same hierarchy as in many IDEs and other programs `Help > [rhings relating to help, info, about, update etc.]`. Maybe instead of naming it "Help" name it "Info & Updates":
```
...
Options ..
Help  - current `More > Help` item would become a category like Settings.
  - social buttons 
  - Version info <hash> [if not under the icon in header]
  - Check For Updates [+refresh icon]
  - What's new [+external URL icon]
  - About [+external URL icon]
  - Online Help [+external URL icon] - current `More > Help` item
  - Open Source Licenses [+external URL icon]
```

- Version under logo in the header looks the best imo (not wasting space)
- "What's new" should open the same dialog that pops up after updating to the newest version instead of redirecting to browser
- Embedded frame with latest "What's new" frame would also look cool
- Whenever app redirects to external link via webview or something, the "external URL icon" should be slightly grayed out next to it
- I'd add "sponsor" button in there
- If embedded dialog/frame is used for "What's New" another link for "Changelog [+external URL icon]"

In my opinion this is more standard to what other programs do

arkon:
Seems silly that this was somehow a controversial change.

Superseded by 8af8c57bb414beafb4eed2bfb448b072f2fe60e4

Soitora:
Only one person voiced against this, which should be disregarded as the complaint isn't really towards a feature @ivaniskandar implemented, but to an already existing design in the About page (The icon section).

I think the version under logo is a great addition design-wise.

PR #5182, opened 2021-05-25
Added radio buttons filter type (implementation) for browsing manga

datreesezcup:
# Summary
This PR adds the app implementation of radio buttons as a type of filter that can be used to filter manga when browsing a source. This is intended to be an alternative to the Select filter, as sometimes it is not practical to use a drop-down for a large series of mutually exclusive options.

The RadioGroup filter acts similar to the Group or Sort filters, where a clickable header expands to show the available options. The filter's options are provided by an array of strings, and the filter's state is Int, indicating the index of the currently selected item.

# Preview

radiofilter_preview

# Addendum
- Since this PR is essentially a feature for extension developers, I realize that adaption of this feature would require an update to the extensions library as well in order to include the stub for the filter.
- The preview image was created by temporarily adding a RadioGroup filter to the Local Source, since it is the only usable source built into the core app.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/model/Filter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/model/Filter.kt
@@ -20,7 +20,7 @@ sealed class Filter<T>(val name: String, var state: T) {
 
     abstract class Group<V>(name: String, state: List<V>) : Filter<List<V>>(name, state)
 
-    abstract class RadioGroup(name: String, val values: Array<String>) : Filter<Int>(name, 0)
+    abstract class RadioGroup(name: String, val values: Array<String>, state: Int = 0) : Filter<Int>(name, state)
 
     abstract class Sort(name: String, val values: Array<String>, state: Selection? = null) :
         Filter<Sort.Selection?>(name, state) {
--- a/app/src/main/java/eu/kanade/tachiyomi/source/model/Filter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/model/Filter.kt
@@ -20,6 +20,8 @@ sealed class Filter<T>(val name: String, var state: T) {
 
     abstract class Group<V>(name: String, state: List<V>) : Filter<List<V>>(name, state)
 
+    abstract class RadioGroup(name: String, val values: Array<String>) : Filter<Int>(name, 0)
+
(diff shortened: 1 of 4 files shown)


CarlosEsco:
While a decent alternative to the drop down, this would need an equivalent update for the stub in the extensions repository, as well as would break backwards compatibility of the extensions, if anyone is using an older version of tachiyomi that does not have this.

h-hyuuga:
> While a decent alternative to the drop down, this would need an equivalent update for the stub in the extensions repository, as well as would break backwards compatibility of the extensions, if anyone is using an older version of tachiyomi that does not have this.

Since the radio group is essentially an expanded Select, would it be feasible to have it subclass that filter?

So old versions of Tachiyomi would just fall back onto a select.

datreesezcup:
> 
> 
> > While a decent alternative to the drop down, this would need an equivalent update for the stub in the extensions repository, as well as would break backwards compatibility of the extensions, if anyone is using an older version of tachiyomi that does not have this.
> 
> Since the radio group is essentially an expanded Select, would it be feasible to have it subclass that filter?
> 
> So old versions of Tachiyomi would just fall back onto a select.

That is a great idea, and after a quick re-classing, it works just fine to make the filter be a subclass of Select.

Commit pushed making this change

arkon:
> sometimes it is not practical to use a drop-down for a large series of mutually exclusive options

Isn't a dropdown the more practical option in that case? A long list of radio buttons isn't ideal. If anything, it'd be more suited for shorter lists of options.

datreesezcup:
> Isn't a dropdown the more practical option in that case? A long list of radio buttons isn't ideal. If anything, it'd be more suited for shorter lists of options

Now that I think about it, I guess that is true. 
(Especially because on a new option being selected, *every* radio item in the group has to be notified of the new state so it can update itself. Hopefully this wouldn't cause issues unless you're dealing with an excessive amount of items, but I haven't thought to check the performance of this)

Plus the Select shows its current value as part of the filter, whereas the RadioGroup does not
(Though off the top of my head I think the header of the RadioGroup could be made to show the current value).

Though one benefit of this radio group system is that options should be able to contain more text before being cut off.
Even the other day while I was working on an extension I found that several of the tags I listed would not fit in the small window the dropdown creates to store text.

I suppose that in the very end it is more of a design preference, since this filter is effectively the exact same as the Select filter, just presented differently.

PR #5187, opened 2021-05-25
Display the currently active restrictions in the library update preference

### LINKED ISSUE #4957: [Feature Request] <Show update restriction>
When you set restrictions to the categories who will be update, the restritions is showed , like this: 
image

However when you set restricions for when the update will happen, like this:
image

The restrictions is not showed:
image

## What/Requirements
This feature is just to show the restricions for when the update will happen, like the frist screenshot

gotson:
Closes #4957

Screenshot 2021-05-25 at 14 59 41

Screenshot 2021-05-25 at 14 59 49

Screenshot 2021-05-25 at 14 59 56

I also pulled the `wifi` and `ac` strings into constants.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -227,7 +227,7 @@ class PreferencesHelper(val context: Context) {
 
     fun libraryUpdateInterval() = flowPrefs.getInt(Keys.libraryUpdateInterval, 24)
 
-    fun libraryUpdateRestriction() = prefs.getStringSet(Keys.libraryUpdateRestriction, setOf("wifi"))
+    fun libraryUpdateRestriction() = flowPrefs.getStringSet(Keys.libraryUpdateRestriction, setOf("wifi"))
 
     fun libraryUpdateCategories() = flowPrefs.getStringSet(Keys.libraryUpdateCategories, emptySet())
     fun libraryUpdateCategoriesExclude() = flowPrefs.getStringSet(Keys.libraryUpdateCategoriesExclude, emptySet())
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -227,7 +227,7 @@ class PreferencesHelper(val context: Context) {
 
     fun libraryUpdateInterval() = flowPrefs.getInt(Keys.libraryUpdateInterval, 24)
 
-    fun libraryUpdateRestriction() = flowPrefs.getStringSet(Keys.libraryUpdateRestriction, setOf("wifi"))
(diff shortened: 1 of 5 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsLibraryController.kt
```diff
@@ -164,6 +165,33 @@ class SettingsLibraryController : SettingsController() {
...
+                            }
+                        }
+                    val restrictionsText = if (restrictions.isEmpty()) {
+                        context.getString(R.string.none)
+                    } else {
+                        restrictions.joinToString()
+                    }
+
+                    summary = buildSpannedString {
+                        append(context.getString(R.string.pref_library_update_restriction_summary))
```
arkon, line 186: Maybe could just get rid of this line? Not sure if it's clear enough without it.

gotson: Up to you really.

arkon: Let's get rid of it.

gotson:
Screenshot 2021-05-26 at 08 26 30

I also removed the english string from `strings.xml`, i believe Weblate will remove it from all translations too afterwards.

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
```diff
@@ -227,7 +227,7 @@ class PreferencesHelper(val context: Context) {
 
     fun libraryUpdateInterval() = flowPrefs.getInt(Keys.libraryUpdateInterval, 24)
 
-    fun libraryUpdateRestriction() = prefs.getStringSet(Keys.libraryUpdateRestriction, setOf("wifi"))
+    fun libraryUpdateRestriction() = flowPrefs.getStringSet(Keys.libraryUpdateRestriction, setOf("wifi"))
```
arkon, line 230: Use the constant here?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsLibraryController.kt
```diff
@@ -164,6 +165,31 @@ class SettingsLibraryController : SettingsController() {
...
+                                else -> it
+                            }
+                        }
+                    val restrictionsText = if (restrictions.isEmpty()) {
+                        context.getString(R.string.none)
+                    } else {
+                        restrictions.joinToString()
+                    }
+
+                    summary = buildSpannedString {
```
arkon, line 185: Don't need a spanned string anymore.

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
gotson: Good catch!

arkon (approved):

arkon:
Thanks!

Issue #5154, opened 2021-05-23
[Bug] Missing places for Ripple


Soitora:
# Unfixed

## No results found
Related file: [`EmptyView.kt` #L60](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/widget/EmptyView.kt#L60)
Setting `setRippleColorResource` works but it requires a Color and not just the theme color attribute.

| No results found |
| --------------------------- |
| <img src="https://user-images.githubusercontent.com/10836780/119270062-2a8cfb80-bbfb-11eb-90fd-fb8ce417c9f4.png" height="200" /> |

Both of the images below are affected somewhat by `Theme.Widget.Button.Action`.
This style is imported here on [`EmptyView.kt` #L46](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/widget/EmptyView.kt#L46)

Except I can't find a way to actually set it for the `EmptyView` one.

| EmptyView | Library Buttons |
| ---------- | ------------------- |
| ![EmptyView](https://user-images.githubusercontent.com/10836780/119589096-67f7b180-bdd2-11eb-8455-8688ef7b009e.png) | ![Library Buttons](https://user-images.githubusercontent.com/10836780/119589132-7e9e0880-bdd2-11eb-9f4f-898e87985ba4.png) |

## Material dialogs buttons
After hours of testing, I've yet to manage to succeed with this.
| Material dialogs buttons |
| --------------------------- |
| <img src="https://user-images.githubusercontent.com/10836780/119270106-5c9e5d80-bbfb-11eb-9527-b4ffbe358aad.png" height="200" /> |

# Fixed

## Dialog highlight
Fixed by #5197.

## Source filter dropdown
Fixed by #5197.

## Tracker login pw revealer
Fixed by #5197.

## Download cancel/etc dialog
Fixed by #5197.

vengiare:
https://github.com/afollestad/material-dialogs/blob/main/documentation/CORE.md#ripple-color

> Related doc: [afollestad/material-dialogs](https://github.com/afollestad/material-dialogs/blob/main/README_OLD.md#global-theming)

Have you looked at [Core tutorial?](https://github.com/afollestad/material-dialogs/blob/main/documentation/CORE.md#ripple-color)

PR #5197, opened 2021-05-26
Allow themes to style more objects

Soitora:
Allows themes to have their accent colors affect more places.
For example, the `Unread` badge will have it's color set by `colorSecondary` and `colorOnSecondary`.
> The default themes this will not have any of these colors changed, but it allows the newly added themes to take advantage of the color differences. No old user should thus be confused.

I added and defined a new **Tertiary** color which allows us to also style the `Download` and `Local` badge.
These badges will be styled by `colorTertiary` and `colorOnTertiary`.

This also styled the **Downloaded only** banner, although that was a 50/50 change for me. (Insignificant)

It creates new `AlertDialog` themes for each theme so it uses the accent color correctly, I tried to set just a global one with using an attribute but that caused the app to crash.

I declared a global `android:colorControlHighlight` for the ancient cases where the `rippleColor` does not set it, this fixes 4 out of 6 cases listed on #5154.

`md_ripple_color` was added and defined, which is one of the fixes for the issue listed above. Unfortunately this only changes the selection ripple color and does not make the action button ripple function.

| Library badges (new) | Library badges (old) |
| ----------------------- | ---------------------- |
| lb1 | lb2 |

| Alert dialog (new) | Alert dialog (old) |
| ------------------- | ------------------- |
| ad1 | ad2 |


app/src/main/res/values/attrs.xml
```diff
@@ -22,6 +22,8 @@
     <attr name="colorFilterActive" format="reference|integer"/>
     <attr name="colorBackgroundSplash" format="reference|integer"/>
     <attr name="colorAccentOnPrimary" format="reference|integer"/>
+    <attr name="colorTertiary" format="reference|integer"/>
+    <attr name="colorOnTertiary" format="reference|integer"/>
```
arkon, line 26: I'm not sure about these colours. They're explicitly used for Download badges/banner, so it may be better to just call it `colorDownloaded` or something so they're not misused elsewhere.

arkon: I don't expect it to be used in the future, but who knows. I'll leave it as is for now.

PR #5230, opened 2021-05-29
Use existing vector icon for splash screen

ivaniskandar:


app/src/main/res/drawable/splash_background.xml
```diff
@@ -3,10 +3,6 @@
 
     <item android:drawable="?attr/colorBackgroundSplash" />
 
-    <item>
-        <bitmap
-            android:gravity="center"
-            android:src="@drawable/splash_icon" />
-    </item>
+    <item android:drawable="@drawable/ic_tachi_splash" />
```
arkon, line 6: Is it even still necessary to have these separate drawables? Could the `ic_tachi_splash.xml` contents just be directly in here?

arkon:
For the record, I tested this in an Android 6.0 emulator and it looked fine.

There was some weird stretching issues when I tried this when first implementing it, but that might've been in a lower API level.

app/src/main/res/drawable/splash_background.xml
ivaniskandar: Ah that's because I needed a separate drawable for S new splash screen stuff. I'll merge it back until compileSdk bumped to 31.

PR #5232, opened 2021-05-29
Added category-wise display setting

ghostbear:
Supersedes #3319

https://user-images.githubusercontent.com/6576096/120068358-913d6980-c080-11eb-976e-ad88baf47e30.mp4

The default category uses the preference, `preferences.libraryDisplayMode()`, instead of the flags because it will insert it into the database otherwise

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -226,6 +226,7 @@
 
     <string name="default_category">Default category</string>
     <string name="default_category_summary">Always ask</string>
+    <string name="categorized_display_settings">Per-category display settings</string>
     <plurals name="num_categories">
         <item quantity="one">%d category</item>
         <item quantity="other">%d categories</item>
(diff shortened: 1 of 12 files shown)


FourTOne5:
Can you also add category-wise sort setting with this. For example Category On Hold will have A-Z and Reading will have Last Chapter.

ghostbear:
> Can you also add category-wise sort setting with this. For example Category On Hold will have A-Z and Reading will have Last Chapter.

The PR is already complicated enough so I would rather add that in a separate PR

FourTOne5:
> 
> The PR is already complicated enough so I would rather add that in a separate PR

Well it doesn't matter if you add it in a separate one. Also for reference you should add the Display Settings to Sort and rename it to Category when you add category-wise sort setting so both of them will stay under a single setting menu

app/src/main/res/values/strings.xml
```diff
@@ -226,6 +226,7 @@
 
     <string name="default_category">Default category</string>
     <string name="default_category_summary">Always ask</string>
+    <string name="categorized_display_settings">Per-category display settings</string>
```
arkon, line 229: This might be a bit misleading since it's only the display mode, but maybe we can change in the future so that the toggle:

- When off, is like existing behaviour
- When on, shows chips at the top of each category for sort and display mode (and removes them from the sheet)

PR #5282, opened 2021-06-03
Legacy backup conversion to Kotlin Serialization

jobobby04:
Converted the Legacy backup restore to Kotlinx Serialization, which was probably a big chunk of the Gson usage left in the main app. 
I cant really say much about the code, it works, but it had to to a lot of weird things to make it fully work with the legacy backup format.


CarlosEsco:
If you just wait til .12x you could just remove this code all together :)

Issue #4904, opened 2021-04-19
[Feature Request] <Reading downloaded mangas when source is dead (SD-Card)>


Rhonos:
It would be awesome when there could be a possibility to read the already downloaded chapters when the source goes "dead". At the moment, I have already downloaded plenty of mangas that I want to read in the future (already finished mangas), on my sd-card, but as soon as the source is down > all the chapters are no longer tagged as "downloaded" and cant be read over the app. so its necessary to switch the source and download it all again.

---

## Why/User Benefit/User Problem
- People do not need to download all chapters again when the source goes "down". 
- At the moment, its realy time consuming for people with large libraries

## What/Requirements
- Maybe, as soon as the source goes "down", the mangas get "tagged" as "offline"

paulo27ms:
Should be marked as a bug tbh. Not being able to read a downloaded chapter after uninstalling a source doesn't make much sense.

PR #5284, opened 2021-06-03
Added warning when uninstalling extension with library manga dependencies

### LINKED ISSUE #3787: [Feature Request] Warn users before deleting obsolete source extension

If you downloaded chapters from a source, and the source becomes obsolete that you uninstall the source, any downloaded chapters won't open (error: Source not found). You can't install the source extension anymore (unless you go to archives of old extensions)

## What/Requirements
- If user has manga from source extension in library, prompt a warning dialog.
- Dialog could suggest user to migrate manga in library to other sources.

datreesezcup:
This PR adds the feature to Tachiyomi in that if a user attempts to uninstall an extension that has sources used by their library manga, a dialog will appear warning the user about it, and indicating what sources are depended on, as well as giving the option to migrate it.

The dialog has 3 options:
- "Uninstall": Continue with extension uninstalling: Android will prompt for uninstalling the extension
- "Cancel": Self-explanatory
- "Migrate": Begins the migration process for the extension:
   - If there is only 1 source dependency, Tachiyomi will open the source migration page (as if the user clicked on a source under the Migration tab)
   - If there is more than 1 source dependency, Tachiyomi will go to the Migration tab
   

This PR closes #3787

### Other Changes
In order to allow the dialog to open the migration tab directly, the BrowseController has been modified to work with the index of what tab should be selected, instead of a boolean specifying that the Extension tab should be selected.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/ExtensionManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/ExtensionManager.kt
@@ -3,6 +3,7 @@ package eu.kanade.tachiyomi.extension
 import android.content.Context
 import android.graphics.drawable.Drawable
 import com.jakewharton.rxrelay.BehaviorRelay
+import eu.kanade.tachiyomi.data.database.DatabaseHelper
 import eu.kanade.tachiyomi.data.preference.PreferencesHelper
 import eu.kanade.tachiyomi.data.preference.plusAssign
 import eu.kanade.tachiyomi.extension.api.ExtensionGithubApi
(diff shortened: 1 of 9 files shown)


arkon:
I want to hold off on this until at least #4904 is addressed (i.e. we maintain a source ID/name mapping in the database, which would allow for reading content when the source is no longer installed).

Saud-97:
The issue holding this was closed by #6389 btw. So this issue is now available to be worked on if needed.

arkon:
Closing since the screens were heavily rewritten so there's a lot of conflicts. The implementation would need to be rewritten.

PR #5311, opened 2021-06-04
Coil: Caching adjustments

ivaniskandar:


arkon:
What's the eviction policy? Maybe we need an option similar to J2K to remove cached covers for non-library items?

ivaniskandar:
> What's the eviction policy? Maybe we need an option similar to J2K to remove cached covers for non-library items?

Currently all manga entries has the same custom caching system which has no eviction policy. See [MangaCoverFetcher](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/coil/MangaCoverFetcher.kt#L65).

I made it that way just so it will be directly compatible with the previous Glide system. But IMO the custom caching should only apply to library entries while we should leave non-library caching to OkHttp which has [easily configurable](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt#L51) caching policy.

arkon:
Yes, I'd agree that non-library covers shouldn't persist on disk forever. Do you want to tackle that as part of this PR?

ivaniskandar:
Yeah sure I'll make a follow-up commit ASAP

ivaniskandar:
Now non-library covers caching will be handled by Coil which is stored in the cache dir. Read more about Coil caching [here](https://coil-kt.github.io/coil/image_loaders/#caching).

PR #5335, opened 2021-06-07
Fix Cover sharing and saving

E3FxGaming:
The newly added manga cover sharing only worked with manga saved to the library (due to the implemented CoverCache only recording covers of library manga).
The changes made with this commit fixes that behavior by implementing a fallback: the cover can now also be retrieved from the Coil memoryCache.


E3FxGaming:
I wish there was a better way to transfer the thumbnailMemoryCacheKey from the MangaInfoHeaderAdapter to the MangaController/MangaPresenter.

If anyone knows of a better way (still pretty new to Android app development), I'd happily implement it.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoHeaderAdapter.kt
```diff
@@ -43,6 +44,9 @@ class MangaInfoHeaderAdapter(
 
     private var initialLoad: Boolean = true
 
+    var thumbnailMemoryCacheKey: MemoryCache.Key? = null
+        private set
```
arkon, line 48: No idea if there's a better way tbh. @ivaniskandar maybe you do?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
```diff
@@ -280,29 +284,85 @@ class MangaPresenter(
...
+        directory.mkdirs()
+        val filename = DiskUtil.buildValidFilename("${manga.title}.${ImageUtil.ImageType.PNG}")
 
         val destFile = File(directory, filename)
-        cover.inputStream().use { input ->
-            destFile.outputStream().use { output ->
-                input.copyTo(output)
-            }
+        destFile.outputStream().use { desFileOutputStream ->
+            coverBitmap.compress(Bitmap.CompressFormat.PNG, 100, desFileOutputStream)
```
arkon, line 365: Why always a PNG rather than the original format?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoHeaderAdapter.kt
ivaniskandar: Not sure if it's a good idea since the image can be scaled by coil as an optimization

E3FxGaming: @ivaniskandar Thank you very much for your input. I didn't know that there could be scaling problems.
I removed the usage of the memory cache and instead start a new Coil request (with the manga as data) + a custom target. This way Coil can do its cover fetching without knowing anything about an ImageView target, which should give us an image without scaling.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
E3FxGaming: Coil only gives me the cover as a Drawable. Drawables are more abstract than a File (from which I could determine the original file type). Thus it's impossible for me to determine the original file format.
I also can't just take the image format from the manga url, since
- the user may have set a custom cover which should be shared, rendering the original manga cover information useless
- the format extension in the url doesn't necessarily represent the actual file type returned by the server when we perform a get request (I would need the mime type from the response, but with Coil I can't even be sure that a network request is made)

Furthermore Bitmaps can only be compressed into
- jpeg - which would be an option, but it's not a lossless format
- png - which is what I'm using
- webp - which got deprecated with API level 30
- WEBP_LOSSLESS, WEBP_LOSSY which both require API level 30 and thus are things we can't use

If knew of a different approach that would give me all the information needed, I'd happily implement it.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -661,12 +665,41 @@ class MangaController :
...
+                        val coverBitmap = (result as BitmapDrawable).bitmap
+                        coverHandler(coverBitmap)
+                    }
+                    .build()
+                context.imageLoader.enqueue(req)
+            }
+        } catch (e: Exception) {
+            Timber.e(e)
+            activity?.toast(R.string.error_sharing_cover)
+        }
```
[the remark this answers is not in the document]
E3FxGaming: I removed the try-/catch code-wrapper.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -661,12 +665,41 @@ class MangaController :
         }
     }
 
+    /**
+     * Fetches the cover with Coil, turns it into Bitmap and does something with it (asynchronous)
+     * @param coverHandler A function that describes what should be done with the Bitmap
+     */
+    private fun useCoverAsBitmap(coverHandler: (Bitmap) -> Unit) {
```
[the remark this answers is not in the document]
E3FxGaming: shareCover and saveCover now pass their activity as context to useCoverAsBitmap.

PR #5345, opened 2021-06-08
Added text to category changing dialog when shown with no categories

datreesezcup:
Currently (as of 0.11.1), if a user long clicks manga in their library, and hits the category button on the action bar, but they don't have any categories setup: they will be met with a "Set Categories" dialog that has no content, a cancel button, and an OK button. I'm certain this could be a source of confusion for new users

This PR modifies the ChangeMangaCategoriesDialog to present the user with the text: 

> You have no categories. Go to \"More\" > \"Categories\" to create them

and only an OK button if the dialog is invoked with an empty category list. (As far as I can tell, the Library is the only place where this dialog can be invoked with an empty category list)


arkon:
I think it'd more ideal if it was something like:

```
Set categories

You don't have any categories yet.

                   Cancel   Edit categories
```

Where the "Edit categories" positive action would dismiss the dialog and navigate directly to that controller.

datreesezcup:
Y'know, I actually tried that at first, but I was having an issue with the LibraryController's "Action Mode" (It wasn't fully disappearing when the new controller was pushed)

I got it working now, just had to clear the selection and invalidate the action mode.

PR #5352, opened 2021-06-08
Add back missing start/end margins in manga info header

ghostbear:


arkon:
Does this handle cases where some of the buttons are hidden (e.g. local manga)?

ghostbear:
No

arkon:
Oof you're right, that wasn't intentional.

ghostbear:
I could probably change it to be in a horizontal list instead

ghostbear:
Now it will always have an end margin

PR #5387, opened 2021-06-12
Extend library search by adding -<source> option

curche:
This PR adds an option search in library using -source 
i.e. search for all manga *not* in a particular source  

A problem when you directly add a negation check is that although
it is working, the genre search kicks in to add back every manga since
-<source> returns true for containsGenre()

Solution chosen was to check source search first, and then if it fails
move to genre check. 
In addition, now you're able to search like this:
Eg: 
 1. -isekai, fantasy, -local, komga
 2. -manganato, -mangakakalot, -local

Checked in following devices:
 - Pixel 3 (blueline)
 - Redmi 7 (onclite)
 
Closes https://github.com/tachiyomiorg/tachiyomi/issues/4846

_______________________________________________________
This was already present in Sy however it heavily relied to methods
defined in exh. This PR tries a different approach

I understand that Multiple layers of if statements might not be always
preferable. If there's any way I can improve this, let me know :d

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryItem.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryItem.kt
@@ -120,16 +120,16 @@ class LibraryItem(
 
     private fun containsSourceOrGenre(tag: String, sourceName: String, genres: List<String>?): Boolean {
         return if (tag.startsWith("-")) {
-            if (sourceName.contains(tag.substringAfter("-"), true)) {
-                false
-            } else {
+            if (!sourceName.contains(tag.substringAfter("-"), true)) {
                 containsGenre(tag, genres)
+            } else {
+                false
             }
         } else {
-            if (sourceName.contains(tag, true)) {
-                true
-            } else {
+            if (!sourceName.contains(tag, true)) {
                 containsGenre(tag, genres)
+            } else {
+                true
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Soitora:
> I understand that Multiple layers of if statements might not be always
> preferable. If there's any way I can improve this, let me know :d

I personally don't think it's too much but an alternative that Android Studio has suggested to me before is using a `when` statement instead

curche:
Thanks for the headsup, Soitora! Thought about it a bit and managed to refactor using `when` and am bit more satisfied with this

Tested on blueline and onclite to confirm the logic is the same
PR is Ready for Review

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryItem.kt
```diff
@@ -105,18 +105,38 @@ class LibraryItem(
...
+     * Checking for genre is done only if the query isn't part of the source name.
+     *
+     * @param query the query to check
+     * @param sourceName name of the manga's source
+     * @param genres list containing manga's genres
+     */
+    private fun containsSourceOrGenre(query: String, sourceName: String, genres: List<String>?): Boolean {
+        val minus = query.startsWith("-")
+        val tag = if (minus) { query.substringAfter("-") } else query
+        val containsGenre by lazy { containsGenre(query, genres) }
```
arkon, line 133: Why the `lazy`? Couldn't this be inlined in L135?

curche: I added that when I was still using multiple `if` statements before. Didn't change this after I switched to `when` in the later commit
I have inlined it

PR #5388, opened 2021-06-12
Fix incorrect toolbar text color after theme change

### LINKED ISSUE #4948: [Bug] Tab header stays original color when phone theme switches if manga is open
## Device information
* Tachiyomi version: 0.10.11
* Android version: 11
* Device: Samsung Galaxy S10

## Steps to reproduce
1. Open up a chapter of any manga.
2. With the manga open switch phone theme mode to opposite of current setting, either dark or light, using Quick Panel icon or auto dark mode setting.
3. Go back to any section Library, Updates, History, Browse, or browse in an extension itself, or More.

### Expected behavior
Header(Library, Updates, etc) should switch to white for dark mode and black for light mode.

### Actual behavior
Header stays original theme color e.g. black when switching from light mode to dark mode.

## Other details
I'm using the default dark mode and light mode haven't tested with other options. Header color changes to correct color after exiting the app and reopening.

ivaniskandar:
Resolves #4948

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
@@ -4,19 +4,18 @@ import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
 import android.app.Activity
 import android.content.Intent
-import android.graphics.Color
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.Menu
 import android.view.MenuInflater
 import android.view.MenuItem
 import android.view.View
+import android.widget.TextView
+import androidx.annotation.FloatRange
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.view.ActionMode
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -294,28 +290,22 @@ class MangaController :
         updateFilterIconState()
     }
 
-    private fun updateToolbarTitleAlpha(alpha: Int? = null) {
+    private fun updateToolbarTitleAlpha(@FloatRange(from = 0.0, to = 1.0) alpha: Float? = null) {
         val scrolledList = binding.fullRecycler ?: binding.infoRecycler!!
+        (activity as? MainActivity)?.binding?.toolbar?.children?.forEach {
+            if (it is TextView) {
```
arkon, line 296: Maybe we can get references to these TextViews just once? Seems like a bad idea to have to get them on every scroll event.

PR #5408, opened 2021-06-16
Sort per category

### LINKED ISSUE #3722: [Feature Request] Different filters and sorts for different categories
Allow to filter categories differently instead of using a global library filter.

Example:
"Updated" Category: Unread filter, Unread sorting
"Reading" Category: No filters, Last read sorting
"Read Next" Category: Downloaded filter, Total chapters sorting
And so on.

With this it would be easier to navigate between categories and find a specific manga without wasting much time.

## What/Requirements
Add a button on filter options to restrict to current category.

ghostbear:
https://user-images.githubusercontent.com/6576096/122224073-0946c480-ceb4-11eb-96b2-d0ef8fe46137.mp4

I have made the sorting mask be able to support more options than currently available ones if any others will ever be added

Converts DisplayMode flags to an enum that replaces old DisplayMode enum. It shouldn't need a migration because they share the same value names. 

Also, fixes a bug with DisplayMode where the adapter would get desynced if one category was changed too much in one go. By reattaching the adapter when categories is updated.

Moving per category sort and display mode to "chips" when it's on will definitely be a separate PR

closes #3722


app/src/main/java/eu/kanade/tachiyomi/Migrations.kt
```diff
@@ -190,6 +199,41 @@ object Migrations {
                     LibraryUpdateJob.setupTask(context, 3)
                 }
             }
+            if (oldVersion < 64) {
```
ghostbear, line 202: When I tested this on the debug build it crashed the app until I moved where the upgrade was called

But I'm not sure if this happens also in the release build

paulo27ms:
Will there be a "filter by category" PR? Just curious, that's definitely the most useful one for me.
Thanks for the options.

ghostbear:
> Will there be a "filter by category" PR? Just curious, that's definitely the most useful one for me.

Probably not. Filters are much more complex compared to display and sort, which will make it harder to implement, and if it were to be implemented it will be harder to add more filters

PR #5421, opened 2021-06-18
Long Press "WebView" button on Manga info page to directly open web browser

datreesezcup:
This PR eliminates a step if a user wants to open a manga's page in a web browser, by allowing them to long press the "WebView" button to directly open their web browser to the manga page.

Otherwise users would either open the page in WebView, then click "Open in Browser" (twice? webview -> chrome custom tabs? -> web browser), or to go to Share -> Copy to Clipboard, and paste into their web browser.


arkon:
I'm a bit hesitant to add this since the whole flow was designed to intentionally force page loads in WebView to potentially resolve any cookies/CAPTCHAs/etc. in a way such that it'd actually help content load in-app. 

The usecase of using the app purely as a catalogue and actually just opening stuff in the browser all the time isn't something that we'd be interested in intentionally supporting.

datreesezcup:
That is a good point, I didn't think about the cookies and captcha's for stuff like that from the details page.
(Plus it might be kind of jarring if a user long presses the button by accident and is suddenly brought to a browser, now that I think about it)...

Apologies for not fully thinking this PR through. Thanks for your input!

Issue #3382, opened 2020-06-25
Issues with Edit categories when you select manga from multiple categories


vengiare:
### Device information
* Tachiyomi version: Stable 0.9.2
* Android version: 9
* Device: Nokia TA-1053

## Steps to reproduce
1. Select manga from multiple categories (Categories "x", "y", "z" for example)
2. Add it to another category (Category "b" for example)

### Expected behavior
Add manga to the category you want while  retaining previous categories of those manga.

### Actual behavior
Those manga get added to the category you want (Category "b") but it will be removed from its previous categories (manga you selected from "x", "y" and "z" will no longer be in those categories)

arkon:
That sounds like the expected behaviour to me. The selection you make is the final result.

vengiare:
No, I want to copy manga from multiple categories to another category, not move them.

arkon:
You can make a case for either UX flow, and the existing one makes more sense IMO. You can still move individual items or items from distinct categories if needed, but I won't be changing this unless there's a good alternative UX that covers what's proposed as well as existing behaviour.

CarlosEsco:
That's not the feature. It's edit categories. Not add more categories.

vengiare:
> That's not the feature. It's edit categories. Not add more categories.

I believe you misunderstood the issue.

> You can make a case for either UX flow, and the existing one makes more sense IMO.
It's weird that it gives different behaviour when you select manga from a single category compared to when you select manga from multiple categories"

> You can still move individual items or items from distinct categories if needed, but I won't be changing this unless there's a good alternative UX that covers what's proposed as well as existing behaviour.

Dunno, probably have a tri-state checkbox with the "partially checked" state for the categories that only have "some" manga selected.

CarlosEsco:
I know the issue.

You have manga 1 with sports
And manga 2 with action

And you want them to be sports, reading and action, reading

But as I said. You are editing categories in the dialog not adding new categories.  Key difference is editing it encompassing adding and removing of categories and adding only adding new categories on top

vengiare:
I'm not trying to "add" categories. The "reading" category you speak of already exists, if that makes sense to you.

arkon:
> Dunno, probably have a tri-state checkbox with the "partially checked" state for the categories that only have "some" manga selected.

Maybe, although the dialog library we're using doesn't allow that currently. I can leave this open, but I don't think it'll change anytime soon.

vengiare:
[...]

PR #5422, opened 2021-06-18
Edit  mangas' Categories in Library using TriState list

### LINKED ISSUE #961: Handling of multiple categories, might remove categories

quangkieu:
Fix #961
Add updateMangasToCategories to build build correct Categories list for
  each manga using Common and Mix list
Update QuadState Multi-Choice to be either Action or Display List
  Display list would have different state sequence from Action
  Uncheck-> Indeterminate (only if initial so)-> Check

Signed-off-by: Quang Kieu quangkieu1993@gmail.com


CarlosEsco:
Can you provide screens of the change

quangkieu:
<details>
<summary>Image from my phone</summary>

Screenshot_20210618-231824_Tachiyomi


</details>

<details>
<summary>Video from Bluestack</summary>


</details>
The video show case the feature of tri-state box and remember its mix status, and how the code behind handle mix, common categories so that we do not loss the info anymore

quangkieu:
I am not at all develop in android/kotlin before. I just copy the code that @arkon mention last time and mix them up to fit.
I would happy if someone could look if any of these follow the prefer practice of android philosophy

arkon:
Before I really dive into it, it'd probably actually be best to just modify the existing dialog rather than creating a new sheet. Were you concerned by differing behaviour between using it in the library and in browse or something?

quangkieu:
So mostly that I do not know how to modify dialog to add tristate so I just copy paste the current implemented classes that using it and expand from there. So I base my design from the library display setting sheet

arkon:
The category-related settings dialogs (global updates and automatic downloads) both implement tristate (actually quadstate) checkboxes, so you could look at that.

e.g. https://github.com/tachiyomiorg/tachiyomi/blob/c3fb5af3fcbeee4f3579435b832907d8165cb42e/app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsLibraryController.kt#L359

quangkieu:
I could look into it but I wonder if you prefer same UI/UI in Library between catalog and display setting or you would like them the same as manga/browse section dialog?
As I could see we slowly move things to sheets new design
Also I would likely need to update listener signature so I could send both common and mix list to build final catalogs per manga

quangkieu:
From digging more into the class and try to implement it, the listItemsQuadStateMultiChoice which using QuadStateMultiChoiceDialogAdapterhas its state sequence hardcoded from Uncheck, Check, Invert. So look like I have to build new dialog class to define the state sequence.
From the outlook it is not bring much to the table and the TabbedBottomSheetDialog which use ExtendedNavigationView has more hook and interface for easy define new classes

quangkieu:
I wonder if there anything I need to change?

arkon:
I don't want to go down the TabbedBottomSheetDialog/ExtendedNavigationView route right now since:
- It kind of looks bad, since it's not really a tabbed experience
- We'd have to transition more things to sheets to be consistent, which I want to avoid right now
- If we _do_ use sheets, I'd rather it be more customized, i.e. similar to what's currently in TachiyomiJ2K

It shouldn't be hard to modify `listItemsQuadStateMultiChoice`/`QuadStateMultiChoiceDialogAdapter` to allow changing the behaviour of what state it goes to (i.e. selected to unselected instead of the current selected to inversed path). May want to wait for #5423 to be completed first though.

vengiare:
I've discussed this in #3382 before and was just wondering (can't test the PR myself) if it handles multiple indeterminate states just fine. Say you select a manga from two different categories.

An example would be

Category A
– Manga 1
Category B
– Manga 2
Category C
– <blank>

And then turn it to

Category A (indeterminate)
– Manga 1
Category B (indeterminate to checked)
– +Manga 1
– Manga 2
Category C (unchecked to checked)
– +Manga 1
– +Manga 2

quangkieu:
New version using QuadStateMultiChoiceDialogAdapter
QuadStateMultiChoiceDialogAdapter now have two types of List: Display, Action
Library categories display list has same behavior as previous version. State sequence: Uncheck-> Indeterminate (only if initial so)-> Check
image

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -533,4 +545,22 @@ class LibraryPresenter(
...
+    /**
+     * Bulk update categories of mangas using old and new common categories.
+     *
+     * @param mangas the list of manga to move.
+     * @param newCommon the selected categories.
+     * @param newMix the list of common categories pre-update
+     */
+    fun updateMangasToCategories(mangas: List<Manga>, newCommon: List<Category>, newMix: List<Category>) {
+        val mc = mutableListOf<MangaCategory>()
+        for (manga in mangas) {
```
arkon, line 558: I think you could do something along the lines of

```
mangas.map {
  val categories = ...
  MangaCategory.create(it, categories) 
}
```

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -626,15 +631,13 @@ class MangaController :
         val categories = presenter.getCategories()
 
         val ids = presenter.getMangaCategoryIds(manga)
-        val preselected = ids.mapNotNull { id ->
-            categories.indexOfFirst { it.id == id }.takeIf { it != -1 }
-        }.toTypedArray()
+        val preselected = categories.map { if (it.id in ids) 2 else 0 }.toTypedArray()
```
arkon, line 634: Same here re: enum usage

arkon (commented):
Looks fine otherwise.

app/src/main/java/eu/kanade/tachiyomi/widget/materialdialogs/QuadStateMultiChoiceDialogAdapter.kt
```diff
@@ -56,6 +67,21 @@ internal class QuadStateMultiChoiceDialogAdapter(
         listener(currentSelection)
     }
 
+    internal fun itemDisplayClicked(index: Int) {
+        val newSelection = this.currentSelection.toMutableList()
+        newSelection[index] = when (currentSelection[index]) {
+            QuadStateTextView.State.UNCHECKED.ordinal -> when (initialSelected[index]) {
+                QuadStateTextView.State.INDETERMINATE.ordinal -> QuadStateTextView.State.INDETERMINATE.ordinal
```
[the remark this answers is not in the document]
quangkieu: So is this resolved then

arkon:
Thanks!

PR #5424, opened 2021-06-19
Add appending SPOILER_ to image names

### LINKED ISSUE #5362: [Feature Request] Append Shared/Saved pages with SPOILER_
When sharing images to discord, it takes a lot of extra steps to name the image for discord spoilers, and discord doesn't have mobile marking spoilers on upload like desktop.
## What/Requirements
There would be a preference for this, and it would append SPOILER_ to the beginning of every shared or saved image

loocool2:
Closes #5362


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -568,9 +568,15 @@ class ReaderPresenter(
...
-        ) + filenameSuffix
+        val filename = if (preferences.addSpoiler()) {
+            DiskUtil.buildValidFilename(
+                "SPOILER_${manga.title} - ${chapter.name}".takeBytes(MAX_FILE_NAME_BYTES - filenameSuffix.byteSize())
+            ) + filenameSuffix
+        } else {
+            DiskUtil.buildValidFilename(
+                "${manga.title} - ${chapter.name}".takeBytes(MAX_FILE_NAME_BYTES - filenameSuffix.byteSize())
+            ) + filenameSuffix
+        }
```
ghostbear, line 579:
```suggestion
        val filenamePrefix = if (preferences.addSpoiler()) "SPOILER_" else ""
        val validFilename = DiskUtil.buildValidFilename(
            "${manga.title} - ${chapter.name}".takeBytes(MAX_FILE_NAME_BYTES - filenameSuffix.byteSize())
        )
        val filename = filenamePrefix + validFilename + filenameSuffix
```

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -558,7 +558,7 @@ class ReaderPresenter(
     /**
      * Saves the image of this [page] in the given [directory] and returns the file location.
      */
-    private fun saveImage(page: ReaderPage, directory: File, manga: Manga): File {
+    private fun saveImage(prefix: String = "", page: ReaderPage, directory: File, manga: Manga): File {
```
ghostbear, line 561: If you make prefix the last parameter it could be excluded from line 610

```kt
// From 
Observable.fromCallable { saveImage("", page, destDir, manga) } 
// To
Observable.fromCallable { saveImage(page, destDir, manga) }
```

or if you use the names for the parameters it could also be excluded 

```kt
Observable.fromCallable { saveImage(
  page = page,
  directory = destDir,
  manga = manga
)
```

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -568,13 +568,16 @@ class ReaderPresenter(
...
         val filenameSuffix = " - ${page.number}.${type.extension}"
-        val filenamePrefix = if (preferences.addSpoiler()) "SPOILER_" else ""
         val validFilename = DiskUtil.buildValidFilename(
-            "${manga.title} - ${chapter.name}".takeBytes(MAX_FILE_NAME_BYTES - filenameSuffix.byteSize())
+            prefix + (
+                "${manga.title} - " +
+                    "${chapter.name}"
+                ).takeBytes(
+                MAX_FILE_NAME_BYTES - filenameSuffix.byteSize()
+            ) + filenameSuffix
```
ghostbear, line 577: Why you split the string to concatenate instead of just using a string template?
```suggestion
            prefix + 
                    ("${manga.title} - ${chapter.name}")
                        .takeBytes(MAX_FILE_NAME_BYTES - filenameSuffix.byteSize()) + 
                    filenameSuffix
```

arkon:
Shouldn't this be optional? I'm trying to think of a good flow that isn't quite as reliant on Discord's naming scheme though.

Maybe something like:

1. Select "Save" from sheet
2. It opens a dialog with:
  - A text input pre-populated with the existing filename convention
  - A switch to prepend with `SPOILER_`
  - Actions: Save and Cancel

Thoughts?

loocool2:
I think most users would use the share button from the sheet instead of save, and having to prompt if I want to add ``SPOILER_`` to every image is annoying in my opinion. Forcing a spoiler mark on every shared image is more convenient and prevents users from sharing spoilers intentionally

NotTsunami:
> Shouldn't this be optional? I'm trying to think of a good flow that isn't quite as reliant on Discord's naming scheme though.
> 
> Maybe something like:
> 
> 1. Select "Save" from sheet
> 2. It opens a dialog with:
> 
> * A text input pre-populated with the existing filename convention
> * A switch to prepend with `SPOILER_`
> * Actions: Save and Cancel
> 
> Thoughts?

I prefer this to the proposed default.

Soitora:
> Shouldn't this be optional? I'm trying to think of a good flow that isn't quite as reliant on Discord's naming scheme though.
> ...
> Thoughts?

Definitely, I can't agree with having all my shared images suddenly becoming marked as spoilers. I regularly share stuff from 4-koma and other types where having them suddenly marked as Spoilers for no reason is a huge drawback.

Something similiar to your suggestion or something else where you can spoiler mark it but it's not entirely forced.

kiriya-aoi:
> Shouldn't this be optional? I'm trying to think of a good flow that isn't quite as reliant on Discord's naming scheme though.
> 
> Maybe something like:
> 
> 1. Select "Save" from sheet
> 2. It opens a dialog with:
> 
> * A text input pre-populated with the existing filename convention
> * A switch to prepend with `SPOILER_`
> * Actions: Save and Cancel
> 
> Thoughts?

A settings toggle seems like the best option to me. What it prepends could then be easily configurable: want it to prepend `SPOILER_` for Discord? Cool. Some other string for some other reason? Also cool, and it would then apply to share+save.

If this toggle appears in the Sharesheet as well, and if it's just a toggle that remembers last-used position (or a toggle+edit filename box), that's fine. It doesn't immediately appear to be possible in the Android Sharesheet though.

If it's always an extra step but the common user frequently shares and needs to switch between those modes, an on-share-action would indeed be best. Based on our Discord server, it seems more like the common user uses screenshots instead of the long-press sheet anyway.

If this were added and made non-configurable, then everyone used to and desiring the current behavior has to go and edit the filename of every single image they share. A "solution" that simply changes who has to edit filenames just swaps one problem for an identical one and is not a real solution. Especially when the actual problem is on Discord for not implementing a spoiler checkbox on mobile...

TL;DR: toggle good, forcing it very bad

vengiare:
Just add a toggleable option to ask for filename upon saving. Adding `SPOILER_` is kinda exclusive to discord and is kinda niche.

vengiare:
Oh also if we need a configurable filename format, might as well go all-in on it. 

(PixEz app)
IMG_20210706_195620

PR #5426, opened 2021-06-19
Apply reader theme to reader progress dialog

datreesezcup:
This PR styles the progress dialog that appears when clicking the previous or next chapter buttons to use a style matching the current reader theme, and is more Tachiyomi-styled.

# Before
White | Gray, Black, Automatic
-------|-------------
old_white_progress | old_dark_progress

# After
White | Gray | Black, Automatic
-------|-------|-------
new_white_progress | new_gray_progress | new_black_progress


Soitora:
Nice! I didn't ever even think of this as the window pops so fast, good job

PR #5436, opened 2021-06-21
Imported implementation for updating library by next expected update from Neko

stinky-lizard:
This sort uses the last 4 updates for a manga to compute an average time between updates and then extrapolates when the next update should occur.

Currently seems to work perfectly. However, I may have silently messed something up along the way.

All code and algorithms are credited to kyjibo on GitHub. The original commit adding this functionality is here: https://github.com/CarlosEsco/Neko/commit/681003926ae1e07b925155d4e1f43972bbe2b843


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateRanker.kt
```diff
@@ -9,9 +12,27 @@ object LibraryUpdateRanker {
...
+        (this::nextFirstRanking)()
     )
 
+    /**
+     * Provides a total ordering over all the Mangas.
+     *
+     * Orders the manga based on the distance between the next expected update and now.
+     * The comparator is reversed, placing the smallest (and thus closest to updating now) first.
+     */
+    //   @RequiresApi(Build.VERSION_CODES.N)
```
arkon, line 25: Can this be removed?

stinky-lizard: Ah, yes, that's a holdover from Neko, which is for 7+. I removed the function that requires this.

PR #5450, opened 2021-06-23
Match ReaderActivity theme with the rest of the app

### LINKED ISSUE #5196: [Feature Request] Reader Sheet should match application theme
Currently the reader does not respect your devices setting.

The reader sheet background color is based on the color you have set to the manga background and not on your devices theme. 

This makes it so that even though you don't like AMOLED for example, you're still forced to use it in the reader sheet unless you pick the Light background.

Before #5193 it was hypocritical and still used a black background even though you had choosen Gray as the background.

The reader sheet also does not respect the accent color set by the used theme.

I've tried myself seeing what I could do to fix this but I don't think this can be fully solved by tweaking `themes.xml` alone.

ivaniskandar:
Resolves #5196

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.data.preference
 
 import android.content.Context
+import android.content.res.Configuration
 import android.os.Environment
 import androidx.core.content.edit
 import androidx.core.net.toUri
@@ -10,6 +11,7 @@ import com.tfcporciuncula.flow.Preference
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.data.database.models.Manga
 import eu.kanade.tachiyomi.data.preference.PreferenceValues.DisplayMode
+import eu.kanade.tachiyomi.data.preference.PreferenceValues.ThemeMode.*
 import eu.kanade.tachiyomi.data.track.TrackService
 import eu.kanade.tachiyomi.data.track.anilist.Anilist
 import eu.kanade.tachiyomi.ui.reader.setting.OrientationType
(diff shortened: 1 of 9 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -355,6 +351,17 @@ class ReaderActivity : BaseRxActivity<ReaderActivityBinding, ReaderPresenter>()
 
         initBottomShortcuts()
 
+        val alpha = 0xE6 // 90%
+        binding.toolbar.background.alpha = alpha
+        binding.toolbarBottom.background.alpha = alpha
+        binding.leftChapter.background.alpha = alpha
+        binding.readerSeekbar.background.alpha = alpha
+        binding.rightChapter.background.alpha = alpha
```
arkon, line 359: Couldn't you do something like

```
listOf(binding.toolbar, binding.toolbarBottom, ...).forEach {
  it.background.alpha = 0.9f
}
```

app/src/main/res/values/themes.xml
```diff
@@ -349,64 +346,6 @@
...
 
-    <!--===============-->
-    <!-- Reader Themes -->
-    <!--===============-->
-
-    <!--== Light Reader base ==-->
-    <style name="Theme.Base.Reader.Light" parent="Theme.Base.Light">
-        <!-- Theme colors -->
-        <item name="colorSurface">@color/md_white_1000</item>
-        <item name="colorToolbar">@color/reader_toolbar</item>
```
arkon, line 360: Should be able to delete these color values.

ivaniskandar:
Also replaced the deprecated `systemUiVisibility` methods while we're at it since the hardcoded flags mess with light theme system bars icons

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
```diff
@@ -43,6 +45,17 @@ fun Preference<Boolean>.toggle(): Boolean {
     return get()
 }
 
+fun PreferencesHelper.isDarkMode(): Boolean {
```
arkon, line 48: Why an extension function? It'd basically fit in with the actual intention of `PreferencesHelper`.

ivaniskandar: Right why did I do that I don't remember, I'll move that one.

PR #5470, opened 2021-06-28
Merge light and dark themes

### LINKED ISSUE #5399: [Feature Request] Add green apple AMOLED version theme
Looks cool. Would be nice to have.

## What/Requirements
Could you use the accent colour used in “green apple” in an AMOLED black theme.

OR
Have a separate selector for mode(light, dark, AMOLED black) and a separate one for accent colour (this would avoid similar future requests)

ivaniskandar:
Allows all theme variants to be used on both light and dark mode. Pure black (AMOLED) backgrounds is now a separate option for dark mode too.

Resolves #5399


app/src/main/res/values/colors.xml
```diff
@@ -43,11 +36,9 @@
     <color name="ripple_colored_midnightdusk">#1FF02475</color>
 
     <!-- AMOLED Theme -->
-    <color name="accent_amoled">#3399FF</color>
     <color name="divider_amoled">@android:color/transparent</color>
-    <color name="surface_amoled">@color/md_black_1000</color>
-    <color name="background_amoled">@color/md_black_1000</color>
-    <color name="ripple_colored_amoled">#1F3399FF</color>
+    <color name="surface_amoled">#000001</color>
+    <color name="background_amoled">#000001</color>
```
arkon, line 41: Why tho

ivaniskandar: Ah it's for the navbar workaround thingy since navbar use surface color, I'll revert the background one

Soitora:
Just FYI all forms of ripples on Dark Blue tool/nav are borked, which is why I wanted it added as a toggle with a notice of deprecation :whip:

ivaniskandar:
> Just FYI all forms of ripples on Dark Blue tool/nav are borked, which is why I wanted it added as a toggle with a notice of deprecation :whip:

Fixed. And I already made the colored bar style modular Just In Case™ so the toggle can be easily added when needed.

arkon:
@ivaniskandar FYI there's a merge conflict.

ivaniskandar:
Yep that's expected. I'm on it.

PR #5479, opened 2021-06-30
Track search dialog adjustments

ivaniskandar:
Based on feedbacks

[Preview](https://user-images.githubusercontent.com/12537387/124310442-602be980-db96-11eb-9c8e-7bd08707a5b3.png)


app/src/main/java/eu/kanade/tachiyomi/util/view/ViewExtensions.kt
```diff
@@ -185,7 +185,7 @@ inline fun ChipGroup.setChips(
 /**
  * Applies elevation overlay to a MaterialCardView
  */
-inline fun MaterialCardView.applyElevationOverlay() {
+fun MaterialCardView.applyElevationOverlay() {
```
ivaniskandar, line 188: Why the inline function breaks after it got merged? That's a mystery.

PR #5488, opened 2021-07-02
Enable users to manually set manga status for library manga

datreesezcup:
The idea with this PR is to let users to have slightly more control over their library (and solving a couple of problems) by manually specifying the status that should be taken into account wherever the app queries the status of a manga (Eg. The "Completed" library filter, the "Only Update Ongoing Manga" setting, etc).

## Implementation
Since status field in the database is a (presumably) 1 byte integer, I have elected to use this value to store both what the _actual_ status of the manga is and what the user set the status to by dedicating 4 bits of the byte to each.
* The lower 4 bits store the status of the manga (as reported by the source)
* The upper 4 bits store the status of the manga as set by the user

Using 4 bits to track the status; there can be up to 16 different statuses for a manga. At least for now, I think this is an acceptable limit, as there are currently only 4 different statuses.
Though while working on this, I did notice that it looks like the 5th and 6th options have been designated already.

Code that needs to get the status of a manga (through the Manga interface) can use a new getter, `publicationStatus`, which returns the status the user set, or what the actual status is if there is no user-set status. There are also getters and setters for the real status and user-set status. All of these getters and setters just perform bitwise-logic on the parent-interface SManga's `status` field.

## Extension Compatibility
If I've done everything right, extensions need no changes at all, as extensions are passed the actual status of the manga, not the user-set status. I chose this to remove any possible side-effects of extensions having a different status than expected.

## UI
On the UI side of things, to change the manga status I have made it so that clicking on the status text of a _favorited_ manga will bring up a dialog of status options (Default, Ongoing, Completed, Licensed), that the user can use to select. Of note, the "Default" option means no user-set value, and the app will use what the actual status is from the source. Additionally, when removing a manga from the library, the user status is cleared,

To indicate to the user that a custom status is applied, the status text becomes _italic_ when there is a custom value set.
| Preview |
|----------|
| status_sample |

### P.S
I apologize for the wall-of-text PR, I just wanted to make sure I covered everything important, since this is the biggest and most complex PR I have created for Tachiyomi so far.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Manga.kt
@@ -19,6 +19,26 @@ interface Manga : SManga {
 
     var viewer_flags: Int
 
+    /** The publication status that should be used when taking actions dependent on a
+     * manga status.
+     * This takes into account the real (source) status, and the user-defined status.
+     *
+     * Setting this value is the same as setting [sourcePublicationStatus]
+     */
+    var publicationStatus: Int
+        get() = if (status >= 0x0F) status shr 4 else status
+        set(value) { status = (status and 0xF0) or value }
+
+    /** The publication status as reported by the source of this manga */
+    var sourcePublicationStatus: Int
+        get() = status and 0x0F
+        set(value) { status = (status and 0xF0) or value }
+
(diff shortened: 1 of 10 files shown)


joelchrono12:
Why does this keep getting moved to the next milestone? lol

PR #5505, opened 2021-07-05
Make Automatic Reader Theme pick background/text color based on dark mode preference

ghostbear:


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -902,6 +905,31 @@ class ReaderActivity : BaseRxActivity<ReaderActivityBinding, ReaderPresenter>()
                 .launchIn(lifecycleScope)
         }
 
+        /**
+         * Picks background color for [ReaderActivity] based on light/dark theme preference
+         */
+        private fun automaticBackgroundColor(): Int {
+            return when (preferences.themeMode().get()) {
```
ivaniskandar, line 912: You can just check against [this method](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt#L284)

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/ReaderTransitionView.kt
```diff
@@ -32,15 +35,41 @@ class ReaderTransitionView @JvmOverloads constructor(context: Context, attrs: At
...
         listOf(binding.upperText, binding.warningText, binding.lowerText).forEach {
             it.setTextColor(color)
         }
     }
 
+    /**
+     * Picks text color for [ReaderActivity] based on light/dark theme preference
+     */
+    private fun automaticTextColor(): Int {
+        return when (preferences.themeMode().get()) {
```
ivaniskandar, line 52: This one too

PR #5516, opened 2021-07-08
Show correct number of items in the library title

jmir1:
When all manga are in a category that isn't the default category, the title just shows the number of items in the default category. This fixes that.

The next comment shows how to reproduce

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
@@ -147,11 +147,8 @@ class LibraryController(
 
         if (preferences.categoryNumberOfItems().get() && libraryMangaRelay.hasValue()) {
             libraryMangaRelay.value.mangas.let { mangaMap ->
-                if (!showCategoryTabs) {
+                if (!showCategoryTabs || adapter?.categories?.size == 1) {
                     title += " (${mangaMap[currentCategory?.id]?.size ?: 0})"
-                } else if (adapter?.categories?.size == 1) {
-                    // Only "Default" category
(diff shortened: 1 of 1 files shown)


jmir1:
https://user-images.githubusercontent.com/43830312/124909549-db542c00-dfea-11eb-9696-3dee4699f14f.mp4

jmir1:
Before:

https://user-images.githubusercontent.com/43830312/124909549-db542c00-dfea-11eb-9696-3dee4699f14f.mp4


After:

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
```diff
@@ -147,11 +147,8 @@ class LibraryController(
 
         if (preferences.categoryNumberOfItems().get() && libraryMangaRelay.hasValue()) {
             libraryMangaRelay.value.mangas.let { mangaMap ->
-                if (!showCategoryTabs) {
+                if (!showCategoryTabs || adapter?.categories?.size == 1) {
```
arkon, line 150: I think this entire check can actually be removed.

jmir1: no it's still necessary. otherwise the title will also show the number of items in addition to the tab titles.

jmir1: Unbenannt

jmir1: ^^^^
that is what would happen.

arkon: Right, good call.

PR #5523, opened 2021-07-10
Allow Search with same query on network state change..

FourTOne5:
Sometime it may be needed to search with same query. And again I don't actually see a real reason for this check.


CarlosEsco:
What reason would you need to search the same query?

FourTOne5:
> 
> 
> What reason would you need to search the same query?

Network State Change?

PR #5534, opened 2021-07-12
Don't show navigation overlay if tap navigation is disabled

ivaniskandar:
This being the reader navigation overlay that's shown when reading mode is changed.


arkon:
Might make more sense to check for `config.tappingEnabled` in the `navigationModeChangedListener`s?

ivaniskandar:
> Might make more sense to check for `config.tappingEnabled` in the `navigationModeChangedListener`s?

Ah right, that'll do.

PR #5554, opened 2021-07-15
Filter source manga by unread status

Vulcannis:
Adds an option to all source filter sheets to filter based upon the unread chapters in the same way as the library filter.

I pulled out a `StandardFilter` interface which should make adding the other library filters pretty easy, but I wasn't interested in those so I only implemented unread. Since the way the unread count is loaded in the library (custom data model class) wasn't easy to re-use, I just created a new basic query. I also needed to distinguish the case of chapters not being loaded into the Chapter table yet vs. having no unread chapters, and this way I don't risk breaking the existing library code. I didn't bother with an `observeChanges*` call because the StorIO documentation is completely unclear on what it does exactly, and I don't believe it buys anything when just doing `executeAsBlocking()` calls.

Repaging is necessary so that the completely filtered pages get handled by the controller properly. Rather than doing some complicated work using `Pager` I just did the simplest thing.

UI-wise the new filter is just stuck at the top of all source filter sheets, using the same GUI as in the Library.

I'm an experienced Java dev but this is my first Kotlin work, so some stuff might not be idiomatic.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/queries/MangaQueries.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/queries/MangaQueries.kt
@@ -64,6 +64,11 @@ interface MangaQueries : DbProvider {
         )
         .prepare()
 
+    fun getMangaUnread(id: Long) = db.get()
+        .cursor()
+        .withQuery(RawQuery.builder().query(mangaUnreadQuery).args(id).build())
+        .prepare()
+
     fun insertManga(manga: Manga) = db.put().`object`(manga).prepare()
 
     fun insertMangas(mangas: List<Manga>) = db.put().objects(mangas).prepare()
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/queries/RawQueries.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/queries/RawQueries.kt
@@ -32,6 +32,22 @@ val libraryQuery =
 """
 
 /**
+ * Query to get the number of unread chapters for a manga, -1 if no chapters have been loaded yet.
+ */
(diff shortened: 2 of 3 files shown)


CarlosEsco:
Maybe i am missing the point of this but do you have a use case where this is useful?
Main issues I see are
1.  Only manga that would have chapters are ones in your library or have been clicked into
2.  The data is not real time so if i clicked into a manga 3 months ago it will still return unread of however many chapters were unread at that time.

Vulcannis:
Not every manga is worth favoriting or managing in the library, especially when you the only thing you care about it is if you've read it already or not. Doujinshi (or more generally, single chapter only sources) would be the most obvious example, but I also find myself searching through short run manwha titles for something new. All this is available in tachiyomi's database but not surfaced unless you bring everything into the library. Personally I keep my library small, with only currently reading stuff and absolute favorites, I'm not interested in curating titles where I only care about "have I read this yet?" And it is very useful when you regularly search some tags for doujinshi and just want to grab some new title, and unless you can memorize several pages of search results you have to instead click into each and every one just to check whether you've read it already.
1. And thus haven't read yet. Code currently handles that case, treats the manga as having been unread, exactly as if you did click on the manga but didn't read it.
2. True for non-doujinshi, but if you cared enough to want to watch for new chapters... well, that's what the library is for, right?

arkon:
We generally don't actually want people relying on the in-app history for non-library items, since that can easily start exploding in size, and there's no guarantee that we wouldn't put an automatic cleanup mechanism in the future.

That said, this sounds quite specific the scenarios you detailed (short, typically read-once things like doujinshi or manwha), so it may be a good fit for [TachiyomiSY](https://github.com/jobobby04/TachiyomiSY) instead.

PR #5565, opened 2021-07-16
Fix splash screen icon on Android 12

ivaniskandar:
This also adds splash screen exit animation on older android version which replicates the animation in Android 12 (I've tried).

The splash screen will be removed after all items inside starting screen is finished showing, also has timeout just in case.

=Previews=
[Android 10](https://user-images.githubusercontent.com/12537387/126042351-bfeaf480-98a2-456e-9a7b-93a8f68cd329.mp4)
[Android 12](https://user-images.githubusercontent.com/12537387/126042456-8e594109-d785-4f4e-a977-f5256c5eecef.mp4)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/SourceController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/SourceController.kt
@@ -31,6 +31,8 @@ import eu.kanade.tachiyomi.ui.browse.BrowseController
 import eu.kanade.tachiyomi.ui.browse.source.browse.BrowseSourceController
 import eu.kanade.tachiyomi.ui.browse.source.globalsearch.GlobalSearchController
 import eu.kanade.tachiyomi.ui.browse.source.latest.LatestUpdatesController
+import eu.kanade.tachiyomi.ui.main.MainActivity
+import eu.kanade.tachiyomi.util.view.onAnimationsFinished
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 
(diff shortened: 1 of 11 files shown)


Soitora:
Send gif

app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/SourceController.kt
```diff
@@ -81,6 +83,9 @@ class SourceController :
         // Create recycler and set adapter.
         binding.recycler.layoutManager = LinearLayoutManager(view.context)
         binding.recycler.adapter = adapter
+        binding.recycler.onAnimationsFinished {
```
arkon, line 86: I assume this is for people that open these controllers directly? Why is it hooked to the recyclerview animation though?

ivaniskandar: Yes it waits until the items animations (the fade-in one) are finished and all the items are completely laid out. [Here's a preview without the wait](https://user-images.githubusercontent.com/12537387/126042600-7dabfaa1-af3d-4714-b3fa-8b8dd1cc96c6.mp4), it is more noticeable on faster devices.

TBH I don't really like the solution but since the result feels right it's here. I completely understand if it doesn't make the cut.

arkon: I was a bit iffy since there's cases where you can go directly to some other views (manga details, downloads, etc.) although I'm not entirely sure if that matters if those the root controllers end up getting created first anyway.

Could potentially refactor it as a lifecycle hook of the `RootController` interface too somehow.

ivaniskandar: Oh I didn't realize `RootController` is a thing. I can try utilizing that but for now all `MainActivity` intents other than the launcher's should dismiss the splash right away [here](https://github.com/tachiyomiorg/tachiyomi/pull/5565/files#diff-50dbf49f0d1782abbaf5c81937b6f56753cb126107f36f0b2f627537188814e1R446).

PR #5567, opened 2021-07-17
Make default tracking status depend if the user has read chapter or not

### LINKED ISSUE #200: Tracking default status Plan to read

ghostbear:
closes #200

This makes so tracking status changes depending on if a user has read chapter(s) for a manga when adding a new tracker to it.
- If a user hasn't read any it will make the tracking status plan to read
- If a user read one or more it will make the tracking status reading
- When a user changes chapter number from 0 it will change tracking status to reading
    - It will not change when changing between e.g. 5 -> 10 or 41 -> 10

When a user reads a chapter it will also change to status to reading

For trackers that have rereading as status, it will leave it as is

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/TrackService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/TrackService.kt
@@ -36,6 +36,10 @@ abstract class TrackService(val id: Int) {
 
     abstract fun getStatus(status: Int): String
 
+    abstract fun getReadingStatus(): Int
+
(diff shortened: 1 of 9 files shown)


arkon (commented):
We should probably refactor the usages of tracking so that this sort of logic happens at a higher level so it doesn't need to be duplicated across every tracker implementation. Not a blocker for this PR though.

Issue #1652, opened 2018-10-10
Refresh tracking metadata for a manga before updating it


paulo27ms:
Sync the manga with the service first and then update it after reading a chapter, especially now that there's no longer a prompt for when you're updating a manga, I believe this would help avoid issues between editing entries outside of Tachiyomi.

Clarity:
If you update a manga on MAL but don't refresh tracking data on Tachiyomi, when you go to read a chapter on Tachiyomi then it sends outdated info (like score) to MAL and you end up with old data.

PR #5568, opened 2021-07-17
Add option to update trackers read progress on Libary update.

FourTOne5:
https://user-images.githubusercontent.com/59261191/126041241-217615b6-3e6a-4d6c-9323-6ad10c03fd95.mp4

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -326,6 +326,10 @@ class LibraryUpdateService(
                                     updateTrackings(manga, loggedServices)
                                 }
 
+                                if (preferences.autoUpdateTrackersReadProgress()) {
+                                    updateTrackingsReadProgres(manga, loggedServices)
+                                }
+
                                 currentlyUpdatingManga.remove(manga)
                                 progressCount.andIncrement
                                 notifier.showProgressNotification(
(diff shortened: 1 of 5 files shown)


arkon:
Related issue: #1652 

I'm not sure what the benefit of this change would be. If it's to always see the proper data, opening the sheet already refreshes it. If it's to ensure it's up to date before updating after reading a chapter, there's no guarantee that a library update was done recently, nor that there was a change in between when it happened and you read a chapter on that particular device.

I think the more sensible method would be to refresh a particular manga's tracking right before it's updated when reading.

FourTOne5:
> 
> I'm not sure what the benefit of this change would be. If it's to always see the proper data, opening the sheet already refreshes it. If it's to ensure it's up to date before updating after reading a chapter, there's no guarantee that a library update was done recently, nor that there was a change in between when it happened and you read a chapter on that particular device.

If you read a bunch of chapter when offline well tachi has no way to update it. It only gets updated if you read a chapter when online or you manually change it  So this will just update tracker read progress if it is less that of the read progress in libary during the global update period.

Well if there is a better suitable approach my ears are open.

arkon:
Ah I think I misunderstood what this PR was even doing.

We should just adopt what J2K is doing with the queued job to run after going online again.

FourTOne5:
> Ah I think I misunderstood what this PR was even doing.

No probs.

> We should just adopt what J2K is doing with the queued job to run after going online again.

Ah yes a queued job will be more suitable I'll take a look then.

PR #5569, opened 2021-07-17
Add Dynamic theme for Android 12

ivaniskandar:
Followed AOSP's DeviceDefault theme except on surface color because it breaks all dialogs.

| Light | Dark |
| ------------- | ------------- |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/base/activity/BaseThemedActivity.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/base/activity/BaseThemedActivity.kt
@@ -40,6 +40,9 @@ abstract class BaseThemedActivity : AppCompatActivity() {
         fun AppCompatActivity.applyThemePreferences(preferences: PreferencesHelper) {
             val resIds = mutableListOf<Int>()
             when (preferences.appTheme().get()) {
+                PreferenceValues.AppTheme.MONET -> {
+                    resIds += R.style.Theme_Tachiyomi_Monet
+                }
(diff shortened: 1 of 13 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/base/activity/BaseThemedActivity.kt
```diff
@@ -40,6 +40,9 @@ abstract class BaseThemedActivity : AppCompatActivity() {
         fun AppCompatActivity.applyThemePreferences(preferences: PreferencesHelper) {
             val resIds = mutableListOf<Int>()
             when (preferences.appTheme().get()) {
+                PreferenceValues.AppTheme.MONET -> {
```
arkon, line 43: It may be worth enforcing the version code check here too in case someone somehow copies over their preferences from an S device.

ivaniskandar: Older versions will get the default theme should they managed to get to this block https://github.com/tachiyomiorg/tachiyomi/pull/5569/files#diff-683326448134b180d2575e28185ff2552667470a4c81d10271c59f0de0a781ffR78

arkon: Hmm right, I guess that inherits `Theme.Tachiyomi`.

PR #5575, opened 2021-07-18
Grouped chapter download list by source

ivaniskandar:
[Preview](https://user-images.githubusercontent.com/12537387/126076382-dfd96f51-2d6e-4fee-9f04-e357b2acd690.png)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadAdapter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadAdapter.kt
@@ -2,13 +2,14 @@ package eu.kanade.tachiyomi.ui.download
 
 import android.view.MenuItem
 import eu.davidea.flexibleadapter.FlexibleAdapter
+import eu.davidea.flexibleadapter.items.AbstractFlexibleItem
 
 /**
  * Adapter storing a list of downloads.
  *
  * @param context the context of the fragment containing this adapter.
  */
-class DownloadAdapter(controller: DownloadController) : FlexibleAdapter<DownloadItem>(
+class DownloadAdapter(controller: DownloadController) : FlexibleAdapter<AbstractFlexibleItem<*>>(
     null,
     controller,
     true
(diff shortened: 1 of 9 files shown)


Riztard:
btw you can use this to show the image `` and maybe add some width if too big

FourTOne5:
Collapsible List?

Riztard:
sort by chapter list
<table>
<tr>
	<td>before
	<td>after
	<td>how it should be
<tr>
	<td>
	<td>
	<td>
</table>

CarlosEsco:
> 
> 
> sort by chapter list
> before 	after 	how it should be(how it work before)
>  	 	

The how it should be here makes no sense.  In what scenario would you want that sort style?  I have 10 manga with chapter 1 from the same source but the source is at the bottom, why do i want them at the top?   How do you handle source 6 having a chapter 6 now?  Is it not grouped together anymore and has 2 entries?    You are adding overhead to have to compare every source and every chapter number, for what benefit?

The current after in this PR makes sense because its sorting the chapters insides the source

CarlosEsco:
> > why do i want them at the top?
> 
> so it sorted close/similar to before?
> 
> > How do you handle source 6 having a chapter 6 now? 
> 
> follow how it currently work?
> 

That sorting mode makes no sense though now.  Unless there is an option to ungroup.  You are now having the same source have multiples groups in the list because ???

I'm trying to understand the scenario where you have multiple sources with multiple chapters 1s and you want those at top?  When they are already going to be at top of the source anyways.

Riztard:
yea maybe option to ungroup is better, so can mix download queue if download from more than 5 source (currently only top 5 source/ cant mix)

arkon:
Does mixing even do anything? It only ever downloads from 5 sources.

Riztard:
> Does mixing even do anything? It only ever downloads from 5 sources.

ya but not always same 5 source

Riztard:
need move to top/bottom for source

ivaniskandar:
Alright I'm on it.

ivaniskandar:
Done. Totally forgot what I did here but it works...

PR #5578, opened 2021-07-19
Auto hide reader menu when user starts reading again

### LINKED ISSUE #3411: [Feature Request] Scrolling should auto-hide controls
When reading if you click anywhere on screen controls show up for switching pages and such. If you continue scrolling these controls wont go away, you have to click again to hide them. 

This is breaking immersion  and its quite bothersome, there should be an auto-hide feature, if you continue scrolling it will automatically go away, if you dont do anything it shouldnt go away.

ghostbear:
closes #3411

This makes so the reader menu auto hides when changing page or scrolling to keep the user immersed in reading

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerPageHolder.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerPageHolder.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.ui.reader.viewer.pager
 
 import android.annotation.SuppressLint
+import android.app.ActionBar
 import android.graphics.PointF
 import android.graphics.drawable.Animatable
 import android.view.GestureDetector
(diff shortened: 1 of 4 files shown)


arkon:
Is it possible to also handle scrolling with a PagerViewer (i.e. dependant on zoom)?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerPageHolder.kt
```diff
@@ -367,6 +368,17 @@ class PagerPageHolder(
                     }
                 }
             )
+            setOnStateChangedListener(
+                object : SubsamplingScaleImageView.OnStateChangedListener {
+                    override fun onScaleChanged(newScale: Float, origin: Int) {
+                        viewer.activity.hideMenu()
```
arkon, line 374: Is it worth only having the listener attached when the menu is open? The idea of these constantly running whenever you zoom/pan seems kind of iffy.

ghostbear: I have added a way to listen when the visibility of the menu changes. That way PageHolder can listen for when the visibility changes and attach/detach the OnStateChangedListener and only listen when needed

PR #5586, opened 2021-07-20
For library update error log, group errors by error string, and then sort the resulting list by source

FlaminSarge:
Format is e.g.
```
! Error 1
  # Source 1
    - Manga 1
! Error 2
  # Source 1
    - Manga 5
  # Source 2
    - Manga 2
    - Manga 3
  # Source 3
    - Manga 4
```


FlaminSarge:
Another option is
```
Error 1
- Source 1
  - Manga 1
```
Thoughts?

CarlosEsco:
My vote is the 2nd one

FlaminSarge:
Updated to the following:
```
! Error
  # Source
    - Manga
```

FlaminSarge:
Screenshot_1626786832
Here it is in action.

arkon:
Maybe they should be sorted too? That way at least the different source not installed errors will be grouped together, despite having an entry per ID.

FlaminSarge:
As a followup I was going to try to group specifically those. Thoughts?

arkon:
I'm fine with that.

PR #5587, opened 2021-07-20
Reader loading progress indicator changes

ivaniskandar:


CarlosEsco:
gif?

ivaniskandar:
It's basically the same reader indicator as before except it has the same style as other progress indicators and not rounded. The rotate animation used is copy-pasta'd from the old one.

PR #5589, opened 2021-07-20
Group 'Source not installed' cases in library update error log

FlaminSarge:
Followup to #5586

Uses "Source not found" string for failedUpdates list.
Exception behaves as normal for other cases.


FlaminSarge:
Error log case and original manga details refresh case screenshots:
Screenshot_1626821096
Screenshot_1626821145

PR #5595, opened 2021-07-21
Allow clearing database data for specific sources

datreesezcup:
This PR changes the "Clear database" button in advanced settings to now open a new screen that lists all of the sources in the database that have non-library manga saved (along with how many entries there are), allowing the user select which sources they want to remove data for.

With one or more sources selected, a FAB appears that when clicked, brings up the "Are you sure" dialog originally on the settings screen, and clicking OK will delete the database info for those sources.

The top toolbar of the screen also includes a "select all" and "select inverse" button for quick selections.

If this list is empty (no non-library manga in database), an EmptyView with the text "Database clean" is displayed.

| Preview |
|----------|
| db_delete_preview |


datreesezcup:
woops, forgot about the change to the new MaterialAlertDialogs.

Will fix and re-open tonight

datreesezcup:
Now it's ready to go!

_Note to self: this is why you frequently rebase on upstream before working on features!_

app/src/main/res/layout/database_source_item.xml
```diff
@@ -0,0 +1,74 @@
...
+        style="@style/TextAppearance.Regular.SubHeading"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="4dp"
+        android:layout_marginEnd="8dp"
+        android:layout_marginTop="8dp"
+        android:ellipsize="middle"
+        android:maxLines="1"
+        android:textSize="16sp"
+        app:layout_constraintHorizontal_bias="0.007"
```
arkon, line 36: What's with these biases? They don't seem particularly necessary judging from the screenshots.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/database/DatabaseSourcesPresenter.kt
```diff
@@ -0,0 +1,42 @@
...
+        db.getAllManga()
+            .asRxObservable()
+            .observeOn(AndroidSchedulers.mainThread())
+            .map(::findSourcesWithDatabaseEntries)
+            .subscribeLatestCache(DatabaseSourcesController::setDatabaseSources)
+    }
+
+    private fun findSourcesWithDatabaseEntries(manga: List<Manga>): List<DatabaseSourceItem> {
+        return manga
+            .filterNot { it.favorite }
```
arkon, line 29: Maybe this should just be part of the DB query that gets executed?

datreesezcup:
holy crap what the heck.
I just pushed like a _ton_ of other commits into this by accident.
Let me get this fixed...

PR #5610, opened 2021-07-26
Add last update time to Updates Tab

### LINKED ISSUE #5466: Add update time to updates tab
### Acknowledgements

- [X] I have updated the app to version **[0.11.1](https://github.com/tachiyomiorg/tachiyomi/releases/tag/v0.11.1)**.

The updates tab could display the last time the library was updated, whether automatically and manually. Ideally, this would help me figure out if the app didn't auto-update my library when it was supposed to for some reason (or if I'm just obsessively updating my library and ddos-ing sources on accident).

### Other details

_No response_

datreesezcup:
This PR inserts a small header in the top of the Updates tab that indicates when the last library update was/

Other small notes:
- This header is not visible if there is no recorded last update time.

Closes #5466

### EDIT: The below image no longer reflects what the current feature looks like. However it is kept here to provide context for the comments below. The feature currently looks [like this](https://github.com/tachiyomiorg/tachiyomi/pull/5610#issuecomment-887098596)
| Preview |
|----------|
| updates_header_preview |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -109,11 +109,6 @@ class LibraryUpdateService(
         const val KEY_TARGET = "target"
 
         /**
-         * Key that defines what triggered the update
-         */
-        const val KEY_TRIGGER = "trigger"
-
-        /**
          * Returns the status of the service.
          *
          * @param context the application context.
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -94,16 +94,6 @@ class LibraryUpdateService(
         TRACKING // Tracking metadata
     }
 
-    /**
-     * Defines what triggered an update request
-     *
-     * This is only used when the update target is [Target.CHAPTERS]
-     */
-    enum class Trigger {
-        MANUAL,
-        AUTOMATIC
-    }
-
     companion object {
 
         private var instance: LibraryUpdateService? = null
@@ -142,11 +132,10 @@ class LibraryUpdateService(
          * @param target defines what should be updated.
          * @return true if service newly started, false otherwise
          */
-        fun start(context: Context, category: Category? = null, target: Target = Target.CHAPTERS, trigger: Trigger = Trigger.AUTOMATIC): Boolean {
+        fun start(context: Context, category: Category? = null, target: Target = Target.CHAPTERS): Boolean {
             return if (!isRunning(context)) {
                 val intent = Intent(context, LibraryUpdateService::class.java).apply {
                     putExtra(KEY_TARGET, target)
-                    putExtra(KEY_TRIGGER, trigger)
                     category?.let { putExtra(KEY_CATEGORY, it.id) }
                 }
                 ContextCompat.startForegroundService(context, intent)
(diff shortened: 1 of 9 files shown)


Soitora:
Maybe lowercase U for update

ivaniskandar:
TBH I don't see the need for the update method there, just the last update time should be enough

FourTOne5:
If it's possible can you make it something like this:

Screenshot_20210726_221104

Where the chapters will be from the Last Update.

datreesezcup:
> Maybe lowercase U for update

Good point; Tachiyomi doesn't use title case in most/all of it's UI, so neither should this

---------

> TBH I don't see the need for the update method there, just the last update time should be enough

I added the method since it was a suggestion of the original feature request, and I did like the idea of knowing what specifically caused it. But then again, yeah, I suppose that there's not really a whole lot of reason to differentiate manual vs automatic;

The original issue had written

> ...whether automatically and manually. Ideally, this would help me figure out if the app didn't auto-update my library when it was supposed to for some reason

But yeah you're right, you would know if didn't update automatically (for some reason) if the last update time is longer than whatever you set as the interval.

----------
> If it's possible can you make it something like this:

I personally think my original way (incorporating the previous feedback) would be better, just because:
- It allows the user to still see when the latest chapter updates happened; having the first "header" there be the update time means you kind of lose that info. (Eg. A user's library may have updated a few min ago, but the last manga that actually had a chapter update was yesterday. The user would not know that anymore because that first header is now just the last time the library had an update).

I don't know, maybe I'm thinking too much about it 🤷

datreesezcup:
Okay, I've made some changes, but I'd like to see what some of you think about a couple of them before I commit

I've incorporated the changes from @ivaniskandar and @Soitora, but I've also made a couple other changes that I think myself are good, being:
- Making the text left-aligned, so it fits in with the style of the app more.
- Changing the text to "Library last updated: \*time\*.
   - I reworded it like that so its more clear that its the time the _library_ last updated, not the time that the latest chapter of a manga came out. Because I could somewhat see that confusion coming up, considering this text is placed directly above the list of recently updated chapters

### Preview
updates_header_2

FourTOne5:
> * It allows the user to still see when the latest chapter updates happened; having the first "header" there be the update time means you kind of lose that info. (Eg. A user's library may have updated a few min ago, but the last manga that actually had a chapter update was yesterday. The user would not know that anymore because that first header is now just the last time the library had an update).
> 
> I don't know, maybe I'm thinking too much about it 🤷

Ah yes that make sense

And about the new preview there is no need to use a colon after "updated" and it's much better as centralized well atleast in my perspective.

ivaniskandar:
That placement looks off to me. IMO center alignment is more fitting for an absolute header, just like the global search button.

Soitora:
Center and keep the colon, probably

datreesezcup:
Alright, I've committed those last changes, and I think this PR is finally ready to go! 

Preview based on latest commit:
updates_header_3

arkon:
Some thoughts on the general change (i.e. I haven't actually looked at the code diff yet):

- Centered LGTM
- Does it distinguish between a full library update (i.e. refresh from Updates), a global update (i.e. respects your category choices), or refreshing an individual category?
- Maybe it could be hidden on initial load, and the user can scroll up to see it? Not sure how others feel about this.

vengiare:
- remove " : " and it's magically better
- toggleable? (on by default)

It's a bit unclear so can anyone clarify.. does this show the time a new chapter is fetched by the app? Chapter upload time in the source? Or is this the time the library was refreshed?

Personally the third one is the most useful here especially for the issue since the user mentioned not knowing if the library was refreshed automatically or not.

Soitora:
> * remove " : " and it's magically better
No

datreesezcup:
> Centered LGTM

Sorry if this should be obvious, but what is LGTM?

------
> Does it distinguish between a full library update (i.e. refresh from Updates), a global update (i.e. respects your category choices), or refreshing an individual category?

Currently, it doesn't distinguish update types. Any library update with the `Target.CHAPTERS` target will update this "last library update" value.

------
> Maybe it could be hidden on initial load, and the user can scroll up to see it? Not sure how others feel about this.

I'd like to hear what others think as well, but my idea is that if we keep the header in this format (simple one-line of text with last update time), it'd probably be fine to not have it hidden (since it's so thin, it takes up very little room). However relating to the above point, if it is desired to also include the update type (full, global, category), then in that case this header would be hidden so you have to scroll up to see it. That way it won't take up a bunch of room on the update screen, unless you specifically scrolled to see it.

vengiare:
> Sorry if this should be obvious, but what is LGTM?

Looks good to me

AqilCont:
When will this be merged? seems like something relatively simple, non-controversial, and handy :<

arkon:
Superseded by 43c195e14a13aff9a3d2552f4f4a08ecee2d0b63

PR #5613, opened 2021-07-27
Edge-to-edge manga details view

ivaniskandar:
Tablet UI and "Blue" theme are not affected. This also makes all controllers doing lift app bar on scroll by default.

[Preview](https://user-images.githubusercontent.com/12537387/127102139-f39de787-2093-4cf9-a858-3fe8d3ba68a4.png)


CrepeTF:
@joseph619 after so many years, it is finally here

joseph619:
Looks great, I wonder if it would be possible to have edge to edge for tablets as well somehow? 

@CrepeTF All thanks to pringle, what a madlad

ivaniskandar:
> Looks great, I wonder if it would be possible to have edge to edge for tablets as well somehow?

It's possible but that will be a huge changes so I'll make the attempt on future PR.

CrepeTF:
Could I get a picture of how it looks in dark theme

arkon (commented):
Seems fairly good from testing it, but it does sometimes fail to make the status bar opaque when scrolling slowly in the manga details view, resulting in seeing the toolbar text in a weird position.

CrepeTF:
@arkon screenshot? Thank

ivaniskandar:
> Seems fairly good from testing it, but it does sometimes fail to make the status bar opaque when scrolling slowly in the manga details view, resulting in seeing the toolbar text in a weird position.

This elevation issue exists even without this, you can try it and see the elevation doesn't change when you scroll slowly without flinging. But well, it's fixed.

PR #5650, opened 2021-08-03
Add ability to tweak auto hide sensitivity in Webtoon Reader

ghostbear:
People that don't like the current sensitivity can now tweak it

Highest basically hides near-instant, high a tiny bit of scroll, low some scroll require, lowest even more scroll required


Riztard:
put the setting under webtoon section?

ghostbear:
> put the setting under `webtoon` section?
> *oh wait, it can be used for `vertical`, so not really `webtoon`

Vertical is just L2R/R2L but vertical and is not affected by on scroll, like Webtoon and Continous Vertical is. So this setting should probably be under webtoon
Note: Continous Vertical is just Webtoon but with padding

Issue #2905, opened 2020-04-19
Offline mode enhancements


arkon:
- [x] More obvious indicator that it's enabled
- [ ] Automatically enable when actually offline 
- [x] Tracker update queue (#2397)

Issue #2397, opened 2019-12-27
[Feature Request] Automatic synchronization after being online again


WasdeAsde:
**Problem:**
After reading a chapter within offline mode with activated synchronization, the read chapter will not be added/updated when coming online again. Therefore, the chapter requires a manual update.

**What should the feature do?**
After a chapter has been read offline, it should be synced automatically with the used synchronization platform (e.g. MyAnimeList, Kitsu) in the background or when opening the app online.

Disposablegloves:
+1 upping this. I read offline then update and do other stuff when I'm online so this feature will be really useful to me.

I suggest keeping read logs offline then only syncing them whenever that manga/the library is updated, instead of whenever you finish reading.

CarlosEsco:
@Disposablegloves  may i ask why you turn off wifi/mobile data when you read downloaded content?

wopian:
+1 as I tend to read the most while on the London Underground these days.

Disposablegloves:
Family wifi just doesn't reach my room and I tend to read before I sleep.

pervertkirito:
Just like how Taiga(windows anime tracking app) works.
its will put on offline queue list if you don't have internet access. Once online, queued list will update to tracking service

serjflint:
When traveling switching to an airplane mode saves a lot of battery. Revisiting all chapters to update tracking is a drag.

CaptainHalflife:
I wish someone can implement this feature as when im on the go i read my downloaded content offline and its a pain to update them manually.

arkon:
#1497

PR #5658, opened 2021-08-05
Remove unnecessary type guards to enable better tracker updating, addresses tracking offline reading concerns

TeamDman:
Addresses #2905 and #2397.

Rather than doing something like queuing tracker changes when offline, being able to sync the library so that trackers are updated based on locally read chapters is even better.

This feature is basically already implemented, but there's some unnecessary checks going on that prevent it from actually working.

The in-app setting is:  
Settings => Library => Automatically update trackers

The `syncChapersWitTrackServiceTwoWay` method doesn't require an `EnhancedTrackService` parameter, so there's no reason to be guarding.

https://github.com/tachiyomiorg/tachiyomi/blob/ff2a4e69526a2edfa5f3bd7570c09204e5ebb9f9/app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterTrackSync.kt#L10-L18

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
@@ -492,9 +492,8 @@ class LibraryUpdateService(
                                 val updatedTrack = service.refresh(track)
                                 db.insertTrack(updatedTrack).executeAsBlocking()
 
-                                if (service is EnhancedTrackService) {
-                                    syncChaptersWithTrackServiceTwoWay(db, db.getChapters(manga).executeAsBlocking(), track, service)
-                                }
+                                syncChaptersWithTrackServiceTwoWay(db, db.getChapters(manga).executeAsBlocking(), track, service)
+
                             } catch (e: Throwable) {
                                 // Ignore errors and continue
                                 Timber.e(e)
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
@@ -807,9 +807,8 @@ class MangaPresenter(
                                 val track = it.service.refresh(it.track!!)
                                 db.insertTrack(track).executeAsBlocking()
 
-                                if (it.service is EnhancedTrackService) {
-                                    syncChaptersWithTrackServiceTwoWay(db, allChapters, track, it.service)
-                                }
+                                syncChaptersWithTrackServiceTwoWay(db, allChapters, track, it.service)
+
                             }
                         }
                         .awaitAll()
@@ -843,9 +842,8 @@ class MangaPresenter(
                     service.bind(item, hasReadChapters)
                     db.insertTrack(item).executeAsBlocking()
 
-                    if (service is EnhancedTrackService) {
-                        syncChaptersWithTrackServiceTwoWay(db, allChapters, item, service)
-                    }
+                    syncChaptersWithTrackServiceTwoWay(db, allChapters, item, service)
+
                 } catch (e: Throwable) {
                     withUIContext { view?.applicationContext?.toast(e.message) }
                 }


FourTOne5:
I added a PR with option to update tracker read progress with library but Arkon wished for offline queue.

arkon:
> Rather than doing something like queuing tracker changes when offline, being able to sync the library so that trackers are updated based on locally read chapters is even better.

Those are 2 separate issues. This PR updates local chapter read statuses based on what's in the tracker(s). What you described is the need to update trackers based on local chapter read statuses while offline.

arkon:
I do think that this probably makes sense though, and should supersede #3705. I'll review it later.

CarlosEsco:
> 
> 
> I do think that this probably makes sense though, and should supersede #3705. I'll review it later.

main thing i see with this, the function as it exists is a guaranteed 1 to 1 mapping, since its the komga extension connecting to the user's komga server.  

Using this method for trackers should probably be opt-in (if it is accepted) because it's probably gonna cause false positives and either mark things read, or completely ignore some chapters.  

The way most trackers and Tachi works is i read chapters 1 in tachi, it updates tracker for chapters 1.  If i read chapter 1.1 in tachi it updates tracker for chapter 1.  The concept of partial chapters doesn't change.

So the below code:
https://github.com/tachiyomiorg/tachiyomi/blob/ff2a4e69526a2edfa5f3bd7570c09204e5ebb9f9/app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterTrackSync.kt#L20

would not work at all for  3rd party tracker.
```
number -> index
1.1 -> 0
1.2 -> 1
1.3 -> 2
2 -> 3
3.1 -> 4
```

tracker lets say has last_read as 2.

1.1, 1.2 will be marked read and 1.3, and 2 will be unread

TeamDman:
> This PR updates local chapter read statuses based on what's in the tracker(s). What you described is the need to update trackers based on local chapter read statuses while offline.

True since it's omnidirectional, but the reason I want it is the other direction: updating trackers based on locally read chapters.

> This also completey ignores specials, and extra chapters that would also throw off the indexes

Took me a few rereads to understand, but I believe you're right.

Will try and update the logic to match existing behaviour.

https://github.com/tachiyomiorg/tachiyomi/blob/c0647c311002e66c54c3f85d1d075cd216124c86/app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt#L693-L728

arkon:
> updating trackers based on locally read chapters.

This already happens for all trackers and this change doesn't affect that though?

TeamDman:
> This already happens for all trackers and this change doesn't affect that though?

If I read a chapter offline, then the remote tracker can't be updated.
Even if the local record for the track is updated, it won't be pushed until I go to that manga and read another chapter.
However, with this change, once I go online and refresh my library then the remote tracker will be updated.

Maybe I'm wrong on some technicalities, but this is the general behaviour I've observed.

arkon:
If the internal remote track models are getting updated despite the calls failing, that sounds like a bug. We shouldn't be drifting between what we think we've sent to trackers and what was actually sent.

TeamDman:
I'm wrong on the internal tracker being updated, sorry for the confusion.

TeamDman:
Probably will add a toggle for preventing it from marking local chapters as read, since sometimes special chapters get posted later. E.g., read 0-24, then 12.5 comes out. Without the toggle, 12.5 would get marked as read.

TeamDman:
Demo of current PR behaviour. Hopefully upload works, preview being wonky.

https://user-images.githubusercontent.com/9356891/128423624-326c8d74-abfe-4753-98f2-c1e5aa57fb8b.mp4

CarlosEsco:
How does it handle these

https://mangadex.org/title/85b51b37-0ce6-4144-a19b-6b064bc2c2ae

https://mangadex.org/title/bd38c075-8b12-450a-aeee-125e3ac5da4b


These are the cases I was concerned with.

TeamDman:
The sync method gets called when:

- refreshing the library
- opening the tracker menu
- associating trackers
- auto-associating trackers when manga is added to your library
	- only for enhanced trackers.

It will only sync if the update trackers toggle is enabled.

The sync will update the tracker to match the highest locally read chapter number. 

The sync will mark local chapters read according to a new preference.

Preference values:

- `always`
	- marks local chapters read if local.ch# <= remote.ch#
	- this truncates with `.toInt()` so should not mark ch4.5 as read if you read ch4.4
- `never`
	- never marks local chapters read, current behaviour before this PR
- `not special`
	- same as `always` but doesn't mark chapters with a decimal as read
	- prevents bonus chapters from being marked read since they usually come out later

## Demos

I used a manga that has decimal chapters, hopefully that's good for the example

### mode=`never`

https://user-images.githubusercontent.com/9356891/128451433-80300898-b71e-48ce-9c68-75dc10af85d4.mp4

### mode=`always`

https://user-images.githubusercontent.com/9356891/128450958-a8d0aec6-b1b6-463f-983b-142c7f48a839.mp4

### mode=`not_special`

https://user-images.githubusercontent.com/9356891/128450954-d1dc3157-3963-4d0a-b7e3-989e02f77187.mp4

### reading offline with mode=`never` then refreshing library to update trackers

The cropping on the video is wack because I cut out some dead time in a video editor.

https://user-images.githubusercontent.com/9356891/128452055-bb0c1246-2495-4315-9e38-540b5986f5b2.mp4

vengiare:
One suggestion I have for this is 

When auto-sync is disabled and a manga is not synced with tracker, it could change tracker icon (in manga info) to "sync problem" icon that has a different color. (orange or red idk)
IMG_20210811_063412
IMG_20210811_063324

And then have a "Sync now" button in tracker sheet, kinda like #3705.

gotson:
This change would probably break Komga tracker which is for now using ordinal position instead of chapter number.

See #5699 for more details.

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterTrackSync.kt
```diff
@@ -15,22 +18,34 @@ import timber.log.Timber
...
+                }
+            }
+            .forEach { it.read = true }
+        db.updateChaptersProgress(chapters).executeAsBlocking()
     }
+    // find last read local chapter number
+    // default to existing tracker value
+    val lastReadLocalChapterNumber = chapters
+        .filter { it.read }
+        .maxByOrNull { it.chapter_number }
```
gotson, line 41: In the current behaviour the local last read is the last read from a sequence of read chapters. This takes the highest of all read chapters.

If reading out of order, this would mark chapters read while they are not, and after the next sync would mark local chapters read too.

For exemple, if you have 5 chapters, and only have locally read number 5, but not number 1 to 4, the local last read would still be 5.

gotson:
@TeamDman all of the work i needed to do has been done and has been merged. Komga tracker in Tachiyomi doesn't rely any more on ordinal position of chapter, and there is no more limitation of using `Int` chapters. The [UI is also functional for Komga tracker](https://github.com/tachiyomiorg/tachiyomi/pull/5917) now, even though [we could make it better](https://github.com/tachiyomiorg/tachiyomi/issues/5699#issuecomment-920514136).

I think you could update your PR to fix the merge conflict 👍🏻

saud-97:
> The syncChapersWitTrackServiceTwoWay method doesn't require an EnhancedTrackService parameter, so there's no reason to be guarding.

I have tested this and enabled the two-way sync for Kitsu tracker. The results were **successfully** being able to sync progress across devices online via tracking sync on update. I don't see why this was not implemented yet, the part of allowing two-way sync for all trackers.

saud-97:
Let me know if you think I should create a PR for this change to supersede this one, although maybe my implementation definitely can be more optimized.

MrChuw:
Giving the user the option to choose a source would be nice.  Example, only update when the chapter changes in MAL and ignore the rest (if you have MAL obviously).  And any update on when it can be added?

TeamDman:
> And any update on when it can be added?

Just fixed the merge conflicts, I have NOT tested it again recently though.

> Giving the user the option to choose a source would be nice

Not sure why this would be necessary, as the proposed code should not cause backtracking when updating trackers based on locally read chapters

AntsyLich:
How will special handle this series https://mangadex.org/title/8791c515-0245-4fcc-81cc-22863a0aba6b?

TeamDman:
> How will special handle this series [mangadex.org/title/8791c515-0245-4fcc-81cc-22863a0aba6b](https://mangadex.org/title/8791c515-0245-4fcc-81cc-22863a0aba6b)?

When the tracker (ch 7) is farther ahead than local (ch 1), "Mark read except specials" produces

image

"Always mark read" produces

image



https://user-images.githubusercontent.com/9356891/182894946-ef39a8e1-629d-4570-9ceb-b675018fa77a.mp4

PR #5695, opened 2021-08-10
Require authentication-confirmation to change biometric lock settings

### LINKED ISSUE #4999: [Bug] 'Require unlock' feature breaks app with no finger print reader
## Device information
* Tachiyomi version: 0.10.12
* Android version: 10
* Device: Galaxy Tab A (SM-T510)

## Steps to reproduce
1. Go to More, Settings, Security
2. Turn on Require unlock
3. Lock when idle doesn't matter as it happens on all values
4. Get out of the app by either going home or locking the device
5. Open app
6. App closes without fully opening

### Expected behavior
Require entering unlock pin same as lock screen

### Actual behavior
The app doesn't open and instead immediately closes.

## Other details
This feature works just fine on my phone with a finger print reader, but not on my tablet which doesn't have a finger print reader. Once this feature is enabled, the only way to get the application to be able to open again is to clear app data. I am unsure how to dump crash logs as the application cannot open until I clear the app data.

If you're experiencing crashes, share the crash logs from More → Settings → Advanced → Dump crash logs.

ivaniskandar:
Also changed the `BiometricPrompt` API with `AuthPrompt`. This should ~prevent~ fix #4999 ~from locking the app permanently~. If not, other apps usually requires confirmation before enabling the lock anyway ¯\\\_(ツ)_/¯

Tested on Android 6 emulator with PIN and Android 12 Pixel 3 with fingerprint.


specterflare:
You dropped this \

To prevent anymore lost limbs throughout GitHub, correctly escape the arms and shoulders by typing the shrug as `¯\\\_(ツ)_/¯`

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsSecurityController.kt
```diff
@@ -17,11 +23,36 @@ class SettingsSecurityController : SettingsController() {
...
+                                isChecked = newValue as Boolean
+                            }
+
+                            override fun onAuthenticationError(
+                                activity: FragmentActivity?,
+                                errorCode: Int,
+                                errString: CharSequence
+                            ) {
+                                super.onAuthenticationError(activity, errorCode, errString)
+                                Toast.makeText(activity, errString, Toast.LENGTH_SHORT).show()
```
arkon, line 50: There's an extension function for toasts.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsSecurityController.kt
```diff
@@ -37,6 +68,33 @@ class SettingsSecurityController : SettingsController() {
...
+                                value = newValue as String
+                            }
+
+                            override fun onAuthenticationError(
+                                activity: FragmentActivity?,
+                                errorCode: Int,
+                                errString: CharSequence
+                            ) {
+                                super.onAuthenticationError(activity, errorCode, errString)
+                                Toast.makeText(activity, errString, Toast.LENGTH_SHORT).show()
```
arkon, line 92: Same here.

app/src/main/java/eu/kanade/tachiyomi/util/system/AuthenticatorUtil.kt
```diff
@@ -1,43 +1,108 @@
...
-    fun isSupported(context: Context): Boolean {
-        return isLegacySecured(context) || getSupportedAuthenticators(context) != 0
     }
 
-    fun isDeviceCredentialAllowed(context: Context): Boolean {
-        return isLegacySecured(context) || (getSupportedAuthenticators(context) and Authenticators.DEVICE_CREDENTIAL != 0)
+    /**
+     * Returns true if Class 2 biometric or credential lock is set and available to use
+     */
+    fun Context.isAuthenticationSupported(): Boolean {
```
arkon, line 49: This is going to break the legacy Android versions use case, isn't it? That's why there was the explicit `isLegacySecured` function before.

ivaniskandar: From [Android Developers](https://developer.android.com/training/sign-in/biometric-auth#declare-supported-authentication-types):
`Note: The following combinations of authenticator types aren't supported on Android 10 (API level 29) and lower: DEVICE_CREDENTIAL and BIOMETRIC_STRONG | DEVICE_CREDENTIAL`

This will work since the combination of `BIOMETRIC_WEAK | DEVICE_CREDENTIAL` is used here. I've tested this on Android 6 emulator with only PIN lock enabled.

arkon: Hmm I must've had issues because I was checking for `BIOMETRIC_STRONG` too. What happens in the case that the device has something that qualifies as `BIOMETRIC_STRONG` enabled with this change anyway?

ivaniskandar: Will also work since strong/class 3 biometrics are guaranteed to meet weak/class 2 type. [Docs for related](https://developer.android.com/reference/androidx/biometric/auth/Class2BiometricOrCredentialAuthExtensionsKt#(androidx.fragment.app.FragmentActivity).startClass2BiometricOrCredentialAuthentication(kotlin.CharSequence,kotlin.CharSequence,kotlin.CharSequence,kotlin.Boolean,java.util.concurrent.Executor,androidx.biometric.auth.AuthPromptCallback))

PR #5715, opened 2021-08-14
DNS-over-HTTPS (Adguard)

Long-FA:
Implement Adguard as the third option for DNS-over-HTTPS, because no one deserves to have their DNS queries logged for personalized advertisment.


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedController.kt
```diff
@@ -146,11 +147,13 @@ class SettingsAdvancedController : SettingsController() {
                     context.getString(R.string.disabled),
                     "Cloudflare",
                     "Google",
+                    "Adguard",
```
Riztard, line 150: capital "G" ?

Long-FA: Hmm okay :eyes:

arkon:
I sort of suspect this is going to confuse people since there's no ads in the app and (most) ad tracking generally doesn't work in sources since there's no JS enabled. I'm fine with adding this from a privacy-focused perspective though.

Long-FA:
I suppose my wording was a bit confusing. This PR is meant to give users an altenative choice for DoH, in case said users are concerned about the not-so-clean history of Cloudflare and Google when it comes to privacy.

PR #5719, opened 2021-08-14
Use relative time in ChapterHolder

ghostbear:
Screenshot_1628970378
Screenshot_1628970388
Screenshot_1628970373


arkon:
Does this still respect the user preference for date format when it's beyond the "relative" threshold?

Soitora:
> Does this still respect the user preference for date format when it's beyond the "relative" threshold?

It does not

Soitora:
Could extensions with a known reliability in displaying chapter upload dates be able to show even more precise data?

Like this on MangaDex: 
image

Shows like this:
image

Maybe if extensions were able to pass a "hasReliableDates" or something idk

ghostbear:
> Does this still respect the user preference for date format when it's beyond the "relative" threshold?

I have made an extension function that gets the relative time instead of using DateUtils and it uses the user preference for date format when it's beyond the "relative" threshold

It also displays "Now", could be something else like "Soon" for example, instead of "in X hours" mentioned in https://github.com/Jays2Kings/tachiyomiJ2K/issues/958

Soitora:
> I have made an extension function that gets the relative time instead of using DateUtils and it uses the user preference for date format when it's beyond the "relative" threshold

No hours/minutes ago?

> 
> It also displays "Now", could be something else like "Soon" for example, instead of "in X hours" mentioned in [Jays2Kings#958](https://github.com/Jays2Kings/tachiyomiJ2K/issues/958)

Maybe "Recently" for anything under 1 hour and then `x hours ago` ? Or for anything under x amount of minutes relatively?

app/src/main/res/values/strings.xml
```diff
@@ -186,6 +186,13 @@
     <string name="pref_label_nsfw_extension">Label in extensions list</string>
     <string name="parental_controls_info">This does not prevent unofficial or potentially incorrectly flagged extensions from surfacing NSFW (18+) content within the app.</string>
 
+    <string name="recently">Recently</string>
+    <string name="today">Today</string>
+    <string name="yesterday">Yesterday</string>
+    <string name="days_ago">%1$.0f days ago</string>
```
arkon, line 192: I'd recommend using a plural for this with the number being the number of days (i.e. 0 = today, 1 = yesterday, other = n days ago). I'm not entirely sure how these sort of things end up getting localized in other languages.

app/src/main/res/values/strings.xml
```diff
@@ -186,6 +186,13 @@
...
+    <string name="relative_format">Shorted relative time</string>
```
arkon, line 193: "Relative timestamps"?

app/src/main/java/eu/kanade/tachiyomi/util/lang/DateExtensions.kt
```diff
@@ -94,3 +96,23 @@ fun Long.toLocalCalendar(): Calendar? {
         )
     }
 }
+
+private const val MILLISECONDS_IN_DAY = 86_400_000.0
+
+fun Date.getRelativeTime(
```
arkon, line 102: `toRelativeString`?

app/src/main/res/values/strings.xml
```diff
@@ -186,6 +186,13 @@
...
+    <string name="relative_format_summary">Excludes \"X days ago\"</string>
```
arkon, line 194: Maybe default it to true and have the summary be an example?

PR #5728, opened 2021-08-16
Fix library icon unchecked state

### LINKED ISSUE #5447: [Bug] Library icon inconsistency
—
## Device information
* Tachiyomi version: r3529
* Android version: 10
* Device: OnePlus 6t

## Steps to reproduce
1. Take note of Library icon
2. Switch language & back
3. Go to back to More
4. Take note of Library icon

### Expected behavior
Consistent Library icon

### Actual behavior
The Library icon has switched due to different unfilled icons used for the base icon and the anim unfilled icon

ivaniskandar:
Path data from AVD version. Fixes #5447


arkon:
I guess I never noticed before, but why are some of these icons so big...

ivaniskandar:
> I guess I never noticed before, but why are some of these icons so big...

It's not the icon, it's just viewport size for path drawing. The actual size is still 24 dips

arkon:
Yeah, but you can generally scale your viewport when working on it so they're 1:1. Doesn't really matter though.

PR #5732, opened 2021-08-16
Fix today still being displayed even though relative time is off

ghostbear:
Fixes so "Today" and "Recently" doesn't get displayed when "Relative time" is off


app/src/main/java/eu/kanade/tachiyomi/util/lang/DateExtensions.kt
```diff
@@ -109,8 +109,8 @@ fun Date.toRelativeString(
     val difference = now.time - this.time
```
arkon, line 109: Might actually be better to just short circuit the function if range is 0.

PR #5739, opened 2021-08-18
Add read/unread and bookmark toggles as swipe actions for chapters

### LINKED ISSUE #3630: Swipe to mark chapters as read
I feel J2K's implementation of the swiping to 'mark as read' is more intuitive but I also like the current hold to select/multiselect. Same goes for the way to download chapters. Having the option to quickly download/swipe individually and multiselecting is something I'd like to see in the future. 

Maybe it could be done so mark as read is swipe left and download is swipe right. I know swipe right is used for bookmarking but I feel it's something used often enough for a swipe feature.

datreesezcup:
- Swipe left to toggle the read/unread status of a chapter
- Swipe right to toggle whether a chapter is bookmarked
- Swipe is disabled while in chapter selection mode

Closes #3630

This was pretty fun to implement!

### Video Preview
https://user-images.githubusercontent.com/25621728/129830844-f94b9540-1bd7-4603-96a6-eb32ea670ffe.mp4

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
@@ -21,6 +21,7 @@ import androidx.core.os.bundleOf
 import androidx.core.view.children
 import androidx.core.view.isVisible
 import androidx.recyclerview.widget.ConcatAdapter
+import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import coil.imageLoader
@@ -62,6 +63,7 @@ import eu.kanade.tachiyomi.ui.library.ChangeMangaCoverDialog
 import eu.kanade.tachiyomi.ui.library.LibraryController
 import eu.kanade.tachiyomi.ui.main.MainActivity
 import eu.kanade.tachiyomi.ui.manga.chapter.ChapterItem
+import eu.kanade.tachiyomi.ui.manga.chapter.ChapterItemTouchHelper
 import eu.kanade.tachiyomi.ui.manga.chapter.ChaptersAdapter
 import eu.kanade.tachiyomi.ui.manga.chapter.ChaptersSettingsSheet
 import eu.kanade.tachiyomi.ui.manga.chapter.DeleteChaptersDialog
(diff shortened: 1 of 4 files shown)


curche:
Nice :tada: 

@joseph619 its finally here!

J2K/Neko also has a simple quick tutorial when you open the chapter list for the first time. This lets users know that the functionality exists else it might go hidden. Would it be possible to add that?

vengiare:
- No animation after swipe looks weird. Could either ~~(steal)~~ copy
–  J2K's animation, item fades in after swipe
–  Gmail's amination, item slides back to place


- Both Google apps and J2K have Undo snackbar after swipe actions. A must have, I believe.


- In Google apps, swiped items are elevated (has shadow) and also has rounded corners. Probably a good suggestion.

IMG_20210819_074524


- Gmail's Swipe actions are configurable. Probably also a good suggestion.

IMG_20210819_075506


- Personally I wouldn't really prefer swipe actions. I feel like the threshold is too sensitive, for me at least. Maybe a toggle would be nice.

datreesezcup:
@curche
> J2K/Neko also has a simple quick tutorial when you open the chapter list for the first time. This lets users know that the functionality exists else it might go hidden. Would it be possible to add that?

I'll at some point see if that's something I could quickly add. Really, I think Tachiyomi could benefit from a sort of "tutorial" animation in several areas (reader, this for example).

---
@vengiare 
> No animation after swipe looks weird

Yeah, that was a side effect of the current implementation of the swipe controller

>Both Google apps and J2K have Undo snackbar after swipe actions. A must have, I believe.

I would disagree (at least as the feature currently stands). At least with Gmail, the actions you can do (or configure to do) are volitile (like deleting emails, archiving, etc). With these Tachiyomi actions (bookmark, read/unread), it'd be just as fast to just swipe again.

> Rounded Corners and Elevation

I'll look into it and see how it looks

>Custom Swipe Actions & Threshold

Yeah, the ability to change the actions might be a good feature to have.
As well as changing the threshold of what is considered a "swipe".

----

I'm going to put this PR into draft and see if I can redo some things. One of the oddities that made this so awkward to implement is how Tachiyomi lays out the manga details screen in comparison to J2K

CarlosEsco:
> I would disagree (at least as the feature currently stands). At least with Gmail, the actions you can do (or configure to do) are volitile (like deleting emails, archiving, etc). With these Tachiyomi actions (bookmark, read/unread), it'd be just as fast to just swipe again.

Tachiyomi action of read is volatile when its a downloaded chapter and a user has delete after mark as read enabled

arkon:
Going to close this since we're in the process of rewriting UI stuff with Compose, but will keep the original issue open since it's still a desirable feature.

PR #5741, opened 2021-08-18
Label one-shots correctly in anilist track search

curche:
The current query uses `type` to list the entry's Type. As per [Anilist graphql](https://anilist.github.io/ApiV2-GraphQL-Docs/) docs, `type` refers to whether an entry is anime/manga. This is redundant (since the query already limits it to `MANGA`). What we actually want is `format` which includes, among other things, whether the media of MANGA type is a One-shot or Manga

This should help in making search a bit better as one no longer needs to rely on the Date to figure out if its the One-shot entry or the Manga entry

Inspired by AL-chan which showed one-shots correctly
Changes to search query and corresponding JSONALManga structure
Tested on Pixel 3 Android 12 Beta

Screenshots:
| Before | After |
|----|----|
| Screenshot_20210818-122014 | Screenshot_20210818-122036 |


curche:
You can view the difference in query in the interactive Anilist GraphQL editor [here](https://anilist.co/graphiql?query=query%20Search(%24query%3A%20String)%20%7B%0A%20%20Page(perPage%3A%2050)%20%7B%0A%20%20%20%20media(search%3A%20%24query%2C%20type%3A%20MANGA%2C%20format_not_in%3A%20%5BNOVEL%5D)%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20title%20%7B%0A%20%20%20%20%20%20%20%20romaji%0A%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20format%0A%20%20%20%20%20%20type%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A&operationName=Search&variables=%7B%0A%20%20%22query%22%3A%20%22blue%20box%22%0A%7D)

One additional thing, the layout uses the string `track_type` which has a value 'Type'. So the screenshot shows 'Type: One-shot'
Could be changed to 'Format: One-shot' but this would change it for all Trackers which I didn't want to do. Would like to know what others think (Maybe a :+1: for changing it to "Format" or :-1: for keeping it as is)

arkon:
I'm fine with either one. Type is fine for now since that's what the app already had, but format does make sense too.

Issue #2507, opened 2020-01-15
[Feature Request] Ability to rename "Default" Category


RuthlessRuler:
### Why/User Benefit/User Problem
Well, it sucks to be unable to rename the default category.

### What/Requirements
When there are more than one Category, the ability to rename the default category should be added.

keinezeit8:
A workaround that works right now, is putting all series into a different category, and name that. The default category vanishes then.

RuthlessRuler:
> A workaround that works right now, is putting all series into a different category, and name that. The default category vanishes then.

Yeah it's an option but when we have 100+ Mangas, It's quite a work to select all as there is no way to select all in the App.(I think it's there in Other Like EH version tho.)

CrepeTF:
There is now a select all in the library.

CrepeTF:
Again, this should be closed due to the above ^

arkon:
I could still see the usecase for having it name-able, or at least localized.

PR #5751, opened 2021-08-19
Make default category translatable

ghostbear:
Also removes duplicate `Default` strings with a common one


arkon:
Closes #2507, kind of.

PR #5758, opened 2021-08-20
Use user preferred title language in Anilist

curche:
Since Anilist is only used by authenticated users, the title language
can be set using the `userPreferred` field

From the [docs](https://anilist.github.io/ApiV2-GraphQL-Docs/)
> MediaTitle.userPreferred: String
> The currently authenticated users preferred title language. Default romaji for non-authenticated

Changed wherever `title>romaji` was being used. Shouldn't have missed
any. `userPreferred` is also available for Staff and Character but not
relevant to Tachiyomi for now.

Users might need to go Logout and log back in on Anilist to see the
change. Actual setting can be found at https://anilist.co/settings/media

closes https://github.com/tachiyomiorg/tachiyomi/issues/5757


ghostbear:
> Users might need to go Logout and log back in on Anilist to see the change

They will not need to logout but they will probably need to update each tracked item to update the title if they want their preferred title to show

curche:
> > Users might need to go Logout and log back in on Anilist to see the change
> 
> They will not need to logout but they will probably need to update each tracked item to update the title if they want their preferred title to show

Oh I was actually unsure of it (hence the might). Wouldn't the refresh occur when the Tracking sheet is opened? OR you can do a full `Refresh Tracking Metadata` ig.

ghostbear:
> > > Users might need to go Logout and log back in on Anilist to see the change
> > 
> > 
> > They will not need to logout but they will probably need to update each tracked item to update the title if they want their preferred title to show
> 
> Oh I was actually unsure of it (hence the might). Wouldn't the refresh occur when the Tracking sheet is opened? OR you can do a full `Refresh Tracking Metadata` ig.

That will not update the title from my testing. You will have to update Anilist.kt `refresh` function to get the new remote title and AnilistModels.kt `toTrack` function to get also get the title from the manga object. Otherwise it will not update the title

curche:
`ALManga.title_romaji` -> `ALManga.title_user_pref` to indicate that the userPreferred title is being used

change one more query (`ALUserManga`) to use `format` instead of `type` (causes NPE; missed this in #5741)

ready for review

PR #5764, opened 2021-08-21
Add a Pull Request template

Soitora:
Not sure what to write for it (it's also the last item on the checklist [here](https://github.com/tachiyomiorg/tachiyomi/community)).

Added some reminders, mainly checking so it works in tablet mode.
Also added a quick template for providing images since that can be dull to write if you forgot how a table is setup.

Feel free to reject it completely or provide feedback on any changes


arkon:
Seems like a decent start. We can add to it if we need to, although I'm not sure what a good standardized checklist would really be.

PR #5766, opened 2021-08-21
Tweak MangaInfoHeader

ghostbear:
- Make margin between the transparent toolbar and cover more match 1.x
- Fixes from when the view was redone with MotionLayout

| Old | New |
|---|---|
| Screenshot_1629584325 | Screenshot_1629585629 |


app/src/main/res/layout/manga_info_header.xml
```diff
@@ -36,7 +36,7 @@
         android:layout_width="100dp"
         android:layout_height="0dp"
         android:layout_marginStart="16dp"
-        android:layout_marginTop="8dp"
+        android:layout_marginTop="24dp"
```
arkon, line 39: I actually feel like this is too much now that it's edge-to-edge and there's overall more content getting shown. I tried 16dp locally and it looked pretty good imo.

ghostbear: 16dp is what 1.x uses so I'm okay with that

vengiare:
The buttons below look odd now. Is bottom margin really needed?

vengiare:
Now it looks cramped compared to current one tho
image

PR #5767, opened 2021-08-21
Align manga info expansion buttons

Soitora:
Aligns the More/Less buttons, at least in English and on the two devices I've tested on. This is not a professional job but I'm not a professional so it checks out 🔫

It was mentioned on Discord that the filter icon doesn't represent the tap target, but I don't think (for now) that it's something that needs to be tweaked.

| New | Old |
| ----- | ---- |
| New | Old |


Soitora:
Superseded by #5771.

PR #5768, opened 2021-08-22
Tracking: replace tick with button

OncePunchedMan:
If wanting to add a manga to a tracker and it showing only one result, users thought it might automatically select this item. The tick will now appear after selecting an item:

- not having an item selected doesn't show the tick

- selecting an item enables tick

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSearchDialog.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSearchDialog.kt
@@ -63,7 +63,6 @@ class TrackSearchDialog : DialogController {
         binding = TrackSearchDialogBinding.inflate(LayoutInflater.from(activity!!))
 
         // Toolbar stuff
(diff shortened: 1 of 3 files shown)


arkon:
I think it was actually proposed at one point, but maybe we should just have a long button at the bottom similar to the default file picker, which would be disabled until something's selected. It seems a bit more intuitive than the current menu item.

OncePunchedMan:
sure

OncePunchedMan:
sure

OncePunchedMan:
Screenshot_20210824_024525

vengiare:
Why is the button enabled when nothing is selected tho

OncePunchedMan:
still WIP

Riztard:
> still WIP

make it a draft?

OncePunchedMan:
this ones ready for merging now

Screenshot_20210824_160924

Screenshot_20210824_160929

vengiare:
Why's it invisible when disabled tho

> Should be greyed out

OncePunchedMan:
Screenshot_20210824_160924

Screenshot_20210824_160929

PR #5771, opened 2021-08-22
New manga info expander

Soitora:
Supersedes #5767.

This is based on the old system we used before "More..." was introduced, it was mainly complained that the old expansion buttons took too much space, but that was because it was set on it's own line. I made this one in-line.

@ghostbear also suggested a scrim fade-up from the bottom instead of a ellipsize, but that won't end well if I personally try to do it.

### Images
| New | Old |
| ------- | ------- |
| New | Old |


app/src/main/res/layout/manga_info_header.xml
```diff
@@ -190,35 +190,27 @@
...
             app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toStartOf="parent"
             app:layout_constraintTop_toBottomOf="@+id/manga_cover"
             tools:text="Collapsed summary content Collapsed summary content Collapsed summary content Collapsed summary content Collapsed summary content Collapsed summary content" />
 
-        <View
-            android:id="@+id/manga_info_toggle_more_scrim"
-            android:layout_width="20dp"
-            android:layout_height="0dp"
-            android:background="@drawable/manga_info_more_gradient"
```
arkon, line 207: You should update the tablet version too, then remove this drawable.

arkon:
Not sure how I feel about the right part not being used when expanded. Was that intentional?

Soitora:
> Not sure how I feel about the right part not being used when expanded. Was that intentional?

What do you mean?

arkon:
Yeah, it continues to not use the space on the right when expanded.

Soitora:
This was indeed not intentional, I have fixed it 👍🏻

app/src/main/res/layout/manga_info_header.xml
```diff
@@ -190,35 +190,27 @@
...
             android:layout_marginTop="8dp"
-            android:layout_marginEnd="16dp"
+            android:layout_marginEnd="38dp"
             android:clickable="true"
             android:focusable="true"
             android:maxLines="2"
+            android:ellipsize="end"
+            android:breakStrategy="simple"
             android:textIsSelectable="false"
             app:layout_constraintEnd_toEndOf="parent"
```
arkon, line 200: I think this should be constrained to the start of `manga_info_toggle_more` rather than adjusting the end margin.

Soitora:
Does something like this look okay with newest commits? I'm really new to constraints so not sure I did it the way you wanted, literally used it first time yesterday and got taught a bit by Ghostbear-sensei

Not sure if there was some constraint I should have used instead of horizontal bias either

image

vengiare:
Can you add images for expanded state?

Riztard:
not sure if the arrow look weird floating in the middle or because "more" only using bottom line

Soitora:
> Can you add images for expanded state?

No, I'm at work for another 10 hours, but it looks fine

Soitora:
> not sure if the arrow look weird floating in the middle or because "more" only using bottom line and so there is a space for another word on top of it

Maybe initially? But using it a bit and it's definitely more enjoyable

> *use primary color for the arrow?

No, it's not a function that warrants it, it's supposed to match the filter button in the chapter list

ivaniskandar:
IMO labeled button is more fitting for this use case...

vengiare:
I don't know how the old system "took too much space." It's just one line. Maybe they're talking about manga with extra lines in the description?

Manga with no extra lines:
IMG_20210823_140731

Manga with extra lines:
IMG_20210823_140731

Soitora:
> I don't know how the old system "took too much space." It's just one line. Maybe they're talking about manga with extra lines in the description?
> 
> Manga with no extra lines:
> IMG_20210823_140731
> 
> Manga with extra lines:
> IMG_20210823_140706
> 
> (Edit: sent wrong pic)

This is not the old system I was referring to lol

vengiare:
Old expander, i mean

Soitora:
> Old expander, i mean

I was referring to what was used in **0.10.0** → **0.10.5**

vengiare:
I mean I'm not sure how one line "took too much space". I don't really see the issue there.

Soitora:
> I mean I'm not sure how one line for the Less text "took too much space". I don't really see the issue there.

If you don't have any reading comprehension then don't bother commenting about something irrelevant.

I have not removed the "Less" part, it's still there, just an icon instead of text.

vengiare:
> it was mainly complained that the old expansion buttons took too much space, but that was because it was set on it's own line.

I'm literally talking about this

Soitora:
> > it was mainly complained that the old expansion buttons took too much space, but that was because it was set on it's own line.
> 
> I'm literally talking about this

Install the aforementioned versions and check for yourself and stop talking about something that's still irrelevant to what I was talking about. Otherwise I'll lock the thread from further discussion.

Soitora:
@ghostbear added fade scrim:

| New scrim | Old scrim |
| ----------- | ----------- |
| Webp net-resizeimage | image |

vengiare:
I wonder if arrow can be on the middle instead cause it looks like it fits there perfectly

CarlosEsco:
I agree if you are going arrow and adding the fade scrim, the arrow in the middle would look better

ivaniskandar:
And I'd say try below the chips since the button also modify the chips layout

vengiare:
> And I'd say try below the chips since the button also modify the chips layout

Hope not, cause something like this would look perfect already
image

Soitora:
I hear you both, I'd like for @arkon to give his thoughts first

Scrim **OR** No scrim
Scrim, button right **OR** Scrim, button centered

Soitora:
> And I'd say try below the chips since the button also modify the chips layout

It makes sense in theory, but not in practice

arkon:
The original shift to the current design (which was lifted from J2K) was intended to save on vertical space, so a horizontal scrim + icon below wouldn't be good for that. The horizontal scrim also seems annoying since it makes it harder to read the 2 lines that _are_ visible.

As for the current text approach, I think it's fine since the tap target isn't actually limited to it, although localization does get wonky when the text becomes longer at times.

I think from a consistency standpoint, the previous 0.10.x design would actually be best since it matches what you'd get from a collapsible filter group too. If we're fine with losing out on some vertical space (which may be fine, given that we cut down on the gap at the top of the screen now), then I'd actually vote for that.

ghostbear:
> The horizontal scrim also seems annoying since it makes it harder to read the 2 lines that are visible.

For this the height value could be tweaked for the scrim to be lower, probably need to tweak the center value in the gradient to make it look good

> although localization does get wonky when the text becomes longer at times.

The chevron fixes this 

> If we're fine with losing out on some vertical space (which may be fine, given that we cut down on the gap at the top of the screen now), then I'd actually vote for that.

I would say add back some vertical spacing, if it's placed in a good spot it will probably not add much vertical spacing (like the position vengaire showed, but if it's added below chips it will add more)

Soitora:
> The original shift to the current design (which was lifted from J2K) was intended to save on vertical space, so a horizontal scrim + icon below wouldn't be good for that. The horizontal scrim also seems annoying since it makes it harder to read the 2 lines that _are_ visible.

Could adjust the scrim maybe?

> As for the current text approach, I think it's fine since the tap target isn't actually limited to it, although localization does get wonky when the text becomes longer at times.

From a interactive point I also think using a chevron arrow like this makes most sense from a user perspective (the collapsible part), and does not look odd in different languages. Design-wise, cleaner.

> I think from a consistency standpoint, the previous 0.10.x design would actually be best since it matches what you'd get from a collapsible filter group too. If we're fine with losing out on some vertical space (which may be fine, given that we cut down on the gap at the top of the screen now), then I'd actually vote for that.

Not totally sure I get the outcome of this message, but wouldn't this original PR (pre-fade) most closely match a collapsible filter group?

I also checked 0.10.5 and we've gained a fair bit of extra vertical space to work with. Current info ends a fair bit before the old one even started (about 2-3 lines of space), a horizontal chevron overlayed over the third line would only add that extra 1 line. Although I'm personally fine with my original approach of a side collapsible as well

arkon:
> Not totally sure I get the outcome of this message, but wouldn't this original PR (pre-fade) most closely match a collapsible filter group?

Sort of, but I meant going back to the 0.10.5 (I think?) design with the heading consistent with the chapters+filter heading lower on that screen.

Soitora:
I did some example builds, please tell your thoughts  regarding it, I made it quick and dirty instead of fooling around in Photoshop, some things might look off but it's the general concept I'm trying to convey.

|  Images  |  Images |
| - | - |
| 2 Lines + Light Fade + Center arrow below 2 Line + Light Fade + Center arrow below | 3 Line + Fade + Side arrow 3 Line + Fade + Side arrow |
| 3 Line + Fade + Center arrow inside 3 Line + Fade + Center arrow inside | 2 Line + Side arrow 2 Line + Side arrow |
| 2 Line + Light fade + Top info bar 2 Line + Light fade + Top info bar | 2 Line + Light Fade + Side arrow 2 Line + Light Fade + Side arrow |

CarlosEsco:
3 Line + Fade + Center arrow inside  is my vote

Riztard:
> 3 Line + Fade + Center arrow inside is my vote

if using that maybe less fade
but i think it waste of space. so i like side arrow more. 

> IMO labeled button is more fitting for this use case...

and i still think the current one is more suitable too like instagram

PR #5785, opened 2021-08-25
Replace NetworkInfo with NetworkCapabilities

TacoTheDank:
I built an APK and played around with downloading. Behavior is the same as far as I can tell. NetworkInfo was deprecated in API 29.

The replacement was introduced in API 21, so no checks are needed (of course).


app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadService.kt
```diff
@@ -145,28 +140,33 @@ class DownloadService : Service() {
...
-            }
-            DISCONNECTED -> {
-                downloadManager.stopDownloads(getString(R.string.download_notifier_no_network))
-            }
-            else -> {
-                /* Do nothing */
+    private fun onNetworkStateChanged() {
+        val manager = connectivityManager
+        val activeNw: Network = manager.activeNetwork ?: return
+        val nwCapabilities = manager.getNetworkCapabilities(activeNw) ?: return
```
ghostbear, line 147: Use pronounceable names? 
`activeNw` -> `activeNetwork`
`nwCapabilities` -> `networkCapabilities`

PR #5795, opened 2021-08-26
Limit the length of Authors and Artists until info is expanded

Soitora:
Caps **Authors** and **Artists** to max `2` lines at an unexpanded state, goes to normal `MAX_INT` if expanded.
I get this is not the best solution to solve how to handle it but it's a start towards a slightly better experience.

### Images
| New (Regular) | New (Expanded) | Old |
| ------- | ------- | ------ |
| Minimized | Expanded | Old screenshot, it doesn't clip but it is crazy long Old |


arkon:
Considering that:
- It's kind of rare to have a ton of authors and artists
- It no longer cuts off

I'm kind of tempted to just not merge/handle this use case?

vengiare:
I like the two lines limit idea (anthologies aren't that uncommon) but I'm not sure about the other one. It would be kinda disorienting if the Info section gets pushed down when expanding (due to artist/author list also expanding)

For me I think it's either a permanent two-line limit, or none at all.

Soitora:
> Considering that:
> - It's kind of rare to have a ton of authors and artists
> - It no longer cuts off
> 
> I'm kind of tempted to just not merge/handle this use case?

You know what, I agree, originally made it just because I saw it cut off anyways. Would be nice if someone did one of the other discussed ideas on how it could be handled.

PR #5796, opened 2021-08-26
Center align title to cover and fix text clipping

Soitora:
Centered the Manga title vertically, no more "no top is better" or "no bottom better", it's just always perfect.
I also fixed content expansion through the help of constraint barriers.

Maybe in the future we could do a better way to present authors/artists, but this combined with #5795 is aimed to make it more pleasant.

| New | Old |
| --- | --- |
| New2 | Old2 |
| New3 | Old3 |
| New4 | Old4 |


ivaniskandar:
For me it's better to snap the infos to the top instead of centering it like how Play Books does it.

Soitora:
> For me it's better to snap the infos to the top instead of centering it like how Play Books does it.

I hear some say top best some bottom best, there's no consensus, I say middle best

vengiare:
I'd rather have it centered than leave a gaping hole in between info and buttons. 

#

> Like how Play Books does it.

The thing with Play Books/Movies is they use a much bigger font so the whitespace is much easier to fill, compared to Tachiyomi.

CrepeTF:
> but no one considered middle

(X) doubt
I already did center on Tako

Soitora:
> > but no one considered middle
> 
> (X) doubt
> I already did center on Tako

Yet you never said anything smh

CrepeTF:
> > > but no one considered middle
> > 
> > (X) doubt
> > I already did center on Tako
> 
> Yet you never said anything smh

Inferior exploring skills smh

Soitora:
How the expansion expands with the help of constraint barriers

https://user-images.githubusercontent.com/10836780/131552135-ce93bcc3-8e21-4d2e-9fad-43ce10e653f6.mp4

Issue #5699, opened 2021-08-11
Using other trackers with the Komga tracker provides unexpected behavior


kiriya-aoi:
### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] I have updated the app to version **[0.12.0](https://github.com/tachiyomiorg/tachiyomi/releases/tag/v0.12.0)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

### Tachiyomi version

0.12.1

### Android version

Android 11

### Device

Oneplus 7 Pro

### Steps to reproduce

1. Use Komga and add a series to your Komga library. In our example, let's say this series has 12 books.
2. Since this manga is probably released as volumes, set the sort number for each volume to the last chapter in that volume. Volume 1 happens to end at chapter 4 for our purposes, so it gets 4. Other volumes have some number greater than that.
    * This is so that in Tachiyomi reading progress can be synced to common manga trackers accurately, instead of using the volume number as the chapter number.
    * Alternately, you could simply be missing chapters 1-3 for some reason. You still don't want tracking to be wrong, so you set the sort number to the correct number for every other chapter and hope you'll add chapter 1-3 to Komga later.
3. Use the Komga source in Tachiyomi and enable tracking with Komga and at least one other tracker. Also set trackers to automatically update when you read a chapter.
4. Add the series from step 1. to your Tachiyomi library and track it with your other tracker (the Komga one should be added automatically).
5. Read the first chapter/book in the list in this series. This is either Volume 1 or Chapter 4, depending on which of the two examples we're following.

### Expected behavior

Komga tracker increments from 0 to 1, as I have read the first chapter/book from that series. This is the same behavior as you get when you have a series tracked only with the Komga tracker.

Non-Komga tracker updates to 4, since that is the sort number it gets from Komga.

### Actual behavior

Non-Komga tracker behaves as expected and updates to chapter 4.

Komga tracker also updates to 4 (or lower, if the series happened to have fewer total books).
[...]

PR #5802, opened 2021-08-27
change Track.last_chapter_read to Float

gotson:
As discussed in #5699 and #5658, the last chapter read being an Int poses a problem for the Komga tracker. Because of this, the way to automatically sync local chapters to remote was done using ordinal chapter index, instead of `chapter_number`. This poses the problem raised in #5699, and also prevents the opening up of `syncChaptersWithTrackServiceTwoWay` to other trackers (proposed in #5658).

The choice to convert the last chapter read to Int is left to the `TrackerService`s. In this PR, everything was converted to Int, so as not to change too many things at the same time.

SQLite doesn't support altering column, so the only way to do so is to:
- rename the table to a temp table
- create the new table
- copy data from the renamed table to the new table
- drop the renamed table


gotson:
> There's a merge conflict, FYI.

I'll have a look, changes go so fast here 😅

PR #5806, opened 2021-08-27
Replace MotionLayout with full screen dialog

ghostbear:
Adds
- Menu for action from previous long clicking cover
- Ability to zoom around

Removes 
- The sick animations
- Some bugs

| Phone | Tablet | 
|---|---|
| Screenshot_1630104813 | Screenshot_1630105126 |

https://user-images.githubusercontent.com/6576096/131196029-43490e7b-6a73-48a8-a8b7-757966e0d1eb.mp4


vengiare:
Looking at it, it appears that it doesn't take navbar into account when centering?

IMG_20210828_150606

#

Also I believe this doesn't remove long press dialog, right?

ghostbear:
> Looking at it, it appears that it doesn't take navbar into account when centering?

Should be fixed this was recorded before bottom inset (E2E) was added 

> Also I believe this doesn't remove long press dialog, right?

No

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -728,6 +732,21 @@ class MangaController :
...
+        val manga = manga ?: return
+        dialog = MangaFullCoverDialog(this, manga)
+        dialog?.addLifecycleListener(
+            object : LifecycleListener() {
+                override fun postDestroy(controller: Controller) {
+                    super.postDestroy(controller)
+                    dialog = null
+                }
+            }
+        )
```
arkon, line 746: Is this worth doing? I wonder if it'd be better to just keep a single `dialog` instance around (after the first initialization) and just destroying it when the MangaController gets destroyed. That way, you don't need to keep recreating/destroying it if someone taps it repeatedly.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -800,6 +819,9 @@ class MangaController :
 
     fun onSetCoverSuccess() {
         mangaInfoAdapter?.notifyDataSetChanged()
+        if (dialog != null) {
```
arkon, line 822: Is this necessary?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaFullCoverDialog.kt
```diff
@@ -0,0 +1,118 @@
...
+            type(navigationBars = true) {
+                // Padding will make to image top align
+                // This is likely an issue with SubsamplingScaleImageView
+                margin(bottom = true)
+            }
+        }
+
+        setImage(manga)
+
+        return AppCompatDialog(activity!!, R.style.ThemeOverlay_Tachiyomi_Dialog_Fullscreen).apply {
```
arkon, line 83: Kind of feel like we should consider abstracting this sort of thing if we're using it in multiple places now.

app/src/main/res/menu/full_cover.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
arkon, line 1: Maybe we should remove the submenu from the manga controller too. Not sure if the redundancy is worth having.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -435,7 +439,6 @@ class MangaController :
 
         // Hide options for non-library manga
         menu.findItem(R.id.action_edit_categories).isVisible = presenter.manga.favorite && presenter.getCategories().isNotEmpty()
-        menu.findItem(R.id.action_edit_cover).isVisible = presenter.manga.favorite
```
arkon, line 438: Need to make sure this logic is handled in the new dialog's menu too.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
ghostbear: > I wonder if it'd be better to just keep a single dialog instance around (after the first initialization) and just destroying it when the MangaController gets destroyed.

Hmm, doesn't seem to get how to do this

> That way, you don't need to keep recreating/destroying it if someone taps it repeatedly.

Currently, it will only be recreated after to user leaves the dialog in other words when it is destroyed.
And the first if statement show handle if someone taps it repeatedly

PR #5824, opened 2021-08-30
Reuse reader's image view in MangaFullCoverDialog

ivaniskandar:
This allows MangaFullCoverDialog to show animated cover and avoid duplication.


app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaFullCoverDialog.kt
```diff
@@ -108,12 +107,54 @@ class MangaFullCoverDialog : DialogController {
...
+                                    return true
+                                }
+
+                                override fun onSingleTapConfirmed(e: MotionEvent?): Boolean {
+                                    dialog?.dismiss()
+                                    return super.onSingleTapConfirmed(e)
+                                }
+                            }
+                        )
+                        setZoomTransitionDuration(500)
```
ghostbear, line 141: Maybe add an extension function for `TRANSITION_ANIMATION_SCALE` as we have for `ANIMATOR_DURATION_SCALE` to disable or slow down the transition

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaFullCoverDialog.kt
```diff
@@ -108,12 +107,54 @@ class MangaFullCoverDialog : DialogController {
...
+                    }
+                } else {
+                    val bitmap = (it as BitmapDrawable).bitmap
+                    SubsamplingScaleImageView(activity!!).apply {
+                        layoutParams = FrameLayout.LayoutParams(
+                            ViewGroup.LayoutParams.MATCH_PARENT,
+                            ViewGroup.LayoutParams.MATCH_PARENT
+                        )
+                        setMinimumDpi(45)
+                        setDoubleTapZoomDuration(500)
```
ghostbear, line 152: > Maybe add an extension function for TRANSITION_ANIMATION_SCALE as we have for ANIMATOR_DURATION_SCALE to disable or slow down the transition

Same here

ivaniskandar:
(Self) reminder to scale the zoom duration of reader outside of this PR.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaFullCoverDialog.kt
```diff
@@ -108,12 +109,54 @@ class MangaFullCoverDialog : DialogController {
...
 
     fun setImage(manga: Manga?) {
-        val manga = manga ?: return
+        if (manga == null) return
         val request = ImageRequest.Builder(applicationContext!!)
             .data(manga)
             .target {
-                val bitmap = (it as BitmapDrawable).bitmap
-                binding?.fullCover?.setImage(ImageSource.cachedBitmap(bitmap))
+                val imageView = if (it is Animatable) {
```
arkon, line 116: I feel like with some of this logic getting duplicate between the reader and here, we should try to extract it out somewhere if possible?

ivaniskandar: Yeah that's possible. I can try wrap it inside a custom view but there's some specific reader configs should I bring them over?

arkon: Doesn't necessarily need to be a full view, it can just be some util functions to get the photoview/ssiv. I didn't look into the different configs between their usages though, so I'm not sure if you'll need a bunch of arguments to allow customization or not.

PR #5834, opened 2021-08-31
Hide language tag when only one language is used

Soitora:
It feels pretty so-so that the language code is displayed for all sources in the Manga info when the user is only ever using a single language.
This fixes that, if only a single source is used then it will display just the name, otherwise it'll show the tag like usual.

In edge cases where a user disables a source after favoriting it it will still show the language tag for that specific entry.

*(👻🐻 helped too much and ended up having almost all of the code written by him...)*

### Images
| Single language | Multiple languages |
| ------- | ------- |
| SingleLang | MultiLang |
| SingleLangSources | MultiLangSources |


Soitora:
Maybe it shouldn't count "All" as a language, I'll fix it tomorrow after work

app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoHeaderAdapter.kt
```diff
@@ -260,7 +260,23 @@ class MangaInfoHeaderAdapter(
...
+                    val enabledLanguages = preferences.enabledLanguages().get()
+                        .filterNot { it == "all" }
+
+                    text = if (enabledLanguages.size == 1) {
+                        // For edge cases where user disables a source they got manga of in their library.
+                        if (source.lang !in enabledLanguages) {
+                            source.toString()
+                        } else {
+                            // Hide the language tag when only one language is used.
+                            source.toString().replace(Regex("\\s\\([^()]*\\)\$"), "")
```
arkon, line 273: You can just use `source.name`

PR #5846, opened 2021-09-04
Komga tracker v2

gotson:
This follows #5802 

- use `chapter_number` instead of ordinal index for `syncChaptersWithTrackServiceTwoWay`. This is related to #5658 and #5699
- use v2 api for Komga tracker for series. This is related to #5699

The new [v2 API](https://github.com/gotson/komga/commit/9d92b2594d5b32df1963671871185fb76b4e7d30) on Komga side has been released in [0.121.0](https://github.com/gotson/komga/releases/tag/v0.121.0) we probably need to wait a week or 2 for people to update.


arkon:
> we probably need to wait a week or 2 for people to update

So I should merge this in around a week?

gotson:
Up to you, updating Komga is usually easy, so if anything breaks for people they just have to upgrade Komga or rollback Tachiyomi.

Given it's only for the preview (I don't know when you will have a new stable release?) it should be fine.

gotson:
Thanks!

PR #5849, opened 2021-09-04
Change ProtoNumber of Backup Models for History and Source to a non-zero digit

### LINKED ISSUE #5334: Backup files contain invalid protobuf field numbers
## Device information
* Tachiyomi version: 0.11.1
* Android version: 11
* Device: Redmi Note 4

## Steps to reproduce
1. Create a backup
2. Try parsing it using `protoc` or other protobuf decoder/library

### Expected behavior
Backup file with data that can be properly read by other protobuf parsers.

### Actual behavior
Decoder errors or corrupt data, depending on the parser implementation.

## Other details
I discovered this while trying to write a script for extracting my read history from a backup file. Both `BackupHistory` and `BackupSource` define a field with `@ProtoNumber(0)`, but [the documentation](https://developers.google.com/protocol-buffers/docs/proto#simple) says the minimum valid field number is 1, in fact, `protoc` will throw a fatal error if you try to compile a message definition that sets the number of a field to 0.

https://github.com/tachiyomiorg/tachiyomi/blob/ec5e6958ef4f12f6cdf6694a37637770ca1e67da/app/src/main/java/eu/kanade/tachiyomi/data/backup/full/models/BackupHistory.kt#L8
https://github.com/tachiyomiorg/tachiyomi/blob/ec5e6958ef4f12f6cdf6694a37637770ca1e67da/app/src/main/java/eu/kanade/tachiyomi/data/backup/full/models/BackupSource.kt#L9

Trying to parse the backup files may generate errors or return corrupt data, depending on the library used. For example, `protoc` 3.16 interprets the affected fields as a sequence of raw bytes:

```
100: "\002\004Guya\010\226\377\245\206\360\334\330\256@"
```

Although `kotlinx.serialization` doesn't seem to care and the app can load the data just fine, I guess it _may_ cause trouble if something changes in a future version. People interested in reading backup files using external tools can either use a library that allows zeros as field numbers or try to patch the file before attempting to load it.

Jacekun:
Changed Proto Backup Models to non-zero ProtoNumber

Fixes #5334


arkon:
We'd have to find a backwards compatible way of fixing this. Breaking backups defeats their purpose.

Jacekun:
There could be a workaround for current backups. Say, install the current version of Tachiyomi, and load the backup.

I can't see any other solution since the protobuf format is highly dependent on ProtoMember Numbers to map its elements.

paulo27ms:
You don't necessarily have to make them all compatible, just go the route of when proto backup first appeared and there was still legacy backups.
The new code backups will be a "v2" and a "v1" (the current) will still be able to be loaded if it's detected that it's not a v2 backup, just keep the old code to load old backups. You could have some code to detect the version of the backup and if it doesn't have a version then it's an old one.
This means new backups won't be compatible with old versions but they don't necessarily have to be as long as no backups that exist right now break in future versions. (We could potentially have an option to generate "v1" backups manually for some backwards compatibility for some time.)
After some time you just remove the code to load old backups like you did for legacy.

Jacekun:
@paulo27ms  That idea works, tho it sounds like another PR. 

As for the checking, I would try something like loading the backup and try to serialize BackupSource with 'v2' first, then if it doesn't map the 'name' element, that could imply that it uses the old zero-based proto number and deserialize it accordingly. This sounds like an extra step and will probably take longer to test for edge-cases but if this works, then amazing.

arkon:
I'm not sure you really need to do that. You can probably just add another field to the `Backup` model with a new list of fixed `BackupSource`s, with the 2 being nullable. The code can then try to get the sources based on the presence of one or the other.

@jobobby04 may be able to speak more towards how to make those kind of changes, since he originally implemented the protobuf backups.

jobobby04:
Adding versions to a protobuf backup would defeat the purpose of the migration to protobuf, this can easily be handled by adding another field + having default values. I talked about this before in the discord and suggested a type of fix, I just haven't done the work and PR'd it 😄 
This is my suggested fix process
This:
```kotlin
data class Parent(
    @ProtoNumber(101) var history: List<BackupHistory> = emptyList(),
)
data class BackupHistory(
    @ProtoNumber(0) var url: String,
    @ProtoNumber(1) var lastRead: Long
) 
```
would become
```kotlin
data class Parent(
    @ProtoNumber(101) val brokenHistory: List<BrokenBackupHistory> = emptyList(),
    @ProtoNumber(102) var history: List<BackupHistory> = emptyList(),
)
data class BrokenBackupHistory(
    @ProtoNumber(0) var url: String,
    @ProtoNumber(1) var lastRead: Long
) 
data class BackupHistory(
    @ProtoNumber(1) var url: Long
    @ProtoNumber(2) var lastRead: String,
) 
```
And then it would be easy to add support for restoring both, like so
```kotlin
val parent = getParent
val history= parent.brokenHistory.map { BackupHistory(it.url, it.lastRead) } + parent.history
```

Jacekun:
@jobobby04  Amazing! I never thought of it like that. Guess I need to learn more about protobuf. But anyways, would it be cool if I implement **that** solution on this PR?

jobobby04:
Yep, go ahead! It's better that it gets done sooner, and I probably won't get to it anytime soon 😄

app/src/main/java/eu/kanade/tachiyomi/data/backup/full/models/BackupSource.kt
```diff
@@ -5,9 +5,13 @@ import kotlinx.serialization.Serializable
 import kotlinx.serialization.protobuf.ProtoNumber
 
 @Serializable
-data class BackupSource(
+data class BrokenBackupSource(
     @ProtoNumber(0) var name: String = "",
     @ProtoNumber(1) var sourceId: Long
+)
+data class BackupSource(
```
arkon, line 12: Same here.

app/src/main/java/eu/kanade/tachiyomi/data/backup/full/models/BackupManga.kt
```diff
@@ -33,8 +33,10 @@ data class BackupManga(
     // Bump by 100 for values that are not saved/implemented in 1.x but are used in 0.x
     @ProtoNumber(100) var favorite: Boolean = true,
     @ProtoNumber(101) var chapterFlags: Int = 0,
-    @ProtoNumber(102) var history: List<BackupHistory> = emptyList(),
+    @ProtoNumber(102) var brokenHistory: List<BrokenBackupHistory> = emptyList(),
     @ProtoNumber(103) var viewer_flags: Int? = null
```
jobobby04, line 37: This needs a comma

arkon:
Thanks!

PR #5853, opened 2021-09-04
Skip bookmark check when cancelling downloads

### LINKED ISSUE #5851: Bookmarked chapters download not being cancelled
### Steps to reproduce

1. Bookmark a chapter
2. Start to download it
3. Try to cancel it

### Expected behavior

The download should be canceled just fine.

### Actual behavior

The download isn't being canceled.

### Crash logs

_No response_

### Tachiyomi version

Preview r3756

### Android version

Android 10

### Device

Moto G7 Power

### Other details

My guess is that canceling from chapter page uses the same method as deleting, so it also checks for the "Don't delete bookmarked chapters" option.
Laziest scenario you could just overload it with an additional param if it's being canceled, which skips that check. That should work in order to change just a few pieces of code. (I think, if Kotlin also allows that)

### Acknowledgements

hyoretsu:
Should close #5851 if nothing went wrong/I didn't do any mistakes.


hyoretsu:
Semantics might be a little off since I'm used to TS which is pretty different from Kotlin

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
```diff
@@ -236,8 +236,13 @@ class DownloadManager(private val context: Context) {
...
      * @param source the source of the chapters.
      */
-    fun deleteChapters(chapters: List<Chapter>, manga: Manga, source: Source): List<Chapter> {
-        val filteredChapters = getChaptersToDelete(chapters)
+    fun deleteChapters(chapters: List<Chapter>, manga: Manga, source: Source, isCancelling: Boolean = false): List<Chapter> {
+        var filteredChapters = chapters
+
+        if (isCancelling == false) {
+            filteredChapters = getChaptersToDelete(chapters)
+        }
```
arkon, line 244:
```kotlin
val filteredChapters = if (isCancelling) {
    chapters
} else {
    getChaptersToDelete(chapters)
}
```

hyoretsu: Ooohh, that's closer to the ternary I originally thought of. I also wasn't sure if simply doing `(isCancelling)` was valid in Kotlin.

arkon: You can just commit again to the same branch. I'll squash them when merging.

arkon:
Thanks!

PR #5854, opened 2021-09-04
Remove the remaining MotionLayout

ghostbear:
Also adds an tap target to dead area around toggle less chevron 
image


app/src/main/res/layout-sw720dp/manga_info_header.xml
```diff
@@ -256,6 +257,19 @@
             app:layout_constraintTop_toBottomOf="@+id/manga_summary_text"
             tools:visibility="gone" />
 
+        <View
```
arkon, line 260: Maybe we can just make `manga_info_toggle_less` full width?

ghostbear: The background (chevron) would be stretched if it were full width. An alternative would be to surround an `ImageView` with another view like a `FrameLayout` that is full width if you fancy that more

arkon: Eh, I didn't realize it was a button; not sure if something like `scaleType` would work for that. I guess this is fine?

ghostbear: That would require it to change to a `ImageButton` from `MaterialButton`. But the chevron are drawable so it makes more sense to have them as `ImageButton`s

Issue #5680, opened 2021-08-09
Per-category Delete After Read (Read/Marked as Read) setting not woking


giosann:
### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] I have updated the app to version **[0.12.0](https://github.com/tachiyomiorg/tachiyomi/releases/tag/v0.12.0)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

### Tachiyomi version

0.12.1 / 0.12.1-3668

### Android version

Android 10 / Android 11

### Device

Galaxy S9+ / Galaxy Tab S6

### Steps to reproduce

Go to Settings - Download 

### Expected behavior

I expect to find "Categories to exclude in removing" as was added in PR https://github.com/tachiyomiorg/tachiyomi/pull/4121

### Actual behavior

The setting is not present

### Other details

This was originally requested with Feature request https://github.com/tachiyomiorg/tachiyomi/issues/1412 and added with pr https://github.com/tachiyomiorg/tachiyomi/pull/4121 but it is not working.

FourTOne5:
PR #4121 wasn't merged so it is likely you won't actually find the option.

Also, reading the PR's conversation it seems like @arkon misunderstood the PR's function.

arkon:
I did misunderstand it, sorry. I'll keep this one open to track it for now.

tobinstultiens:
Yeah and I also completely forgot about it since I was quite busy at the time.
But if you need help with this issue then I can make it again for the current version.

arkon:
@tobinstultiens That'd be great!

PR #5857, opened 2021-09-05
Allow exclusion on Delete After Read per category

tobinstultiens:
Created a settings that allows you to set certain categories to ignore delete after read.
Related to #5680 and #4121


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -409,8 +409,20 @@ class ReaderPresenter(
         val currentChapterPosition = chapterList.indexOf(currentChapter)
         val removeAfterReadSlots = preferences.removeAfterReadSlots()
         val chapterToDelete = chapterList.getOrNull(currentChapterPosition - removeAfterReadSlots)
+
+        // Retrieve the categories that are set to exclude from being deleted on read
```
arkon, line 413: I think it'd make more sense to put this in `DownloadManager`.

arkon (commented):
Shouldn't this also apply for users manually marking chapters as read too?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
tobinstultiens: The download manager was the exact location I was looking for.
I have now placed it there although I needed to add an injection of the db and by doing this it now also works for the marked as read.

tobinstultiens:
Now it also ignores it when it is manually marked as read.

arkon:
Thanks!

PR #5869, opened 2021-09-07
During migration, only do MangaController replacement if previous controller is also MangaController

FlaminSarge:
If previous controller is instead a MigrationController/other, push the new MangaController onto the stack instead.


FlaminSarge:
(Is there a better way to check the top controller instead of backstack.last()?)

arkon:
> (Is there a better way to check the top controller instead of backstack.last()?)

Not that I know of, but you could add an extension function to make it sound less janky.

FlaminSarge:
> > (Is there a better way to check the top controller instead of backstack.last()?)
> 
> Not that I know of, but you could add an extension function to make it sound less janky.

Eh it's probably fine as-is. If we have more extensions to add to Router we'll create RouterExtensions under util pkg and make a getTopController over there.

PR #5905, opened 2021-09-13
Allow preferences to multi-line

Soitora:
I kinda wanted to improve preferences and got reminded that text isn't multiline, but now it can. Although I have my doubts that I did it the most optimal or complete way, inb4 long review of places I missed.

Preferably try to not make title multi-line in English, but some translations are hard to fit in the same size

### Images
| New | Old |
| ------- | ------- |
| New | Old |


arkon:
Honestly don't even remember if there's other places that'd need the change, but we can change those if we come across them later.

PR #5923, opened 2021-09-17
Adjust gray reader background color, make it the default option.

Long-FA:
### Description
Now that gray reader background is an option, there is no reason not to use it over black because:
- It works well for both white and black manga pages.
- It reduces eye strain for long-term reading.

Still, the current gray color #212025 is still considered too dark. I propose two candicates from Material Design:
- Light Gray #757575, gives the [same contrast ratio](https://contrastchecker.online/) for white and black pages.
- Gray #424242, a bit darker than #757575 but still acceptable.

For reference, GIMP's Dark theme is #454545 and Gray is #777777.

### Images

Current gray #212025
| White | Black |
| ------- | ------- |
|  |  |

Gray #424242
| White | Black |
| ------- | ------- |
|  |  |

Light Gray #757575
| White | Black |
| ------- | ------- |
|  |  |

Open for debate.


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsReaderController.kt
```diff
@@ -97,7 +97,7 @@ class SettingsReaderController : SettingsController() {
                 titleRes = R.string.pref_reader_theme
                 entriesRes = arrayOf(R.string.black_background, R.string.gray_background, R.string.white_background, R.string.automatic_background)
                 entryValues = arrayOf("1", "2", "0", "3")
-                defaultValue = "1"
+                defaultValue = "2"
```
ghostbear, line 100: FYI this doesn't change the default value, you will have to change it in `PreferenceHelper` as well.

However, it will probably be an unwanted change for a lot of people who haven't changed their background from black. I would say don't change the default value and keep gray as an option for users that want to have a gray background

Long-FA: Thanks for the head up, i'm still learning the basics of coding. Should be this value, isn't it? 5e38bb9

vengiare:
White, Gray, and Black is just the same colors as app themes; Light, Dark (except Tako and Midnight Dusk), and Pure Black.

I understand the idea but a lighter shade of gray can just be added instead of completely replacing the current one. Could just rename the current one to Dark gray really.

arkon:
I don't have much of an opinion on the particular shade of grey, but I don't think the default should be grey. I view black is sort of "not adding a background around the page", whereas white or grey sort of do.

arkon:
Does anyone else support the change in the shade of grey? If not, we can close this. If there is, then we'd need to at least keep the existing default of black.

Riztard:
i think its ok change to Gray #424242, but not default

vengiare:
Don't really know if reader's grey should differ from app theme's grey.

PR #5930, opened 2021-09-18
rewrite getFormat the kotlin way

AriaMoradi:
Rewrote `LocalSource.getFormat(File)` when adopting it into Tachidesk, looks nice so why not upstream it?


app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -272,17 +272,17 @@ class LocalSource(private val context: Context) : CatalogueSource {
...
+        return file.run {
+            when {
+                isDirectory -> Format.Directory(file)
+                extension.equals("zip", true) -> Format.Zip(file)
+                extension.equals("cbz", true) -> Format.Zip(file)
+                extension.equals("rar", true) -> Format.Rar(file)
+                extension.equals("cbr", true) -> Format.Rar(file)
+                extension.equals("epub", true) -> Format.Epub(file)
+
+                else -> throw Exception("Invalid chapter format")
```
scb261, line 284: Shouldn't it use `R.string.local_invalid_format`?

arkon: Sigh this is what I get for looking at PRs on a phone.

AriaMoradi: I'm really sorry, I copied the code over and forgot that Android uses R

PR #5931, opened 2021-09-18
App-wide typography adjustments

ivaniskandar:
Also adjust chapter item to properly accommodate larger font and/or display size.

| c1 | c2 |
| ------- | ------- |
| image | image |
| image | image |
| image | image |
| image | image |


arkon:
There's some merge conflicts, btw. I think this is generally fine, although there may be some backlash around the library compact font change.

ivaniskandar:
I'll fix the conflicts. About the title font change, should I add an option for it? I can possibly add it in Library->Display

arkon:
> About the title font change, should I add an option for it?

For now, I'd say no. If there's sufficient enough backlash, we could maybe consider it.

PR #5956, opened 2021-09-22
Reattach after slight delay instead on every database update

ghostbear:
Currently when updating the library and one has per category display setting enabled it causes scroll lag due to the adapter being reattached.

To fix this a delay is used instead of updating it when the database changes, only when per category display setting is enabled. By delaying the reattach when the display setting gets changed it does not cause the weird behavior fixed by #5640

https://user-images.githubusercontent.com/6576096/134382158-22c28979-e44b-4d28-b437-7a4ac558f3bb.mp4 

https://user-images.githubusercontent.com/6576096/134382179-f1481936-7618-40aa-8676-873715c0099e.mp4


dabreadman:
LGTM!

PR #6000, opened 2021-10-01
Ability to order sources by library count when migrating 

### LINKED ISSUE #4703: [Feature Request] order sources by library count when migrating
The recent issues that multiple sources have had have given a good example as to why users should try and avoid having all of their manga from a single source. In order to make this easier to avoid for the users I would like to request the following:

1: Add an option to sort the result list based when searching for a new manga or when migrating to have the result list be ordered by the amount of manga that the user already has from that source.
2: When looking at the sources (in the migration menu for example) there is a number showing the amount of manga the user has from that source. 

The reason for these requests is to make it easier for a user to see if the majority of their library comes from a single source. For I believe that if the user can quickly see if they have a large number of manga from a single source it makes it easier for them to pro-actively migrate and spread it over multiple sources. This should reduce the amount of migrations needed if a source becomes unavailable and puts less strain of the source.
And by having the result list being ordered in a way that puts sources that have less manga from the library associated with them higher up it would promote the use of those sources over sources that are already in use for other manga, thus again helping the user with selecting multiple sources.

ghostbear:
closes #4703


app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/sources/MigrationSourcesController.kt
```diff
@@ -58,6 +64,7 @@ class MigrationSourcesController :
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         when (item.itemId) {
             R.id.action_source_migration_help -> activity?.openInBrowser(HELP_URL)
+            R.id.action_filter -> migrationBottomSheet.show()
```
arkon, line 67: I'm actually not really sure this needs to be a full-on sheet. Just a plain menu like in the download queue is probably fine?

PR #6011, opened 2021-10-02
Manga description adjustments

ivaniskandar:
- Animated state changes
- Adjust scrim position to fully show 2 lines when shrunk
- Set minLines to avoid scrim hiding oneliner

https://user-images.githubusercontent.com/12537387/135721506-529161fa-cae2-4e53-a41c-bc4f3b8069cc.mp4


CVIUS:
Is it possible to animate the chevron like how J2K does (flipping chevron), instead of rotating it 180°?

Soitora:
https://gfycat.com/ShimmeringBleakFrog

vengiare:
Wonder why trimming is done mid-animation
#
I guess a hacky way to deal with "less than 3 line" descriptions would be to add an empty line to them when collapsed

IMG_20211003_060536
IMG_20211003_060628

Though I don't really think there's gonna be an elegant fix with how the chevron is sandwiched between description and tags rn

CVIUS:
> **3**. https://gfycat.com/DisfiguredEnragedBinturong

I think I've already seen this animation in some app/s. And it is way better than 180° rotation IMO.

vengiare:
> I think I've already seen this animation in some app/s. And it is way better than 180° rotation IMO.

Google Clock easiest example

app/src/main/java/eu/kanade/tachiyomi/widget/MangaSummaryView.kt
```diff
@@ -0,0 +1,188 @@
...
+            if (field != value) {
+                field = if (value.isNullOrBlank()) {
+                    context.getString(R.string.unknown)
+                } else {
+                    value
+                }
+                binding.descriptionText.text = field
+                recalculateHeights = true
+                doOnNextLayout {
+                    updateExpandState()
```
ghostbear, line 60: You could check if the `lineCount` is less than or equal to 3 to then expand the description instead of doing `minLines`

Or instead of expanding you could try to hide the toggle and scrim, show the expanded pills, and disabled the click listeners that set the state to expanded

```suggestion
                    if (binding.descriptionText.lineCount <= 3) {
                        expanded = true
                    } else {
                        updateExpandState()
                    }
```

ivaniskandar: > you could try to hide the toggle and scrim, show the expanded pills, and disabled the click listeners that set the state to expanded

I've tried this before but it's possible one manga have huge amount of tags, so I'm not entirely sure if it's wise to expand it by default.

PR #6013, opened 2021-10-02
Distinguish Sources with additional settings in Extensions List

curche:
This PR adds a plus sign at the end to form "Settings+" for extensions that have
additional settings like Komga/Mangadex. The "+" signifies that there are more
settings inside the extension. 

The "+" is added if and only if:
 1. Extension is Installed
 2. is a `ConfigurableSource` (Eg: Komga, MangaDex)
 3. doesn't have a pending Update
 4. is currently in an IDLE state i.e. not downloading/pending/installing

In addition, while by default, the text "Settings+/Settings" is centred, 
I tried to keep it right-aligned which felt nicer to me.

#### Before
| How it appears in Main | How it appears in Sy | 
| ------- | ------- |
| Screenshot_20211002-203648_Tachiyomi_1 | Screenshot_20211002-202453_TachiyomiSY_1 | 

#### This PR
| Main | Extension in non-idle state |
| ------ | ----- |
| Screenshot_20211002-203610_Tachiyomi_1 | Screenshot_20211002-203728_Tachiyomi_1 |

### Food for thought
Considering that the app has translations, it must be asked whether "+" signifies the
same meaning in other languages as it does in English: 
i.e. "Settings+" means "there are additional settings than the normal ones for this"  

This came about when I was looking at the Android Studio Suggestion/Warning:
 > `text = "$text+"` should be replaced with 
  a resource usage (`R.string`)
  or Supressed (with `-at-SuppressLint("SetTextI18n")` like Sy does it)
  
### Notes
- Yoinked from Sy when I was fixing a bug related to this and thought of adding this to 
main as well
- Tested on Pixel 3 (blueline, A12 beta) and Redmi 7 (onclite, A11 LineageOS)
- couldn't think of a better PR title :d


arkon:
I've turned this down in the past since it looks bad. I'm open to ideas on how to better surface that there's more extension, but this isn't it.

PR #6021, opened 2021-10-03
Sort and remove duplicates in genres

FourTOne5:
Co-authored-by: ivaniskandar <12537387+ivaniskandar@users.noreply.github.com>


arkon:
I get the distinct part, but why sort? Wouldn't some sites order them by "weight"?

FourTOne5:
> I get the distinct part, but why sort? Wouldn't some sites order them by "weight"?

Hmm.. I didn't consider it. I'll remove it.

PR #6023, opened 2021-10-03
Enable "ALL" in Browse by default

curche:
some extensions, including self-hosted ones, have the "ALL" label and
sometimes users get confused with not having enabled "ALL" after
installing new extensions

Ideally would need a migration to force enable for existing updates I think
https://github.com/tachiyomiorg/tachiyomi/blob/6d650518a14953a1d5e74d1e56de841492929fb6/app/src/main/java/eu/kanade/tachiyomi/Migrations.kt#L32
```kotlin
if (oldVersion < 69) {
  val prefs = PreferenceManager.getDefaultSharedPreferences(context)
  val defEnabledLanguages = setOf("all", "en", Locale.getDefault().language)
  val enabledLanguages = prefs.getStringSet(PreferenceKeys.enabledLanguages, defEnabledLanguages)
  if ("all" !in enabledLanguages!!) {
    enabledLanguages += "all"
        prefs.edit {
            putStringSet(PreferenceKeys.enabledLanguages, enabledLanguages)
        }
    }
}
```

Tested on a fresh install on Mi A2 (jasmine-sprout, Android 11)


arkon:
Probably fine to not migrate for installations where that setting's already been modified (which would at least indicate that they've gone to that screen to change stuff before, so they may be aware of it by now).

PR #6024, opened 2021-10-03
Fix Local and Other lang sources showing up in seperate Other Categories

curche:
This PR fixes the following:
 - moves all sources with lang "other" into the same category "Other" as which LocalSource uses 
   - previously only LocalSource (lang = "") became the in-app string `R.other_source` 
   - while sources with lang = Other became just an "Other" category which used the ext string rather than the translatable in-app string
   - now both get merged into a single category
 - also arranged the `when` by constants, strings, else

At first I tried to change LocalSource lang to "other" but that ended up being too many changes (esp in `SourcePresenter.kt`). This one is just a single commit which is cleaner 

### Images
| Before | Now |
| ------- | ------- |
| Screenshot_20211003-173052_TachiyomiSY_Other_Other | IMG_20211003_225806 |


app/src/main/java/eu/kanade/tachiyomi/util/system/LocaleHelper.kt
```diff
@@ -16,9 +16,10 @@ object LocaleHelper {
      */
     fun getSourceDisplayName(lang: String?, context: Context): String {
         return when (lang) {
-            "" -> context.getString(R.string.other_source)
             SourcePresenter.LAST_USED_KEY -> context.getString(R.string.last_used_source)
             SourcePresenter.PINNED_KEY -> context.getString(R.string.pinned_sources)
+            "" -> context.getString(R.string.other_source)
+            "other" -> context.getString(R.string.other_source)
```
ghostbear, line 22:
```suggestion
            "", "other" -> context.getString(R.string.other_source)
```

curche: nice. I actually marked it as draft because my build I was testing wasn't working second time I tried. I'll add this once I figure out what went wrong where.

curche:
Apparently, unless I set `lang` for LocalSource as `other`, it tries to be a different category from sources with `lang` as `other`.
Also, this means I had to take care of duplicate entries for LocalSource in SourcePresenter.kt

Ready for review. Please suggest improvements if any

app/src/main/java/eu/kanade/tachiyomi/util/system/LocaleHelper.kt
```diff
@@ -16,9 +16,9 @@ object LocaleHelper {
      */
     fun getSourceDisplayName(lang: String?, context: Context): String {
         return when (lang) {
-            "" -> context.getString(R.string.other_source)
             SourcePresenter.LAST_USED_KEY -> context.getString(R.string.last_used_source)
             SourcePresenter.PINNED_KEY -> context.getString(R.string.pinned_sources)
+            "", "other" -> context.getString(R.string.other_source)
```
ghostbear, line 21: Maybe remove `""` as there are no extensions that use it anymore or?

ghostbear (approved):
Other than the little question about `""` this LGTM 👍🏽

app/src/main/java/eu/kanade/tachiyomi/util/system/LocaleHelper.kt
curche: fair enough

arkon:
Eh, it's probably fine for now? We can revisit it if the behaviour seems weird to people later.

PR #6027, opened 2021-10-04
INTERNAL PLAYER

CJved:
Option to adjust screen ratio and and basically a better internal player UI


alessandrojean:
1. This isn't Aniyomi repository.
2. Don't open PRs instead of issues.

PR #6049, opened 2021-10-05
[5753] - Add pending downloads count on Download queue

### LINKED ISSUE #5753: Add pending downloads count on Download queue
### Acknowledgements

One on the title bar, like "Download queue (#items)"

IMG_20210820_093601

#

One on the More screen, a pending downloads counter badge on the right side. (Replacing secondary "999 remaining" text)

Reasoning for this one is that you can't see how many remaining chapters are there when downloads are paused.

image

(edited with a little more info)

### Other details

_No response_

Platiplus:
Closes #5753 

@vengiare was asking for a way to find out how many downloads were pending even when they were paused, so I add a condition to show pending downloads even when they are paused on the More Options Download Queue section.

### Images
Old
image

New
image

I put the information on the label instead of a badge on the side because I thought it would be better, and also, no idea if we can use a badge on a preference, or if this would be bad UI.

Also, I tried to update the title in the DownloadController with each update on the screen, but it didn't work, any help or tips would be appreciated.


app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
```diff
@@ -170,10 +170,15 @@ class MoreController :
...
+        var numberOfPendingDownloads = resources?.getQuantityString(R.plurals.download_queue_summary, downloadQueueSize, downloadQueueSize)
+
         preference.summary = when {
-            downloadQueueSize == 0 -> null
-            !isDownloading -> resources?.getString(R.string.paused)
-            else -> resources?.getQuantityString(R.plurals.download_queue_summary, downloadQueueSize, downloadQueueSize)
+            !pendingDownloadExists -> null
+            !isDownloading && !pendingDownloadExists -> pauseMessage
+            !isDownloading && pendingDownloadExists -> "$pauseMessage - $numberOfPendingDownloads"
+            else -> numberOfPendingDownloads
```
ghostbear, line 181: You could do a string plural like the `R.plurals.download_queue_summary` that way you would also not need to add another case to the when statement 

```kt
preference.summary = when {
  downloadQueueSize == 0 -> null
  !isDownloading -> resources?.getQuantityString(R.plurals.download_paused_summary, downloadQueueSize, downloadQueueSize)
  else -> resources?.getQuantityString(R.plurals.download_queue_summary, downloadQueueSize, downloadQueueSize)
}
```

And then in `strings.xml` add 
```
<plurals name="download_paused_summary">
  <item quantity="zero">Paused</item>
  <item quantity="other">Paused - %1$s remaining</item>
</plurals>
```

Platiplus: Thanks for the tip. I'm not an experienced Kotlin developer myself, but since I love the app and Hackoberfest is happening I thought of giving it a go! :zap:

app/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -444,6 +444,10 @@
     <string name="local_source_help_guide">Guia sobre a fonte local</string>
     <string name="restore_duration">%02d min e %02d seg</string>
     <string name="downloaded_only_summary">Filtra todos os mangás em sua biblioteca</string>
+    <plurals name="download_paused_summary">
```
alessandrojean, line 447: Better to remove them and let them be translated through Weblate, let only the English ones.

Platiplus: Done :zap:

app/src/main/java/eu/kanade/tachiyomi/ui/more/MoreController.kt
```diff
@@ -170,9 +170,11 @@ class MoreController :
     }
 
     private fun updateDownloadQueueSummary(preference: Preference) {
+        var pendingDownloadExists = downloadQueueSize != 0
+
         preference.summary = when {
-            downloadQueueSize == 0 -> null
-            !isDownloading -> resources?.getString(R.string.paused)
+            !pendingDownloadExists -> null
```
ghostbear, line 176: The variable here is not really needed, could be reverted back to
```suggestion
           downloadQueueSize == 0 -> null
```

app/src/main/res/values/strings.xml
```diff
@@ -497,6 +497,10 @@
     <string name="pref_incognito_mode_summary">Pauses reading history</string>
     <string name="notification_incognito_text">Disable incognito mode</string>
     <string name="downloaded_only_summary">Filters all manga in your library</string>
+    <plurals name="download_paused_summary">
+        <item quantity="zero">Paused</item>
```
arkon, line 501: @ghostbear Sorry, I actually misled you in an earlier PR. This apparently doesn't work for some locales (including English, at least from on my device).

Related commit: https://github.com/tachiyomiorg/tachiyomi/commit/071bad123208274a8d4858038b4073a0ce208aa3

@Platiplus You'll have to go back to another way of doing this. 😅

ghostbear: > @ghostbear Sorry, I actually misled you in an earlier PR. This apparently doesn't work for some locales (including English, at least from on my device).


Though they follow the quantity but it's depended on the language, I see


> @arkon The way I was doing is fine? Or should I look for another solution?


Yeah, the way you were doing was fine then, so you can revert back to d460b1f0b927c836439ea5257c5c10ace847e4ab. Sorry for misleading you 🙇‍♂️

PR #6112, opened 2021-10-15
[5774] - Make long pressing manga in Global search act like in Source view

Platiplus:
@vengiare asked for a way to add or remove items from the library on global search along with other things.

Description:

1. Use global search to look for a Manga
2. Manga when long pressed acts the same way as Source view
* Long press to Add to library (then show Set categories dialog)
* If manga is "In Library", show "Remove from library" dialog

I think I'm up to something here, but probably need someone's help with this. Appreciate feedback.


arkon:
I need to sit down at some point and really think about this, but my initial concern is the different behaviour between global search and the migration flow, which look the same.

Platiplus:
Gonna close this until further discussion.

PR #6177, opened 2021-10-30
Allow manga titles to update from source if they are not in library

FlaminSarge:
The previous rationale for not allowing manga titles to update (at all) was that it would be confusing for users if a manga's title arbitrarily changed when the source changed it. Presumably, users would care less about this arbitrary change for manga that is not in library, so this provides a path for getting a manga's title updated, and prevents incorrect titles from persisting in the DB for manga that get title updates but aren't in library.


FlaminSarge:
There may need to be a discussion around updating titles for manga that are in library, but that can probably be deferred. In most sources, a title update means a URL update anyways, so this change just fixes up some title handling for the few sources that use IDs in URLs instead of titles.

arkon:
This behaviour already exists in some (all?) forks, so it's probably fine.

Issue #5916, opened 2021-09-16
Force tablet UI when unfolded for folding phone support


ArcdoesMC:
### Describe your suggested feature

Add an option to force tablet ui for landscape and/or portrait mode when unfolded on folding phones like zfold 2 or 3. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.12.2](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

PR #6208, opened 2021-11-07
Added tabletUI option: "Automatic"

OncePunchedMan:
added automatic tablet ui option; useful for foldables like the Z Fold 3

tested in emulator and on my phone

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -326,10 +326,7 @@ class PreferencesHelper(val context: Context) {
 
     fun incognitoMode() = flowPrefs.getBoolean(Keys.incognitoMode, false)
 
-    fun tabletUiMode() = flowPrefs.getEnum(
-        Keys.tabletUiMode,
-        Values.TabletUiMode.AUTOMATIC
-    )
+    fun tabletUiMode() = flowPrefs.getEnum(Keys.tabletUiMode, Values.TabletUiMode.AUTOMATIC)
 
     fun extensionInstaller() = flowPrefs.getEnum(
         Keys.extensionInstaller,
--- a/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
@@ -328,7 +328,7 @@ class PreferencesHelper(val context: Context) {
 
     fun tabletUiMode() = flowPrefs.getEnum(
         Keys.tabletUiMode,
-        if (context.applicationContext.isTablet()) Values.TabletUiMode.AUTOMATIC else Values.TabletUiMode.NEVER
+        Values.TabletUiMode.AUTOMATIC
     )
 
     fun extensionInstaller() = flowPrefs.getEnum(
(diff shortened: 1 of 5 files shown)


arkon:
I expected the "Never" forcing option to behave like this. Does it not?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedController.kt
```diff
@@ -247,7 +247,7 @@ class SettingsAdvancedController : SettingsController() {
                 key = Keys.tabletUiMode
                 titleRes = R.string.pref_tablet_ui_mode
                 summary = "%s"
-                entriesRes = arrayOf(R.string.lock_always, R.string.landscape, R.string.lock_never)
+                entriesRes = arrayOf(R.string.automatic_background, R.string.lock_always, R.string.landscape, R.string.lock_never)
                 entryValues = PreferenceValues.TabletUiMode.values().map { it.name }.toTypedArray()
                 defaultValue = if (context.isTablet()) {
```
ivaniskandar, line 252: This and [this](https://github.com/OncePunchedMan/TachiyomiSY/blob/added_automatic_tabletui_option/app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt#L331) should be changed to `PreferenceValues.TabletUiMode.AUTOMATIC`

OncePunchedMan: yep, done

OncePunchedMan:
Will close: #5916

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
```diff
@@ -328,7 +328,7 @@ class PreferencesHelper(val context: Context) {
 
     fun tabletUiMode() = flowPrefs.getEnum(
         Keys.tabletUiMode,
-        if (context.applicationContext.isTablet()) Values.TabletUiMode.ALWAYS else Values.TabletUiMode.NEVER
+        if (context.applicationContext.isTablet()) Values.TabletUiMode.AUTOMATIC else Values.TabletUiMode.NEVER
```
ivaniskandar, line 331:
```suggestion
        Values.TabletUiMode.AUTOMATIC
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedController.kt
```diff
@@ -247,10 +247,10 @@ class SettingsAdvancedController : SettingsController() {
...
```
ivaniskandar, line 252:
```suggestion
                defaultValue = PreferenceValues.TabletUiMode.AUTOMATIC.name
```

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
OncePunchedMan: dunno if we should make it the default for normal phones too

OncePunchedMan: ye can't do that since it will apply tabletUI in landscape. at least in my emulator

ivaniskandar: But this check is redundant when AUTOMATIC will also disable tablet ui on small screens [here](https://github.com/tachiyomiorg/tachiyomi/pull/6208/files#diff-d1f7e6cd0e06fcf594323c5d5c7c38b8bd803478fa421088fc99cad3d00f26d3R319)

OncePunchedMan: ye, that's true

PR #6219, opened 2021-11-11
When retrieving manga details from source, ensure title is pulled from source rather than local DB

### LINKED ISSUE #3608: [Bug] Title does not update from source along with other metadata.
### Device information
* Tachiyomi version: v0.10.3
* Android version: 8.0
* Device: Galaxy Tab S

## Steps to reproduce
1. Have a manga in your library.
2. Given source updates the title of the manga.
3. Refresh to update cover/metadata.

### Expected behavior
Title should update if changed to reflect the title used in the source.

### Actual behavior
Title remains the same until the manga is removed from library and cache is fully cleared.

### Other details
This is a fairly rare problem, but here's the real world case: The main title for "Ganbare, Douki-chan" was originally "Senpai is Mine" and MangaDex has updated their main display title to reflect this change but Tachiyomi won't update the title without fully removing the manga from the database and then doing a DB clean to remove not-in-library titles (wiping out any non-library progress in the process). It should probably update in line with the source with less friction to the end-user.

Not related to #2190

FlaminSarge:
Alongside a previous change, this ensures that manga titles for non-Library manga are always up to date, where previously an old/outdated title could be persisted in the DB and never updated until DB clear (#6177)

Closes #3608

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/AppModule.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/AppModule.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi
 
 import android.app.Application
+import android.os.Build
 import androidx.core.content.ContextCompat
 import androidx.sqlite.db.SupportSQLiteOpenHelper
 import androidx.sqlite.db.framework.FrameworkSQLiteOpenHelperFactory
@@ -24,6 +25,7 @@ import eu.kanade.tachiyomi.data.track.job.DelayedTrackingStore
 import eu.kanade.tachiyomi.extension.ExtensionManager
 import eu.kanade.tachiyomi.network.NetworkHelper
 import eu.kanade.tachiyomi.source.SourceManager
+import io.requery.android.database.sqlite.RequerySQLiteOpenHelperFactory
 import kotlinx.serialization.json.Json
 import uy.kohesive.injekt.api.InjektModule
 import uy.kohesive.injekt.api.InjektRegistrar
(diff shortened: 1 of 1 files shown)


FlaminSarge:
I checked that an outdated title was updated properly when swipe-refreshing a manga details page when the manga was not in library, and that the new behavior did not occur if the manga was in library (to prevent the app from arbitrarily changing a user's library titles). In addition, every place that uses getMangaDetails subsequently uses manga.copyFrom(networkManga.toSManga()) to pull data from the resulting MangaInfo object, so this updated title behavior will only be picked up for cases that specifically update the manga title (alongside copyFrom) in the caller as well.

I have open questions about why these are set up the way they are, but that can be a discussion for later.

arkon:
As others mentioned on Discord, you'd probably want to rename the downloads directory if present.

quangkieu:
Hey I wonder if this is relative to kissmanga keep missing cover whenever initialized or favorite?

app/src/main/java/eu/kanade/tachiyomi/source/Source.kt
```diff
@@ -67,6 +67,7 @@ interface Source : tachiyomi.source.Source {
         val sManga = manga.toSManga()
         val networkManga = fetchMangaDetails(sManga).awaitSingle()
         sManga.copyFrom(networkManga)
+        sManga.title = networkManga.title
```
FourTOne5, line 70: What not do it inside the `.copyFrom()` function?

FlaminSarge: I was under the impression that copyFrom was also used in places that wouldn't necessarily want to copy the title. My initial impl had a param on copyFrom for copying title that defaulted to false, and that version is probably also valid. If you feel strongly about it I can update to that instead.

FlaminSarge:
Still pending changes for renaming downloaded folders, moving to Draft for now

app/src/main/java/eu/kanade/tachiyomi/source/Source.kt
FourTOne5: Check if the network manga title is blank?

PR #6225, opened 2021-11-12
[Draft] Add filter, sort to result from Browse source [Help Request]

quangkieu:
Add filter, sort to result from Browse source

  If your changes are visual, please provide images below:

### Images
image
image
image


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
```diff
@@ -373,31 +631,180 @@ open class BrowseSourceController(bundle: Bundle) :
...
+        presenter.insertCategory(flagCategory)
+        flagCategory = presenter.getCategories().first { it == flagCategory }
+        val mangas = adapter.currentItems.map { it as SourceItem; it.manga }
+        presenter.updateMangasToCategories(mangas, listOf(flagCategory), emptyList())
+        var running = isRunning()
+        if (running) {
+            context.toast(R.string.notification_fetch_already_running, Toast.LENGTH_LONG)
+            return
+        }
+        LibraryUpdateService.start(context, flagCategory)
```
quangkieu, line 666: So I wonder if anyone familiar with Android service that could help me on this feature.
As most Item we pull from source would only be initialized but does not have full detail like chapters details to compare
So this feature would pull chapters detail and its related metadata and we could do the sort/filter on those data
But I keep getting error 
```Context.startForegroundService() did not then call Service.startForeground(): ServiceRecord{d65e155 u0 eu.kanade.tachiyomi.qk.test.debug/eu.kanade.tachiyomi.data.library.LibraryUpdateService}```

arkon:
I'm not sure client-side filtering/sorting is something that we really want to do, considering how heavy it is to get the required amount of data to do it.

quangkieu:
I think that most people would only look at the first 5 pages of latest or popular. And this extra fetch chapters is manual, not auto. And if you talk about date size, it should be the same as initializing as we just request same page and parse new detail

quangkieu:
We could put warning about the size too when user fetch chapters. I could add another button to clean up non-favorite in current view

quangkieu:
There is error in tab name. It should be Request, Result, Display instead of Filter, Sort, Display

CarlosEsco:
I feel this is a bit excessive and agree with Arkon

quangkieu:
Oh right now, it just using local ready metadata. And it would disable to fetch next page as I could not make it work either way

arkon:
Can you better summarize the proposed changes for this PR? Is this just client-side filtering (of title, author, and artist (?)) and sorting?

I _might_ be fine with it depending on how simple the UX is, but I still suspect it'd be highly confusing since it'll depend on the available pages of data, nor am I sure how it'll work when also doing a source-based search?

quangkieu:
yeah I am working on rebase this and get some demo clip

arkon:
I'm going to close this since the underlying screens are going to be rewritten with Compose, but I also wanted to move them to be paginated instead of infinitely scrolling, which would lend itself to something like this in a much simpler way (I think).

This may be worth revisiting after the Compose/pagination changes are done.

PR #6241, opened 2021-11-16
Add feature to clear database manga by source

datreesezcup:
### Overview
This PR changes the "Clear database" button in advanced settings to now open a new screen that lists all of the sources in the database that have non-library manga saved (along with how many entries there are), allowing the user select which sources they want to remove data for.

With one or more sources selected, a FAB appears that when clicked, brings up the "Are you sure" dialog originally on the settings screen, and clicking OK will delete the database info for those sources.

The top toolbar of the screen also includes a "select all" and "select inverse" button for quick selections.

If this list is empty (no non-library manga in database), an EmptyView with the text "Database clean" is displayed.
| Preview |
| -------- |
| clear_database_by_source |

----
### V2 of this PR
Awhile ago [I made a PR that did this same feature](https://github.com/tachiyomiorg/tachiyomi/pull/5595), however while merging upstream changes into it, something went wrong and my PR became a mess of ~150 other commits, and I closed it. Finally I have got around to remaking this PR.

The only noteworthy change included in this PR is that the counting of non-library manga per source is handled as part of the database query, instead of in the presenter. Most of the other changes from the other PR are just different names for files, and whatever changes needed to be made to make the PR work for the most recent commit at the time of writing this (25784d1)


vengiare:
I feel like it should be using either,
> **Source (##)**
> **Full lang**

Which is used in migrate
Screenshot_20211117-032928

Or,
> **Source (short lang)**
> \#\# of manga not in library

Which is used in manga info
Screenshot_20211117-033103

app/src/main/java/eu/kanade/tachiyomi/ui/setting/database/ClearDatabaseController.kt
```diff
@@ -0,0 +1,188 @@
...
+
+        menu?.apply {
+            get(0).isVisible = size > 0
+            get(1).isVisible = size > 0
+        }
+    }
+
+    override fun onItemClick(view: View?, position: Int): Boolean {
+        toggleSelection(position)
+        adapter?.notifyItemChanged(position, Payload.SELECTION)
```
ghostbear, line 125: As we only use `toggleSelection` once after the changes in `onOptionsItemSelected` we can move the toggle code here
```suggestion
        val adapter = adapter ?: return false
        adapter.toggleSelection(position)
        adapter.notifyItemChanged(position, Payload.SELECTION)
        updateFab()
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/database/ClearDatabaseController.kt
```diff
@@ -0,0 +1,188 @@
...
+        actionFab = null
+    }
+
+    private fun addSelection(position: Int) {
+        val adapter = adapter ?: return
+        adapter.addSelection(position)
+        updateFab()
+    }
+
+    private fun toggleSelection(position: Int) {
```
ghostbear, line 163: Not needed anymore

app/src/main/java/eu/kanade/tachiyomi/ui/setting/database/ClearDatabaseController.kt
```diff
@@ -0,0 +1,188 @@
...
+        super.onDestroyView(view)
+    }
+
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
+        inflater.inflate(R.menu.generic_selection, menu)
+        this.menu = menu
+        menu.apply {
+            get(0).isVisible = adapter?.itemCount ?: 0 > 0
+            get(1).isVisible = adapter?.itemCount ?: 0 > 0
+        }
```
ghostbear, line 89: Can just loop over them to change the visibility accordingly
```suggestion
        menu.forEach { menuItem -> menuItem.isVisible = (adapter?.itemCount ?: 0) > 0 }
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/database/ClearDatabaseController.kt
```diff
@@ -0,0 +1,188 @@
...
+        if (size > 0) {
+            binding.emptyView.hide()
+        } else {
+            binding.emptyView.show(activity!!.getString(R.string.database_clean))
+        }
+
+        menu?.apply {
+            get(0).isVisible = size > 0
+            get(1).isVisible = size > 0
+        }
```
ghostbear, line 120: Can just loop over them to change the visibility accordingly
```suggestion
        menu?.forEach { menuItem -> menuItem.isVisible = size > 0 }
```

app/src/main/java/eu/kanade/tachiyomi/ui/setting/database/ClearDatabaseController.kt
```diff
@@ -0,0 +1,188 @@
...
+    private var recycler: RecyclerView? = null
+    private var adapter: FlexibleAdapter<ClearDatabaseSourceItem>? = null
+
+    private var menu: Menu? = null
+
+    var actionFab: ExtendedFloatingActionButton? = null
+    var actionFabScrollListener: RecyclerView.OnScrollListener? = null
+
+    private val selectedSources: List<ClearDatabaseSourceItem>
+        get() = adapter?.selectedPositions?.mapNotNull(adapter!!::getItem) ?: emptyList()
```
ghostbear, line 43: Is this really needed it is only used once in `clearDatabaseForSources`?

app/src/main/java/eu/kanade/tachiyomi/data/database/models/SourceCount.kt
```diff
@@ -0,0 +1,3 @@
+package eu.kanade.tachiyomi.data.database.models
+
+data class SourceCount(val source: Long, val count: Int)
```
ghostbear, line 3: Could be more descriptive
```suggestion
data class SourceIdWithCount(val sourceId: Long, val count: Int)
```

app/src/main/java/eu/kanade/tachiyomi/data/database/queries/MangaQueries.kt
```diff
@@ -123,12 +137,12 @@ interface MangaQueries : DbProvider {
 
     fun deleteMangas(mangas: List<Manga>) = db.delete().objects(mangas).prepare()
 
-    fun deleteMangasNotInLibrary() = db.delete()
+    fun deleteMangasNotInLibraryBySources(sources: List<Long>) = db.delete()
```
ghostbear, line 140:
```suggestion
    fun deleteMangasNotInLibraryBySourceIds(sourceIds: List<Long>) = db.delete()
```

datreesezcup:
OK, code has been cleaned up a bit, based on the code reviews by @ghostbear. Thanks for pointing out those issues, I should've caught those when I was refactoring this PR the first time.

### @vengiare's suggestion
I did change the language section of each item to the short form, since both:
* The short form is how it is shown when placed next to the source name (as pointed out in the manga header)
* Reduces the likelihood of the text being cut with ellipsis due to long source name + language name

| Short Lang Preview |
| --------------------- |
| updated_clear_database_item |

I didn't like the migration-screen style items, as I worried that it might be slightly more confusing. But that might just be me.

app/src/main/java/eu/kanade/tachiyomi/data/database/resolvers/SourceIdMangaCountGetResolver.kt
```diff
@@ -0,0 +1,23 @@
...
+import android.annotation.SuppressLint
+import android.database.Cursor
+import com.pushtorefresh.storio.sqlite.operations.get.DefaultGetResolver
+import eu.kanade.tachiyomi.data.database.models.SourceIdMangaCount
+import eu.kanade.tachiyomi.data.database.tables.MangaTable
+
+class SourceIdMangaCountGetResolver : DefaultGetResolver<SourceIdMangaCount>() {
+
+    companion object {
+        val INSTANCE = SourceIdMangaCountGetResolver()
```
arkon, line 12: Why is a singleton instance needed?

datreesezcup: While I was looking at a few of the other GetResolvers used in Tachiyomi, they had a singleton instance that was used as the parameter for `withGetResolver` in database queries:
https://github.com/tachiyomiorg/tachiyomi/blob/bdef2cfdfb50800c35c028a7d41343a99bf95f60/app/src/main/java/eu/kanade/tachiyomi/data/database/resolvers/MangaChapterGetResolver.kt#L12

https://github.com/tachiyomiorg/tachiyomi/blob/bdef2cfdfb50800c35c028a7d41343a99bf95f60/app/src/main/java/eu/kanade/tachiyomi/data/database/resolvers/MangaChapterHistoryGetResolver.kt#L12

https://github.com/tachiyomiorg/tachiyomi/blob/98822a39d9524b08f5954f6bb58ac1d9c704b76d/app/src/main/java/eu/kanade/tachiyomi/data/database/resolvers/LibraryMangaGetResolver.kt#L12

arkon: Ah fair enough. I haven't worked with them much so I wasn't even aware.

arkon:
Sort of wonder if this should've had a button along the bottom similar to tracking search instead of a FAB?

Riztard:
> Sort of wonder if this should've had a button along the bottom similar to tracking search instead of a FAB?

maybe

PR #6286, opened 2021-11-27
MinMaxNumberPicker: Set IME input to use number only

### LINKED ISSUE #6265: Cosmetic: Use Numpad

In places where numbers are used more often than letters (for example, in the tracking menu, when entering scores), the numeric keypad/numpad should be used rather than the full keyboard. This is a built-in part of Android as an (input method?) 

### Other details

_No response_

### Acknowledgements

ivaniskandar:
Resolves #6265


app/src/main/java/eu/kanade/tachiyomi/widget/MinMaxNumberPicker.kt
```diff
@@ -1,13 +1,27 @@
...
 
 class MinMaxNumberPicker @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null) :
     NumberPicker(context, attrs) {
 
+    override fun setDisplayedValues(displayedValues: Array<out String>?) {
+        super.setDisplayedValues(displayedValues)
+
+        // Disable keyboard input when a value that can't be auto-filled with number exists
+        val notNumberValue = displayedValues?.find { it.getOrNull(0)?.digitToIntOrNull() == null }
+        if (notNumberValue != null) {
```
arkon, line 20:
```suggestion
        val notNumberValue = displayedValues?.any { it.getOrNull(0)?.digitToIntOrNull() == null }
        if (notNumberValue) {
```

ivaniskandar: Uhh say we're checking the values of library grid (1-10 and a "default"). The notNumberValue will be the string "default".

If you want to directly get a bool value I can move the null check.

arkon: Ah nevermind, I misunderstood what it was doing.

PR #6289, opened 2021-11-27
Automatically set tracker as completed after reading the last chapter

### LINKED ISSUE #3058: Auto-update tracker status, start/end dates
Quality of Life change. We already have auto-update on the chapters read, since in 0.9.0 the MAL start/finish date feature is added. It is expected for this feature to be automatic too.

### What/Requirements
Start date.
- User goes into tracking and adds MAL title.
- User starts ready a chapter and it has been marked as read
- If MAL start date exists then ignore.
- If MAL start date is empty, update with current time.

### LINKED ISSUE #5779: Change tracking status to complete and add completed date automatically on finishing the last chapter for AniList and MAL

When you finish the last chapter that a tracker has recorded in it's database, it could automatically change the reading status and add the finished date. i.e. Anilist has 39 chapters recorded for a finished manga, you finish chapter 39 and it'll automatically change the status to "Completed" and change the finished date to whatever day it is.

### Other details

I noticed Kitsu already has this feature

### Acknowledgements

jmir1:
closes #3058

This PR sets the tracker as completed and sets the `finished_reading_date` if the chapter was completed in the reader and the chapter number of the read chapter is equal to the total chapters (according to the tracker).

It also sets the `started_reading_date` when starting a manga

In action: https://files.catbox.moe/q5gevs.mp4


azealyx:
> - the manga status is "completed"

Is this really needed? the 3rd condition does the same job I believe (and is already used in tracker sheet)

#

Oh and yeah, this behavior [already exists for the tracker sheet](https://github.com/tachiyomiorg/tachiyomi/blob/10ddeeb7999214f1ed2fa579dadab7e9e379c1a6/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt#L902), maybe just copy parts of that or something

#

Also maybe a toggle in Settings would be nice. Something like "Mark as complete after reading last chapter". Personally I kinda want to mark chapters as read **manually** cause there usually are differences like these.

Common
Screenshot_20211128-124553

Extreme case
Screenshot_20211128-123650

vengiare:
> - the manga status is "completed"

Is this really needed? the 3rd condition does the same job I believe (and is already used in tracker sheet)

#

Oh and yeah, this behavior [already exists for the tracker sheet](https://github.com/tachiyomiorg/tachiyomi/blob/10ddeeb7999214f1ed2fa579dadab7e9e379c1a6/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt#L902), maybe just copy parts of that or something

#

Also maybe a toggle in Settings -> Tracker would be nice. Something like "Mark as complete after reading last chapter". Personally I kinda want to mark chapters as read **manually** cause there usually are differences like these.

Common
Screenshot_20211128-124553

Extreme case
Screenshot_20211128-123650

jmir1:
> Is this really needed? the 3rd condition does the same job I believe (and is already used in tracker sheet)

Good point, it would be better to check for the status on the track service, instead of relying on the source.

> Oh and yeah, this behavior [already exists for the tracker sheet](https://github.com/tachiyomiorg/tachiyomi/blob/10ddeeb7999214f1ed2fa579dadab7e9e379c1a6/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt#L902), maybe just copy parts of that or something

Huh, I didn't even notice that existed. I will adjust my implementation to be more consistent with that.



As for the cases in which the track service has more chapters than the source, that's not a problem, because then this feature will do nothing because it can never reach a high enough chapter number, if you know what I mean.

arkon:
For reference, @curche dug up what J2K/Neko does: https://github.com/Jays2Kings/tachiyomiJ2K/blob/28321c7830f65e63fa4e374d749153447af5aafe/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt#L140

```kotlin
if (track.status == READING && track.total_chapters != 0 && track.last_chapter_read == track.total_chapters) {
     track.status = COMPLETED
}
```

I agree that it's probably better to not rely on the internal "completed" status, since it can be pretty inconsistent across sources.

PR #6292, opened 2021-11-28
Remove unnecessary bottom bar styling

CVIUS:
Making it M3 default


arkon:
This is the same as J2K, meant to keep it down to a somewhat reasonable height to prioritize the content.

I'd be more tempted to use the default if other apps did, but it's highly inconsistent everywhere already anyway.

arkon:
Closing this for now, but we can revive it if we decide to use the default someday.

PR #6304, opened 2021-12-02
Calculate maxConcurrent for search dynamically

makemefeelgr8:
Search is taking ages in case there are over 100 extensions installed.
Flat search example: 200 extensions, 10 seconds each results in over half an hour of waiting.
Obviously, the app executes 5 requests concurrently, and this is supposed to reduce search time to about 6 minutes. But, there is a catch: request timeouts and slow sources. As soon as there are 4 slow running sources that block execution, the actual search will be performed in only 1 thread. And in case there are 5 slow sources, the search will take ages to complete.

So, to counter this, I figured one can send a bigger amount of requests at a time. I've made the app determine amount of requests based on amount of sources.

I'm not an android developer, and I've never worked with kotlin before. This change made the app perform way better on my device, so I've decided to share it with the community.


paulo27ms:
Honestly, why not all of them if it doesn't crash your phone.
It's not like you're actually going to click anything while searching unless everything is extremely slow, the constant reordering of the sources makes that almost impossible.

makemefeelgr8:
I'm running a beast of a device with lots of horsepower, so it can easily handle loading all the records at once. But I wouldn't recommend publishing it like that. 
People are using all sorts of devices, some of them are old tablets. I assume, sending 200 requests at once will make them unresponsive at the very least.
Also, we have to consider the network load. Users love to read on the go, possibly with nothing but a weak 3G signal around. So, running 200 concurrent GETs will cause most of them to timeout without even starting.
I was thinking to make it an optional feature, like add a "Search X sources at once" slider in settings, but I don't have enough experience to do that. And most of the users will just rely on default settings anyway.
Hence, my suggestion is to merge it the way it is right now, to ensure it doesn't cause new issues. And afterwards one can rewrite it properly.

CarlosEsco:
Where's the tachi ddos image when you need it

curche:
> Where's the tachi ddos image when you need it

image

arkon:
I'm not sure I'm comfortable with this change. There's definitely a _lot_ of users with weak devices that already struggle with concurrently searching 5 sources, so anything like this would need to be a bit smarter (e.g. forcing timeouts or something).

PR #6318, opened 2021-12-05
Add sui support

D34DPlayer:
Sui is a Magisk module that provides the Shizuku API for rooted devices and will eventually replace Shizuku for those.
As of right now, you can only choose the Shizuku installer if the Shizuku application is installed, so I tweaked the check to also look for Sui.


arkon:
I'm not super familiar with Shizuku or Sui. Do we need to update the wording in the app to reflect both at some point?

D34DPlayer:
Right now, if neither Shizuku or Sui are installed, a popup appears explaining how to install Shizuku.

I think this behavior should be kept as Shizuku is easier to install, and once installed it'll inform root users about Sui anyways.

The only thing that I guess could be changed is renaming Shizuku to Shizuku API in the selection field.

ivaniskandar:
Is this tested? Referring to the API guide it's missing the init call.

arkon:
It's actually automatically initialized by Shizuku: https://github.com/RikkaApps/Shizuku-API/blob/58e25c450800bf56791eaaaa88fdb1afbfb30b97/provider/src/main/java/rikka/shizuku/ShizukuProvider.java#L167

ivaniskandar:
Ah you're right. It's all good then

PR #6344, opened 2021-12-13
Tweak Midnight Dusk colors

OncePunchedMan:
There were still some colors i didn't change and i went back to the original pink. I also removed the elevation color to prevent weird looking reader settings

dark | pure black
---|---
Screenshot_20211213_110345 | Screenshot_20211213_110404


CrepeTF:
thank bb

PR #6359, opened 2021-12-18
Add support to searching with source name, id and baseUrl for a extension's sources.

FourTOne5:
So I was thinking but should I make it so that it'll only check via id if we do a search like `id:2`


arkon:
It's probably fine. I imagine most people would end up searching the IDs, so needing to tell them that there's a prefix is just going to be a burden for support. Most IDs are also long now, so there's little chances for "false positives".

Issue #918, opened 2017-07-31
Compress (ZIP / RAR / 7z) mangas to avoid Android performance problems.


444nonymous:
I know this has been requested in the past and has been considered to be implemented in the future.

As a feature, having mangas zipped by chapters (or one zip per manga with subfolders per chapters) is not a priority. It's a great feature for archiving purposes and collecting / sorting mangas, but not intermediately necessary.

On the other hand I have real problems with my phone. It took me a while to understand that they were caused by Tachiyomi. It's a 2016 Samsung A5, 8 cores, 2GB RAM. I was experiencing very slow boots and some apps did not load or gave error at startup. So I had to reboot.

It's caused by the memory card. Sandisk original (Amazon), brand new and fast, it can record video at the highest resolution. I only keep like 20-30 mangas at most, but they have +100.000 individual jpg files. Even if all folders are with .nomedia to avoid any problems, it takes a lot of time to mount the memory card. And I don't even have any app running from the card. The phone just hangs for a lot of time when mounting it.

So please, reconsider the zip feature as a priority to avoid performance issues. My phone boots in 30 seconds with a full memory card with a few files (large videos). But it takes minutes and reboots with a mostly empty card with tens of thousands of files. Since Tachiyomi is going to download so many files, I think it's necessary to store them. Any fast no-compression storage file format would do, after all jpeg is already compressed, so no cpu processing or delay in storing / accesing manga images.

Thanks!

inorichi:
It shouldn't be too difficult to add an option to compress everything. However, I think only zip will be supported. Rar compression is propietary and requires written permission from its creator, and 7z is not very popular.

444nonymous:
I think zip with minimal o no compression (store option) would do. Fastest to store and read and less CPU load.  It would be a waste trying to re-compress image formats like jpg. With ultra compression options in rar5 and 7z you can get a 20% size reduction. Normal zip is even worse.

A good option would be to download and store by chapters. Since that's how Tachiyomi works.
NOW: /Mangas/Dragon Ball/Chapter 1/image1.jpg, image2.jpg, etc.
ZIP: /Mangas/Dragon Ball/Chapter1.cbz, Chapter2.cbz, etc.

[...]

PR #6360, opened 2021-12-18
add compress to cbz on download

Seishirou101:
Adds ability to compress to cbz after downloading, taken from sy.

Things tested:

- downloads and converts folder to cbz
- able to read cbz file online and offline
- delete cbz when button is used and delete after read
- correctly show that download is deleted afterwards
- download cache recognizes it removed/added after restart of app


AriaMoradi:
duplicate of #918

PR #6364, opened 2021-12-19
Change wording of "update_never"

FourTOne5:


Riztard:
merge with the other wording pr?

FourTOne5:
> merge with the other wording pr?

The two pr is for two seperate things.....

arkon:
We use "Off" in other places too, so it should at least be consistent across all of those.

PR #6376, opened 2021-12-21
Sort manga by last read for "Unread" sort mode when unread count is equal

### LINKED ISSUE #6355: Sort read manga by last read for "Unread" sort mode

right now if i select the "Unread" to sort my manga it will start showing the unread manga on top. but after the unread it starts being "Alphabetically" which is very annoying. i would like to request a new "Sort" or a "modification" to the currents. 

i would like it to show "Unread" manga then "Last read" manga. right now it is "Unread" then "Alphabetically". if i want to reread the chapter i have to go and search for it among my many manga. i know there is a search feature but sometimes the manga title is so confusing i wouldn't know how to write it without copy pasting. so i want it to be "Unread" manga then "Last read" manga.

### Other details

_No response_

### Acknowledgements

FourTOne5:
Closes #6355


THElegend5:
I actually prefer the current system of sort , as I kind of know where something would be , would prefer if this is added as a alternative instead.

CarlosEsco:
I don't agree with this change.  As alphabetical is a common secondary sort.  Last read doesn't make sense as a secondary sort and the original other should have using a different sort method in the first place

arkon:
I feel like it _sort of_ makes sense for unread sort specifically, but I'm a bit impartial since I don't personally use it.

PR #6381, opened 2021-12-22
Hide scrim and toggle when less than 3 lines

ghostbear:
| Before (less than 3 lines) | After (less than 3 lines) | Before | After |
| ------- | ------- | ------- | ------- |
| Screenshot_1640207795 | Screenshot_1640207571 | Screenshot_1640207792 | Screenshot_1640207578 |
| Screenshot_1640207845 | Screenshot_1640207603 | Screenshot_1640207854 | Screenshot_1640207586 |


Riztard:
btw there is some manga with tons of tag, what happen if you cant hide it :v

ghostbear:
> btw there is some source with tons of tag, what happen if you cant hide it :v

I don't know, compile it and test yourself and tell me the result. Because I have never seen a manga with tons of tags and a tiny description

Sygmus-1897:
> > btw there is some source with tons of tag, what happen if you cant hide it :v
> 
> I don't know, compile it and test yourself and tell me the result. Because I have never seen a manga with tons of tags and a tiny description 

They are very common in nhentai

ghostbear:
> > I don't know, compile it and test yourself and tell me the result. Because I have never seen a manga with tons of tags and a tiny description
> 
> They are very common in nhentai

After clicking roughly 40 random entries from both popular and latest. It seems that there is a pattern in NHentai's description that makes it more than 3 lines each time which will make it expandable.

FourTOne5:
@ghostbear can you check with this one https://mangadex.org/title/dc481ed5-4925-4f3c-bac1-e73f1e7ed338 1 line description with tons of genre

ghostbear:
Wouldn't say it has too many tags but it has a lot of tags. 

| Screenshot_1640249073 | Screenshot_1640207792 |
|---|---|

FourTOne5:
umm does clicking on the description hide the tags (for pic 1)

ghostbear:
No, the description is not expandable so clicking does nothing

FourTOne5:
So we won't be able to collapse the tags inthis case?

ghostbear:
No. Why would you need to collapse it?

FourTOne5:
So I don't have to see all the tags? I basically don't care about the tags unless I'm browsing. Also isn't tags and description seperate things? Just cause there is less description tags shouldn't just become uncollapsable.

Sygmus-1897:
Agree. I feel like description and tags should be treated separately

ve1mi:
the most popular Russian sources have tons of tags, you really need to hide them too
Screenshot_20211223-185835_Tachiyomi
Screenshot_20211223-185824_Tachiyomi

vengiare:
I guess maybe hide scrim and toggle when description is less than 3 lines **and tags take up less than 3 (or less than 4) lines**.

Soitora:
> the most popular Russian sources have tons of tags, you really need to hide them too
>

In these examples nothing would change, as is current; it would not show since descriptions is long

ve1mi:
> > the most popular Russian sources have tons of tags, you really need to hide them too
> 
> In these examples nothing would change, as is current; it would not show since descriptions is long

I mean the tags themselves are a small problem: they can't be disabled or reduced if there are a lot of them. I want to read the description, but the tags aren't interesting to me at all

arkon:
I'm uncomfortable with the tags potentially not being collapsible.

PR #6382, opened 2021-12-24
Add "Started" library filter and library update restriction

FourTOne5:


app/src/main/java/eu/kanade/tachiyomi/data/database/tables/MangaTable.kt
```diff
@@ -41,6 +41,8 @@ object MangaTable {
 
     const val COL_UNREAD = "unread"
 
+    const val COL_READ = "read"
```
arkon, line 44: This seems misleading because this isn't actually a column in the table.

FourTOne5: I'm not quite sure what you meant but do I have change it to something like `COL_READ_COUNT`

PR #6413, opened 2021-12-30
Fixes descriptive notification message for errors

### LINKED ISSUE #6401: More descriptive download errors

Whenever I get a download error all  it says is "Chapter 1 no page found" or "Chapter 1 http  error 404",  but it never actually tells me for which manga I'm downloading.
image
image
image

### Other details

The reason this is a big issue is because I travel a lot and I don't have access to the internet a lot so I downloaded bulk . It's kind of hard to see the issue if I don't know which manga is causing the issue. Thanks for taking it consideration my thoughts.

### Acknowledgements

MohitMandalia:
Download Notification now shows the downloaded chapter and which manga it belongs to Fixes #6401
Screenshot_20211229-215529_2


app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadNotifier.kt
```diff
@@ -209,12 +209,11 @@ internal class DownloadNotifier(private val context: Context) {
...
      * @param chapter string containing chapter title.
      */
-    fun onError(error: String? = null, chapter: String? = null) {
+    fun onError(error: String? = null, chapter: String? = null, mangaTitle: String? = null) {
         // Create notification
         with(errorNotificationBuilder) {
             setContentTitle(
-                chapter
-                    ?: context.getString(R.string.download_notifier_downloader_title)
+                mangaTitle?.plus(" : $chapter") ?: context.getString(R.string.download_notifier_downloader_title)
```
FourTOne5, line 216: only put " : " if `chapter` is not null.

MohitMandalia: Okay

MohitMandalia: So if chapter is null should manga title be displayed? what should be the result output if chapter is null.

currently if chapter is not null then output will be `manga : chapter`

FourTOne5: i my opinion if manga is null then `R string.download_notifier_downloader_title` and if chapter is null `manga` else `manga - chapter` will look good. But I'll say you should wait for arkon's decision.

MohitMandalia: okay I've update the code but I'll wait for @arkon  for the decision code looks something like (chapter not null then `manga - chapter` else `manga`) else `downloader`

MohitMandalia: okay I've update the code but I'll wait for @arkon  for the decision code looks something like (chapter not null then `manga - chapter` else `manga`) else `downloader`

arkon: `onError` is only ever called if there's only a message or if there's a (nullable) message + manga chapter, in which case both `chapter` and `mangaTitle` shouldn't actually be `null`.

If there's `mangaTitle` or `chapter` present, do "manga - chapter", otherwise just fallback to `R string.download_notifier_downloader_title`.

MohitMandalia: I figured bcoz if there is manga title there should be a chapter right so I think based on your comment the current code should be ok right? or should i do something like this (chapter not null then manga - chapter else manga) else downloader

arkon: Yes, it's probably fine since `chapter` _shouldn't_ be null if `mangaTitle` isn't null.

However, please do remove the leading space before the colon.

MohitMandalia: okay will commit it by just removing these spaces.

MohitMandalia: @arkon one more thing to discuss about the app should we like discuss here or some other place? I have an issue about the app before raising an issue I just wanted to discuss that.

arkon: Either create a new issue or discuss it on Discord.

PR #6418, opened 2021-12-30
Update wording of "pref_remove_after_marked_as_read"

FourTOne5:


CarlosEsco:
Does this need clarification?

FourTOne5:
> Does this need clarification?

yes ~~Saying from personal experience~~

CarlosEsco:
There is no other way to mark as read though?

FourTOne5:
> There is no other way to mark as read though?

After reading a chapter it gets marked as read (which is done automatically)

CarlosEsco:
The setting right below that specifies marking after reading though

FourTOne5:
> The setting right below that specifies marking after reading though

It says `After reading automatically delete` or am I missing something?

CarlosEsco:
Sorry mistyped. I meant the one below is obviously for the after reading delete.  So the current one is when a user marks it

FourTOne5:
> Sorry mistyped. I meant the one below is obviously for the after reading delete. So the current one is when a user marks it

Well it doesn't bother to add a extra word to help dumb people ~~like me~~

PR #6435, opened 2022-01-04
Reduce stutter when entering Browse screen

ivaniskandar:
Had a hard time justifying this band-aid fix because the gain is negligible until I tested it with a weaker device.


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/SourceController.kt
```diff
@@ -92,6 +92,7 @@ class SourceController :
 
         // Update list on extension changes (e.g. new installation)
         (parentController as BrowseController).extensionListUpdateRelay
+            .skip(2)
```
arkon, line 95: Same here.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionController.kt
```diff
@@ -143,6 +144,7 @@ open class ExtensionController :
         }
 
         searchView.queryTextChanges()
+            .drop(1)
```
arkon, line 147: Can you add a comment as to why?

arkon (changes requested):

PR #6439, opened 2022-01-05
fix crash in ReaderActivity

jmir1:
building from after commit 0b9d436753f22479254dff46b5b1292ba45addcf, the ReaderActivity crashes upon opening.

we need to call `requestFeature()` before accessing the contentView in any way, because right now we get this error: `android.util.AndroidRuntimeException: requestFeature() must be called before adding content`


arkon:
Cheers, I had this staged locally. :P

jmir1:
at least you didn't publish a release with this bug like a certain someone.................

PR #6443, opened 2022-01-05
Place TabletUI setting under appearance

OncePunchedMan:
makes more sense to me having it under appearance


vengiare:
I believe having it in advanced was deliberate. 

Also in my opinion this is a setting that you'll touch only if you're having problems anyway.

arkon:
Yeah it was mostly intentional since you could force tablet UI then end up with being unable to access some side nav items if your screen's too small.

Open to other opinions.

PR #6445, opened 2022-01-07
add Albanian string file

ndriqa:
This is just the addition of the Albanian string file for the translation of the app


arkon:
Hi, translations are done over on Weblate. See https://tachiyomi.org/help/contribution/#translation

We also discourage machine translations.

PR #6463, opened 2022-01-13
fix: handle Komga tracks during manga migration

### LINKED ISSUE #6354: Komga tracker and migration
### Steps to reproduce

1. Have a manga from a Komga server, with Komga tracking enabled.
2. Migrate this manga to the same or a different source and opt to include trackers in migration.

### Expected behavior

Komga tracker is updated to reflect the new entry (for Komga -> Komga), or is removed when migrating to non-Komga sources.

### Actual behavior

Komga tracker is copied over directly to the new entry.

### Crash logs

_No response_

### Tachiyomi version

0.12.3

### Android version

Android 11

### Device

Galaxy Tab S7

### Other details

The main motivation for this is that sometimes changes to my server or to Komga itself require migration from an old Series to a new Series. Since the motivation for migrating generally also means that the old Series link (and tracker) no longer functions, the user ends up with a non-functioning Komga tracker despite migration.
The ideal case would be that the tracker automatically switches to match the Komga Series it is connected to, but a "good enough" solution in my opinion would be to not migrate the Komga tracker at all. This would make it apparent from the manga details that a tracker was missing, and the tracker filter in the library could be used to identify Komga entries that still needed their tracker updated.

### Acknowledgements

gotson:
This will better handle migration of tracks for Komga sources.

closes #6354


arkon (changes requested):
Would be more ideal to implement logic within the individual trackers to handle what to do on migration instead.

gotson:
> Would be more ideal to implement logic within the individual trackers to handle what to do on migration instead.

Why not, however we would either need to pull a `TrackService`, which is not readily available here, or to use static methods.

Ideally domain logic should not be scattered in the app, with the UI layer performing db persistence, and go through domain services, but the state of the app is far from this and refactoring would be huge.

We also don't know how other trackers, if ever, would handle migration. That could be premature optimization to pull and abstract that behaviour.

If you want we can pull this in the `EnhancedTracker` interface as static methods? We can probably assume that all enhanced tracker will need the same treatment, since they don't need matching, they are tied up by the `Manga` url.

gotson:
@arkon any preference on the above ?

arkon:
The static method in `EnhancedTrackService` is probably good enough for the current requirements.

gotson:
> The static method in `EnhancedTrackService` is probably good enough for the current requirements.

Turned out more complex than i thought, so i inject the enhanced services in the `SearchPresenter`, and the logic is moved inside `EnhancedTrackService`, to be implemented by each tracker.

arkon (approved):

PR #6471, opened 2022-01-15
Update core-splashscreen

ivaniskandar:


arkon:
@ivaniskandar Not sure if you noticed that the build failed.

ivaniskandar:
oopsie i forgot to push the fix

Issue #2161, opened 2019-08-03
[Feature Request] Literally scroll through pages instead of navigating to other pages if needed


0xAdk:
**Idea: The ability to use the volume buttons to through scroll pages for optimal laziness**

**Just to clarify** 

- Since there was a similarly named request #95 that lead to the volume key navigation feature. What I am requesting is the ability to Scroll the page (not "turn" it).

**How I see it**
- This would be a child option of the previously mentioned feature. 
When enabled pressing the volume keys while the page has content that is hidden off screen on only one of the screen's sides, the viewer smoothly translates a preset amount (Something like a whole, 1/2, or 1/4 of a screen's width / height) before "turning" to the next page

**Little note** for ease of implementation making it only compatible with "smart fit" scale type might make it more simple as you know what say you need to scroll.

PhaZ90771:
This would be especially useful for webtoon series, since the page split is often through a panel in order to keep the page height consistent.

arkon:
S-Pen integration also relevant to this: #4235

arkon:
Navigate to pan addresses this.

PR #6481, opened 2022-01-17
Navigate to pan / landscape zoom

### LINKED ISSUE #6432: In Smart Fit, Zoom out briefly to show full landscape page before zooming in

When viewing landscape images that don't fit screen, the Smart Fit reader will zoom on the left/right side of the image, in order to fit the height.

This can be a problem when navigating, if one is not aware that the image is landscape, tapping the edge of the screen will go to the next page (related issue #2161).

The Smart Fit reader could show the full image briefly when opening landscape page, then zoom in as usual. That way the user would be aware that this is a landscape page, which needs panning to be seen in its entirety.

At the moment i have to use Screen Fit instead, and perform manual zoom when viewing a landscape image.

### Other details

This was a feature that was available in the ComicRack for Android application.

### Acknowledgements

gotson:
This brings 2 changes for the Paged Reader:
1. Touch to pan: when the image is zoomed, touching will pan the image instead of navigating to the previous/next image. Should address #2161 but only for the Paged Reader.
2. Landscape overview: when in Fit Screen, and the image is landscape (width > height), then the image will be shown entirely (normal), but then it will zoom to fit height. Closes #6432.

~~There are some open questions before we can integrate this:~~

~~1. Touch to pan: should it be a reader option that can be turned on/off ?~~
~~2. Landscape overview: should it be a reader option that can be turned on/off, or should it be a new scale type _Fit screen (with landscape overview)_~~

See [Video](https://imgur.com/a/gl0rayg) of the feature in action.


gotson:
@arkon if you could have a look at the open points, that would be great. Any other feedback is welcome, or if there's anything I should try to port to the webtoon viewer too? I never use webtoon so I don't really now of there's any similar need there.

gotson:
Note also that landscape overview doesn't work on the first image when opening the Manga (whether it's the first page or the last page read), this has to do with events being fired on page change, but I couldn't find when/how to trigger manually for the first page shown.

gotson:
I renamed the touch to pan feature as navigate to pan, as it also works with volume button navigation.

gotson:
I've changed the reader settings for _Landscape image zoom_ to be only displayed when the _Image scale_ is set to _Fit Screen_.

I have also managed to fix the following:
- zoom if the first image is landscape
- properly reload the page holders when landscape zoom is enabled or disabled
- zoom when the image finally loads

Should be ready to review @arkon

gotson:
Both options are enabled by default, after discussing on Discord, which seems to be a good idea to gather feedback with the preview. We can always change the default before releasing in stable.

PR #6504, opened 2022-01-26
fallback to chrome when no default browser is found

CarlosEsco:
fall back to chrome no default browser is found, so verified extensions at least try chrome.


arkon:
We may want to actually query for options and present a picker for Android 12+?

Was also noted that the app opening loop is happening on older versions too (at least Android 11), so we might need to handle that too (could be separately if you don't want it to be in this PR).

CarlosEsco:
Updated to handle android 6+ https://developer.android.com/training/app-links#android-app-links

arkon:
Closing this since we discovered on Discord that it was only an issue in Neko due to the lack of `QUERY_ALL_PACKAGES` permission.

I'll do the version check change separately.

CarlosEsco:
I did end up switching to launching the intent with the match all

Issue #6517, opened 2022-01-30
No title library grid


Horodep:
### Describe your suggested feature

As font in new version became a bit clumsy and really affects the appearance of covers in some locales may be it would be a good option to let users hide titles in the compact view. Most of the titles are anyway dublicated on the cover.
<img src="https://user-images.githubusercontent.com/8913017/151689218-96dc36c2-8a19-4474-b908-c438cab63160.png" width="300" >

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.12.3](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

Issue #3272, opened 2020-06-04
[Feature Request] Option to Hide Text in Library


shashank-p:
### Why/User Benefit/User Problem
(explain why this feature should be added)
Cleaner UI with only manga covers without any Text on or below it in library
 
### What/Requirements
(explain how this feature would behave)
Checkbox to hide manga title text in filter from library

arkon:
Going to turn this down since there's occasional issues with covers not showing up properly and I want to avoid having too much feature/options bloat.

PR #6528, opened 2022-01-31
Cover only grid added to library 

micaelagimenez:
Resolves [#6517 ](https://github.com/tachiyomiorg/tachiyomi/issues/6517)

cover grid


vengiare:
Does it address the problem mentioned in the previous closed issue (#3272) about the same feature? Or is it not that much of a problem now?

> there's occasional issues with covers not showing up properly

micaelagimenez:
@vengiare I hadn't seen that issue (#3272) but I didn't notice any issue myself with the covers when I tried this feature.

vengiare:
Ideal solution for that problem would be to show title when cover hasn't loaded yet, but that could be complicated.

Horodep:
> Ideal solution for that problem would be to show title when cover hasn't loaded yet, but that could be complicated.

I may say some nonsense but doesn't changing the order layer solves the issue? 
I mean.. we can always draw text, but then draw cover over the text. If there is no cover, the text won't be hidden. And there will be no complicated bugs about determining if cover if displayed or not.

micaelagimenez:
I added a new commit with an else added to the holder. This way, in case the image is null or empty, the title is displayed.

arkon:
Does this actually make any sense to have when browsing?

FourTOne5:
Shouldn't "Cover Only Grid" sound better?

CarlosEsco:
> Does this actually make any sense to have when browsing?

Not at all.  Should be a library only thing

arkon:
@micaelagimenez Do you mind updating it so it's only an option in the library?

micaelagimenez:
Sure! I can do that and change the name if everyone prefers

app/src/main/res/values-es/strings.xml
```diff
@@ -222,7 +222,7 @@
     <string name="action_display">Mostrar</string>
     <string name="action_display_grid">Rejilla compacta</string>
     <string name="action_display_list">Lista</string>
-    <string name="action_display_no_title_grid">Rejilla sin títulos</string>
+    <string name="action_display_cover_only_grid">Rejilla con portada</string>
```
FourTOne5, line 225: Only add `en` strings others are translated and added via weblate

FourTOne5, line 225: Only add `en` strings others are translated and added via weblate

Horodep:
> No title grid renamed and now only available in library
You forgot to rename PR itself.

PR #6531, opened 2022-02-01
Unexpected Error

Razorior:
I am Razorior and i m not able to download episodes due an error/bug.
It's shows, Downloader ( could not download chapter due to unexpected error) 
What should I do now??


loocool2:
We are not Aniyomi.

Issue #5591, opened 2021-07-21
Warning dialog if a manga is already in the library & has the exact title regardless of the source


CVIUS:
### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.11.1](https://github.com/tachiyomiorg/tachiyomi/releases/tag/v0.11.1)**.
- [X] I will fill out all of the requested information in this form.

### Describe your suggested feature

A detection mechanism where it'll detect if a manga is already in the library/category & has the exact title regardless of the source. Duplicate manga sometimes can be found in the library since one cannot remember all the manga titles he/she have. This also happen when a user is browsing from different sources. 

A warning dialog will pop saying "This manga is already in the library, do you still want to add this?" 

Speaking of "In library" badge, it is only present in respective source from where the user first found & added it( the manga).

CVIUS:
closed by #6579

PR #6579, opened 2022-02-06
Detect identical mangas when adding to library

foxscore:
### Summary
When adding a manga to your library, the app will go through the library and compare the names of all mangas until either a match is found, or the end of the list is reached.
1) Nothing got tagged, so it will just be added to the library
2) A manga with the same name AND from a different source was found.
    A dialog box will appear, informing the user about this and giving them 3 options:
    - "Add"
    - "Cancel"
    - and "Show manga"

I have tested it myself, and it appears to be working. However, I have not been able to test the German translation, since I can't seem to be able to change the language of the app 🤔 

The detection method is very simple (just comparing the names, ignoring letter casing). So it will neither trigger if the names are in different languages, nor if they are just slightly off.
It's not perfect, but it's a start.

### Images
<details>
<summary>Screenshot_20220206_021348.png</summary>

</details>

### Quick note
This is my first time working with Kotlin/Android Studio, and I'm very much not familiar with either.
It's also my second ever time contributing to something on GitHub, so if I did something wrong, please tell me so that I may learn from it.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
@@ -2,6 +2,7 @@ package eu.kanade.tachiyomi.ui.manga
 
 import android.app.Activity
 import android.app.ActivityOptions
+import android.app.AlertDialog
 import android.content.Context
 import android.content.Intent
 import android.graphics.Bitmap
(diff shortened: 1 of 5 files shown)


CVIUS:
Closes #5591

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -516,8 +517,53 @@ class MangaController :
...
+            // Source check, just in case we fetched the library list before it got removed
+            origin.source != other.source
+    }
+
+    private fun getSourceFromLong(sourceId: Long): Source {
+        return Injekt.get<SourceManager>().getOrStub(sourceId)
+    }
+
+    private fun showAddDuplicateDialog(manga: Manga, other: Manga, source: Source) {
+        AlertDialog.Builder(activity).apply {
```
arkon, line 556: Use `MaterialAlertDialogBuilder` instead

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -516,8 +517,53 @@ class MangaController :
...
         }
+
+        return result
+    }
+
+    private fun isSameManga(origin: Manga, other: Manga): Boolean {
+        return origin.title.equals(other.title, ignoreCase = true) &&
+            // Source check, just in case we fetched the library list before it got removed
+            origin.source != other.source
+    }
```
arkon, line 549: This should probably just be a database query rather doing it in Kotlin.

app/src/main/res/values-de/strings.xml
```diff
@@ -701,4 +701,6 @@
     <string name="publishing_finished">Veröffentlichung abgeschlossen</string>
     <string name="action_faq_and_guides">Häufige Fragen und Leitfäden</string>
     <string name="backup_restore_invalid_uri">Fehler: leerer URI</string>
+    <string name="confirm_manga_add_duplicate">Es gibt bereits ein Manga mit diesem Namen in deiner Bibliothek, aber von einer anderen Quelle (%1$s).\\n\\nSoll es dennoch hinzugefügt werden?</string>
```
arkon, line 704: Translations are managed in Weblate, so only add English strings in the PR.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
foxscore: Done as requested, but I had to wrap the function in `activity?.let { ... }`, because MaterialAlertDialogBuilder does not allow nullables. *(At least that solution is what AndroidStudio recommended, and it appears to be working)*

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
foxscore: This commit a42bcfd758d5e0dfdb31b01989103bc6cb0c4267 seems to work, but I don't know if there are any underlying issues with how I've done it.

CVIUS:
#5591 can be closed🙏

arkon:
Cleaned up and merged to master as 71ddb1657440d54178f10fd62e3d0c66ee95a5a0

PR #6606, opened 2022-02-08
Auto Install after an app update is downloaded (for A12+)

curche:
Based on the same feature as in J2k. (Ref: https://github.com/Jays2Kings/tachiyomiJ2K/commit/4a7317ad7032e97f8db61d5631ad8a2728037071).

Also includes an option in Notification for cancelling the downloading notification in case the connection to GitHub breaks. (Ref https://github.com/Jays2Kings/tachiyomiJ2K/commit/4dea924337ffd4a01342fa0b48da47c221d2b897)

Tested on MiA2 (jasmine-sprout) running Stable Custom ROM Android 12 
 - New Setting comes up under Settings > General
 - Tested by decreasing the version number in build.gradle.kts, enabling the app_update_checker for debug build and trying to let the app find the new preview update
 - App manages to find the new update and starts downloading it. The installation does ofc fail because of different signature. However, no error is shown and the update notif just goes away after completion (regardless of failure/success)
 - More testing required to ensure I haven't missed anything (just migrated code from j2k while assuming how tachi might be doing it)
 - The Cancel button works correctly but sometimes comes up Twice in the Notification :thinking: 
 - All Credits to Jay2Kings for adding the feature to TachiyomiJ2K first


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsGeneralController.kt
```diff
@@ -54,5 +57,19 @@ class SettingsGeneralController : SettingsController() {
...
+            intListPreference() {
+                key = Keys.autoUpdateApp
+                titleRes = R.string.pref_app_auto_update
+                entriesRes = arrayOf(
+                    R.string.pref_app_auto_update_any_network,
+                    R.string.pref_app_auto_update_wifi_only,
+                    R.string.pref_app_auto_update_never,
+                )
+                entryValues = arrayOf("0", "1", "2")
+                defaultValue = "1" // only on unmetered
```
NotTsunami, line 70: Please make this strictly opt-in.

arkon:
I'd actually prefer the behaviour to be:

- App prompts to download update (same as before)
- App downloads APK, then prompts to install (same as before)
- Installation happens without the Package Installer prompt in A12+ where applicable

I say this because I find it annoying that the app suddenly dies when sometime after the initial download is started.

CrepeTF:
How does all this work on data, some users may have limited data so it'd be wise to have an option to only download over Wifi

arkon:
Going to close this since I still find it incredibly awkward when J2K/Neko auto-update, which as a user means the app suddenly kills itself and can't be reopened while it's updating in the background with no visible indication of it doing so/its progress.

PR #6607, opened 2022-02-08
Fix issues with invalid Strings XML entries

Jeehut:
According to [the docs](https://developer.android.com/guide/topics/resources/string-resource) it's not allowed to have empty start-tags for strings or item entries, which I fixed here. Before, there were multiple Strings files with an opening XML tag (such as `<string name="manga"/>`) but no translation content. This fixes them all.

### DIFF
--- a/app/src/main/res/values-eo/strings.xml
+++ b/app/src/main/res/values-eo/strings.xml
@@ -569,7 +569,7 @@
     <string name="pref_library_update_refresh_metadata_summary">Kontroli ĉu estas nova kovrio kaj detaloj kiam ĝisdatigas bibliotekon</string>
     <string name="cover_saved">Kovrilo konservita</string>
     <string name="manga_cover">Kovrilo</string>
-    <string name="licensed"/>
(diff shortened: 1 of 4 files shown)


FourTOne5:
Hi, translations are done over on Weblate. See https://tachiyomi.org/help/contribution/#translation

Jeehut:
@FourTOne5 Then it seems Weblates output is not conforming to the spec.

arkon:
That's been the case before. We (or you) can open issues with them if needed.

Jeehut:
@arkon I'm not particularly interested in using this app (I don't even use an Android phone). I just happened to detect this because I'm writing an Android Strings file parser for my current project and integrated this apps Strings files in my unit tests (among others). And I thought it's probably a good idea to inform you about it and maybe provide a fix.

How you handle this and if you want to fix it at all is out of my scope, I basically just wanted to inform you. :)

arkon:
Tracking in https://github.com/WeblateOrg/weblate/issues/7248

PR #6620, opened 2022-02-10
Update logic in not showing Blank Upload Date

FourTOne5:


app/src/main/java/eu/kanade/tachiyomi/data/database/DbOpenCallback.kt
```diff
@@ -20,7 +20,7 @@ class DbOpenCallback : SupportSQLiteOpenHelper.Callback(DATABASE_VERSION) {
         /**
          * Version of the database.
          */
-        const val DATABASE_VERSION = 14
+        const val DATABASE_VERSION = 13
```
ghostbear, line 23: This will make the database inaccessible for anyone who has gotten the database upgraded to version 14. It is better to increase the version and revert whatever you want to revert with a SQL query. 

The silent error this causes
```
Caused by: android.database.sqlite.SQLiteException: Can't downgrade database from version 14 to 13
        at androidx.sqlite.db.SupportSQLiteOpenHelper$Callback.onDowngrade(SupportSQLiteOpenHelper.java:205)
        at io.requery.android.database.sqlite.RequerySQLiteOpenHelperFactory$CallbackSQLiteOpenHelper.onDowngrade(RequerySQLiteOpenHelperFactory.java:59)
        at io.requery.android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:271)
        at io.requery.android.database.sqlite.SQLiteOpenHelper.getReadableDatabase(SQLiteOpenHelper.java:199)
        at io.requery.android.database.sqlite.SQLiteOpenHelper.getReadableDatabase(SQLiteOpenHelper.java:44)
        at com.pushtorefresh.storio.sqlite.impl.DefaultStorIOSQLite$LowLevelImpl.query(DefaultStorIOSQLite.java:372)
        at com.pushtorefresh.storio.sqlite.operations.get.DefaultGetResolver.performGet(DefaultGetResolver.java:34)
```

FourTOne5: Thanks for the info.

FourTOne5:
> I personally don't think we should do this. if the data you're given is bad you shouldn't account for it in the app. It makes sense if the extension doesn't give an upload date but when given bad data you should fix it in the ext

It'll be now blank on bad data.

FourTOne5:
it'll be better how thing currently is.

PR #6628, opened 2022-02-11
Update Theme Preview Items

CrepeTF:
Updated the look of the theme previews

### Images
| Before | After |
| ------- | ------- |
| Before buying Korone NFTs | After buying Korone NFTs |


app/src/main/res/layout/pref_theme_item.xml
```diff
@@ -2,7 +2,7 @@
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
-    android:layout_width="118dp"
+    android:layout_width="114dp"
```
arkon, line 5: You probably need to update the number in `ThemesPreference` too.

CrepeTF: Nice, didn't catch that

app/src/main/res/layout/pref_theme_item.xml
```diff
@@ -15,102 +15,115 @@
...
+            <com.google.android.material.card.MaterialCardView
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:layout_marginStart="4dp"
                 android:layout_marginTop="4dp"
-                android:background="@drawable/rounded_rectangle"
                 app:layout_constraintStart_toStartOf="@+id/cover_container"
-                app:layout_constraintTop_toTopOf="@+id/cover_container">
+                app:layout_constraintTop_toTopOf="@+id/cover_container"
+                app:cardCornerRadius="6dp">
```
arkon, line 68: I'm not sure how I feel about the radii for the cover/badges. The previous implementation actually represented the real UI.

CrepeTF: but generally when things become smaller, it may be better to exaggerate the shape

PR #6651, opened 2022-02-14
LibraryUpdateNotifier: Remove counter

OncePunchedMan:
Many users complain about how it's checking less manga than before. Since we want to prevent broken or removed sources due to heavy load caused by Tachi, this is a small change with big impact.

We already got the new update restrictions so why not support these settings so they stay default for longer than one google search


arkon:
I plan on addressing to make it more obvious why certain entries are skipped.

OncePunchedMan:
> I plan on addressing this differently to make it more obvious why certain entries are skipped.

ight

PR #6679, opened 2022-02-19
Show number of all unique mangas in all categories

### LINKED ISSUE #6522: Show number of all unique mangas in library

When you have the tab settings of "Show category tabs" and "Show number of items" both on, it should show number of items for the library at the top right after "Library", like it does for the categories. It should also change the number with the filter options, like the category numbers already do.

### Other details

the reason i'm suggesting this is i sometimes want to know my total manga but because you can put manga in mulitple categories at once, you sometimes can't simply add the amount of manga in each category seperatly. 
Example:
Screenshot_20220130-062903_Tachiyomi

### Acknowledgements

ComBeat:
Closes #6522

A method was implemented that counts all mangas in the library uniquely and that number is shown in the action bar, even if categories and the number of items are enabled.

> Would look better if the number was a more darkened color and maybe a variant of the font
Rather than in parenthesis and same color

I've tried to implement this but haven't found a solution. Since the main part of the issue is implemented, I decided to open the pr.

default


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
```diff
@@ -151,15 +151,38 @@ class LibraryController(
...
     }
 
+    /**
+     * Iterates through all mangas in all categories.
+     * If the manga is not already in the list, add them.
+     *
+     * @param mangaMap The Map with all the categories and mangas.
+     * @return the count of all unique mangas.
+     */
+    private fun getTotalLibraryCount(mangaMap: Map<Int, List<LibraryItem>>): Int {
```
FourTOne5, line 172: Instead of recalculating it every time "Title" is updated i think it'll be better to include the value in `Library` class.

vengiare:
Can I ask what does it look like on the Library sheet

ComBeat:
> Can I ask what does it look like on the Library sheet

@vengiare Look at the screenshot above and then look at the title bar. The number right next to "Library" is what has been implemented. In this case there are 12 unique mangas in the library.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
ComBeat: I agree. I'll do that.

vengiare:
I mean how is the feature enabled? What does the toggle look like in the filter sheet.

#

It would also have conflicts with this specific configuration, which show manga count for the current category.

Screenshot_20220222-021319

ComBeat:
> I mean how is the feature enabled? What does the toggle look like in the filter sheet.

So far there is no seperate setting for this. This feature can be enabled  with "Show number of items".
> It would also have conflicts with this specific configuration, which show manga count for the current category.

I've tested this pr and so far there have been no conflicts with that specific configuration.
With that configuration it looks like this.
image

vengiare:
Maybe something like this?

image
image

ComBeat:
That's a good idea. I will implement that in the next few days.

ComBeat:
I've moved the library count to the class `Library` and added a setting for enabling/disabling the library count.

ComBeat:
> Maybe something like this?

Right now it's implemented exactly like you suggested in the screenshot, with `Show number of library items` being disabled if `Show number of items per category` is disabled.
But when I think about it, it makes more sense to not let getting it disabled and instead handle them as 2 independent settings. 
What do you think @vengiare?

FourTOne5:
Any non-eng strings are handled by weblate revert changes on those.

ComBeat:
> Any non-eng strings are handled by weblate revert changes on those.

Done.

vengiare:
> with `Show number of library items` being disabled if `Show number of items per category` is disabled.

Wait what? My idea was `Show number of library items` will be disabled if >> `Show category tabs` << is disabled.

#

Also on second thought, "Show number of library items" can be renamed to "Show total library count" or something,

ComBeat:
> My idea was Show number of library items will be disabled if >> Show category tabs << is disabled.

I've misread your post. Sorry.
I've implemented it now exactly like you suggested.

image
image
image

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
```diff
@@ -242,7 +242,9 @@ class PreferencesHelper(val context: Context) {
 
     fun categoryTabs() = flowPrefs.getBoolean("display_category_tabs", true)
 
-    fun categoryNumberOfItems() = flowPrefs.getBoolean("display_number_of_items", false)
+    fun categoryNumberOfItems() = flowPrefs.getBoolean("display_number_of_category_items", false)
```
FourTOne5, line 245: since the key was changed from `display_number_of_items` to `display_number_of_category_items` you might want to migrate the value in `Migrations.kt`

app/src/main/res/values/strings.xml
```diff
@@ -95,7 +95,8 @@
     <string name="action_display_local_badge">Local manga</string>
     <string name="action_display_language_badge">Language</string>
     <string name="action_display_show_tabs">Show category tabs</string>
-    <string name="action_display_show_number_of_items">Show number of items</string>
+    <string name="action_display_show_number_of_category_items">Show number of items per category</string>
+    <string name="action_display_show_number_of_library_items">Show total number of library items</string>
```
FourTOne5, line 99: `Show number of items in library`?

app/src/main/res/values/strings.xml
```diff
@@ -95,7 +95,8 @@
     <string name="action_display_local_badge">Local manga</string>
     <string name="action_display_language_badge">Language</string>
     <string name="action_display_show_tabs">Show category tabs</string>
-    <string name="action_display_show_number_of_items">Show number of items</string>
+    <string name="action_display_show_number_of_category_items">Show number of items per category</string>
```
FourTOne5, line 98: `Show number of items in category`?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
```diff
@@ -140,19 +140,26 @@ class LibraryController(
...
         } else {
             currentCategory?.name
         }
 
-        if (preferences.categoryNumberOfItems().get() && libraryMangaRelay.hasValue()) {
+        if (libraryMangaRelay.hasValue()) {
             libraryMangaRelay.value.mangas.let { mangaMap ->
-                if (!showCategoryTabs || adapter?.categories?.size == 1) {
-                    title += " (${mangaMap[currentCategory?.id]?.size ?: 0})"
+                title += if (title == resources?.getString(R.string.label_library)) {
```
FourTOne5, line 153: Checking if the string is same might not be a good idea.

FourTOne5 (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
ComBeat: So far I have no idea how to do that, but I will continue to look into it.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
ComBeat: I'm also not quite content with this solution. I'll try something different.

app/src/main/res/values/strings.xml
ComBeat: It's shorter than the current phrase.
I will apply that.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
ComBeat: It's now checked against the preference `preferences.categoryTabs().get()`.

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
ComBeat: @FourTOne5 I integrated it into `Migrations.kt`. Please take a look at it.

app/src/main/java/eu/kanade/tachiyomi/Migrations.kt
```diff
@@ -265,6 +265,12 @@ object Migrations {
                     preferences.navigationModePager().set(5)
                     preferences.navigationModeWebtoon().set(5)
                 }
+                if (prefs.getBoolean("display_number_of_category_items", false)) {
+                    preferences.categoryNumberOfItems().set(false)
+                }
+                if (prefs.getBoolean("display_number_of_library_items", false)) {
+                    preferences.libraryNumberOfItems().set(false)
+                }
```
FourTOne5, line 273:
```suggestion
                val displayNumberOfItems = flowPrefs.getBoolean("display_number_of_items", false)
                if (displayNumberOfItems)
                    preferences.categoryNumberOfItems().set(false)
                }               
```
Also move this another if block after bumping app version

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySettingsSheet.kt
```diff
@@ -394,31 +394,38 @@ class LibrarySettingsSheet(
...
                 }
                 adapter.notifyItemChanged(item)
             }
         }
 
         inner class TabsGroup : Group {
             private val showTabs = Item.CheckboxGroup(R.string.action_display_show_tabs, this)
-            private val showNumberOfItems = Item.CheckboxGroup(R.string.action_display_show_number_of_items, this)
+            private val showNumberOfLibraryItems = Item.CheckboxGroup(R.string.action_display_show_number_of_library_items, this)
+            private val showNumberOfCategoryItems = Item.CheckboxGroup(R.string.action_display_show_number_of_category_items, this)
```
arkon, line 405: I'm actually not sure if I want to have separate settings for this. This change would be a lot simpler without it.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -356,10 +356,31 @@ class LibraryPresenter(
...
     }
 
+    /**
+     * Iterates through all mangas in all categories.
+     * If the manga is not already in the list, add them.
+     *
+     * @param mangaMap The Map with all the categories and mangas.
+     * @return the number of all unique mangas.
+     */
+    private fun getTotalLibraryCount(mangaMap: Map<Int, List<LibraryItem>>): Int {
```
arkon, line 370: You should be able to just get all the map values (`mangaMap.values`) as a flattened list, then get all the distinct items (`distinctBy`).

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySettingsSheet.kt
```diff
@@ -394,31 +394,38 @@ class LibrarySettingsSheet(
                     unreadBadge -> preferences.unreadBadge().set((item.checked))
                     localBadge -> preferences.localBadge().set((item.checked))
                     languageBadge -> preferences.languageBadge().set((item.checked))
-                    else -> {}
```
arkon, line 397: Don't remove this, the compiler warns about it.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryAdapter.kt
```diff
@@ -36,6 +36,14 @@ class LibraryAdapter(
      */
     private var itemsPerCategory: List<Int> = emptyList()
 
+    var itemsInLibary: Int = 0
+        set(value) {
+            if (field != value) {
+                field = value
+                notifyDataSetChanged()
```
arkon, line 43: Is this actually necessary?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySettingsSheet.kt
ComBeat: Fixed

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySettingsSheet.kt
ComBeat: It'd be definitely simpler.  
It's already implemented, but if you as a member say, that one setting is better, I'll revert it.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryAdapter.kt
ComBeat: The compiler complained, if there was no set block.
I just removed the set block and the compiler isn't complaining anymore.
I remove it.

app/src/main/java/eu/kanade/tachiyomi/Migrations.kt
ComBeat: Are you sure that ```display_number_of_items``` works?
Or did you mean ```display_number_of_category_items```?
In ```PreferencesHelper.kt``` the preference ```display_number_of_items``` was renamed to ```display_number_of_category_items```.

arkon: No need to rename the preference if we don't introduce a new one.

ComBeat: With this pr ```display_number_of_library_items``` is new.

arkon: Right, but I suggested not adding a new setting.

arkon:
Superseded by 8a3a7418d055b500517ef6c9c5623328ae01bc10

PR #6681, opened 2022-02-20
adding CC feature in app

royrapstar:
I have a request  you to  that , add feature of close caption (CC) means subtitles. like youtube have auto generated subtitles features for eng dub animes .


curche:
1. Tachiyomi does not support anime, and has no plans to support anime. In addition Tachiyomi is not affiliated with Aniyomi https://github.com/jmir1/aniyomi
2. You're supposed to create your request under **Issues** not **Pull Request**

Issue #5058, opened 2021-05-11
Local manga doesn't load Title from details.json file


laurinvesely:
## Why/User Benefit/User Problem
The user wants to add a local manga that has a special character (for example ':') in its title.
The manga folder naming cant support special characters so title is added as title in details.json.
Tachiyomi only shows the folder name as title of the manga instead of title set in details.json.

## What/Requirements
Tachiyomi should use details.json title for the manga title and if it doesnt exist then use folder name.

curche:
> Tachiyomi should use details.json title for the manga title and if it doesnt exist then use folder name.

Afaik this is how it does work. 

> The manga folder naming cant support special characters so title is added as title in details.json.

Tried at my end and works fine. 
Some users have reported similar problems before (#4803) and I think this is caused due to the device filesystem not accepting it

laurinvesely:
> Tried at my end and works fine.

Have you tried creating a folder with characters like `* < > : ? " / \ |` ? this is not possible on most smartphones and neighter on OS like Windows 10.

> Afaik this is how it does work.

I use latest Tachiyomi on Samsung Galaxy S10e and this does not work as intended on the phone. Title attribute in json has no impact on manga name in Tachiyomi.

curche:
Checked again, can confirm that the title is not getting updated from *.json file for local manga in preview version r3368
Sorry for the confusion. Can be marked as a [Bug].

https://user-images.githubusercontent.com/72807749/118523572-df866a80-b75a-11eb-86c0-148faf6c438c.mov

laurinvesely:
> 
> 
> Checked again, can confirm that the title is not getting updated from *.json file for local manga in preview version r3368
> Sorry for the confusion. Can be marked as a [Bug].
> MNML-May17-094903PM.mov

it seems like i cant change the label to bug

curche:
/edit-title "Local manga doesn't load Title from details.json file"

arkon:
We actually just never update the title, which is tracked in #3608.

PR #6688, opened 2022-02-21
#5058 fix issue local manga not using title from json file

laurinvesely:
Signed-off-by: Laurin Vesely <laurin@vesely-online.de>

#5058 fix issue local manga not using title from json file
When opening a manga from local source the json details including the title will be read.
MangaPresenter.fetchMangaFromSource calls SManga.copyFrom() which now also copies the title that was previously taken from json.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/source/model/SManga.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/source/model/SManga.kt
@@ -24,6 +24,10 @@ interface SManga : Serializable {
     var initialized: Boolean
 
     fun copyFrom(other: SManga) {
+        if (other.title != null) {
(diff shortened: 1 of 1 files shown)


arkon:
This has the same issue with #6219, i.e. it needs to account for moved downloads (which isn't relevant for local manga, but this function is used for non-local manga too).

FlaminSarge:
Question regarding the issue with download folder name for both issues, would it instead be cleaner to maintain the downloaded folder name for each manga in the DB rather than attempting to move it any time the title changes? We can decouple title string updates from download folder if we do that.

laurinvesely:
i will be looking into this problem... closing the pull request for now.

PR #6718, opened 2022-03-02
Use existing worker for manual backup creation

ivaniskandar:
This will show the "creating backup" notification when auto backup is
running. Complete or error notification will continue to be shown only on
manual job.


app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupCreatorJob.kt
```diff
@@ -45,12 +71,26 @@ class BackupCreatorJob(private val context: Context, workerParams: WorkerParamet
...
                 )
                     .addTag(TAG)
+                    .setInputData(workDataOf(IS_AUTO_BACKUP_KEY to true))
                     .build()
 
-                WorkManager.getInstance(context).enqueueUniquePeriodicWork(TAG, ExistingPeriodicWorkPolicy.REPLACE, request)
+                workManager.enqueueUniquePeriodicWork(TAG, ExistingPeriodicWorkPolicy.REPLACE, request)
             } else {
-                WorkManager.getInstance(context).cancelAllWorkByTag(TAG)
+                workManager.cancelAllWorkByTag(TAG)
```
arkon, line 79: Won't this potentially kill a manual backup job if you decide to toggle the interval setting while it's running since they're sharing the same tag?

ivaniskandar: This only called when user disables the auto backup but yeah you're right I'll do some tweak

PR #6725, opened 2022-03-03
Coil 2.x upgrade

ivaniskandar:
[changelog](https://github.com/coil-kt/coil/blob/main/CHANGELOG.md)


app/build.gradle.kts
```diff
@@ -220,7 +220,7 @@ dependencies {
     implementation("com.github.inorichi.injekt:injekt-core:65b0440")
 
     // Image loading
-    val coilVersion = "1.4.0"
+    val coilVersion = "2.0.0-rc01"
```
arkon, line 223: Will need to rebase on top of ghostbear's Gradle version catalog change.

PR #6726, opened 2022-03-03
Remove unused string

Riztard:
removing `Only update: ` string


Riztard:
wait it was changed but didnt remove unused string

PR #6728, opened 2022-03-03
Use Version Catalog & clean up Gradle files

ghostbear:


build.gradle.kts
```diff
@@ -1,21 +1,22 @@
...
-        mavenCentral()
-        google()
-        maven { setUrl("https://www.jitpack.io") }
-    }
+plugins {
+    alias(androidx.plugins.application) apply false
+    alias(androidx.plugins.library) apply false
+    alias(kotlinx.plugins.android) apply false
+    alias(libs.plugins.kotlinter)
+    alias(libs.plugins.versionsx)
```
ghostbear, line 15: Just gonna mention that `libs`, `kotlinx` and `androidx` creates a false-positive here in Android Studio but the project sync and build just fine

https://youtrack.jetbrains.com/issue/KTIJ-19369

PR #6765, opened 2022-03-10
Avoid throw as it is slow expensive operations

quangkieu:
Just rework some call to avoid "throw" operations which is slow and expensive


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -313,32 +313,34 @@ class LibraryUpdateService(
...
+                                                }
                                             }
-
-                                            // Convert to the manga that contains new chapters
-                                            newUpdates.add(
-                                                manga to newChapters.sortedByDescending { ch -> ch.source_order }
-                                                    .toTypedArray()
-                                            )
                                         }
-                                    } catch (e: SkipUpdateException) {
```
arkon, line 340: Should be able to remove `SkipUpdateException` itself too.

arkon (changes requested):

arkon:
Merged as 9547311d7d40c98b951e679b7a5e4967cf73ee0d

PR #6772, opened 2022-03-13
add display mode "List Medium"

Riztard:
not sure about the string name
List(S), List(M) / List small, List medium / List-small, List-medium/ other

<table>
<tr>
	<td>
	<td>
</table>

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
@@ -205,7 +205,7 @@ open class BrowseSourceController(bundle: Bundle) :
             binding.catalogueView.removeView(oldRecycler)
         }
 
-        val recycler = if (preferences.sourceDisplayMode().get() == DisplayModeSetting.LIST || preferences.sourceDisplayMode().get() == DisplayModeSetting.LIST_MEDIUM) {
+        val recycler = if (preferences.sourceDisplayMode().get() in listOf(DisplayModeSetting.LIST, DisplayModeSetting.LIST_MEDIUM)) {
             RecyclerView(view.context).apply {
                 id = R.id.recycler
                 layoutManager = LinearLayoutManager(context)
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
@@ -205,7 +205,7 @@ open class BrowseSourceController(bundle: Bundle) :
             binding.catalogueView.removeView(oldRecycler)
         }
 
-        val recycler = if (preferences.sourceDisplayMode().get() == DisplayModeSetting.LIST) {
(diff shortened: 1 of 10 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/library/setting/DisplayModeSetting.kt
```diff
@@ -4,6 +4,7 @@ enum class DisplayModeSetting(val flag: Int) {
     COMPACT_GRID(0b00000000),
     COMFORTABLE_GRID(0b00000001),
     LIST(0b00000010),
+    LIST_MEDIUM(0b00000100),
```
ghostbear, line 7: Just gonna mention that this is outside of the mask, so it will not work. You will have to do a DB migration on all categories to bit shift the whole flag to the left and also update all the mask values and flag values for each setting related to the category flag

vengiare:
Does this use the same ratio as Grid mode, Manga page covers?

> not sure about the string name

use Comfortable/Compact List 🤔

FourTOne5:
> use Comfortable/Compact List 🤔

The original one is already compact so maybe go for Comfortable

vengiare:
I meant rename the current one as "Compact list" then Comfortable for this PR

FourTOne5:
> I meant rename the current one as "Compact list" then Comfortable for this PR

👌

app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
```diff
@@ -205,7 +205,7 @@ open class BrowseSourceController(bundle: Bundle) :
             binding.catalogueView.removeView(oldRecycler)
         }
 
-        val recycler = if (preferences.sourceDisplayMode().get() == DisplayModeSetting.LIST) {
+        val recycler = if (preferences.sourceDisplayMode().get() == DisplayModeSetting.LIST || preferences.sourceDisplayMode().get() == DisplayModeSetting.LIST_MEDIUM) {
```
FourTOne5, line 208: listOf?

Riztard: ah miss that

app/src/main/java/eu/kanade/tachiyomi/ui/library/setting/DisplayModeSetting.kt
Riztard: yea i have no idea how that bits thing works, so maybe it would be better to let you handle it. maybe another pr or to here *wink*

Riztard:
> Does this use the same ratio as Grid mode, History, Manga page covers?
nah

Riztard:
> use Comfortable/Compact List 🤔

maybe

Soitora:
One should not be named Comfortable List.

Although List and Compact List works I guess

CarlosEsco:
99% of time the reason people have list in the first place is they want a condensed view, so  you are really catering to the minority here

Riztard:
> 99% of time the reason people have list in the first place is they want a condensed view, so you are really catering to the minority here

just think of it as different type, something like history tab style

CarlosEsco:
> > 99% of time the reason people have list in the first place is they want a condensed view, so you are really catering to the minority here
> 
> just think of it as different type, something like history tab style

Doesn't mean feature creep is warranted though

arkon:
Perhaps best for a fork.

Riztard:
image

PR #6787, opened 2022-03-16
Share logic for saving page/cover

ghostbear:


app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
```diff
@@ -110,6 +108,8 @@ class MangaPresenter(
 
     private val loggedServices by lazy { trackManager.services.filter { it.isLogged } }
 
+    private val imageSaver by lazy { ImageSaver(preferences.context) }
```
arkon, line 111: Could maybe add it to `AppModule` and inject it instead?

app/src/main/java/eu/kanade/tachiyomi/data/saver/ImageSaver.kt
```diff
@@ -0,0 +1,149 @@
...
+        return destFile.getUriCompat(context)
+    }
+}
+
+sealed class Image(
+    open val name: String,
+    open val location: Location
+) {
+    data class Cover(
+        val bitmap: Bitmap,
```
arkon, line 94: I _think_ Coil 2.x might allow us to get non-bitmap versions of the images, but we could look into that separately.

app/src/main/java/eu/kanade/tachiyomi/data/saver/ImageSaver.kt
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    object Cache : Location()
+
+    fun directory(context: Context): File {
+        return when (this) {
+            Cache -> context.cacheImageDir
+            is Pictures -> {
+                val file = File(
+                    Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_PICTURES),
```
arkon, line 136: Was part of the motivation of this PR to remove usage of deprecated APIs?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
```diff
@@ -338,44 +338,13 @@ class MangaPresenter(
...
 
     /**
-     * Save manga cover Bitmap to temporary share directory.
+     * Save manga cover Bitmap to picture or temporary share directory.
      *
-     * @param context for the temporary share directory
-     * @param coverBitmap the cover to save (as Bitmap)
-     * @return cover File in temporary share directory
+     * @param image the image with specified location
+     * @return flow Flow which emits the Uri which specifies where the image is saved when
```
arkon, line 344: Why a flow and not just a suspension function?

app/src/main/java/eu/kanade/tachiyomi/data/saver/ImageSaver.kt
ghostbear: Yeah, but they are still used as MediaStore before API level 29 doesn't have all functions needed to remove them entirely

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
arkon: Doesn't it seem awkward to keep the flow around until the scope's cancelled despite it not emitting multiple items?

Issue #6788, opened 2022-03-18
Add option to only update series where the latest chapter has been read


soruh:
### Describe your suggested feature

Since the recent changes in #6599 I have been missing some updates in series I am actively reading. I noticed that this was due to single unread chapters at the start of the series (mostly duplicated chapters). I think it would make sense to add an option to update all series where the newest chapter is read instead of either updating all series or updating only series where every chapter has been read. I also think that this would be a better default than the current one without putting unreasonably more load on sources.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.13.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

alessandrojean:
You can just mark the previous chapters as read (assuming you have already read them before).

soruh:
that's true, but I'd rather not manually go throught every single series in my library to find which ones habe duplicate chapters. At that point I will just turn on sync for everything.

alessandrojean:
Closing since the solution is mark the things you already have read as read.

soruh:
No I cannot. The unread filter includes series I have not finshed/started yet and ones that have new chapters so I will still need to go through a ton of series. I guess if there was an option to filter the library by the lateat chapter being read this would be feasable but at that point that could also just be an option for sync as I proposed.

soruh:
Damn, okay. I guess I will fork and try to implement it myself.

soruh:
P.S.: if your idea of good UX is "I categorically refuse to add this tiny feature. The user should do that work manually" idk. what to tell you.

Issue #6599, opened 2022-02-07
Global/library update not updating all items


arkon:
Starting in Tachiyomi 0.13.0, library updates may update less entries by default.

To learn why, please read the FAQ: https://tachiyomi.org/help/faq/#why-does-global-update-skip-some-entries

---

Related discussions:
- #6649

arkon:
Starting in 0.13.2, the update progress notification shows the full amount, but an additional informational notification is shown afterwards with the number of skipped items. If you don't want to see the skipped number, you can hide it from More > Settings > General > Manage notifications.

PR #6792, opened 2022-03-18
add option to skip updating mangas where the latest chapter is unread

soruh:
This adds the setting I suggested in #6788. I am sorry I got a bit frustrated in that issue.
I implemented the change myself and would appreciate if you could merge it, especially since there aren't really any downsides to doing so.

I also think it would make sense to make this the default instead of "all chapters read" because I think it will reduce the number of normal users wondering why they are missing updates but I won't argue on that matter.

Thank you.

P.S.: this is missing translations, please let me know if I should have put a dummy value for missing translations or if this is fine.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/queries/RawQueries.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/queries/RawQueries.kt
@@ -14,7 +14,7 @@ val libraryQuery =
     """
     SELECT M.*, COALESCE(MC.${MangaCategory.COL_CATEGORY_ID}, 0) AS ${Manga.COL_CATEGORY}
     FROM (
-        SELECT ${Manga.TABLE}.*, COALESCE(C.unreadCount, 0) AS ${Manga.COMPUTED_COL_UNREAD_COUNT}, COALESCE(R.readCount, 0) AS ${Manga.COMPUTED_COL_READ_COUNT}
+        SELECT ${Manga.TABLE}.*, COALESCE(C.unreadCount, 0) AS ${Manga.COMPUTED_COL_UNREAD_COUNT}, COALESCE(R.readCount, 0) AS ${Manga.COMPUTED_COL_READ_COUNT}, COALESCE(L.isRead, 0) AS ${Manga.COMPUTED_COL_LATEST_READ}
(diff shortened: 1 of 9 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/database/queries/RawQueries.kt
```diff
@@ -30,6 +30,13 @@ val libraryQuery =
             GROUP BY ${Chapter.COL_MANGA_ID}
         ) AS R
         ON ${Manga.COL_ID} = R.${Chapter.COL_MANGA_ID}
+        LEFT JOIN (
+            SELECT ${Chapter.COL_MANGA_ID}, ${Chapter.COL_READ} AS isRead
+            FROM ${Chapter.TABLE}
+            ORDER BY ${Chapter.COL_CHAPTER_NUMBER} DESC
```
FourTOne5, line 36: Chapter with highest number =/= latest chapter there are cases where manga resets every volume.

soruh: Hmm, I used the id first but I think that would wrongly choose old extras from the middle right?
I'm not sure how the chapters are ordered. I would choose the latest chapter to be the one at the top of the chapter list and during my search I found a "sort_by(chapter_number)". What would be the correct ordering that the chapter listing uses?

CarlosEsco:
So instead of just spending ten minutes marking previous as read, you spent 30+ solving a problem nobody else has, with a solution that doesn't quite work. I also thought you were gonna make your own fork?

soruh:
Yes, I spent 20mins once instead of looking through every series in my library and having to make sure I don't accidentally forget to mark a duplicate chapter (which are fairly common btw.) as read every time I use the app.

The reactions #6599 kinda sounds like I am not the only one that has this problem but I only shared this in case someone else needs this. I would of course prefer if this was merged so I don't need to keep rebasing my version but I will use this either way.

And yes, to make a Pull Request you do need to fork the repository.

arkon:
I'm fine with this in concept, but the idea of "latest chapter" isn't very well defined.

arkon:
#6916 would be more ideal.

PR #6794, opened 2022-03-19
Add "Move all chapters from series to top" option to download context menu

frankete333:
Added an option that allows you to move all the chapters from your download queue of a given series to the top.
Its a small chunk of code but any aditional changes are welcome.
I dont know how the translations are managed, i added the base value in english and in spanish, so any help with this is appreciated.


alessandrojean:
> i added the base value in english and in spanish, so any help with this is appreciated.

Just add the English one, the Spanish and the others will be added through Weblate later.

frankete333:
> > i added the base value in english and in spanish, so any help with this is appreciated.
> 
> Just add the English one, the Spanish and the others will be added through Weblate later.

Nice, thnks

app/src/main/res/values-es/strings.xml
```diff
@@ -421,6 +421,7 @@
     <string name="action_newest">Lo más nuevo</string>
     <string name="action_oldest">Lo más antiguo</string>
     <string name="action_move_to_top">Mover al principio</string>
+    <string name="action_move_to_top_all_for_series">Mover al principio toda esta serie</string>
```
arkon, line 424: Do you mind removing this? Non-English strings need to come from Weblate instead.

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadController.kt
```diff
@@ -366,6 +366,22 @@ class DownloadController :
                     }
                     presenter.reorder(newDownloads)
                 }
+                R.id.move_to_top_series -> {
+                    val newDownloads = mutableListOf<Download>()
+                    var counter = 0
+                    adapter?.currentItems
+                        ?.filterIsInstance<DownloadItem>()
+                        ?.map {
```
arkon, line 374: You should be able to use `partition` to get the items that are the same manga and those that aren't, then just reorder it as a list of those two concatenated (they ought to be in the right order).

arkon (changes requested):

PR #6858, opened 2022-04-03
Moved the main activity fab to the center horizontally.

Jeluxe:
Moved the main activity fab to the center horizontally.
Instead of adjusting only to right-handed or left-handed, in the center it can be adjusted to both ways, because we are all using either left or right hand from time to time.
So I think it can be more comfortable this way for all the users, and not only to right-handed.

studio64_rGZ7iXAiOI
studio64_uoMPfAVzIS
studio64_UmFV2kCxDa
studio64_R3C8jZ6oUe


arkon:
See https://github.com/tachiyomiorg/tachiyomi/issues/6844#issuecomment-1086875239

If there's a better idea around replacing the FAB entirely with something that works better at different screen sizes, then that should be what's considered.

PR #6861, opened 2022-04-03
feat: add read last read chapter shortcut

### LINKED ISSUE #6784: Double tap history button to read last read chapter

Double tapping history button will:
1) Open the last read chapter, if it's not read fully (like 4/5 pages read).
2) Open next chapter of the series, if the chapter was read fully.
3) Will move onto the next entry in history, if the series has chapter fully read and has no more chapters remaining.

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.13.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

Pierre-Monier:
Fixes #6784 

I add another shortcut, and a method on the HistoryController to open the ReaderActivity with the last read chapter

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
@@ -173,9 +173,9 @@ class HistoryController :
     private fun showResumeLastReadChapterErrorToast(position: Int) {
         val mangaName = getMangaName(position)
         val resResources = resources
-        
+
         if (mangaName != null && resResources != null) {
-            activity?.toast("${resResources.getString(R.string.no_last_read_chapter)} $mangaName")
+            activity?.toast(String.format(resResources.getString(R.string.no_last_read_chapter), mangaName))
         } else {
             activity?.toast(R.string.cant_open_last_read_chapter)
         }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.ui.recent.history
 
 import android.app.Dialog
+import android.content.Intent
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.Menu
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -705,6 +705,7 @@
     <string name="pref_clear_history">Clear history</string>
     <string name="clear_history_completed">History deleted</string>
     <string name="clear_history_confirmation">Are you sure? All history will be lost.</string>
+    <string name="cant_open_last_read_chapter">Unable to open the last read chapter</string>
```
arkon, line 708: It either opens an in-progress or the next unread chapter, so this is a bit misleading. It should be something like "No unread chapter for <manga name>".

arkon (changes requested):

Riztard:
link the issue to this pr btw

Pierre-Monier:
> link the issue to this pr btw

I can't link the issue manually, and I don't put any automatic link keyword on the description. I try to edit the description to add a "fix", which is a keyword that automatically link issue to pull request but it doesn't seems to work

*just a test*

Fix [this issue](https://github.com/tachiyomiorg/tachiyomi/issues/460)

app/src/main/res/values/strings.xml
Pierre-Monier: I put the "No unread chapter for" message if we can access the manga name, otherwise it displays "Unable to open last read chapter"

FourTOne5:
> > link the issue to this pr btw
> 
> I can't link the issue manually, and I don't put any automatic link keyword on the description. I try to edit the description to add a "fix", which is a keyword that automatically link issue to pull request but it doesn't seem to work

Change it to `Fixes #6784` instead of Fix this [issue](https://github.com/tachiyomiorg/tachiyomi/issues/6784)

app/src/main/res/values/strings.xml
```diff
@@ -705,6 +705,8 @@
     <string name="pref_clear_history">Clear history</string>
     <string name="clear_history_completed">History deleted</string>
     <string name="clear_history_confirmation">Are you sure? All history will be lost.</string>
+    <string name="cant_open_last_read_chapter">Unable to open last read chapter</string>
+    <string name="no_last_read_chapter">No unread chapter for</string>
```
arkon, line 709: You should use a placeholder in the string instead of concatenating the strings so that this can be localized properly.

arkon (changes requested):

app/src/main/res/values/strings.xml
Pierre-Monier: fix with last commit

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
```diff
@@ -157,19 +158,57 @@ class HistoryController :
...
+            showResumeLastReadChapterErrorToast(position)
+        }
+    }
+
+    private fun showResumeLastReadChapterErrorToast(position: Int) {
+        val mangaName = getMangaName(position)
+        val resResources = resources
+
+        if (mangaName != null && resResources != null) {
+            activity?.toast(String.format(resResources.getString(R.string.no_last_read_chapter), mangaName))
```
arkon, line 178: Don't use `String.format`; `getString` takes in the args directly.

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
```diff
@@ -157,19 +158,57 @@ class HistoryController :
...
+            showResumeLastReadChapterErrorToast(position)
+            resumeLastChapterRead(position = position + 1)
+        } else {
+            showResumeLastReadChapterErrorToast(position)
+        }
+    }
+
+    private fun showResumeLastReadChapterErrorToast(position: Int) {
+        val mangaName = getMangaName(position)
+        val resResources = resources
```
arkon, line 175: You can just use `activity`. You can do `val activity = activity ?: return` as the first line to avoid it being nullable when you use it.

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
Pierre-Monier: fix in last commit

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
Pierre-Monier: fix in last commit

arkon:
Sorry, you'll need to rebase on top of some other history controller changes. 😅

arkon:
Superseded by #7230

PR #6869, opened 2022-04-06
Move delete action to match placement in library_selection.xml

ghostbear:
Move delete action to far right in chapter_selection.xml and updates_chapter_selection.xml, for consistency with library_selection.xml


arkon:
The chapter menu items actually shift around more depending on what's selected, so I'm not entirely sure if this really improves it. I'll merge it anyway though.

PR #6876, opened 2022-04-07
Add option to disable navigation layout

### LINKED ISSUE #4820: [Feature Request] Option to completely disable navigation layout overlay.
If users want to change their reader mode "on the fly" (as in, using the sub-menu in a manga's reader), the navigation layout overlay is shown regardless if the option "show nagivation layout overlay" is enabled or disabled.

This becomes a problem if users need to change this often (i.e. they read lots of manga so have their 'default' optimised for this, but also read lots of webtoon and need to manually set that type of content to the 'webtoon' reading mode in each individual sub-menu). If so, they are constantly shown the navigation layout overlay as normal after every manual change.

## What/Requirements
Add a second toggle for "Show tap zones when reading mode is changed in reader".

manatails:
I added a navigation layout "Disabled" that doesn't contain any regions.

In many popular vertical reading apps like Naver Webtoon, navigation is entirely done by scrolling.
Users who are accustomed to such navigation system might get some unintentional page movements with current layouts, so I am proposing to add this option.

Closes #4820

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/navigation/DisabledNavigation.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/navigation/DisabledNavigation.kt
@@ -1,0 +1,18 @@
+package eu.kanade.tachiyomi.ui.reader.viewer.navigation
+
+import eu.kanade.tachiyomi.ui.reader.viewer.ViewerNavigation
+
+/**
+ * Visualization of default state without any inversion
+ * +---+---+---+
+ * | M | M | M |   P: Previous
+ * +---+---+---+
+ * | M | M | M |   M: Menu
+ * +---+---+---+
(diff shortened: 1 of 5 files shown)


CVIUS:
Just turn off tapping?

manatails:
> 

Ok, I just discovered about that option.

But there are too many problems with that I think.

1. It is difficult to find that option as it is only available in settings->reader menu not directly from the manga reader
2. Poor translation makes it difficult to tell what it actually does
3. it globally turns off all navigation layouts, you cannot set it individually
4. even when it is turned off manga reader still asks/shows which layout to use, it is confusing.

I think issue #4820 can also be addressed if we simply add "Disabled" as another layout rather than making it an extra option

arkon:
This seems like a reasonable replacement for the tapping option. I can remove that option afterwards in favor of users using this afterwards.

PR #6885, opened 2022-04-08
MangaCoverFetcher: Handle moving cover cache after adding to library

ivaniskandar:
Move cover cache to separate cache dir after the parent manga is added to library/favorited


app/src/main/java/eu/kanade/tachiyomi/data/coil/MangaCoverFetcher.kt
```diff
@@ -185,6 +181,48 @@ class MangaCoverFetcher(
...
+        } catch (e: Exception) {
+            logcat(LogPriority.ERROR, e) { "Failed to write snapshot data to cover cache ${cacheFile.name}" }
+            null
+        }
+    }
+
+    private fun writeResponseToCoverCache(response: Response, cacheFile: File?): File? {
+        if (cacheFile == null || !options.diskCachePolicy.writeEnabled) return null
+        return try {
+            response.peekBody(Long.MAX_VALUE).source().use { input ->
```
arkon, line 203: Why the cap?

ivaniskandar: That's to create a lightweight copy of the response body. Since it may need to be read again to write DiskCache.

PR #6893, opened 2022-04-10
feat(library-update): can update even if job is already running

Pierre-Monier:
Fix [this issue](https://github.com/tachiyomiorg/tachiyomi/issues/460). 

I change the `mangaToUpdate` from a `List` to a `Subject`. This way when we addMangaToQueue it's handle by the current running update function.

*Note that I'm new to Kotlin, and I'm not 100% sure that the async operation are handle the perfect way, but I'm sure it's working*


FourTOne5:
This approach can lead to Failed update notification being overwritten and same manga being updated twice leading to unnecessary calls. 

You should probably change it to how it used to work previously.. new updates are added to current job instead of making a queue.

arkon:
Closing this for now, I feel like the library update service needs a larger refactoring in general.

PR #6922, opened 2022-04-14
Migrate History screen to Compose

ghostbear:
- Migrate History screen to Compose
- Moves most logic from the Presenter into domain and data layer
- Setup for other screen being able to migrate to Compose with Theme


CarlosEsco:
So at what point does 1.x get abandoned? lol

app/src/main/java/eu/kanade/data/history/local/HistoryPagingSource.kt
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    override suspend fun load(params: LoadParams<Int>): LoadResult.Page<Int, MangaChapterHistory> {
+        val nextPageNumber = params.key ?: 0
+        logcat { "Loading page $nextPageNumber" }
+
+        val response = repository.getHistory(PAGE_SIZE, nextPageNumber, query)
+
+        val nextKey = if (response.size == 25) {
```
arkon, line 27:
```suggestion
        val nextKey = if (response.size == PAGE_SIZE) {
```

app/src/main/java/eu/kanade/data/history/repository/HistoryRepositoryImpl.kt
```diff
@@ -0,0 +1,137 @@
...
+import rx.schedulers.Schedulers
+import java.util.*
+
+class HistoryRepositoryImpl(
+    private val db: DatabaseHelper
+) : HistoryRepository {
+
+    /**
+     * Used to observe changes in the History table
+     * as RxJava isn't supported in Paging 3
```
arkon, line 24: Haven't looked at it yet, but I assume this goes away with your SQLDelight PR?

app/src/main/java/eu/kanade/data/history/repository/HistoryRepositoryImpl.kt
```diff
@@ -0,0 +1,137 @@
...
+            .observeChangesInTable(HistoryTable.TABLE)
+            .observeOn(Schedulers.io())
+            .subscribe {
+                pagingSource.invalidate()
+            }
+        return pagingSource
+    }
+
+    override suspend fun getHistory(limit: Int, page: Int, query: String) = coroutineScope {
+        withContext(Dispatchers.IO) {
```
arkon, line 44: Could probably just use `launchIO`?

app/src/main/java/eu/kanade/data/history/repository/HistoryRepositoryImpl.kt
```diff
@@ -0,0 +1,137 @@
...
+            // Set date limit for recent manga
```
arkon, line 45: We could add the JDK desugaring stuff if you want to use newer date/time APIs.

app/src/main/java/eu/kanade/data/history/repository/HistoryRepositoryImpl.kt
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    override suspend fun resetHistoryByMangaId(mangaId: Long): Boolean = coroutineScope {
+        withContext(Dispatchers.IO) {
+            try {
+                val history = db.getHistoryByMangaId(mangaId)
+                    .executeAsBlocking()
+                history.forEach { it.last_read = 0 }
+                db.upsertHistoryLastRead(history)
+                    .executeAsBlocking()
```
arkon, line 116: Could call the above `resetHistory`?

app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
```diff
@@ -0,0 +1,293 @@
...
+@Composable
+fun HistoryHeader(
+    modifier: Modifier = Modifier,
+    date: Date,
+    relativeTime: Int,
+    dateFormat: DateFormat,
+) {
+    Text(
+        modifier = modifier
+            .padding(horizontal = 16.dp, vertical = 8.dp),
```
arkon, line 175: We should probably declare the horizontal value as a constant somewhere so we can easily change handle larger paddings for tablets later.

app/src/main/res/layout/history_controller.xml
```diff
@@ -1,33 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
-<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
-    xmlns:tools="http://schemas.android.com/tools"
+<androidx.compose.ui.platform.ComposeView xmlns:android="http://schemas.android.com/apk/res/android"
```
arkon, line 2: Could probably just rename this to something more generic since we'll need the same thing for other things.

app/src/main/res/values/strings.xml
```diff
@@ -701,6 +701,7 @@
     <string name="updating_library">Updating library</string>
 
     <!-- History fragment -->
+    <string name="history_prefix">Ch.</string>
```
arkon, line 704: I'm not entirely sure if there's locales where this wouldn't be a prefix for whatever reason. Would it be safer to have something like `Ch. %1$s -`?

gradle/compose.versions.toml
```diff
@@ -0,0 +1,10 @@
+[versions]
+compose = "1.2.0-alpha07"
+
+[libraries]
+activity = "androidx.activity:activity-compose:1.4.0" # unused
```
arkon, line 5: Is this needed yet then?

app/src/main/java/eu/kanade/data/history/local/HistoryPagingSource.kt
ghostbear: This whole classes get yeeted in SQLDelight PR

app/src/main/java/eu/kanade/data/history/repository/HistoryRepositoryImpl.kt
ghostbear: Yes, it gets replaced the SQLDelight PR

app/src/main/java/eu/kanade/data/history/repository/HistoryRepositoryImpl.kt
ghostbear: Is replaced in SQLDelight PR

app/src/main/java/eu/kanade/data/history/repository/HistoryRepositoryImpl.kt
ghostbear: This data limit is replaced in SQLDelight PR as it seems to only be a filter to remove `last_read` that is set to `0`

app/src/main/java/eu/kanade/data/history/repository/HistoryRepositoryImpl.kt
ghostbear: Is replaced by a single database update query in SQLDelight PR

arkon:
This is a good start so I'll merge it for now. There seems to be some issues around the appbar elevation state and general performance, but we can iron that out after.

PR #6933, opened 2022-04-15
Migrate History screen database calls to SQLDelight

ghostbear:
To be merged after #6922

- Move a few queries over to SQLDelight from History table
- Move all migrations to SQLDelight
- Move all tables to SQLDelight

Related #6951
- [x]  Fixed eviscerate ripple
- [x]  Fixed padding of reset checkbox option in remove dialog


ghostbear:
Happy reviewing ヾ(≧▽≦*)o

app/src/main/java/eu/kanade/data/TransactionContext.kt
```diff
@@ -0,0 +1,160 @@
...
+internal suspend fun AndroidDatabaseHandler.getCurrentDatabaseContext(): CoroutineContext {
+    return coroutineContext[TransactionElement]?.transactionDispatcher ?: queryDispatcher
+}
+
+/**
+ * Calls the specified suspending [block] in a database transaction. The transaction will be
+ * marked as successful unless an exception is thrown in the suspending [block] or the coroutine
+ * is cancelled.
+ *
+ * Room will only perform at most one transaction at a time, additional transactions are queued
```
arkon, line 29: Should probably update the various comments about Room.

app/src/main/java/eu/kanade/data/history/HistoryMapper.kt
```diff
@@ -0,0 +1,24 @@
+package eu.kanade.data.history
+
+import eu.kanade.domain.history.model.History
+import eu.kanade.domain.history.model.HistoryWithRelations
+
+val historyMapper: (Long, Long, Long?, Long?) -> History = { id, chapterId, readAt, _ ->
```
arkon, line 6: What's the point of the last unused arg?

app/src/main/java/eu/kanade/data/AndroidDatabaseHandler.kt
```diff
@@ -0,0 +1,94 @@
...
+
+    override fun <T : Any> subscribeToOne(block: Database.() -> Query<T>): Flow<T> {
+        return block(db).asFlow().mapToOne(queryDispatcher)
+    }
+
+    override fun <T : Any> subscribeToOneOrNull(block: Database.() -> Query<T>): Flow<T?> {
+        return block(db).asFlow().mapToOneOrNull(queryDispatcher)
+    }
+
+    fun <T : Any> subscribeToPagingSource(
```
arkon, line 64: Should we just add this to the base interface too? We'd likely end up using this a lot throughout the app and we could try to refactor it again if we decide to make it multiplatform later.

app/src/main/java/eu/kanade/data/manga/MangaMapper.kt
```diff
@@ -0,0 +1,26 @@
+package eu.kanade.data.manga
+
+import eu.kanade.domain.manga.model.Manga
+
+val mangaMapper: (Long, Long, String, String?, String?, String?, String?, String, Long, String?, Boolean, Long?, Long?, Boolean, Long, Long, Long, Long) -> Manga =
+    { id, source, url, artist, author, description, genre, title, status, thumbnailUrl, favorite, lastUpdate, _, initialized, viewer, chapterFlags, coverLastModified, dateAdded ->
```
arkon, line 6: Same here with the unused arg.

app/src/main/java/eu/kanade/domain/history/model/History.kt
```diff
@@ -0,0 +1,7 @@
+package eu.kanade.domain.history.model
+
+data class History(
+    val id: Long?,
+    val chapterId: Long,
+    val readAt: Long
```
arkon, line 6: Maybe we should use actual `Date`/`Instant`s or something? I guess that'd make the migrations difficult though.

app/src/main/java/eu/kanade/domain/history/model/HistoryWithRelations.kt
```diff
@@ -0,0 +1,11 @@
+package eu.kanade.domain.history.model
+
+data class HistoryWithRelations(
```
arkon, line 3: Would it make sense to extend `History`?

app/src/main/java/eu/kanade/domain/history/model/History.kt
```diff
@@ -0,0 +1,7 @@
+package eu.kanade.domain.history.model
+
+data class History(
```
arkon, line 3: Is this even used?

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -0,0 +1,36 @@
...
+    val dateAdded: Long,
+    val viewerFlags: Long,
+    val chapterFlags: Long,
+    val coverLastModified: Long,
+    val url: String,
+    val title: String,
+    val artist: String?,
+    val author: String?,
+    val description: String?,
+    val genre: String?,
```
arkon, line 17: How constrained are we in staying with the existing DAO models? Would this be a good time to do some changes like making this a list?

app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
```diff
@@ -215,23 +216,23 @@ fun HistoryItem(
...
                             append(
                                 stringResource(
                                     R.string.history_prefix,
-                                    chapterFormatter.format(history.chapter.chapter_number)
+                                    chapterFormatter.format(history.chapterNumber)
                                 )
                             )
                         }
-                        append(Date(history.history.last_read).toTimestampString())
+                        append(Date(history.readAt ?: 0).toTimestampString())
```
arkon, line 235: Would this ever actually be a valid thing to show?

app/src/main/java/eu/kanade/tachiyomi/data/database/queries/HistoryQueries.kt
```diff
@@ -4,39 +4,11 @@ import com.pushtorefresh.storio.sqlite.queries.DeleteQuery
...
 import eu.kanade.tachiyomi.data.database.models.History
-import eu.kanade.tachiyomi.data.database.models.MangaChapterHistory
 import eu.kanade.tachiyomi.data.database.resolvers.HistoryUpsertResolver
-import eu.kanade.tachiyomi.data.database.resolvers.MangaChapterHistoryGetResolver
 import eu.kanade.tachiyomi.data.database.tables.HistoryTable
-import java.util.Date
 
 interface HistoryQueries : DbProvider {
 
-    /**
```
arkon, line 15: I assume it's okay that we're still interfacing with the same data with both SQLDelight and this old code so long as the schemas match up?

app/src/main/java/eu/kanade/tachiyomi/data/database/DbOpenCallback.kt
```diff
@@ -20,81 +14,12 @@ class DbOpenCallback : SupportSQLiteOpenHelper.Callback(DATABASE_VERSION) {
...
     }
 
-    override fun onCreate(db: SupportSQLiteDatabase) = with(db) {
-        execSQL(MangaTable.createTableQuery)
-        execSQL(ChapterTable.createTableQuery)
-        execSQL(TrackTable.createTableQuery)
-        execSQL(CategoryTable.createTableQuery)
-        execSQL(MangaCategoryTable.createTableQuery)
-        execSQL(HistoryTable.createTableQuery)
+    override fun onCreate(db: SupportSQLiteDatabase) {} // SQLDelight handles this
```
arkon, line 20: May have been nice to have had moved all this stuff in a separate PR first.

arkon (commented):
Initial set of comments. I haven't actually reviewed the moved migration SQL yet.

app/src/main/java/eu/kanade/data/history/HistoryMapper.kt
ghostbear: When querying `SELECT * FROM history` the mapper is required to have an arg for each field in the table but as the `time_read` is not implemented I choose not to include it in the domain model and that is why it's unused here

app/src/main/java/eu/kanade/data/manga/MangaMapper.kt
ghostbear: Similar to what I commented on the history mapper this field is not used anymore/is deprecated from what understand but when querying `SELECT * FROM mangas` it's required to have all fields from the table

app/src/main/java/eu/kanade/domain/history/model/History.kt
ghostbear: Will look into a `ColumnAdapter` for this that way the coulmn should be able to remain as a `Integer` but SQLDelight returns it as a `Date`

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
ghostbear: If we use `ColumnAdapter` to convert it between a string (when stored) and a list (when used) it shouldn't interfere with the existing DAO models. But if we want to change things in the database we most likely need to update the existing DAO models

app/src/main/java/eu/kanade/tachiyomi/data/database/queries/HistoryQueries.kt
ghostbear: Yeah, as long as the database version and schemas match up it should be fine

app/src/main/java/eu/kanade/domain/history/model/History.kt
ghostbear: Not right now but it will most likely be in the future

app/src/main/java/eu/kanade/domain/history/model/HistoryWithRelations.kt
ghostbear: `data class` are final and can't be extended from. Would require an `interface` for both `data class` then

app/src/main/sqldelight/view/historyView.sq
```diff
@@ -0,0 +1,45 @@
+CREATE VIEW historyView AS
```
ghostbear, line 1: This is now a view as the `HistoryWithRelations` queries made it so SQLDelight wasn't able to generate an Adapter for History which is needed for the DateColumnAdapter

app/src/main/java/eu/kanade/tachiyomi/App.kt
```diff
@@ -77,6 +79,12 @@ open class App : Application(), DefaultLifecycleObserver, ImageLoaderFactory {
         Injekt.importModule(AppModule(this))
         Injekt.importModule(DomainModule())
 
+        val database = Injekt.get<SqlDriver>()
+
+        val query = database.executeQuery(1, "PRAGMA schema_version;", 0)
+        query.next()
+        Log.d("Application", "Current database version is ${query.getLong(0)}")
```
arkon, line 86: Remove or use `logcat`?

arkon (approved):
Seems to work now. 👍

PR #6947, opened 2022-04-16
Add elevation to navigation rails

CrepeTF:
Makes navigation rail full height and adds elevation

uwu ghostbear-senpai owo


CrepeTF:
Worked fine on Tako.





Probably.

PR #6971, opened 2022-04-21
Remove resume button

CrepeTF:
Makes the whole item the resume button, and you press the cover to get to manga info

### Images
| Before | After |
| ------- | ------- |
| crinj old item with resume button | chad no dumb play button |


app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
```diff
@@ -197,13 +193,15 @@ fun HistoryItem(
...
+            .clickable(onClick = onClickResume)
             .height(96.dp)
             .padding(horizontal = horizontalPadding, vertical = 8.dp),
         verticalAlignment = Alignment.CenterVertically,
     ) {
         MangaCover(
-            modifier = Modifier.fillMaxHeight(),
+            modifier = Modifier
+                .fillMaxHeight()
+                .clickable(onClick = onClickItem),
```
arkon, line 204: Should probably rename these.

CrepeTF:
Is that good?

CrepeTF:
I have no fucken clue how git works

Issue #3079, opened 2020-05-05
[Bug] Chapter transitions not skipped for single-page series


Kabutak:
### Device information
* Tachiyomi version: Stable 0.9.2 (existed on Stable 0.9.0 as well)
* Android version: 7.0
* Device: Samsung Galaxy S5 Neo

## Steps to reproduce
1. Start reading a downloaded series which has only one page per chapter (e.g. Cyanide and Happiness or another webcomic)
2. Read through several chapters (with Always show chapter transitions off)

### Expected behavior
Chapter transitions should be skipped, since each page is downloaded and loaded before changing to the next one.

### Actual behavior
The transition is skipped the first time, but after that will not be skipped unless the user returns to the chapter list and hits resume again, upon which it will again skip only one transition. It also skips if you flip between two chapters after the initial transition, but that is intended behaviour and unrelated.

### Other details
Related to #2728, which was resolved previously, but this is a different issue.

It is likely a race condition or similar related to the fact that the first page is also the last page, so it may be skipping preloading or another trigger related to the transitions, but only when already reading. When resuming it seems to properly detect that it is the last page and it should begin to load the next chapter. The check is probably in the actual chapter transition itself.

arkon:
I dug a little bit into it. It seems like the next (or previous) chapters aren't yet in the state of `ReaderChapter.State.Loaded` since they're not preloaded.

Haven't actually figured out how to handle that though.

scb261:
I hope that I understood it correctly (wanted to try to write a fix, but I don't fully understand the code). Took several hours to download and launch all this stuff. I will write about https://github.com/inorichi/tachiyomi/blob/dev/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt , but I guess it also applies to WebtoonViewer.kt

When first (and at the same time the only one) page is loaded, at the line 163 expression `page.chapter == adapter.currentChapter` is `false`. Not sure what it does (checks if it's still the same chapter?), but it prevents preload.

When I removed this part (so it would be only `inPreloadRange && allowPreload`) and looked at the variables, I understood that the one that we want to preload is next to the next chapter (since the one that will appear on screen right now is considered next for now). And so removing `page.chapter == adapter.currentChapter` doesn't help because it just tries to preload next one twice.

I think, at the line 168 you need to add next code:
```
else if ([next chapter page (similar to line 147)] && [nextChapter.numberOfPages == 1]) {
[...]

PR #6998, opened 2022-04-24
Fix chapter transition setting for one page chapters

ItsLogic:
Fixes the chapter transition setting on one page chapters (#3079) Only impacts paged reading methods so webtoon still has the same behaviour. I didnt notice any obvious bugs with this implentation (it is only a single if statement so there shouldnt be anything)

### Video

https://user-images.githubusercontent.com/38233332/164960345-f7d35896-2629-465b-83cf-fa73723cdd46.mp4


BartTerpstra:
works for right to left, left to right, vertical.
does not seem to affect webtoon and continious vertical (this is also not required)

Issue #6951, opened 2022-04-17
Compose rewrite issues


arkon:
This is a tracking issue for various problems with the Jetpack Compose rewrite.

General:
- [x] No fast scroll
- [ ] Different ripple color
- [ ] Performance issues
- [x] No appbar elevation lift on content scroll
- [x] Inconsistent incognito/downloaded only banner visibility
- [x] Inconsistent controller/screen/scaffolds
- [ ] Missing range selection in library
- [ ] No tooltips (e.g., appbar actions; https://issuetracker.google.com/issues/171674404)

History:
- [x] Date grouping is repeated
- [x] Appbar elevation isn't reset when scrolled back to the top
- [x] General performance issues (should look into `@Stable`/`@Immutable` usage)
- [x] Padding of reset checkbox option in remove dialog
- [x] Broken/failed to load covers don't show the placeholder
- [x] Jumping to next section when an item is deleted
- [x] Nav bar inset padding needs to be adjusted
- [x] History/library data being lost on SQL upgrade for some users

CrepeTF:
Eviscerate ripple

![Screenshot_20220417-174710_Tachibear.png](https://user-images.githubusercontent.com/70870719/163724393-bb01b533-75ad-421f-bcb6-bad9bc62850b.png)

ghostbear:
> * [ ]  Jumping to next section when an item is deleted

Might be related to https://github.com/cashapp/sqldelight/issues/2591 and/or https://github.com/cashapp/sqldelight/issues/2434 as it also crashes when deleting on last page

stevenyomi:
Currently having a bunch of issues related to missing state updates:
- #7418
- #7834
- #7866

arkon:
Breaking out the library selection issue to #8071.

The other remaining ones are either already individual issues are broad enough that we'll address them when possible (e.g. tooltip support).

PR #7022, opened 2022-04-27
Add error state to MangaCover composable 

ghostbear:
- Add error drawable when url isn't able to be loaded
- Tweak usage of MangaCover

Related to #6951


app/src/main/java/eu/kanade/presentation/components/MangaCover.kt
```diff
@@ -11,29 +11,31 @@ import androidx.compose.ui.graphics.painter.ColorPainter
...
-            .aspectRatio(aspect.ratio)
-            .clip(shape),
-        contentScale = ContentScale.Crop
-    )
+    @Composable
+    operator fun invoke(
+        modifier: Modifier = Modifier,
+        data: String?,
+        contentDescription: String = "",
+        shape: Shape? = null
```
arkon, line 26: Why not keep the default here?

ghostbear: Quirkiness of the invoke operator, and probably a bug. Had `shape` return a default value but it became null even though it had a default value. I just realized `contentDescription` is actually also null by default even though it has `""`

PR #7047, opened 2022-04-30
Notify user of error during a page download failure

saud-97:
If a page download faced an error it will skip the page but without notifying the user of what is the error faced. Eventually the chapter download status will be ERROR status but no way the user will be able to know what happened.

### To reproduce the bug
Donwload a chapter that has some pages with dead links to get http error 404

### This is how the notification looks

image


app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -401,6 +401,7 @@ class Downloader(
             .onErrorReturn {
                 page.progress = 0
                 page.status = Page.ERROR
+                notifier.onError(it.message, download.chapter.name)
```
arkon, line 404: May want to have the manga title too, similar to the progress notification.

saud-97: Done

PR #7052, opened 2022-05-01
Properly show history state

FourTOne5:
And also make `HistoryState` similar to `MigrateState`


app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
```diff
@@ -92,3 +94,8 @@ class HistoryController : ComposeController<HistoryPresenter>(), RootController
         }
     }
 }
+
+sealed class HistoryUiModel {
```
arkon, line 98: We should probably be consistent with where we put these. Probably makes the most sense to be by the composable?

app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
```diff
@@ -66,37 +67,34 @@ fun HistoryScreen(
...
-            CircularProgressIndicator()
-        }
-        history.itemCount == 0 -> {
-            EmptyScreen(
-                textResource = R.string.information_no_recent_manga
-            )
-        }
-        else -> {
+    when (state) {
+        is HistoryState.Loading -> CircularProgressIndicator()
```
arkon, line 71: Should this be a `LoadingScreen` instead?

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
```diff
@@ -41,39 +42,43 @@ class HistoryPresenter(
 ) : BasePresenter<HistoryController>() {
 
     private var _query: MutableStateFlow<String> = MutableStateFlow("")
-    private var _state: MutableStateFlow<HistoryState> = MutableStateFlow(HistoryState.EMPTY)
-    val state: StateFlow<HistoryState> = _state
+    private var _state: MutableStateFlow<HistoryState> = MutableStateFlow(HistoryState.Loading)
```
arkon, line 45: This can probably be a `val` actually.

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryController.kt
FourTOne5: Yeah makes sense

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
```diff
@@ -40,40 +42,44 @@ class HistoryPresenter(
...
+                getHistory.subscribe(query)
+                    .cachedIn(presenterScope)
+                    .catch { exception ->
+                        _state.emit(HistoryState.Error(exception))
+                    }
+                    .map { pagingData ->
+                        pagingData.toHistoryUiModels()
+                    }.let { uiModelsPagingDataFlow ->
+                        _state.emit(HistoryState.Success(uiModelsPagingDataFlow))
                     }
```
ghostbear, line 63: `cachedIn` should cache the transformation
```suggestion
            _query.collectLatest { query ->
                getHistory.subscribe(query)
                    .catch { exception ->
                        _state.emit(HistoryState.Error(exception))
                    }
                    .map { pagingData ->
                        pagingData.toHistoryUiModels()
                    }
                    .cachedIn(presenterScope)
                    .let { uiModelsPagingDataFlow ->
                        _state.emit(HistoryState.Success(uiModelsPagingDataFlow))
                    }
```

PR #7053, opened 2022-05-01
Make `SourceState` similar to `MigrateState`

FourTOne5:


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/SourcePresenter.kt
```diff
@@ -28,23 +28,24 @@ class SourcePresenter(
...
+                .collectLatest { sourceList ->
+                    val uiModels = sourceList.toSourceUiModels()
+                    _state.emit(SourceState.Success(uiModels))
                 }
-                .collectLatest(::collectLatestSources)
         }
     }
 
-    private fun collectLatestSources(sources: List<Source>) {
+    private fun List<Source>.toSourceUiModels(): List<SourceUiModel> {
```
ghostbear, line 48: Why the change to an extension function?

FourTOne5: just felt like it...

ghostbear: You could make `collectLatestSources` a suspend function rather than making an extension function. As I guess this change is in part with having to use `_state.emit`

PR #7062, opened 2022-05-02
Add option to reset viewer flags

### LINKED ISSUE #3277: "Reset reader settings for individual manga" option in Settings > Advanced

FourTOne5:
Closes #3277


app/src/main/java/eu/kanade/data/manga/MangaRepositoryImpl.kt
```diff
@@ -12,4 +13,14 @@ class MangaRepositoryImpl(
     override fun getFavoritesBySourceId(sourceId: Long): Flow<List<Manga>> {
         return databaseHandler.subscribeToList { mangasQueries.getFavoriteBySourceId(sourceId, mangaMapper) }
     }
+
+    override suspend fun resetViewerFlags(): Boolean {
+        return try {
+            databaseHandler.await { mangasQueries.resetViewerFlags() }
+            true
+        } catch (e: Exception) {
+            logcat(throwable = e)
```
arkon, line 22: This should actually have the level set to error. I missed that in the other PRs.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedController.kt
```diff
@@ -309,6 +320,16 @@ class SettingsAdvancedController : SettingsController() {
             activity?.toast(R.string.cache_delete_error)
         }
     }
+
+    private fun resetViewerFlags() {
+        if (activity == null) return
```
arkon, line 325: `val activity = activity ?: return`

Then it won't be nullable below.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedController.kt
```diff
@@ -309,6 +320,16 @@ class SettingsAdvancedController : SettingsController() {
...
+        val isSuccesful = runBlocking { mangaRepository.resetViewerFlags() }
```
arkon, line 326: May want to run this on the IO thread instead.

arkon (commented):
Would it make sense to reset the default settings too, or have the resetting respect what's actually set as default?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedController.kt
```diff
@@ -276,37 +288,52 @@ class SettingsAdvancedController : SettingsController() {
...
+            activity.toast(R.string.cache_delete_error)
+        }
+    }
+
+    private fun resetViewerFlags() {
+        val activity = activity ?: return
+        launchIO {
+            val isSuccesful = mangaRepository.resetViewerFlags()
+            withUIContext {
+                val resouurceString = if (isSuccesful) {
```
arkon, line 330: Typo

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedController.kt
```diff
@@ -276,37 +288,52 @@ class SettingsAdvancedController : SettingsController() {
...
             webview.setDefaultSettings()
             webview.clearCache(true)
             webview.clearFormData()
             webview.clearHistory()
             webview.clearSslPreferences()
             WebStorage.getInstance().deleteAllData()
-            activity?.applicationInfo?.dataDir?.let { File("$it/app_webview/").deleteRecursively() }
-            activity?.toast(R.string.webview_data_deleted)
+            activity.applicationInfo?.dataDir?.let { File("$it/app_webview/").deleteRecursively() }
+            activity.toast(R.string.webview_data_deleted)
```
arkon, line 318: It actually may be possible for the activity to be gone by this point since it does take a bit of time to delete the data.

app/src/main/res/values/strings.xml
```diff
@@ -483,6 +483,10 @@
     <string name="pref_refresh_library_covers">Refresh library manga covers</string>
     <string name="pref_refresh_library_tracking">Refresh tracking</string>
     <string name="pref_refresh_library_tracking_summary">Updates status, score and last chapter read from the tracking services</string>
+    <string name="pref_reset_viewer_flags">Reset Viewer Flags</string>
+    <string name="pref_reset_viewer_flags_summary">Resets reading mode and orientation of all series</string>
+    <string name="pref_reset_viewer_flags_succesful">Viewer flags reset was succesful</string>
```
arkon, line 488: The "was succesful" part is redundant. Also, it's "successful".

app/src/main/res/values/strings.xml
```diff
@@ -483,6 +483,10 @@
     <string name="pref_refresh_library_covers">Refresh library manga covers</string>
     <string name="pref_refresh_library_tracking">Refresh tracking</string>
     <string name="pref_refresh_library_tracking_summary">Updates status, score and last chapter read from the tracking services</string>
+    <string name="pref_reset_viewer_flags">Reset Viewer Flags</string>
```
arkon, line 486: Don't capitalize each word

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsAdvancedController.kt
FourTOne5: Even if it's stored in another variable?

arkon: I'm not entirely sure what happens in the case that it's referencing a dead activity.

PR #7063, opened 2022-05-02
Fix removing manga from library reverts during global update

### LINKED ISSUE #3281: [Bug] Removing manga from library reverts while global update is ongoing
### Device information
* Tachiyomi version: 0.9.2-2373
* Android version: 10 (SDK 29)
* Device: Mi 9T Pro

## Steps to reproduce
1. Tap global update in library
2. Select one manga and remove it from library while update is ongoing

### Expected behavior
Removed from library

### Actual behavior
After removing from library, wait for a moment and it will indicate again `In library`. When you tap again it will display a dialog of categories (do not select anything, just ignore). When you go back to the library, the manga that you remove was there and when you select it again, it was `In library`.

FourTOne5:
Fixes #3281


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -174,7 +174,7 @@ class LibraryUpdateService(
      */
     override fun onDestroy() {
         updateJob?.cancel()
-        ioScope?.cancel()
+        ioScope.cancel()
```
arkon, line 177: This has actually been null and caused crashes before, despite what the lifecycles would lead you to believe.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -415,8 +413,12 @@ class LibraryUpdateService(
                 sManga.thumbnail_url = manga.thumbnail_url
             }
 
-            manga.copyFrom(sManga)
-            db.insertManga(manga).executeAsBlocking()
+            // Get manga from database to account for if it was removed from library or database
+            val dbManga = db.getManga(manga.id!!).executeAsBlocking()
```
arkon, line 417: Can just return empty lists here rather than continuing to get the chapters.

Should also:
- Do a short-circuit at the beginning of `updateManga` to avoid updating at all if it's no longer in the library at that point
- Just check prior to calling `syncChaptersWithSource`, since there's no point in updating the chapter data if it's not in the library anymore

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -305,39 +305,42 @@ class LibraryUpdateService(
...
+                                            return@withUpdateNotification
+                                        }
+                                        if (MANGA_HAS_UNREAD in restrictions && mangaWithNotif.unreadCount != 0) {
+                                            skippedUpdates.add(mangaWithNotif to getString(R.string.skipped_reason_not_caught_up))
+                                            return@withUpdateNotification
+                                        }
+                                        if (MANGA_NON_READ in restrictions && mangaWithNotif.totalChapters > 0 && !mangaWithNotif.hasStarted) {
+                                            skippedUpdates.add(mangaWithNotif to getString(R.string.skipped_reason_not_started))
+                                            return@withUpdateNotification
+                                        }
```
arkon, line 328: Should probably still categorize the entries for everything it fails to pass instead of short-circuiting at each one.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -445,16 +461,16 @@ class LibraryUpdateService(
                                     currentlyUpdatingManga,
                                     progressCount,
                                     manga,
-                                ) { manga ->
-                                    sourceManager.get(manga.source)?.let { source ->
+                                ) { mangaWithNotif ->
```
arkon, line 464: Try to avoid having so many unrelated changes in a PR in the future so that they're easier/quicker to review.

FourTOne5: i'll keep that in mind

PR #7069, opened 2022-05-03
make daily backup default for fresh installs

OncePunchedMan:
It will ask for storage permission at start up (Library, since default start screen) instead of in Browse and sets the automatic backup to daily and will set the default backup path

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
@@ -1,5 +1,6 @@
 package eu.kanade.tachiyomi.ui.library
 
+import android.Manifest.permission.WRITE_EXTERNAL_STORAGE
 import android.content.res.Configuration
 import android.os.Bundle
 import android.view.LayoutInflater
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
```diff
@@ -169,6 +171,7 @@ class LibraryController(
 
     override fun onViewCreated(view: View) {
         super.onViewCreated(view)
+        requestPermissionsSafe(arrayOf(WRITE_EXTERNAL_STORAGE), 301)
```
ghostbear, line 174: Would probably be better to have this in the MainActivity since any screen can be opened with App shortcuts

FourTOne5:
I think you will also have to start backupJob

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
OncePunchedMan: hm true. but you'd need to choose that which in my opinion would be a very rare case

OncePunchedMan:
> I think you will also have to start backupJob

an empty backup or what? the process is already running in the background

FourTOne5:
From backup settings
```kt
onChange { newValue ->
    val interval = (newValue as String).toInt()
        BackupCreatorJob.setupTask(context, interval)
    true
}
```

arkon:
As discussed on Discord, it'd make more sense to try to save backups in a location that doesn't require explicit storage permissions (such as the Documents folder). I'm going to close this, but feel free to reopen/open another one if you decide to make that change.

PR #7075, opened 2022-05-04
fix downloader stuck as foreground service

shadow578:
After updating to Android 12 (or maybe it was with a app update, can't really tell) i started getting a persistent "App is running in backgrund" notification, and had to force- stop tachiyomi to get rid of it. 
As it seems, this is because the download service is started as a background service, and gets stuck when stopping downloads due to no suitable network being available. With my change, the download service is stopped in this case. 

| Notification: Tachiyomi running in background | 
| ------- | 
| tachi-running-in-bg | 

---

Note: I'm currently testing this change and will mark the PR ready if testing goes well

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
@@ -82,8 +82,8 @@ class DownloadManager(
      *
      * @param reason an optional reason for being stopped, used to notify the user.
      */
-    fun stopDownloads(reason: String? = null) {
-        downloader.stop(reason)
+    fun stopDownloads(reason: String? = null, suppressCompletionNotification: Boolean = false) {
+        downloader.stop(reason, suppressCompletionNotification)
     }
 
     /**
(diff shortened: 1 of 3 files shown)


arkon:
I think this happened before Android 12 too, and it also happens when a progress change hasn't occurred yet (e.g., you start downloading a chapter but the first page takes a long time to complete; it'll get replaced with the normal progress notification after that).

saud-97:
I can confirm, I have android 9 and noticed this persistent notification in multiple forks. Also, I used emulator with ethernet connection so this is most likely caused by source issue not related to 

> no suitable network being available

arkon:
Feel free to open another PR if you decide to tackle it again, but I feel like the downloader notification logic needs cleaning up in general.

shadow578:
Hi,
sorry if i'm misunderstanding, but why was the PR closed? Is there something wrong with the changes i suggested? 

> I feel like the downloader notification logic needs cleaning up in general

I do agree on that, but thus far my changes seem to have fixed the issue i was trying to solve. It's just that testing if everything works as expected takes quite long since i had to wait for actual manga to update to reproduce the issue. 

Please let me know how to proceed (or where the issue was) so i can learn for future PRs

arkon:
This won't really fix the placeholder notification being shown in other cases. I'd rather there be a more significant overhaul to the logic than trying to monkey patch individual edge cases.

PR #3296, opened 2020-06-08
steady chapter downloads feature

lmj0011:
closes https://github.com/inorichi/tachiyomi/issues/3295

PR #7091, opened 2022-05-07
Auto download new chapters : Limit and skip titles with unread chapters

nzoba:
- Changes the `Download new chapters` setting from a switch to an intList `Download new chapters until limit (per title)` to choose how many chapters you want to download per manga. This limit includes the already downloaded unread chapters of a manga. For example if the setting is `3 unread chapters`, and there already are 2 downloaded unread chapters for the manga, if 5 new chapters come only the first one will be downloaded.
- Adds a setting `Skip titles with unread chapter(s)` for `Download new chapters`, useful if you disable (temporarily or not) the `Skip titles with unread chapters` from library update or if you manually update a chapter from manga details


Riztard:
might as well add restrictions like wifi only?

nzoba:
I added all the restrictions from library update (only on wifi, charging and battery low)

FourTOne5:
Can you send a pic on how it looks like?

app/src/main/java/eu/kanade/tachiyomi/util/MangaExtensions.kt
```diff
@@ -56,8 +65,15 @@ fun Manga.shouldDownloadNewChapters(db: DatabaseHelper, prefs: PreferencesHelper
...
 
     // Boolean to determine if user wants to automatically download new chapters.
-    val downloadNewChapter = prefs.downloadNewChapter().get()
-    if (!downloadNewChapter) return false
+    val downloadNewChapter = prefs.downloadNewChapters().get()
+    if (downloadNewChapter == 0) return false
+
+    val restrictions = prefs.downloadNewDeviceRestriction().get()
+    if (DEVICE_ONLY_ON_WIFI in restrictions && !prefs.context.isConnectedToWifi()) return false
+    val batteryManager = prefs.context.getSystemService(Context.BATTERY_SERVICE) as BatteryManager
```
FourTOne5, line 73: In `app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt` add
```kt
val Context.batteryManager: BatteryManager
    get() = getSystemService()!!
``` 
and use that

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -326,7 +329,16 @@ class LibraryUpdateService(
...
-                                                        downloadChapters(mangaWithNotif, newChapters)
+                                                        val chaptersDatabase = db.getChapters(mangaWithNotif).executeAsBlocking()
+                                                            .map { ChapterItem(it, mangaWithNotif) }
+                                                        chaptersDatabase.filter {
+                                                            downloadManager.isChapterDownloaded(it, mangaWithNotif)
+                                                        }.forEach { it.status = Download.State.DOWNLOADED }
+
+                                                        val chaptersToDownload = mangaWithNotif.getChaptersToDownload(
+                                                            newChapters, chaptersDatabase, preferences
+                                                        )
```
FourTOne5, line 340: i think it'll be better to move the whole thing to `app/src/main/java/eu/kanade/tachiyomi/util/MangaExtensions.kt`

FourTOne5 (changes requested):

nzoba:
I juste changed `Download new chapters until limit (per title)` and added `Auto download device restrictions` + `Skip titles with unread chapter(s)`

drawing

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
nzoba: Not sure I understand what you mean, are you talking about `chaptersDatabase` and checking which chapters are downloaded ? If so, it's necessary only for `LibraryUpdateService` not for `MangaPresenter` since the downloaded status is already present, so it wouldn't make sense to move it to `MangaExtensions`

app/src/main/java/eu/kanade/tachiyomi/Migrations.kt
```diff
@@ -276,6 +276,11 @@ object Migrations {
                     else -> oldSortingMode
                 }
                 preferences.librarySortingMode().set(newSortingMode)
+
+                val downloadNew = prefs.getBoolean("download_new", false)
+                if (downloadNew) {
+                    preferences.downloadNewChapters().set(-1)
+                }
```
FourTOne5, line 283: Bump app version by 1 and move this to another `if` block

app/src/main/java/eu/kanade/tachiyomi/util/MangaExtensions.kt
```diff
@@ -56,8 +65,15 @@ fun Manga.shouldDownloadNewChapters(db: DatabaseHelper, prefs: PreferencesHelper
...
-    if (!downloadNewChapter) return false
+    val downloadNewChapter = prefs.downloadNewChapters().get()
+    if (downloadNewChapter == 0) return false
+
+    val restrictions = prefs.downloadNewDeviceRestriction().get()
+    if (DEVICE_ONLY_ON_WIFI in restrictions && !prefs.context.isConnectedToWifi()) return false
+    val batteryManager = prefs.context.batteryManager
+    if (DEVICE_CHARGING in restrictions && !batteryManager.isCharging) return false
+    val batteryLevel = batteryManager.getIntProperty(BATTERY_PROPERTY_CAPACITY)
+    if (DEVICE_BATTERY_NOT_LOW in restrictions && batteryLevel <= 15) return false
```
FourTOne5, line 76: These device restrictions makes it so that the new chapters doesn't even get added to Downloader and this is a bad Idea. It'll probably be better to make the restrictions work on the whole downloader and not here
2

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -326,7 +329,16 @@ class LibraryUpdateService(
 
                                                 if (newChapters.isNotEmpty()) {
                                                     if (mangaWithNotif.shouldDownloadNewChapters(db, preferences)) {
-                                                        downloadChapters(mangaWithNotif, newChapters)
+                                                        val chaptersDatabase = db.getChapters(mangaWithNotif).executeAsBlocking()
+                                                            .map { ChapterItem(it, mangaWithNotif) }
```
FourTOne5, line 333: Is there any need to map these as `ChapterItem` when you are already providing the manga?

FourTOne5 (changes requested):

app/src/main/java/eu/kanade/tachiyomi/util/MangaExtensions.kt
nzoba: Well I'm not sure, I think we want the restrictions only for the auto-download of new chapters not each download. The benefits of doing it here is that this method is only for the new chapters. And do we really want to add theses chapters in the Downloader if we don't want to auto-download in these cases ?

FourTOne5: Just because my device is not connected to wifi doesn't mean it won't be connected to one sometime later. Not adding the new chapters to download just because a device isn't connected to wifi will confuse people using this. Same for the battery related restrictions

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
nzoba: Yes, if we don't map to `ChapterItem` we can't set the status parameter (from `BaseChapterItem`). I need it to know how many chapters are already downloaded.
I could probably get this count before and pass it to the method but I'm not sure it would be better since I would need to do it 2 times (for `LibraryUpdate` and `MangaPresenter`)

app/src/main/java/eu/kanade/tachiyomi/util/MangaExtensions.kt
nzoba: Oh yeah I didn't think about that, I thought about this as `disable auto-download when` not as `wait for`. The problem is that I'm not sure we could easily differentiate a manual from an auto download once they are in the downloads queue. And we may want to have different restrictions for them

I see 2 "easy" possible solutions :
1.  Rename the restriction setting to something like `Disable auto-download when`
2. Or have the same restrictions for manual and auto download (do we want that ?)

A complex solution would be to differentiate in the downloads queue but it would need to, as far as I know, either have 2 queues or change how the queue work to know the type of download. Not sure we want that and if so I would probably not do it in this PR, it's far too much work for a single PR

jferdous70:
I believe this supersedes https://github.com/tachiyomiorg/tachiyomi/pull/3296 so that one can be closed

FourTOne5:
> I believe this supersedes #3296 so that one can be closed

it does not both are separate things

jferdous70:
> > I believe this supersedes #3296 so that one can be closed
> 
> it does not both are separate things

sorry, I misread.

nzoba:
I fixed the conflicts and added an `isAutoAndRestricted` boolean to differentiate a stucked auto download from other downloads. So if an auto download is added to the queue and the restrictions are not met, the auto download won't start and an `AUTO` tag will be visible. If restrictions are met it will be converted to a normal download

You can convert the auto download to a normal one or force start it manually.
The check for the restrictions is only done when you start downloading, you add a new download or the only remaining downloads are stucked auto

| Download queue | Download notification |
| --- | --- |
| Download queue | Download notification |

Riztard:
hmm confict

Ajxrhaider:
Is it possible for the download ahead limit to be increased or customized

PR #7098, opened 2022-05-08
Fixed splitting logic and CBZ compression logic

saud-97:
### Changes
- Moved splitting logic to happen inside `getOrDownloadImage`
- Added a condition to check and skip splitting if a splitting was already done before on the page - to fix downloads geting stuck on pause.
- Modified error handling for splitting to handle OOM error caused by `BitmapFactory.decodeFile` in case of extremely large images
- Allowed the download to continue and skip the splitting ( and show an error notification) if splitting failed instead of throwing an error and terminating the download.
- Moved CBZ logic to avoid marking a download as complete before the CBZ compression was completed

### Side note:
I only cought the exception for OOM. However the solution to fix OOM is implemented in https://github.com/tachiyomiorg/tachiyomi/pull/7099 by using `BitmapRegionDecoder` instead of `BitmapFactory`


app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -401,12 +396,17 @@ class Downloader(
         }
 
         return pageObservable
-            // When the image is ready, set image path, progress (just in case) and status
+            // When the page is ready, set page path, progress (just in case) and status
             .doOnNext { file ->
                 page.uri = file.uri
                 page.progress = 100
                 download.downloadedImages++
                 page.status = Page.READY
```
FourTOne5, line 404: Should move the 3 lines abovater splitting so download GTA completed after splitting

saud-97: Hmm, sure.

saud-97: Done

arkon:
I ended up merging #7099 first since it had larger changes. We can test that first and update this PR as needed.

saud-97:
Sure, there are still bugs remained unfixed even after the new PR was merged.

Here are the remaining applicable changes from this PR.

- Moved splitting logic to happen inside `getOrDownloadImage`
- Added a condition to check and skip splitting if a splitting was already done before on the page - to fix downloads geting stuck on pause.
- Allowed the download to continue and skip the splitting ( and show an error notification) if splitting failed instead of throwing an error and terminating the download.
- Moved CBZ logic to avoid marking a download as complete before the CBZ compression was completed

Done by #7099 
- Modified error handling for splitting to handle OOM error caused by `BitmapFactory.decodeFile` in case of extremely large images

saud-97:
Let me know once that PRs testing is done so I rebase on it and update this PR.

FourTOne5:
> Let me know once that PR's testing is done so I rebase on it and update this PR.

Arkon probably misunderstood your pr since #7099 and this doesn't have much relation

saud-97:
> Arkon probably misunderstood your pr since #7099 and this doesn't have much relation

I'm confused honestly, since the PR description should be clear as it explains the difference between them.

arkon:
I meant in regards to merge conflicts, which this PR still needs to fix...

saud-97:
I have merged and updated the PR

### Changes Applied:

- Moved splitting logic to happen inside `getOrDownloadImage`
- Added a condition to check and skip splitting if a splitting was already done before on the page to fix downloads getting stuck on pause.
- Moved CBZ logic to avoid marking a download as complete before the CBZ compression was completed

### Changes cancelled: 
- Allowed the download to continue and skip the splitting ( and show an error notification) if splitting failed instead of throwing an error and terminating the download.

I cancelled this since there was no `notifier ` object in ImageUtil, and I'm not sure if you would agree that it should be done or not. So currently if splitting failed it will throw an error and will make the `download.status =  Download.State.ERROR`

PR #7100, opened 2022-05-09
Check for app updates by comparing SemVer

curche:
Instead of just checking whether the current app version *matches* with latest app version in GitHub Releases, compare the SemVer from the tag names to check whether the latter is greater and the app needs an update

the comparison is done MAJOR > MINOR > PATCH manner
SemVer is obtained by parsing it out from versionTag/Version_name
you can view the sample code in the Kotlin playground [here](https://pl.kotl.in/erDE5Ifgy)

Tested on app (`.debug`; with Updater enabled) installed to Android 12 with the following values
> versionCode = 82
versionName = "0.13.6"

and Checking for Update shows "No New Updates"

Reference: SemVer Spec No.11 https://semver.org/#spec-item-11


app/src/main/java/eu/kanade/tachiyomi/data/updater/AppUpdateChecker.kt
```diff
@@ -56,6 +56,7 @@ class AppUpdateChecker {
     private fun isNewVersion(versionTag: String): Boolean {
         // Removes prefixes like "r" or "v"
         val newVersion = versionTag.replace("[^\\d.]".toRegex(), "")
+        val oldVersion = BuildConfig.VERSION_NAME.replace("[^\\d.]".toRegex(), "")
```
arkon, line 59: Can move this into the `else` block.

PR #7106, opened 2022-05-11
Minor cleanup in `ImageUtil` and `SourceFilterPresenter`

FourTOne5:


app/src/main/java/eu/kanade/tachiyomi/util/system/ImageUtil.kt
```diff
@@ -190,7 +190,7 @@ object ImageUtil {
     /**
      * Splits tall images to improve performance of reader
      */
-    fun splitTallImage(imageFile: UniFile, imageFilePath: String) {
+    fun splitTallImage(imageFile: UniFile, imageFilePath: String): Boolean {
         if (isAnimatedAndSupported(imageFile.openInputStream()) || !isTallImage(imageFile.openInputStream())) {
             return
```
saud-97, line 195: Return true here.
```suggestion
            return true
```

PR #7110, opened 2022-05-12
Adjust download menu in manga page

CVIUS:
Screenshot_20220512_103724
Screenshot_20220512_110853


saud-97:
I think most people will always use the unread options, I don't see why it's a good idea to increase number of clicks needed to reach there.

arkon:
Agreed, the menu's short enough as is. Feel free to open another PR if you think the wording should be updated.

PR #7119, opened 2022-05-13
Switch to DownloadPageLoader when previous/next chapter is downloaded

nzoba:
A minor addition to automatically switch to `DownloadPageLoader` if the previous/next chapter's download finishes while reading


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -345,6 +346,14 @@ class ReaderPresenter(
      * that the user doesn't have to wait too long to continue reading.
      */
     private fun preload(chapter: ReaderChapter) {
+        if (chapter.pageLoader is HttpPageLoader) {
+            val manga = manga ?: return
```
saud-97, line 350: if it returned here wouldn't skipping the rest of the method potentially may cause issues?

nzoba: I don't think this can really happen tbh, `manga` should always be instantiated at this point, it's more of a safety to avoid `!!`. But I agree that if it was null it could be problematic so it might have been a better idea to use `manga?.let`

saud-97: yeah either that or move it to another method and just place a call here. so if you returned from that method the workflow continues normally.

PR #7126, opened 2022-05-14
Settings cleanup

CVIUS:
### Changes
• Only show "Refresh tracking" in advanced settings if has any tracker logged in
• Hide "Invert tap zones" if tap zones is set to "Disabled"


arkon (commented):
Do you want to make the hiding invert tap change in `ReaderReadingModeSettings` too?

CVIUS:
> Do you want to make the hiding invert tap change in `ReaderReadingModeSettings` too?

Done

Issue #5252, opened 2021-06-01
[Feature Request] <prioritize default language extensions>


SrW0nka:
## Why/User Benefit/User Problem
This will help to found manga in your native language, as long as you have sources with other languages 

## What/Requirement
Put on top and prioritize the loading of default language extensions, for now the most near feature is pin the sources, but that's not my point 

Note: If you don't wanna add this by default, how about a option to set the main language extension in browse settings? (Like the "default category" option in library)

PR #7127, opened 2022-05-14
Enable sub-languages on app first run

alessandrojean:
Following the discussions in tachiyomiorg/tachiyomi-extensions#11840, this PR addresses the issue where the sub-languages aren't enabled on devices that have the same major language from the extensions.

As an example, if the user has a `zh` device language, the app will also enable `zh-Hans`, `zh-Hant` and any `zh-XYZ` that exists in the sources list from the extensions' repository JSON. This same issue also affects users with `pt` and `pt-BR`.

This will only happen in the first run. If the user have already changed the filters or is already an existing user, the enabled languages will be the same to respect the user choices done before this change.

I also changed the `Locale` parsing from the string code as it seems to be broken with codes that uses script variation in the second position instead of the country code. Using `zh-Hans` as an example again, previously it was creating a `language="zh", country="Hans"` locale, causing it to be displayed as `中文 (HANS)`. Now it is correctly creating a `language=zh, country="", variation="Hans"` that is displayed as `中文 (简体中文)` by using the built-in `Locale.forLanguageTag` method.

This is my first time playing with the app code, and I am not familiarized with it. If there's something that need to be changed or can be improved, I am open to suggestions, just let me know.


alessandrojean:
@kasperskier I also changed the language display name to also correctly displays Simplified (`中文 (简体中文)`) or Traditional (`中文 (繁體中文)`) instead of `中文 (HANS)`/`中文 (HANT)`. Now, I assume it's much more clear to Chinese speakers of what language/variant it is, since not everyone is familiar with `HANS`/`HANT` naming, but any additional feedback is appreciated too.

app/src/main/java/eu/kanade/tachiyomi/util/system/LocaleHelper.kt
```diff
@@ -37,20 +37,15 @@ object LocaleHelper {
...
      */
-    private fun getLocale(lang: String): Locale {
-        val sp = lang.split("_", "-")
-        return when (sp.size) {
-            2 -> Locale(sp[0], sp[1])
-            3 -> Locale(sp[0], sp[1], sp[2])
-            else -> Locale(lang)
-        }
+    fun getDefaultEnabledLanguages(): Set<String> {
+        return setOf("all", "en", Locale.getDefault().language)
```
kasperskier, line 49: I would insist on not turning on `all` and `en` unless the default language does not apply as per https://github.com/tachiyomiorg/tachiyomi-extensions/pull/11840#issuecomment-1126774583.

alessandrojean: @arkon What's your instance on this?

My opinion is that at least `all` should be enabled to show the multilang extensions. But I also understand `en` being enabled by default since the vast majority of users uses the app in English or uses English sources. Again, this is my opinion. Anyway I think this would be better to be addressed in other issue or PR, this one is just intended to fix an issue.

kasperskier: I guess this is already better because selected languages will appear on top of the filter list, so keeping it as is would be ok.

Another approach is to pin the system locale(s) on top. I don't know if it's hard to implement.

alessandrojean: > Another approach is to pin the system locale(s) on top. I don't know if it's hard to implement.

There's an open issue about this already (#5252).

arkon: > My opinion is that at least `all` should be enabled to show the multilang extensions.

Agreed, since it's not really intuitive to show those unfortunately.

> since the vast majority of users uses the app in English or uses English sources

This still holds true. If a minority non-English speaking user wants to, they can still explicitly disable it.

app/src/main/java/eu/kanade/tachiyomi/extension/ExtensionManager.kt
```diff
@@ -162,10 +168,48 @@ class ExtensionManager(
...
+     * the device language is inside that major group. As an example, if a user
+     * has a zh device language, the app will also enable zh-Hans and zh-Hant.
+     *
+     * If the user have already changed the enabledLanguages preference value once,
+     * the new languages will not be added to respect the user enabled choices.
+     */
+    private fun enableAdditionalSubLanguages(extensions: List<Extension.Available>) {
+        val prefs = PreferenceManager.getDefaultSharedPreferences(context)
+        val alreadyEnabled = prefs.getBoolean(PreferenceKeys.subLanguagesEnabledOnFirstRun, false)
+        val enabledLanguages = prefs.getStringSet(PreferenceKeys.enabledLanguages, null)
```
arkon, line 189: Use the helper `enabledLanguages()` function instead. You should be able to check if it's `set()` as needed.

app/src/main/java/eu/kanade/tachiyomi/extension/ExtensionManager.kt
```diff
@@ -162,10 +168,48 @@ class ExtensionManager(
...
+
+        val deviceLanguage = Locale.getDefault().language
+        val defaultLanguages = LocaleHelper.getDefaultEnabledLanguages()
+        val languagesToEnable = availableLanguages.filter {
+            it != deviceLanguage && it.startsWith(deviceLanguage)
+        }
+
+        prefs.edit {
+            putBoolean(PreferenceKeys.subLanguagesEnabledOnFirstRun, true)
+            putStringSet(PreferenceKeys.enabledLanguages, defaultLanguages + languagesToEnable)
```
arkon, line 209: Likewise, use the `enabledLanguages()` from the helper to avoid us drifting. I'd like to avoid more manual preference reading/editing where possible in case we ever decide to swap out the internals.

app/src/main/java/eu/kanade/tachiyomi/extension/ExtensionManager.kt
```diff
@@ -162,10 +168,48 @@ class ExtensionManager(
...
+     * the issue where users still need to enable some specific languages even when
+     * the device language is inside that major group. As an example, if a user
+     * has a zh device language, the app will also enable zh-Hans and zh-Hant.
+     *
+     * If the user have already changed the enabledLanguages preference value once,
+     * the new languages will not be added to respect the user enabled choices.
+     */
+    private fun enableAdditionalSubLanguages(extensions: List<Extension.Available>) {
+        val prefs = PreferenceManager.getDefaultSharedPreferences(context)
+        val alreadyEnabled = prefs.getBoolean(PreferenceKeys.subLanguagesEnabledOnFirstRun, false)
```
arkon, line 188: We can add a flag to `ExtensionManager` too so we don't need to keep reading from preferences to short circuit all of this (considering `findAvailableExtensions()` can be run multiple times during a single app session).

alessandrojean: Solved by creating an inner variable that reads from the preference once on init.

PR #7141, opened 2022-05-15
Add top padding to screens with header

CVIUS:


vengiare:
History needs this too iirc

app/src/main/java/eu/kanade/presentation/extension/ExtensionScreen.kt
```diff
@@ -111,6 +111,7 @@ fun ExtensionContent(
     val (trustState, setTrustState) = remember { mutableStateOf<Extension.Untrusted?>(null) }
     LazyColumn(
         contentPadding = WindowInsets.navigationBars.asPaddingValues(),
+        modifier = Modifier.padding(top = 8.dp),
```
ghostbear, line 114: Additional padding should be added with the `contentPadding`

CVIUS:
> History needs this too iirc

Alreadt included

app/src/main/java/eu/kanade/presentation/extension/ExtensionScreen.kt
CVIUS: Have no idea how to, with insets included. And after observing the behavior, ig additional padding should be added with the `contentPadding` to ALL screen with header since the items get behind the padding.

PR #7143, opened 2022-05-15
Remove explicit clean task from gradle

jimman2003:
I checked gradle's task list, clean was still there
I also did a clean build for a sanity check


arkon:
Is this just what the default clean task is? It's honestly just a leftover from the original project template.

jimman2003:
I think so... At least it worked locally so..

jimman2003:
maybe gradle 7 made it a default task?

ghostbear:
I tried this and this makes it so the root build directory isn't removed

jimman2003:
hmm..let me figure this out..

jimman2003:
Yeah, I failed to notice the difference of the root build dir vs app/build.. 
turns out , that directory still remains :(

jimman2003:
This PR is not needed,after all...

Issue #7140, opened 2022-05-15
Splitting Height for Tall Images


saud-97:
### Currently

Split height is equal to screen height. Often splitted chapters have too much page count compared to when the chapter is properly paged. For example, a chapter having 20 real pages can be considered as 50 pages after splitting.

### Suggestion 
I suggest to split based on the common webtoon images ratio. The math of split height would be calculated depending on the ratio of webtoons and the page width to determine what each page should have as height logically.

This will help to make the splitted page height to have more chance to actually split actual pages inside the chapter so to make it potentially more accurate in a sense. Therefore when someone tries to share a page he would have higher chance that the image he shares actually has a complete page rather than 2 pages with a weird cutoff between them.

saud-97:
Please replace bug tag with feature request tag.

arkon:
There is no common webtoon image height. Closing since it was already adjusted based in max screen dimension.

PR #7148, opened 2022-05-16
Fixed Splitting Height for Tall Images

saud-97:
Fixed Splitting Height for Tall Images #7140
Read the issue for an alternative suggestion.


app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
```diff
@@ -162,8 +163,8 @@ fun Context.hasPermission(permission: String) = ContextCompat.checkSelfPermissio
     }
 }
 
-val getDisplayHeightInPx: Int
-    get() = Resources.getSystem().displayMetrics.heightPixels
+val getDisplayMaxHeightInPx: Int
```
FourTOne5, line 166: No need to rename

app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
```diff
@@ -162,8 +163,8 @@ fun Context.hasPermission(permission: String) = ContextCompat.checkSelfPermissio
...
+    get() = max(Resources.getSystem().displayMetrics.heightPixels, Resources.getSystem().displayMetrics.widthPixels)
```
FourTOne5, line 167:
```suggestion
    get() = Resources.getSystem().displayMetrics.let { max(it.heightPixels, it.widthPixels) }
```

FourTOne5 (changes requested):

app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
FourTOne5: You know what rename it to this
```suggestion
val displayBiggerSideSizeInPx: Int
```

saud-97: I think max in the name shows that the value is not always the same reference so it will be clearer.

Issue #6751, opened 2022-03-08
[Local Manga] Can not set cover image


LastAi:
### Steps to reproduce

Add a manga to local that does not include a cover image.
In Tachiyomi browse the manga select a page, hold on screen wait for the options.
[Set as cover]
[Share]
[Save]

Select [Set as cover]
"Use this image as cover art?"
OK

the prompt 'Cover Set' breifly shows up 

### Expected behavior

The cover should be set in the Chapter Select or browse.

### Actual behavior

nothing is there, neither in the chapter select or in the browse.

### Tachiyomi version

Stable 0.13.1

### Android version

Android 12

### Device

Pixel 6

### Other details

This may be similar to issue #6737
but is not the same. Difference being that the cover IS NEVER set even after restart if the manga didn't have any cover.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] I have updated the app to version **[0.13.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

Sunspark-007:
The issue might be due to the fact that the cover is created and saved, but on the internal storage instead of the sd card. The library entry still points to the sd card, so none is displayed.

PR #7163, opened 2022-05-18
Local Source cleanup

### LINKED ISSUE #5319: [Feature Request] Generate cover.jpg in the sd card local directory instad of always saving it at /internal/Tachiyomi/local
Right now when a user adds a local manga at /sdcard/Tachiyomi/local Tachiyomi creates a new directory at /internal/Tachiyomi/local with the same name as the manga directory on the sdcard and saves the cover.jpg there. If a user deletes a local manga from the sdcard without deleting it from the internal storage the manga will still be shown in Tachiyomi just without the chapters. To really remove a manga a user has to search through their internal storage  and delete the manga from there as well.

## What/Requirements
save the cover.jpg into the directory were the manga is saved instead of saving it to the path /internal/Tachiyomi/local

FourTOne5:
Also fixes that generated cover.jpg are always saved in `/internal/Tachiyomi/local/Manga Name` instead of actual Manga folder
Fixes #5319


specterflare:
Also closes #6751

FourTOne5:
> Also closes #6751

It does not

specterflare:
Is it not the same as #5319?

FourTOne5:
> 

The 

> Is it not the same as #5319?
> 
> From 6751: The issue might be due to the fact that the cover is created and saved, but on the internal storage instead of the sd card. The library entry still points to the sd card, so none is displayed.
> 
> From 5319: Generate cover.jpg in the sd card local directory instad of always saving it at /internal/Tachiyomi/local

The issue there is that the updated isn't reflected until a app restart

specterflare:
Oh ok.

PR #7169, opened 2022-05-19
Migrate extensions language filter screen to compose

FourTOne5:


app/src/main/java/eu/kanade/presentation/browse/ExtensionLangFilterScreen.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    LazyColumn(
+        modifier = Modifier.nestedScroll(nestedScrollInterop),
+        contentPadding = WindowInsets.navigationBars.asPaddingValues(),
+    ) {
+        items(
+            items = items,
+        ) { model ->
+            ExtensionFilterItem(
+                modifier = Modifier.animateItemPlacement(tween(1000, easing = LinearOutSlowInEasing)),
```
arkon, line 67: Why different tweening only here?

FourTOne5: The items placement were changing quickly so added it to make the change look smoother.

PR #7188, opened 2022-05-23
Fix global update ignoring network constraint

### LINKED ISSUE #7016: global updates ignoring WLAN only setting
### Steps to reproduce

1. Navigate to 'Settings' -> Library
2. Set automatic update to '12 hours'
3. Set restriction to only update on WLAN connection

### Expected behavior

The library should only update when the phone is connected to WLAN, not over mobile data. If the 12 hour window is up it should push the update back until WLAN is available.

### Actual behavior

The library keeps updating and disregarding the setting of the restriction whenever the 12 hour window is up.

### Crash logs

_No response_

### Tachiyomi version

0.13.4

### Android version

Android 10

### Device

Oneplus 7 Pro

### Other details

_No response_

### Acknowledgements

shadow578:
Fixes automatic library updates ignoring network constraint and closes #7016 .

Note: I've added a constraint for unmetered network to the work request. That way, the job will run as soon as a unmetered (== wifi) connection is available instead of waiting for the next run in however many hours the user set it to.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
@@ -22,8 +22,9 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
 
     override fun doWork(): Result {
         val preferences = Injekt.get<PreferencesHelper>()
-        if (requiresWifiConnection(preferences) && !context.isConnectedToWifi()) {
-            Result.failure()
+        val restrictions = preferences.libraryUpdateDeviceRestriction().get()
+        if ((DEVICE_ONLY_ON_WIFI in restrictions) && !context.isConnectedToWifi()) {
+            return Result.failure()
         }
 
         return if (LibraryUpdateService.start(context)) {
@@ -42,7 +43,7 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
             if (interval > 0) {
                 val restrictions = preferences.libraryUpdateDeviceRestriction().get()
                 val constraints = Constraints.Builder()
-                    .setRequiredNetworkType(NetworkType.CONNECTED)
+                    .setRequiredNetworkType(if (DEVICE_ONLY_ON_WIFI in restrictions) { NetworkType.UNMETERED } else { NetworkType.CONNECTED })
                     .setRequiresCharging(DEVICE_CHARGING in restrictions)
                     .setRequiresBatteryNotLow(DEVICE_BATTERY_NOT_LOW in restrictions)
                     .build()
@@ -61,11 +62,6 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
             } else {
                 WorkManager.getInstance(context).cancelAllWorkByTag(TAG)
             }
-        }
-
-        fun requiresWifiConnection(preferences: PreferencesHelper): Boolean {
-            val restrictions = preferences.libraryUpdateDeviceRestriction().get()
-            return DEVICE_ONLY_ON_WIFI in restrictions
         }
     }
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
@@ -8,10 +8,7 @@ import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.Worker
 import androidx.work.WorkerParameters
-import eu.kanade.tachiyomi.data.preference.DEVICE_BATTERY_NOT_LOW
-import eu.kanade.tachiyomi.data.preference.DEVICE_CHARGING
-import eu.kanade.tachiyomi.data.preference.DEVICE_ONLY_ON_WIFI
-import eu.kanade.tachiyomi.data.preference.PreferencesHelper
+import eu.kanade.tachiyomi.data.preference.*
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
```diff
@@ -42,7 +43,7 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
             if (interval > 0) {
                 val restrictions = preferences.libraryUpdateDeviceRestriction().get()
                 val constraints = Constraints.Builder()
-                    .setRequiredNetworkType(NetworkType.CONNECTED)
+                    .setRequiredNetworkType(if (DEVICE_ONLY_ON_WIFI in restrictions) { NetworkType.UNMETERED } else { NetworkType.CONNECTED })
```
arkon, line 46: Metered (or unmetered) networks don't actually strictly correlate with it being Wi-Fi or not, hence the separate `isConnectedToWifi` check earlier on. You can just leave this as-is.

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
shadow578: I'm aware that the unmetered network constraint doesn't correlate to the network being WiFi. However, as i explained in the PR text, i added this so that WorkManager will schedule the job to run only on unmetered connections (which, for most users) will be their wifi connection. 
Otherwise, the job could get started on a mobile connection, where it will then fail. This might lead to a situation where the job never runs (e.g. when the update job is scheduled for every 24h, and every time the job is scheduled the device is off of wifi).

I'm assuming that if you're going to the effort of constraining tachiyomi to only use wifi, you wouln't want it to use a metered wifi (which is a option you have to set manually) either.

shadow578: I really think we should use the network constraint that workmanager offers to ensure jobs are only scheduled when they make sense. 
Perhaps i could add an additional option to settings so that users can explicitly enable the constraint in this line, instead of implicitly enabling it for them. Something along the lines of 'Only update on unmetered Networks' (in the same menu where one can set the option to only update over wifi, that is)

arkon: I doubt most users would really understand what that means (it isn't even clear in the developer documentation), but it may be useful for some people.

I'm not super convinced it'll really affect most job schedules since having a metered Wi-Fi connection is probably rare unless a VPN is being used as well, but go for it.

shadow578:
Added explicit constraint, see attached screenshot
update-constraints

app/src/main/res/values/strings.xml
```diff
@@ -221,6 +221,7 @@
     <string name="update_weekly">Weekly</string>
     <string name="pref_library_update_restriction">Automatic updates device restrictions</string>
     <string name="connected_to_wifi">Only on Wi-Fi</string>
+    <string name="network_not_metered">Only on unmetered Network</string>
```
arkon, line 224:
```suggestion
    <string name="network_not_metered">Only on unmetered network</string>
```

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
```diff
@@ -8,10 +8,7 @@ import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.Worker
 import androidx.work.WorkerParameters
-import eu.kanade.tachiyomi.data.preference.DEVICE_BATTERY_NOT_LOW
-import eu.kanade.tachiyomi.data.preference.DEVICE_CHARGING
-import eu.kanade.tachiyomi.data.preference.DEVICE_ONLY_ON_WIFI
-import eu.kanade.tachiyomi.data.preference.PreferencesHelper
+import eu.kanade.tachiyomi.data.preference.*
```
arkon, line 11: Don't use wildcard imports.

app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsLibraryController.kt
```diff
@@ -11,13 +11,7 @@ import eu.kanade.tachiyomi.R
...
 import eu.kanade.tachiyomi.data.database.models.Category
 import eu.kanade.tachiyomi.data.library.LibraryUpdateJob
-import eu.kanade.tachiyomi.data.preference.DEVICE_BATTERY_NOT_LOW
-import eu.kanade.tachiyomi.data.preference.DEVICE_CHARGING
-import eu.kanade.tachiyomi.data.preference.DEVICE_ONLY_ON_WIFI
-import eu.kanade.tachiyomi.data.preference.MANGA_HAS_UNREAD
-import eu.kanade.tachiyomi.data.preference.MANGA_NON_COMPLETED
-import eu.kanade.tachiyomi.data.preference.MANGA_NON_READ
-import eu.kanade.tachiyomi.data.preference.PreferencesHelper
+import eu.kanade.tachiyomi.data.preference.*
```
arkon, line 14: Same here re: wildcard imports.

arkon (changes requested):
Minor nitpicks but otherwise LGTM.

Saud-97:
> Added explicit constraint, see attached screenshot update-constraints

I suggest to unify the language used and make it `While condition" or drop the while word to match the rest.

arkon:
I'll take care of the minor comments separately.

PR #7192, opened 2022-05-24
ChapterSourceSync: treat negative timestamps as fallback

### LINKED ISSUE #7162: ChapterSourceSync: add fallback upload date logic

Some sources only provide upload date of the lastest chapter. In this case, extensions either have to ignore it, or falsely set upload date of all existing chapters to it: https://github.com/tachiyomiorg/tachiyomi-extensions/pull/9242

Adding a fallback upload date will solve this issue:
```kotlin
// only needed for SChapter etc, not needed for db Chapter
val upload_date_fallback: Long?
```

And fallback new chapters' upload date to it:
https://github.com/tachiyomiorg/tachiyomi/blob/6580f5771f634b0e2c25f8cd42fa1596b2ea4e1c/app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterSourceSync.kt#L72-L76
```diff
-   sourceChapter.date_upload = Date().time
+   sourceChapter.date_upload = (???).date_upload_fallback ?: Date().time
```

so that extensions may set it to the known upload time and `upload_date = 0` when parsing chapters to avoid falsely updating all existing chapters to the lastest upload time, bypassing this check:
https://github.com/tachiyomiorg/tachiyomi/blob/6580f5771f634b0e2c25f8cd42fa1596b2ea4e1c/app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterSourceSync.kt#L87-L89

The update logic seems to be using a db Chapter object, and I think it's unnecessary to add such field to it. Not sure how to implement this.

### Other details

_No response_

### Acknowledgements

kasperskier:
Closes #7162.

This allows sources that only know the upload date of the latest chapters to apply it only to the new chapters instead of falsely updating all chapters. **(by explicitly setting the upload date to its opposite number)**

This shouldn't break ~~anything~~. Looks a bit hacky though.

**Side effect: new chapters whose date is falsely parse to a negative number will show up as a bad date in the chapter list, instead of having empty date.**

The original idea in #7162 was to add a new field called `date_upload_fallback`, which is more robust. But I found that `Chapter` also inherits `SChapter` so this field will be saved into the database, which is not necessary.

Let me know if there's a better approach.

- [ ] If this gets merged, add it to extension repo's [CONTRIBUTING.md](https://github.com/tachiyomiorg/tachiyomi-extensions/blob/master/CONTRIBUTING.md)


kasperskier:
Wait, will negative timestamps break in older versions of the app? 🤔 Is it necessary to make it compatible with older versions?

Or, can extensions know the app version and set negative timestamps if the version is new enough?

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterSourceSync.kt
```diff
@@ -73,6 +73,9 @@ fun syncChaptersWithSource(
         if (dbChapter == null) {
             if (sourceChapter.date_upload == 0L) {
                 sourceChapter.date_upload = Date().time
+            } else if (sourceChapter.date_upload < 0L) {
+                // Treat negative timestamps as fallback.
+                sourceChapter.date_upload = -sourceChapter.date_upload
```
FourTOne5, line 78: negative timestamps are result of broken parsing and can result in some date 10 years later or ago

kasperskier: Current version does not filter them out, so they will cause wrong data to be shown anyway.

FourTOne5: When viewing chapters of manga negative upload date show up as blank

FourTOne5: I slightly misunderstood the code this actually sounds like a good idea

Seishirou101: You shouldn't account for bad data and try and fix it but instead just fix the bad data on the ext side. Especially in this case if it was negative and wrong, why do you think making it positive would make the bad data good? For all we know the ext could have parsed some random number and it just happened to be negative.

kasperskier: I probably didn't explain it clearly in the post. My idea is that: extensions can **explicitly** set the upload date to its opposite number to make it a "fallback" date which will not be updated to existing chapters. Wrong data is still wrong. Side effect is that they will show bad date in the chapter list instead of being empty (filtered by UI).

Seishirou101: Do you have an example of an extension that does this? That doesn't sound like something we should ever do. If a source provides a negative number then the extension will have to account for it, but at that point its fine because a human will be looking at it and hopefully figure out a pattern if they consistently do it.

kasperskier: @Seishirou101 No, this PR attempts to enable extensions to do this. Extensions that would now output negative numbers can be updated to clamp that number to zero, which is what they should do. If this PR gets merged, negative numbers will mean "fallback" dates.

kasperskier:
Update to use a safer threshold (negative 2<sup>31</sup> ms) in case anyone hardcoded numbers like -1.

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterSourceSync.kt
Seishirou101: I'm not really convinced, assuming you can use a fallback then your latest chapters will show that date correct? what would the chapters without dates do then? you'd would still need to do something extension wise, and at that point you may as well fix it there imo. You're in a very bad place as your source does not provide upload dates for all their chapters. I personally think the 1st scenario is the "best" for the situation, but you'll just have to wait for arkon and see what they think

kasperskier: At least that book will show up in the right place in your bookshelf.

Another approach is to let the extension set the upload date for the latest chapter only and -1 for all previous chapters. For current versions, this will override all previous dates (if set). The second part of this commit will make the app skip updating those.

FourTOne5: Why not just set 0 for old chapters and let app handle the rest. I'll do a PR so that the upload date doesn't exceed the max upload date of all chapter if it's 0

kasperskier: > Why not just set 0 for old chapters and let app handle the rest. I'll do a PR so that the upload date doesn't exceed the max upload date of all chapter if it's 0

Yeah that's a good idea!

kasperskier: @FourTOne5 See #7197.

kasperskier:
Replaced by #7197.

PR #7196, opened 2022-05-25
New: Migrating titles maintains custom covers

### LINKED ISSUE #7189: Migrating titles maintains custom covers

Hi! I think Tachiyomi should be able to maintain my custom covers when I migrate from dead extensions. When ReadComicOnline died, for example, I had all of my comics with nice, bespoke HD covers for each one, which often contrast starkly with the very blurry and low quality default covers the websites have available. But when migrating them, I had to save them beforehand, and update all of the comics, again, one by one, with each cover I had presaved. I think it would be cool if the custom covers stayed the same when migrating, and it doesn't sound too hard to implement either, so I figured I should ask here.

P.S. Thank you very much for your time and attention, I've been using your app for over 5 years and I absolutely adore it!

### Other details

_No response_

### Acknowledgements

Saud-97:
Implemented Migrating titles maintains custom covers. Closes #7189


saud-97:
I'll

app/src/main/res/values/strings.xml
```diff
@@ -623,6 +623,7 @@
     <string name="download_custom">Custom</string>
     <string name="download_all">All</string>
     <string name="download_unread">Unread</string>
+    <string name="custom_covers">Custom Covers</string>
```
arkon, line 626:
```suggestion
    <string name="custom_covers">Custom covers</string>
```

app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/MigrationFlags.kt
```diff
@@ -4,17 +4,18 @@ import eu.kanade.tachiyomi.R
...
+    private const val CATEGORIES = 0b0010
+    private const val TRACK = 0b0100
+    private const val CUSTOM_COVERS = 0b1000
 
     private const val CHAPTERS2 = 0x1
     private const val CATEGORIES2 = 0x2
     private const val TRACK2 = 0x4
 
-    val titles get() = arrayOf(R.string.chapters, R.string.categories, R.string.track)
+    val titles get() = arrayOf(R.string.chapters, R.string.categories, R.string.track, R.string.custom_covers)
```
arkon, line 16: It'd be ideal if the option wasn't presented if there isn't actually a custom cover. I guess the same applies for trackers too though, so maybe we can just do that as a followup.

app/src/main/res/values/strings.xml
Saud-97: done

app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/MigrationFlags.kt
Saud-97: I will try and see if I was able to do this.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/MigrationFlags.kt
```diff
@@ -7,15 +7,15 @@ object MigrationFlags {
     private const val CHAPTERS = 0b0001
     private const val CATEGORIES = 0b0010
     private const val TRACK = 0b0100
-    private const val CUSTOM_COVERS = 0b1000
+    private const val CUSTOM_COVER = 0b1000
```
AntsyLich, line 10: Why not `0b0101`

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/MigrationFlags.kt
Saud-97: I have done this and tested it.
I found 3 issues with how Tachiyomi handle the logic.

First issue is that tracking option will show if the manga has any track data in db that means it would show even when the manga is no longer tracked currently and even if the tracking services is not logged in.

Second issue is that old logic depends on index on saving flags enabled status rather than flag value like 0b0001. That would cause incorrect enabled status when indices change.

```
                    // Save current settings for the next time
                    val selectedIndices = mutableListOf<Int>()
                    selected.forEachIndexed { i, b -> if (b) selectedIndices.add(i) }
```

app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/MigrationFlags.kt
Saud-97: > Why not `0b0101`

That would make `CHAPTERS` and `TRACK` be mistakenly counted as enabled because that's how flags work iirc :)

app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/MigrationFlags.kt
arkon: > First issue is that tracking option will show if the manga has any track data in db that means it would show even when the manga is no longer tracked currently and even if the tracking services is not logged in.

That's technically fine. We end up backing up/restoring that sort of data anyway, and it typically works if you just re-login.

Kediny:
Thank you very much!! 😁

PR #7200, opened 2022-05-26
Improved Extensions Language filtering

Saud-97:
I noticed that when looking for a language during Extensions installation with only selecting a specific language.
The current behavior will show extensions that are ONLY in the selected languages and it will not show any extensions exist in ALL although they have the selected language. 

Since this behavior is misleading and is conflicting with how sources view group sources by language regardless if it is from ALL or not. I created this PR to fix the behavior.

For example when selecting HANS only it will still show some extensions from ALL that have HANS language. Instead of showing only the last extension. 

## Before:

image

## After: 

 image


arkon:
I don't understand what you're searching?

saud-97:
Let me explain a bit, currently what someone has ALL language disabled, extensions that are in ALL will not show as available for install, even if they contain the languages the user is looking for.

Before when someone intends to install an extension with the following filters
HANS = ON and ALL = OFF

Before the PR it will show

-HANS
--BILBILI MANGA

After the PR it will show

-ALL
--Bato.to
--and others that match user filtered languages

-HANS
--BILBILI MANGA

arkon:
It might make more sense to stop showing a separate "All" section for extensions and instead duplicate it in the different language sections based on the sources' languages. In parallel, wording it as something like "Multiple" would better indicate that it supports multiple languages.

saud-97:
I planned to do show duplicated in each language but unfortunately I didn't know how to modify the groupBy to do it. since I will need to do a loop ` it.sources.forEach { it.lang }` and combine the loop with groupBy somehow overlapping.
 https://github.com/tachiyomiorg/tachiyomi/blob/9f42306f79b9faa601016e1b826acd9f575ec976/app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt#L91-L100

I tried to think how I can do it maybe in multiple steps and then combine the overlapping maps but I don't really know how to exactly do it so I left it to your discretion.

Also I agree the naming of ALL should be changed to Multilanguage or similar meaningful names.

saud-97:
I will try and see if I can do the overlapping maps again.

saud-97:
Well I spent sometime trying to find a solution but I wasn't able to. I thought about making the` available<Extensions.Available>` to be flattened, i.e., move the internal extensions and make it 1 level list. But I wasn't able to change the list type from `AvailableExtensionSources `to `Extensions.Available` to make it work in the code.

If someone want to understand more on how the extensions json and sources mapping were made have a look at https://raw.githubusercontent.com/tachiyomiorg/tachiyomi-extensions/repo/index.json

arkon:
Closing this since the PR as-is isn't useful. Feel free to log the ideas/findings in an issue.

Saud-97:
Can someone re-open this PR. I was able to define the logic for correctly showing sources under their own language instead of putting them under ALL. After the last changes the PR should be ready for review again.

Saud-97:
The only remaining thing is to refactor the app and the update the **ALL** logic or if only minimum changes are needed then just hide the ALL filter from Extensions Page to not show it explicitly and just show it inside other languages groups. These are optional changes I can think of but not required.

app/src/main/java/eu/kanade/tachiyomi/extension/model/Extension.kt
```diff
@@ -62,4 +62,21 @@ data class AvailableExtensionSources(
     val name: String,
     val id: Long,
     val baseUrl: String,
-)
+    val lang: String,
+) {
+    fun toExtension(ext: Extension.Available, index: Int): Extension.Available {
+        return Extension.Available(
```
nzoba, line 68: You could use `ext.copy(...)` to avoid setting each parameter. And maybe use `ext.pkgName + lang` instead of an index ?

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
```diff
@@ -88,7 +92,22 @@ class ExtensionsPresenter(
             ) { query, (installed, untrusted, available), updates, downloads ->
                 isRefreshing = false
 
-                val languagesWithExtensions = available
+                val flattenedAvailable = emptyList<Extension.Available>().toMutableSet()
+                val activeLanguages = preferences.enabledLanguages().get()
+
+                available.forEach { ext ->
```
nzoba, line 98: Imo, `flatMap` is better in this case. Probably something like that :
```
  val flattenedAvailable = available.flatMap { ext ->
      ext.sources.filter { it.lang in activeLanguages }
          .map {
          if (ext.lang == it.lang) ext else it.toExtension(ext)
      }
  }
```

app/src/main/java/eu/kanade/tachiyomi/extension/model/Extension.kt
Saud-97: Yes I will use copy, I am still getting used to kotlin so I forgot copy method.
Regarding `lang `instead of index it wouldn't work since some extensions have repeated `lang `values.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
Saud-97: Thanks for the suggestion.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
```diff
@@ -88,7 +92,14 @@ class ExtensionsPresenter(
             ) { query, (installed, untrusted, available), updates, downloads ->
                 isRefreshing = false
 
-                val languagesWithExtensions = available
+                val activeLanguages = preferences.enabledLanguages().get()
+                val flattenedAvailable = available.flatMap { ext ->
+                    ext.sources.filter { it.lang in activeLanguages }.mapIndexed { i, it ->
+                        if (ext.lang == it.lang) ext
+                        else ext.copy(name = it.name, lang = it.lang, pkgName = ext.pkgName + i)
```
AntsyLich, line 99:
```suggestion
                        else ext.copy(name = it.name, lang = it.lang, pkgName = ext.pkgName.replace(".${ext.lang}.", ".${it.lang}.")
```

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
```diff
@@ -88,7 +92,14 @@ class ExtensionsPresenter(
...
+                    }
+                }.toSet()
```
AntsyLich, line 101: Why not a list?

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
```diff
@@ -25,13 +26,16 @@ import kotlinx.coroutines.flow.update
...
 import uy.kohesive.injekt.api.get
+import uy.kohesive.injekt.injectLazy
 
 class ExtensionsPresenter(
     private val extensionManager: ExtensionManager = Injekt.get(),
     private val getExtensionUpdates: GetExtensionUpdates = Injekt.get(),
     private val getExtensions: GetExtensions = Injekt.get(),
 ) : BasePresenter<ExtensionsController>() {
 
+    private val preferences: PreferencesHelper by injectLazy()
```
AntsyLich, line 37: Move to class constructor

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
Saud-97: Because some extenstions have duplicated sources of the same lang 
`                        if (ext.lang == it.lang) ext
`
Like Komga and few others. Set is to dedup the repeated langs to avoid app crashes. Since  compose rows have `pkgName` as the primary ID
I could check for dups manually or let the set handle it.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
Saud-97: What's the benefits? other than possibly having issues with deduping duplicated pkg names because `it.lang` is not always unique.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
Saud-97: I tried initially but because I received an error I moved it. Let me check again and solve the error if possible.

Saud-97:
I have Updated the discussed changes and the PR description now shows comparison screenshots.

app/src/main/java/eu/kanade/domain/extension/interactor/GetExtensions.kt
```diff
@@ -38,9 +38,17 @@ class GetExtensions(
...
-                }
+                }.flatMap { extension ->
+                    extension.sources.filter { it.lang in _activeLanguages }.sortedBy { it.name }.map {
+                        extension.copy(
+                            name = if (it.lang == extension.lang) extension.name else it.name,
+                            lang = it.lang,
+                            pkgName = extension.pkgName + ".${it.lang}",
+                        )
+                    }
+                }.toSet()
```
AntsyLich, line 51: `.distinct()`?

app/src/main/java/eu/kanade/domain/extension/interactor/GetExtensions.kt
```diff
@@ -38,9 +38,17 @@ class GetExtensions(
...
                         (showNsfwSources || extension.isNsfw.not())
-                }
+                }.flatMap { extension ->
+                    extension.sources.filter { it.lang in _activeLanguages }.sortedBy { it.name }.map {
+                        extension.copy(
+                            name = if (it.lang == extension.lang) extension.name else it.name,
+                            lang = it.lang,
+                            pkgName = extension.pkgName + ".${it.lang}",
+                        )
+                    }
```
AntsyLich, line 50:
```suggestion
                    val extSources = extension.sources.filter { it.lang in _activeLanguages }.sortedBy { it.name }
                    if (extSources.size == 1) {
                        listOf(extension)
                    } else {
                        extSources.map {
                            extension.copy(
                                name = if (it.lang == extension.lang) extension.name else it.name,
                                lang = it.lang,
                                pkgName = extension.pkgName + ".${it.lang}",
                            )
                        }
                    }
                    
```

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
```diff
@@ -92,23 +92,28 @@ class ExtensionsPresenter(
...
                     .toSortedMap()
-                    .flatMap { (key, value) ->
-                        listOf(
-                            ExtensionUiModel.Header.Text(key),
-                            *value.map(extensionMapper(downloads)).toTypedArray(),
+                    .flatMap { (lang, sources) ->
+                        val sources = sources.groupBy { it.pkgName.contains(".all.") }.toSortedMap()
+	                        .flatMap { (all, sources) ->
+		                        if (all) sources.map { it.copy(lang = "all") } else sources
+	                        }
```
AntsyLich, line 99: Doesn't seem necessary?

AntsyLich (changes requested):

app/src/main/java/eu/kanade/domain/extension/interactor/GetExtensions.kt
Saud-97: I read on stackoverflow a comment saying toSet() has less overhead. But since its marginal I will switch to distinct for readablity.

app/src/main/java/eu/kanade/domain/extension/interactor/GetExtensions.kt
Saud-97: I don't really see what's the point in checking the size here. If you want me to not `copy `an object if it was a primary source I can just check via `(it.lang == extension.lang)`. I removed it for readability, I guess I will put it back for performance.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/ExtensionsPresenter.kt
Saud-97: Refer to discussion in #programming channel. It's needed to differentiate between partial sources and primary sources and for sorting purposes.

app/src/main/java/eu/kanade/tachiyomi/data/preference/PreferencesHelper.kt
```diff
@@ -168,7 +168,7 @@ class PreferencesHelper(val context: Context) {
 
     fun sourceDisplayMode() = flowPrefs.getEnum("pref_display_mode_catalogue", DisplayModeSetting.COMPACT_GRID)
 
-    fun enabledLanguages() = flowPrefs.getStringSet("source_languages", setOf("all", "en", Locale.getDefault().language))
+    fun enabledLanguages() = flowPrefs.getStringSet("source_languages", setOf("en", Locale.getDefault().language))
```
AntsyLich, line 171: extensions like komaga, mango will be hidden then

Saud-97: Isn't "all" enabled by default to show extensions that multi languages extensions that may possibly support "en"? 

I honestly don't think most users will even be aware about Komga.
Regarding mango I can see it has a partial source "en" so it will still be shown regardless of "all" being enabled or not.

```
  {
    "name": "Tachiyomi: Mango",
    "pkg": "eu.kanade.tachiyomi.extension.all.mango",
    "apk": "tachiyomi-all.mango-v1.3.9.apk",
    "lang": "all",
    "code": 9,
    "version": "1.3.9",
    "nsfw": 0,
    "hasReadme": 0,
    "hasChangelog": 1,
    "sources": [
      {
        "name": "Mango",
        "lang": "en",
        "id": "2882993841250976719",
        "baseUrl": "",
        "versionId": 1,
        "hasCloudflare": 0
      }
    ]
  },
```

Saud-97: > Was just saying as an example. But keep it enabled

I honestly disagree.

Currently the app by default enabled All, en, and current system locale. After my change en and current system local will even show extensions that are overlapped from All. So whats the point of showing All extensions again as duplicates if they don't explicitly include the languages the user most likely would be interested in?

I have checked which extensions that only has All or other as a lang which were only 10 out of 63 extensions:
```
    "apk": "tachiyomi-all.buondua-v1.3.1.apk",
    "apk": "tachiyomi-all.everiaclub-v1.3.2.apk",
    "apk": "tachiyomi-all.foolslidecustomizable-v1.3.3.apk",
    "apk": "tachiyomi-all.hentaishark-v1.3.5.apk",
    "apk": "tachiyomi-all.junmeitu-v1.3.2.apk",
    "apk": "tachiyomi-all.kavita-v1.3.4.apk",
    "apk": "tachiyomi-all.komga-v1.3.40.apk",
    "apk": "tachiyomi-all.lanraragi-v1.3.12.apk",
    "apk": "tachiyomi-all.xinmeitulu-v1.3.3.apk",
    "apk": "tachiyomi-all.yaoimangaonline-v1.3.2.apk",
``` 

Do you still think its worth showing 63 extensions possibly duplicated x3 times (cuz of possible overlapping with other locales) to be shown as default?

Saud-97: If you still think those 10 sources should show by default. I would suggest to include en locale in these sources as part of their lang set similar to Tachidesk ext. Since en will be enabled by default this will force them to show.

```
  {
    "name": "Tachiyomi: Tachidesk",
    "pkg": "eu.kanade.tachiyomi.extension.all.tachidesk",
    "apk": "tachiyomi-all.tachidesk-v1.3.5.apk",
    "lang": "all",
    "code": 5,
    "version": "1.3.5",
    "nsfw": 1,
    "hasReadme": 0,
    "hasChangelog": 0,
    "sources": [
      {
        "name": "Tachidesk",
        "lang": "en",
        "id": "3100117499901280806",
        "baseUrl": "",
        "versionId": 1,
        "hasCloudflare": 0
      }
    ]
  },
```

AntsyLich: Unlike tachidesk the others can't filter series via language so can't

PR #7205, opened 2022-05-27
MangaUpdates tweaks

ghostbear:
- Change status to reading when user read chapter
- Use decimal point for rating (isn't a user preference from what I understand)
- Parse HTML description to text  (I asked about if the API could provide it and the MU DEV said some endpoint do allow for it to be, search endpoint doesn't, and it may become Markdown in the future)


app/src/main/java/eu/kanade/tachiyomi/data/track/mangaupdates/MangaUpdates.kt
```diff
@@ -52,11 +52,28 @@ class MangaUpdates(private val context: Context, id: Int) : TrackService(id) {
...
+    override fun getScoreList(): List<String> {
+        val values = mutableListOf<String>()
+        for (i in 0..10) {
+            values.add("$i.0")
+            if (i == 10) break
+            for (j in 1 until 10) {
+                values.add("$i.$j")
+            }
+        }
+        return values
```
arkon, line 64:
```kotlin
(0..9).flatMap { i -> (0..9).map { j -> "$i.$j" } } + listOf("10.0")
```

Should also pull this out so it doesn't need to get recreated every time.

arkon (changes requested):

Saud-97:
Please also parse HTML of title. Consider checking if other fields need any extra parsing similarly to ensure correctness.

image

app/src/main/java/eu/kanade/tachiyomi/data/track/mangaupdates/dto/Record.kt
```diff
@@ -28,7 +29,7 @@ fun Record.toTrackSearch(id: Int): TrackSearch {
         title = this@toTrackSearch.title ?: ""
```
Saud-97, line 29:
```suggestion
        title = Jsoup.parse(this@toTrackSearch.title ?: "").text()
```

Saud-97 (changes requested):

PR #7210, opened 2022-05-28
Corrected register tracking logic by updating last read chapter

Saud-97:
Closes https://github.com/tachiyomiorg/tachiyomi/issues/7209

I have done the update after binding the tracker because some (all?) trackers default to a value upon first register so it is safer to update remote after registering.


arkon:
Doesn't this fail to account for the existing progress from trackers being ahead of what's locally marked as read?

Saud-97:
> Doesn't this fail to account for the existing progress from trackers being ahead of what's locally marked as read?

Yes you are right. I fixed it.

Saud-97:
This is a simple 5 lines PR, is there any need to change anything or shall I close?

arkon:
MangaUpdates relies on database changes that aren't trivially backported back to stable.

arkon:
Superseded by 488d8ab8cfbee2cab6121154b546f4764852f088

PR #7226, opened 2022-05-29
Download ahead while reading

### LINKED ISSUE #3295: [Feature Request] Have the next chapter automatically download while reading.
The option to have the next chapter **download** while reading will improve the manga reading experience in Tachiyomi (particularly for readers normally on a Wi-Fi connection). From my experience, I encounter a couple of inconveniences when the [HttpPageLoader](https://github.com/inorichi/tachiyomi/blob/ec9d592cf1c120500e11cb39f92c485aa3d66e4e/app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/HttpPageLoader.kt) is used to preload pages in a chapter:

1. If the chapter has a lot of pages, lengthy loading times and timeouts are likely to occur.
2. resuming a chapter after opening the app causes the loader to fetch pages again, causing a few seconds of delay before the page shows.

Other benefits I see this feature having is alleviating the reader from having to remember to download X amount of chapters before continuing with their favorite manga and also decreasing the chance a manga host throttling or breaking client connection, causing those dreaded download errors, because of bulk chapter downloading.

### What/Requirements

Add a new Downloads option to enable this feature (off by default), see "Steady Downloads":

Then whenever you start a manga chapter, instead of the usual preload, the next chapter is now downloaded:

nzoba:
Closes #3295

~~New setting to download the next unread chapters after exiting the reader~~

~~I already added it to J2K (cf. [PR](https://github.com/Jays2Kings/tachiyomiJ2K/pull/1174)). At first it was **while** reading but after discussing it with Jay we decided to only download **after** reading to avoid potential double traffic (downloading + fetching a chapter)~~

**EDIT:** Much has changed since the opening of the PR. Look at the message's history if you want to see the old screenshots. Here are the current changes:
1. The next unread chapters are automatically downloaded **while** reading after 20% of a chapter is already read. It works only on manga in library and if the current chapter plus the next one are **already** downloaded. The number of chapters to download depends of the setting
2. Prevents both fetching and downloading the same chapter :
    a. When you start reading a chapter **pending but not yet downloaded**, it stops (removes) the chapter's download
    b. If you exit the reader mid-chapter, the chapter is added back to the queue at first position (downloaded immediatly). The queue will restart only if it was already running or if the queue was empty. If it was paused it won't
    c. Same if you exit at the end of the chapter or continue to the next one. The only difference is if you have the setting `After reading auto delete` set to `last chapter`, in this case the chapter won't be downloaded

Download ahead setting


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -177,8 +180,15 @@ class ReaderPresenter(
         super.onDestroy()
         val currentChapters = viewerChaptersRelay.value
         if (currentChapters != null) {
+            val isChapterDownloaded = currentChapters.currChapter.pageLoader is DownloadPageLoader
+            val currentChapter = currentChapters.currChapter
+            val totalPages = currentChapter.pages?.size ?: 0
             currentChapters.unref()
-            saveReadingProgress(currentChapters.currChapter)
+            saveReadingProgress(currentChapter)
+            if (totalPages != 0 && (currentChapter.chapter.last_page_read + 1.0) / totalPages > 0.33 || isAnyPrevChapterDownloaded) {
```
Saud-97, line 188: I don't see the need to check current progress to be at least 33% to queue the next chapters download. Keeping the second condition only will be better imo. Keeping the 33% condition will be meaningful if the feature is download "ahead" during reading.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -413,6 +426,43 @@ class ReaderPresenter(
         }
     }
 
+    private fun downloadNextChapters(isChapterDownloaded: Boolean) {
+        val manga = manga ?: return
+        val chaptersNumberToDownload = preferences.autoDownloadAfterReading().get()
+        if (chaptersNumberToDownload == 0 || !manga.favorite) return
```
Saud-97, line 432: Is the second condition `!manga.favorite` needed?

app/src/main/res/values/strings.xml
```diff
@@ -408,11 +408,18 @@
     <string name="pref_category_auto_download">Auto-download</string>
     <string name="pref_download_new">Download new chapters</string>
     <string name="pref_download_new_categories_details">Manga in excluded categories will not be downloaded even if they are also in included categories.</string>
+    <string name="download_ahead">Download ahead</string>
+    <string name="auto_download_after_reading">Download next unread chapters after reading</string>
+    <plurals name="next_unread_chapters">
+        <item quantity="one">Next unread chapter</item>
+        <item quantity="other">Next %d unread</item>
+    </plurals>
+    <string name="download_ahead_info">Download ahead only applies to entries in library and if the last chapter read was downloaded</string>
```
Saud-97, line 417: It would be better if this was reworded to be clearer and mention that it only triggers on reader exit.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
Riztard: right, if for example we set to "2" it wont download the third one right
or is there any example case where it needed that

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
Riztard: i think so, people might just want to check something or only reading 1/few chapter if not in library

Riztard:
wait this didnt trigger at the end of chapter transition? but on reader exit?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
nzoba: Yeah, as Riztard said
Almost the same for the current progress at 33% (random value), it's to avoid downloading in case of miss-clicking or if you just want to check something at the beginning of a chapter

nzoba:
Well yes, I know it's far from being optimal and less useful than `while reading` but it's either that or :
1. We prevent the user from going to the next chapter while it's not downloaded (= bad UX)
2. We accept a potential double traffic that may be quite bad for the sources, especially the slower ones (= bad for sources)

If you have any idea to avoid these 2 problems while downloading directly in the reader, I would happily take it

Saud-97:
> Well yes, I know it's far from being optimal and less useful than `while reading` but it's either that or :
> 
> 1. We prevent the user from going to the next chapter while it's not downloaded (= bad UX)
> 2. We accept a potential double traffic that may be quite bad for the sources, especially the slower ones (= bad for sources)
> 
> If you have any idea to avoid these 2 problems while downloading directly in the reader, I would happily take it

So far the best possible outcome is that if we could solve the double traffic by for example download ahead the next.next chapter and implement a download kill switch when the reader catches up to the downloaded images. Of course this will also require your other PR to be implemented https://github.com/Jays2Kings/tachiyomiJ2K/pull/1176. Or if there is any other similar ways to handle the duplicate requests. 

Honestly in current status I don't really see the benefit of having the feature without it working as "during" instead of after. You are just saving the user from manually clicking the download action when he already left the reader.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -413,6 +426,43 @@ class ReaderPresenter(
...
+            downloadChapters(chaptersToDownload)
+        }
+    }
+
+    private fun getNextUnreadChaptersSorted(includeCurrentChapter: Boolean): List<ChapterItem> {
+        val currentChapterId = getCurrentChapter()?.chapter?.id
+
+        val chapters = chapterList.map { ChapterItem(it.chapter, manga!!) }
+            .filter { !it.read || it.id == currentChapterId }
+            .distinctBy { it.name }
```
AntsyLich, line 451: Why?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -810,5 +860,10 @@ class ReaderPresenter(
     companion object {
         // Safe theoretical max filename size is 255 bytes and 1 char = 2-4 bytes (UTF-8)
         private const val MAX_FILE_NAME_BYTES = 250
+
+        /**
+         * To check if any chapter of the current reading session was downloaded
+         */
+        private var isAnyPrevChapterDownloaded = false
```
AntsyLich, line 867: why in companion object?

app/src/main/res/values/strings.xml
```diff
@@ -408,11 +408,18 @@
     <string name="pref_category_auto_download">Auto-download</string>
     <string name="pref_download_new">Download new chapters</string>
     <string name="pref_download_new_categories_details">Manga in excluded categories will not be downloaded even if they are also in included categories.</string>
+    <string name="download_ahead">Download ahead</string>
+    <string name="auto_download_after_reading">Download next unread chapters after reading</string>
+    <plurals name="next_unread_chapters">
+        <item quantity="one">Next unread chapter</item>
+        <item quantity="other">Next %d unread</item>
+    </plurals>
+    <string name="download_ahead_info">Download ahead is triggered after exiting the reader, it only applies to entries in library and if the most recently read chapter was downloaded</string>
```
AntsyLich, line 417:
```suggestion
    <string name="download_ahead">Download ahead</string>
    <string name="auto_download_after_reading">After reading download</string>
    <plurals name="next_unread_chapters">
        <item quantity="one">Next unread chapter</item>
        <item quantity="other">Next %d unread chapters</item>
    </plurals>
    <string name="download_ahead_info">Chapters are added to download queue after exiting the reader.\nOnly works on entries in library and if the previous read chapter was downloaded</string>
```

AntsyLich (changes requested):

Riztard:
> Well yes, I know it's far from being optimal and less useful than `while reading` but it's either that or :
> 
> 1. We prevent the user from going to the next chapter while it's not downloaded (= bad UX)
> 2. We accept a potential double traffic that may be quite bad for the sources, especially the slower ones (= bad for sources)
> 
> If you have any idea to avoid these 2 problems while downloading directly in the reader, I would happily take it

make it only if the next chapter is downloaded?

if need to exit the reader, people can just download the next 5/10 ch and it will be simpler right?

Saud-97:
Yeah it looks much better now. And to mitigate the double traffic issue even more I recommend doing a check if the current chapter being loaded is also still being downloaded to pause the download for that chapter and resume it on reader exit (similar to your previous after reading commit). Since when it resumes it will just download using the cache instead of going to the source for a second time and cause extra traffic.

Riztard:
> Yeah it looks much better now. And to mitigate the double traffic issue even more I recommend doing a check if the current chapter being loaded is also still being downloaded to pause the download for that chapter and resume it on reader exit (similar to your previous after reading commit). Since when it resumes it will just download using the cache instead of going to the source for a second time and cause extra traffic.

just ignore? so it will load current chapter and download 3rd chapter

vengiare:
I feel like feature should be called `While reading automatically download`, just like `After reading automatically delete`. Current one can be confusing.

Though I don't know if the dialog also should use numbers as words (like `After reading automatically delete` dialog) or plain numbers (like `Manga screen > Download dropdown`)

jferdous70:
> I feel like feature should be called `While reading automatically download`, just like `After reading automatically delete`. Current one can be confusing.
> 
> Though I don't know if the dialog also should use numbers as words (like `After reading automatically delete` dialog) or plain numbers (like `Manga page > Download dropdown`)

I feel your suggestion sounds much worse than OP, much more awkward. Don't think there's a reason to change this. 

In fact, _After reading automaticaly delete_ sounds bad too. Should change that instead imo

vengiare:
> "Download next unread chapters while reading" "Next 3 chapters" 

This just sounds really confusing. Whereas

> "After reading automatically delete" "Third to last read chapter"
> "While reading automatically download" "Next 3 chapters"

... makes sense to read right away.

nzoba:
I agree, `While reading automatically download` alone sounds like Yoda's talking but it makes sense if you read with the options.
Imo, `Auto download while reading` looks better but I don't really mind

Btw I added a few things :
1. When you start reading a chapter **pending but not yet downloaded**, it stops (removes) the chapter's download
2. If you exit the reader mid-chapter, the chapter is added back to the queue directly after the chapter currently being downloaded
3. If you exit at the end of the chapter or continue to the next one same as 2. Except if you have the setting `After reading auto delete` set to last chapter, in this case the chapter won't be downloaded

AntsyLich:
Supersedes #3296
Closes #3295

Riztard:
> I agree, `While reading automatically download` alone sounds like Yoda's talking but it makes sense if you read with the options. Imo, `Auto download while reading` looks better but I don't really mind
> 
> Btw I added a few things :
> 
> 1. When you start reading a chapter **pending but not yet downloaded**, it stops (removes) the chapter's download
> 2. If you exit the reader mid-chapter, the chapter is added back to the queue directly after the chapter currently being downloaded
> 3. If you exit at the end of the chapter or continue to the next one same as 2. Except if you have the setting `After reading auto delete` set to last chapter, in this case the chapter won't be downloaded

what will happen in case like this:
1. there is some failed/paused chapter(another manga/same) in the download queue. 
- will it restart while adding to queue? 
- will the chapter added going to be 1st in the list so it can start immediately? (if there is some pending download from same manga)
2. i forgot number 2 😂

nzoba:
It will restart if it vas already running or if the queue was empty. If it was paused it won't
Yes it's added at 1st position

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -179,6 +184,10 @@ class ReaderPresenter(
         if (currentChapters != null) {
             currentChapters.unref()
             saveReadingProgress(currentChapters.currChapter)
+            if (chapterDownload != null) {
+                downloadManager.addDownloadsToStartOfQueue(listOf(chapterDownload!!))
+                chapterDownload = null
+            }
```
AntsyLich, line 190:
```suggestion
            chapterDownload?.let {
                downloadManager.addDownloadsToStartOfQueue(listOf(it))
                it = null
            }
```

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -403,23 +413,62 @@ class ReaderPresenter(
...
+            downloadNextChapters()
+        }
+    }
+
+    private fun downloadNextChapters() {
+        val manga = manga ?: return
+        val nextChapter = viewerChaptersRelay.value?.nextChapter?.chapter ?: return
+        val chaptersNumberToDownload = preferences.autoDownloadWhileReading().get()
+        if (chaptersNumberToDownload == 0 || !manga.favorite) return
+        if (getCurrentChapter()?.pageLoader !is DownloadPageLoader) return
```
AntsyLich, line 436: move this to top too for early exit

AntsyLich (changes requested):

Sazaju:
Nice to see it going again. Thanks a lot. {^_^}b

AntsyLich:
Can extract download icon in transition view as a separate pr

nzoba:
> Can extract download icon in transition view as a separate pr

Done

Riztard:
also btw what will happen if the chapter filter is applied?
will it download 2a or 3

for example:
```
ch 1
ch 2
ch 2a (hidden by filter)
ch 3
```

nzoba:
> also btw what will happen if the chapter filter is applied? will it download 2a or 3
> 
> for example:
> 
> ```
> ch 1
> ch 2
> ch 2a (hidden by filter)
> ch 3
> ```

Only the visible chapters are automatically downloaded, the hidden chapters are ignored. In your example, 2a won't be downloaded but 3 will be

AntsyLich:
So it's like downloading next chapter from download menu at action bar

nzoba:
Exactly, same logic

arynthernium:
It's been a month, have there been any updates to this? I don't want to sound impatient, just wondering where activity is on this.

nzoba:
No idea, the feature itself is done, now it's up to the devs to decide if it needs further reviews or can be merged
They are probably quite busy rewriting the app architecture for now, so it may take a while

arkon:
I think I skimmed through it a few days ago and it seemed fine. Can you just resolve the merge conflicts?

nzoba:
> I think I skimmed through it a few days ago and it seemed fine. Can you just resolve the merge conflicts?

Done

arynthernium:
Awesome. Looks good to merge whenever the devs take a look. Thanks for this feature- it's one of the things I've looked for in tachiyomi extensively, and this makes me glad tachiyomi is open source and has the developer community surrounding it. This likely wouldn't ever be possible if it weren't community built. I look forward to seeing this in the app :)

PR #7241, opened 2022-06-02
Make `syncChaptersWithSource` use sqldelight

AntsyLich:
Will break chapter list live update on current UI


AntsyLich:
Gonna rebase on ivaniskandar's wip so he has less rebasing to do

PR #7295, opened 2022-06-12
Use SQLDelight in Backup/Restore

ghostbear:
To be merged after #7293

Tested with a backup created on 0.13.4, works.
Also tested with a backup created with the changes by this PR on 0.13.4, works.
Newer backup will probably break on older version of the app when the user has values greater than 2,147,483,647


app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupCreatorJob.kt
```diff
@@ -36,9 +37,11 @@ class BackupCreatorJob(private val context: Context, workerParams: WorkerParamet
 
         context.notificationManager.notify(Notifications.ID_BACKUP_PROGRESS, notifier.showBackupProgress().build())
         return try {
-            val location = FullBackupManager(context).createBackup(uri, flags, isAutoBackup)
-            if (!isAutoBackup) notifier.showBackupComplete(UniFile.fromUri(context, location.toUri()))
-            Result.success()
+            runBlocking {
```
arkon, line 40: Could make this into a `CoroutineWorker`

arkon (approved):

PR #7308, opened 2022-06-14
Communication for the readers

Toweshi:
Let us make an account and make a comment section and discussion chats in every mangas that can add friend every reader thank you


arkon:
1. Make an issue for requests, not a pull request.
2. Already declined before.

PR #7311, opened 2022-06-15
Add ability to filter by scanlator (closes #969)

### LINKED ISSUE #969: Filter chapters list by scanlator
Since scanlator was added on chapter list can we have a way to hide/select specific groups?

CVIUS:
Taken from SY 

I haven't tested thoroughly in backup/restore, but filtering works😂
The filter doesn't change state tho. I give up (╯° °)╯︵ ┻━┻


AntsyLich:
> Also test things. We dont need more @AntsyLich style PRs

It was only one time zoomeeyyyees emoji

PR #7317, opened 2022-06-16
Fix freezing on migrating manga

### LINKED ISSUE #7305: Migrating causes Tachiyomi too freeze.
### Steps to reproduce

1. Migrate a manga.
2. On the migration screen choose a manga to migrate to.
3. Click migrate.

### Expected behavior

Mifration completes without any issue.

### Actual behavior

The screen gets stuck on loading indefinitely and it seems like you can click around and open other manga while this happens. If you try to go back Tachiyomi freezes and after a while you get the app not responding dialogue.

### Crash logs

_No response_

### Tachiyomi version

0.13.4-4466

### Android version

Android version: 11 (SDK 30) 

### Device

Redmi Note 9 Pro Max

### Other details

_No response_

### Acknowledgements

jmir1:
Fixes #7305


jobobby04:
This is probably a hit or miss fix based on what I just said in the issue report. Since everything is already happening in the IO dispatcher its just happens to not switch threads since the TransactionDispatcher uses the IO thread pool.

jmir1:
> This is probably a hit or miss fix based on what I just said in the issue report. Since everything is already happening in the IO dispatcher its just happens to not switch threads since the TransactionDispatcher uses the IO thread pool.



> This is probably a hit or miss fix based on what I just said in the issue report. Since everything is already happening in the IO dispatcher its just happens to not switch threads since the TransactionDispatcher uses the IO thread pool.

then wouldn't the solution be to lift the call to `syncEpisodesWithSource` out of the transaction?
or is that risky somehow?

jobobby04:
That's probably the best solution. Or remove the db.inTransaction. Either would work fine I think. I don't think the db.inTransaction is required in the first place since migration is its own menu and there isn't a need to worry about conflicting inserts between different places.

PR #7321, opened 2022-06-17
Use custom QueryPagingSource

ghostbear:
- Adds placeholders to make the list jump around less
- Fixes the issue where SQLDelights QueryPagingSource would throw IndexOutOfBounds

https://user-images.githubusercontent.com/6576096/174432964-a0145fac-6339-4158-b633-78e2d568196a.mp4


app/src/main/java/eu/kanade/presentation/history/components/HistoryItem.kt
```diff
@@ -0,0 +1,137 @@
...
+            Icon(
+                imageVector = Icons.Outlined.Delete,
+                contentDescription = stringResource(R.string.action_delete),
+                tint = MaterialTheme.colorScheme.onSurface,
+            )
+        }
+    }
+}
+
+@Composable
```
arkon, line 97: FWIW: https://google.github.io/accompanist/placeholder/#shimmer

app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
```diff
@@ -107,6 +98,34 @@ fun HistoryContent(
     var removeState by remember { mutableStateOf<HistoryWithRelations?>(null) }
 
     val scrollState = rememberLazyListState()
+
+    val gradient = listOf(
```
arkon, line 102: Probably want to pull these out to somewhere generic for reuse if we keep these.

app/src/main/java/eu/kanade/data/QueryPagingSource.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    override fun getRefreshKey(state: PagingState<Long, RowType>): Long? {
+        return state.anchorPosition?.let { anchorPosition ->
+            val anchorPage = state.closestPageToPosition(anchorPosition)
+            anchorPage?.prevKey ?: anchorPage?.nextKey
+        }
+    }
+
+    override fun queryResultsChanged() {
+        invalidate()
```
arkon, line 72: Is it necessary to invalidate everything? Does this trigger even if it's a single item being changed (state, not position)?

app/src/main/java/eu/kanade/presentation/history/components/HistoryItem.kt
ghostbear: I feel like doing it manually is fine

app/src/main/java/eu/kanade/data/QueryPagingSource.kt
ghostbear: Yes as some things may be on a page edge. It only triggers on database changes so if an item is inserted, updated deleted, etc

PR #7330, opened 2022-06-18
Convert edit cover function to suspend function

ivaniskandar:


app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
```diff
@@ -299,31 +301,20 @@ class MangaPresenter(
...
-                    } else if (manga.favorite) {
-                        coverCache.setCustomCoverToCache(manga, it)
-                        manga.updateCoverLastModified(db)
-                    }
-                    true
+    fun editCover(context: Context, data: Uri) {
+        presenterScope.launchIO {
+            context.contentResolver.openInputStream(data)?.use {
+                try {
+                    val result = manga.toDomainManga()!!.editCover(context, it, coverCache = coverCache)
```
arkon, line 311: Does the `coverCache` actually need to be passed if it's being injected anyway?

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -692,36 +693,28 @@ class ReaderPresenter(
...
-                            coverCache.setCustomCoverToCache(manga, it)
-                            manga.updateCoverLastModified(db)
-                            SetAsCoverResult.Success
-                        } else {
-                            SetAsCoverResult.AddToLibraryFirst
-                        }
-                    }
+        presenterScope.launchIO {
+            val result = try {
+                manga.editCover(context, stream(), coverCache = coverCache)
```
arkon, line 703: Same here.

arkon (approved):

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
ivaniskandar: Right I guess it doesn't. I'll remove it

PR #7342, opened 2022-06-19
ChaptersSettingsSheet: Single source of truth and use new manga class

ivaniskandar:
Currently breaks initial settings state until the source of truth is properly updated which may be closed in #7244.


app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -45,15 +47,99 @@ data class Manga(
...
+        const val CHAPTER_DISPLAY_NAME = 0x00000000L
+        const val CHAPTER_DISPLAY_NUMBER = 0x00100000L
+        const val CHAPTER_DISPLAY_MASK = 0x00100000L
+    }
+}
+
+enum class TriStateFilter {
+    DISABLED, // Disable filter
+    ENABLED_IS, // Enabled with "is" filter
+    ENABLED_NOT, // Enabled with "not" filter
```
arkon, line 135: `INCLUDE`/`EXCLUDE`? I'm not sure these names are that intuitive, or if it's just because native English is more broken than logical English.

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -45,15 +47,99 @@ data class Manga(
...
+
+        const val CHAPTER_SHOW_BOOKMARKED = 0x00000020L
+        const val CHAPTER_SHOW_NOT_BOOKMARKED = 0x00000040L
+        const val CHAPTER_BOOKMARKED_MASK = 0x00000060L
+
         const val CHAPTER_SORTING_SOURCE = 0x00000000L
         const val CHAPTER_SORTING_NUMBER = 0x00000100L
         const val CHAPTER_SORTING_UPLOAD_DATE = 0x00000200L
         const val CHAPTER_SORTING_MASK = 0x00000300L
+        const val CHAPTER_SORTING_DESC = 0x00000000L
```
arkon, line 124: What's the difference between this and `CHAPTER_SORT_DESC`?

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
ivaniskandar: I was just copy-pasting the stuff 🤣 I agree it's confusing but idk what to use either.

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
ivaniskandar: One is for the sorting mode and another one is the sorting direction. Again, this was a direct copy-paste from the old class but it can be changed.

PR #7345, opened 2022-06-20
Regenerate drawables

TacoTheDank:
- Regenerated all icons to latest material design using the Android Studio vector tool
- Replaced materialdesignicons.com versions of the double left and right navigation drawables with official Google Material versions
- Removed duplicate overflow drawable (duplicate of more_vert)

Use https://yqnn.github.io/svg-path-editor/ and paste the SVG paths to see the differences (if you want).


TacoTheDank:
Alrighty 👍

PR #7354, opened 2022-06-21
ChapterDownloadView: Convert to compose

ivaniskandar:
Removed long click action


app/src/main/java/eu/kanade/presentation/components/ChapterDownloadIndicator.kt
```diff
@@ -0,0 +1,145 @@
...
+    modifier: Modifier = Modifier,
+    downloadState: Download.State,
+    downloadProgress: Int,
+    onClick: (ChapterDownloadAction) -> Unit,
+) {
+    Box(modifier = modifier, contentAlignment = Alignment.Center) {
+        CompositionLocalProvider(LocalMinimumTouchTargetEnforcement provides false) {
+            val isDownloaded = downloadState == Download.State.DOWNLOADED
+            val isDownloading = downloadState != Download.State.NOT_DOWNLOADED
+            val (menuExpanded, onExpanded) = remember(downloadState) { mutableStateOf(false) }
```
ghostbear, line 44:
```suggestion
            val isMenuExpanded by remember(downloadState) { mutableStateOf(false) }
```

app/src/main/java/eu/kanade/presentation/components/ChapterDownloadIndicator.kt
```diff
@@ -0,0 +1,145 @@
...
+                    if (isDownloading) {
+                        val indeterminate = downloadState == Download.State.QUEUE ||
+                            (downloadState == Download.State.DOWNLOADING && downloadProgress == 0)
+                        if (indeterminate) {
+                            arrowColor = strokeColor
+                            CircularProgressIndicator(
+                                modifier = progressIndicatorModifier,
+                                color = strokeColor,
+                                strokeWidth = IndicatorStrokeWidth,
+                            )
```
ghostbear, line 89: Couldn't this be extracted with all the other `CircularProgressIndicator`s to a composable function?

ivaniskandar: These doesn't have any common similarities so I don't think so

PR #7366, opened 2022-06-25
Use sqldelight for direct db calls in `MangaPresenter`

AntsyLich:


app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
```diff
@@ -678,16 +709,21 @@ class MangaPresenter(
...
-                                db.insertTrack(track).executeAsBlocking()
+                                val track = it.track ?: return@async null
+
+                                val updatedTrack = it.service.refresh(track)
 
-                                if (it.service is EnhancedTrackService) {
+                                val domainTrack = updatedTrack.toDomainTrack() ?: return@async null
+                                insertTrack.await(domainTrack)
+
+                                (it.service as? EnhancedTrackService).let { _ ->
```
arkon, line 721: Shouldn't it be `?.let`?

AntsyLich:
Pr still needs testing. Forgot to make this a draft since I was in a hurry.

PR #7389, opened 2022-06-27
RateLimitInterceptor: ignore canceled calls

stevenyomi:
This prevents large numbers of canceled thumbnail requests from jamming the interceptor and slowing down subsequent requests.

Probably backport to Stable?


arkon:
Probably want to do it for the specific host interceptor too.

stevenyomi:
> Probably want to do it for the specific host interceptor too.

Done.

PR #7411, opened 2022-06-29
DownloadProvider: Only provide necessary info and not whole chapter/m…

AntsyLich:
…anga class


app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
```diff
@@ -65,23 +65,31 @@ class DownloadCache(
...
-            val source = sourceManager.getOrStub(manga.source)
-            return provider.findChapterDir(chapter, manga, source) != null
+            val source = sourceManager.getOrStub(sourceId)
+            return provider.findChapterDir(chapterName, chapterScanlator, mangaTitle, source) != null
         }
 
         checkRenew()
 
-        val sourceDir = rootDir.files[manga.source]
+        val sourceDir = rootDir.files[sourceId]
```
arkon, line 88: You need to update `renew` so that the keys are actually the source ID. That said, it should probably still be the name given that the actual folder names are the source names and not the IDs.

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
AntsyLich: `manga.source` is the source id

PR #7412, opened 2022-06-29
Reimplement chapter download indicator longpress

AntsyLich:


arkon:
For reference: https://issuetracker.google.com/issues/201550983

PR #7454, opened 2022-07-04
Use Compose for Category screen

ghostbear:
| Categories | Create Dialog | Rename Dialog | Delete Dialog |
|-------|-------|-------|-------|
| image | image | image | image |

Currently missing drag n drop but could be added either with an external well-maintained library or wait until Compose support it officially


vengiare:
Can you tap the title to edit category name, or is it explicitly just the edit button?

PR #7475, opened 2022-07-07
Reset lastPageRead when chapter is marked unread

ghostbear:
Creates interactor to change read status on chapters as it currently behaves differently on different screens and it even behaves differently than it did before Manga screen rewrite and places where SQLDelight is used

Behaviour before any changes https://github.com/tachiyomiorg/tachiyomi/blob/e3b1053c03da17c8c1b66f1914251707134e84a9/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt#L458-L464


app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
```diff
@@ -525,12 +527,10 @@ class MangaPresenter(
      */
     fun markChaptersRead(chapters: List<DomainChapter>, read: Boolean) {
         presenterScope.launchIO {
-            val modified = chapters.filterNot { it.read == read }
-            modified
-                .map { ChapterUpdate(id = it.id, read = read) }
-                .let { updateChapter.awaitAll(it) }
+            val filtered = chapters.filterNot { it.read == read }
+            setReadStatus.awaitAll(filtered, read)
             if (read && preferences.removeAfterMarkedAsRead()) {
```
arkon, line 532: Isn't this handled in SetReadStatus?

ghostbear: Yes, and I fixed it right before this comment

PR #7493, opened 2022-07-09
Reinstate in-app language picker

ivaniskandar:
Also add required config for android 13 system picker

~Draft because~ minor layout direction issue on compose screens. Current workaround is to do full restart/manual activity recreate ([b/236538894](https://issuetracker.google.com/issues/236538894))


arkon:
Will have to rebase since I merged #7754.

PR #7533, opened 2022-07-15
Add Kdrama

beyond3dartist:
please ad Kdrama as well from kissasian


CarlosEsco:
This isn't a video player app.

PR #7535, opened 2022-07-15
MangaPresenter: Incognito and downloaded only label related tweaks

AntsyLich:
When manga screen takes time to load incognito and downloaded only label doesn't show up. It's fixed in the pr.


arkon:
I feel like we need to handle the toolbar/associated things differently once everything's closer to being Compose-only, but I guess this is fine in the short term...

PR #7551, opened 2022-07-17
Fix concurrent download

f1998f1998:
Fix concurrent download by adding .subscribeOn(Schedulers.io())


f1998f1998:
need to lower tjreads

PR #7552, opened 2022-07-17
fix concurrent download

f1998f1998:
Fix concurrent download by adding .subscribeOn(Schedulers.io()
and lower concurrency to 2.


app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -342,7 +342,7 @@ class Downloader(
             .flatMap { download.source.fetchAllImageUrlsFromPageList(it) }
             // Start downloading images, consider we can have downloaded images already
             // Concurrently do 5 pages at a time
```
vetleledaal, line 344:
```suggestion
            // Concurrently do 2 pages at a time
```

PR #7557, opened 2022-07-18
Use Compose for Library screen

ghostbear:
- Move Pager to Compose
- Move AppBar to Compose
- Use Stable interface for state
- Use pills for no. of manga in category instead of (x)
- Tab indicator changed uses the MD3 library so could be changed in a future update of it

| - | - | - |
|---|---|---|
| image | image | image |


app/src/main/java/eu/kanade/presentation/library/components/LibraryCompactGrid.kt
```diff
@@ -29,10 +34,23 @@ fun LibraryCompactGrid(
...
     onClick: (LibraryManga) -> Unit,
     onLongClick: (LibraryManga) -> Unit,
+    searchQuery: String?,
+    onGlobalSearchClicked: () -> Unit,
 ) {
     LazyLibraryGrid(
         columns = columns,
     ) {
+        item(span = { GridItemSpan(maxLineSpan) }) {
+            if (searchQuery.isNullOrEmpty().not()) {
```
arkon, line 44: Considering this is repeated, it could be pulled out as a composable.

app/src/main/java/eu/kanade/presentation/library/components/LibraryContent.kt
```diff
@@ -0,0 +1,126 @@
...
+                onToggleSelection(manga)
+            }
+        }
+        val onLongClickManga = { manga: LibraryManga ->
+            onToggleSelection(manga)
+        }
+
+        SwipeRefresh(
+            state = rememberSwipeRefreshState(isRefreshing = false),
+            modifier = Modifier.nestedScroll(nestedScrollInterop),
```
arkon, line 86: Do we still need this?

app/src/main/java/eu/kanade/presentation/library/components/LibraryContent.kt
```diff
@@ -0,0 +1,126 @@
...
+            onRefresh = onRefresh,
+            indicator = { s, trigger ->
+                SwipeRefreshIndicator(
+                    state = s,
+                    refreshTriggerDistance = trigger,
+                )
+            },
+        ) {
+            if (state.searchQuery.isNullOrEmpty() && isLibraryEmpty) {
+                val context = LocalContext.current
```
arkon, line 96: Can use `LocalUriHandler` instead.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -411,18 +425,18 @@ class LibraryPresenter(
...
     }
 
     /**
      * Get the categories from the database.
      *
      * @return an observable of the categories.
      */
-    private fun getCategoriesObservable(): Observable<List<Category>> {
-        return getCategories.subscribe().asObservable()
+    private fun getCategoriesObservable(): Flow<List<Category>> {
```
arkon, line 438: Should rename these functions.

arkon (commented):
Some minor comments, but I'll merge this first so that I don't need to re-review a bunch of stuff again.

PR #7575, opened 2022-07-21
Add downloaded icon in TransitionView when chapter is downloaded

nzoba:
Add icon in the `TransitionView` to know if the next/previous chapter is downloaded without exiting the reader (almost like in the J2K [PR](https://github.com/Jays2Kings/tachiyomiJ2K/pull/1174)). The icon is visible next to the chapter's name if it is downloaded.

TransitionView icon


Riztard:
what happen if the chapter title is very long

nzoba:
It adds a new line, the icon is just appended to the end of the chapter title like text.

TransitionView icon with long title

CrepeTF:
I recommend using a different icon since this one implies an action. The icon named offline_pin would work better imo

nzoba:
Sure, I don't mind

TransitionView changed icon

app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/ReaderTransitionView.kt
```diff
@@ -21,31 +30,44 @@ class ReaderTransitionView @JvmOverloads constructor(context: Context, attrs: At
         layoutParams = LayoutParams(LayoutParams.WRAP_CONTENT, LayoutParams.WRAP_CONTENT)
     }
 
-    fun bind(transition: ChapterTransition) {
+    fun bind(transition: ChapterTransition, downloadManager: DownloadManager, manga: Manga?) {
```
arkon, line 33: This definitely feels a bit weird in how it's being passed around, but it'll probably end up getting rewritten at some point soon anyhow.

PR #7600, opened 2022-07-24
Add fast scroller to Library screen

ghostbear:


app/src/main/java/eu/kanade/presentation/components/LazyGrid.kt
```diff
@@ -0,0 +1,62 @@
...
+import androidx.compose.foundation.lazy.grid.rememberLazyGridState
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun FastScrollLazyVericalGrid(
```
ivaniskandar, line 20: verical

Issue #7401, opened 2022-06-28
Covers have shrunk


specterflare:
### Steps to reproduce

Open up any manga

### Expected behavior

The cover should be at the size it always is

### Actual behavior

The cover looks much smaller
![Screenshot_20220627-214632](https://user-images.githubusercontent.com/61708534/176069666-85e4c173-b49c-4429-bffa-4ed593965ad3.png)
Than it used to
![Screenshot_20220607-214839](https://user-images.githubusercontent.com/61708534/176069786-bf07157e-82a7-4b07-82d8-c0e89ee06afb.png)

### Tachiyomi version

r4524

### Android version

10.0

### Device

Vankyo P40 Tablet

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] I have updated the app to version **[0.13.4](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

AntsyLich:
It's the opposite for me.

r4524:
![Screenshot_20220628-005046](https://user-images.githubusercontent.com/59261191/176103240-fad66675-aadc-4d83-9074-7c2f7c32b48e.png)


0.13.4:
![Screenshot_20220628-004744](https://user-images.githubusercontent.com/59261191/176103098-2403003c-3439-4046-8c9b-4d92e9956ffd.png)

specterflare:
Could one of the repository maintainers comment on wether this is indeed a bug or an intended design change? If not the latter, it seems like a non-trivial issue.

arkon:
The now different size in portrait isn't intentional.

PR #7625, opened 2022-07-27
MangaScreen large UI tweaks

ivaniskandar:
* Fix column sizing (equal width until left column width is 450dp)
* Changed cover fill fraction (#7401)

Not entirely related
* Removed bottom rounded bg of bottom action menu


app/src/main/java/eu/kanade/presentation/components/MangaBottomActionMenu.kt
```diff
@@ -67,7 +68,7 @@ fun MangaBottomActionMenu(
         val scope = rememberCoroutineScope()
         Surface(
             modifier = modifier,
-            shape = MaterialTheme.shapes.large,
+            shape = MaterialTheme.shapes.large.copy(bottomEnd = ZeroCornerSize, bottomStart = ZeroCornerSize),
```
ghostbear, line 71: You could also add this tweak to LibraryBottomActionMenu

AntsyLich: Why not just merge those two by adding another icon for categories

arkon: We should really consolidate all of the various action menus/app bars/etc., but that can happen separately.

ghostbear: > Why not just merge those two by adding another icon for categories

I think it would be smarter to extract the haptic part of this composable, instead of having one massive composable

PR #7639, opened 2022-07-29
Use Compose on Clear Database screen

ghostbear:


arkon:
I forgot I was even working on this at one point ☠️

PR #7655, opened 2022-07-31
Filter out empty genres before saving manga to database

stevenyomi:
Currently empty genres will be shown. `SManga::toMangaInfo()` is the only source that could possibly cause this problem.

Previously:
https://github.com/tachiyomiorg/tachiyomi/blob/26ddc6e3aa1a48f091fedc0cac32dbca6c1834ff/app/src/main/java/eu/kanade/tachiyomi/ui/manga/info/MangaInfoHeaderAdapter.kt#L293

Currently:
https://github.com/tachiyomiorg/tachiyomi/blob/483b204fb5b6790fb6fe59ffd80345cd484ca241/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt#L362


53jk1:
Are you sure that if you remove an anonymous function from one place and put it in another, one of the functionalities will not stop working because of this?

PR #7659, opened 2022-07-31
Remove deprecated LibrarySort

ghostbear:


53jk1 (approved):

app/src/main/java/eu/kanade/tachiyomi/Migrations.kt
```diff
@@ -105,9 +104,9 @@ object Migrations {
                 val oldSortingMode = prefs.getInt(PreferenceKeys.librarySortingMode, 0)
 
                 @Suppress("DEPRECATION")
```
stevenyomi, line 106: I guess this `@Suppress` can be removed? Also the one in line 203.

arkon (commented):
It'd be worth just leaving comments on what the numbers mean for better context.

app/src/main/java/eu/kanade/tachiyomi/Migrations.kt
```diff
@@ -104,10 +103,9 @@ object Migrations {
                 // Reset sorting preference if using removed sort by source
                 val oldSortingMode = prefs.getInt(PreferenceKeys.librarySortingMode, 0)
 
-                @Suppress("DEPRECATION")
-                if (oldSortingMode == LibrarySort.SOURCE) {
+                if (oldSortingMode == 5) {
```
ghostbear, line 106:
```suggestion
                if (oldSortingMode == 5 /* SOURCE */) {
```

app/src/main/java/eu/kanade/tachiyomi/Migrations.kt
```diff
@@ -104,10 +103,9 @@ object Migrations {
...
                     prefs.edit {
-                        putInt(PreferenceKeys.librarySortingMode, LibrarySort.ALPHA)
+                        putInt(PreferenceKeys.librarySortingMode, 0)
```
ghostbear, line 108:
```suggestion
                        putInt(PreferenceKeys.librarySortingMode, 0 /* ALPHABETICAL */)
```

ghostbear (commented):
I will leave you the `when` for comments as what on the right is more or less what the number mean

PR #7663, opened 2022-07-31
Reimplement appbar color overlay on scroll

ivaniskandar:
only on updates and history screen for now, but the required changes on app bar
is there.

also fix missing incognito-downloaded mode indicator on history screen


app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
```diff
@@ -45,11 +46,16 @@ class HistoryPresenter(
     private val deleteHistoryTable: DeleteHistoryTable = Injekt.get(),
     private val removeHistoryById: RemoveHistoryById = Injekt.get(),
     private val removeHistoryByMangaId: RemoveHistoryByMangaId = Injekt.get(),
+    preferences: PreferencesHelper = Injekt.get(),
```
arkon, line 49: Can be private?

app/src/main/java/eu/kanade/presentation/components/AppBar.kt
```diff
@@ -51,19 +48,63 @@ fun AppBar(
...
+        incognitoMode = incognitoMode,
+        scrollBehavior = scrollBehavior,
+    )
+}
+
+@Composable
+fun AppBar(
+    modifier: Modifier = Modifier,
+    // Title
+    titleContent: @Composable () -> Unit,
```
arkon, line 85: Is this in preparation for different content or something?

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
ivaniskandar: oopsie

app/src/main/java/eu/kanade/presentation/components/AppBar.kt
ivaniskandar: It's currently used for history search

app/src/main/java/eu/kanade/tachiyomi/ui/recent/history/HistoryPresenter.kt
ivaniskandar: wait no this is just a parameter not a property

arkon: Whoops, you right

PR #7664, opened 2022-07-31
fix(if): added checker if the file still exists

53jk1:


arkon:
This is realistically already handled semi-elegantly by the reader, so the old TODO can just be removed.

ghostbear (commented):
I just tried this and I get errors on every page even though the files exist

image

PR #7665, opened 2022-07-31
fix(delete): fake status update

53jk1:


ghostbear (commented):
This removes the items from the updates list literally. Which is not supposed to happen. The item should remain and its downloaded state change to not downloaded which the code you removed handled just fine

PR #7672, opened 2022-08-02
Save read duration to backup

nzoba:
Probably not necessary to add it to `BrokenBackupHistory` but well :shrug:


app/src/main/java/eu/kanade/tachiyomi/data/backup/full/models/BackupHistory.kt
```diff
@@ -7,10 +7,12 @@ import kotlinx.serialization.protobuf.ProtoNumber
 data class BrokenBackupHistory(
     @ProtoNumber(0) var url: String,
     @ProtoNumber(1) var lastRead: Long,
+    @ProtoNumber(2) var readDuration: Long,
```
ghostbear, line 10:
```suggestion
    @ProtoNumber(2) var readDuration: Long = 0,
```

app/src/main/java/eu/kanade/tachiyomi/data/backup/full/models/BackupHistory.kt
```diff
@@ -7,10 +7,12 @@ import kotlinx.serialization.protobuf.ProtoNumber
...
 )
 
 @Serializable
 data class BackupHistory(
     @ProtoNumber(1) var url: String,
     @ProtoNumber(2) var lastRead: Long,
+    @ProtoNumber(3) var readDuration: Long,
```
ghostbear, line 17:
```suggestion
    @ProtoNumber(3) var readDuration: Long = 0,
```

ghostbear (changes requested):
We will need to give the new a default value otherwise it will break old backups

image

PR #7676, opened 2022-08-02
Move default category into database

ghostbear:
- DisplayMode and Sort(Mode/Direction)Mode are all handled in the database and the preferences are used when turning off per category
- Remove the old StorIO model for Category
- Per category is enabled if the flags are different from one another


app/src/main/java/eu/kanade/tachiyomi/ui/setting/SettingsLibraryController.kt
```diff
@@ -125,6 +128,14 @@ class SettingsLibraryController : SettingsController() {
             switchPreference {
                 bindTo(preferences.categorizedDisplaySettings())
                 titleRes = R.string.categorized_display_settings
+
+                preferences.categorizedDisplaySettings().asFlow()
+                    .onEach {
+                        if (it.not()) {
+                            resetCategoryFlags.await()
```
arkon, line 135: This is new behaviour that I'm not totally sure is definitely desired? Does it seem valid for people to want to retain the per-category settings even when toggled?

ghostbear: It's an effort to simplify the getting and setting of the display and the sort settings. Instead of having one flag in shared preferences and one in the database, there is only one flag which is the one in the database. While the shared preferences one acts as a reset flag (the flag the user used before starting to use per-category) when the user goes from per-category to non per-category.

I want to think that there are two people. People who use per-category and people who don't use per category. So I would say that there isn't a need to retain the per-category settings when toggled.

Issue #7549, opened 2022-07-16
Double tap Browse to go to last used source


vengiare:
### Describe your suggested feature

Title. Should either "not work" or "default to Local source" if there is no last used source (app first run)

### Other details

Browse is the last navbar item to not have a double tap action so I'm just throwing this idea out there

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.13.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

Safayy:
I'm not sure if it works the same for new features, but as per contributing guidelines I'd like to state I'm working on this.

arkon:
Duplicate of #6515

PR #7712, opened 2022-08-09
Double tap Browse to go to most recent source

Safayy:
Added the feature suggested in issue #7549.


AntsyLich:
Should just go to extensions tab directly instead of opening last used source

Safayy:
I feel like last used source is used far more often than extension settings and therefore fits a shortcut better

app/src/main/java/eu/kanade/tachiyomi/ui/main/MainActivity.kt
```diff
@@ -186,6 +186,13 @@ class MainActivity : BaseActivity() {
                             }
                         }
                     }
+                    R.id.nav_browse -> {
+                        if (router.backstackSize == 1) {
+                            if (preferences.lastUsedSource().defaultValue != preferences.lastUsedSource().get()) {
+                                router.pushController(BrowseSourceController(preferences.lastUsedSource().get()))
```
arkon, line 192: This would be problematic if the last used source is no longer available (not loaded due to 18+ setting, uninstalled, etc.). It'd be ideal to call something in `BrowseController` which actually includes logic to check that.

arkon:
Closing for now, feel free to open another PR if you revisit it.

PR #7715, opened 2022-08-10
Change return value of `SyncChaptersWithSource.await()`

AntsyLich:
`updatedToAdd.subtract(reAdded).toList()` never worked as at this point `updatedToAdd` contained ids from db where `reAdded` had default one. Was the same case before the rewrite.

Removed `toDelete` from return value as it was not being used anywhere


app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
```diff
@@ -174,7 +182,8 @@ class SyncChaptersWithSource(
         // Note that last_update actually represents last time the chapter list changed at all
         updateManga.awaitUpdateLastUpdate(manga.id)
 
-        @Suppress("ConvertArgumentToSet") // See tachiyomiorg/tachiyomi#6372.
-        return Pair(updatedToAdd.subtract(reAdded).toList(), toDelete.subtract(reAdded).toList())
+        val reAddedUrls = reAdded.map { it.url }
```
stevenyomi, line 185:
```suggestion
        val reAddedUrls = reAdded.map { it.url }.toHashSet()
```

PR #7722, opened 2022-08-11
Add back reset action to the Library and History Search toolbars

ghostbear:


app/src/main/java/eu/kanade/presentation/history/components/HistoryToolbar.kt
```diff
@@ -98,6 +102,13 @@ fun HistorySearchToolbar(
         },
         navigationIcon = Icons.Outlined.ArrowBack,
         navigateUp = onClickCloseSearch,
+        actions = {
+            AnimatedVisibility(visible = searchQuery.isNotEmpty()) {
+                IconButton(onClick = onClickResetSearch) {
+                    Icon(Icons.Outlined.Close, contentDescription = "reset")
```
arkon, line 108: Localize the descriptions pls

Issue #6539, opened 2022-02-01
Add a `dontUpdate`/`isOneShot` flag to `SManga` to avoid updating single-chapter titles


alessandrojean:
### Describe your suggested feature

Some sources like `nhentai` and `Hitomi` have unique chapters in the titles that doesn't need to be checked if there's new chapters in the global and manual update.

It would be nice if the `SManga` class had a `dontUpdate` or `isOneShot` flag that, if set to `true`, the app would skip these titles from the updates as it's not necessary since they're single-chapter.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.13.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

arkon:
I imagine this should be an attribute of an SManga?

alessandrojean:
Works too, like a new status maybe?

PR #7730, opened 2022-08-12
Add DownloadDisabledSource interface

AntsyLich:
To disable downloading on some ~~horny~~ sources.


jobobby04:
What ~horny~ sources would use this?

alessandrojean:
Maybe also implement #6539 too?

vetleledaal:
Think this feature might be controversial, leading into fragmentation. Leaving this separate from #6539 is probably better.

PR #7760, opened 2022-08-14
Added tickers in the reader page slider

Re-Tails:
Adds tickers to the page slider in reader view. Default setting is off, under the reader page


arkon (commented):
I think it's okay to just not have a setting and always show it. You can just set it in the XML instead.

PR #7768, opened 2022-08-15
Displays next unread chapter on resume button

Re-Tails:
Displays the chapter number of the next unread chapter when resuming, resolves https://github.com/tachiyomiorg/tachiyomi/issues/7572.


app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -693,3 +700,8 @@ private fun onChapterItemClick(
         else -> onChapterClicked(chapterItem.chapter)
     }
 }
+
+private val chapterFormatter = DecimalFormat(
+    "#.###",
+    DecimalFormatSymbols().apply { decimalSeparator = '.' },
```
alessandrojean, line 706: This isn't ready for i18n, there's some locales that uses `,` instead of `.` as the decimal separator.

AntsyLich:
you forgot to add the button that lets me disable this

arkon:
I'm not sure it actually makes sense to increase the amount of text in the FAB, which will be pretty bad in more verbose languages / smaller screens.

Re-Tails:
> I'm not sure it actually makes sense to increase the amount of text in the FAB, which will be pretty bad in more verbose languages / smaller screens.

An alternative is to just have the chapter number only, without the resume text. That won't look great though admittedly

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
Re-Tails: How are we planning to handle i18n? Used this formatter from app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateNotifier.kt

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -279,7 +285,7 @@ private fun MangaScreenSmallImpl(
                         } else {
                             R.string.action_start
                         }
-                        Text(text = stringResource(id))
+                        Text(text = (stringResource(id) + " " + chapterFormatter.format(nextUnreadChapterNumber.invoke())))
```
stevenyomi, line 288: Some languages like Chinese and Japanese have text both before and after the number, e.g. "第 ## 章".

You also need to handle invalid negative chapter numbers.

stevenyomi (changes requested):

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
Re-Tails: So something like `<resume_start> ## <resume_end>`, yep gonna add in handling fallback for invalid chapter numbers.

stevenyomi: Just do `String.format("Resume ch. %s", chapterNumberString)` and let translators put the `%s` in the right place.

Re-Tails: Oh dang, didn't realise `%s` work for these strings as well. Thanks

arkon:
Someone on Discord suggested adopting the J2K button between the info and chapters, along with a suggestion of showing the FAB only when that button isn't visible.

We could also have a subtitle below the "x chapters" heading that tells you what's next or what you're currently on?

arkon:
Closing for now, I'm not really sold on any of the solutions so far.

PR #7781, opened 2022-08-17
Remove 1.x source models

stevenyomi:


stevenyomi:
> Are you hoping to evolve the existing models in the future, or still swap over to new models?

I guess swapping to the new ones eventually, when the new extension system is polished and it's clear how to migrate to it.

Currently removing these might help with performance a bit.

PR #7783, opened 2022-08-17
Fix notification and add visual cues when triggering a second update

### LINKED ISSUE #7770: Chapter Update Toast UI Bug
### Steps to reproduce

1. Have at least two categories with different entries in them, for this example one with X entries and the other with Y entries.
2. Reload to auto update the first category
4. Wait for some entries to have updated
5. UI will say something along the lines of n/X
6. Now reload to auto update the other category
8. UI bug happens

### Expected behavior

The UI should now show n/X+Y to show the correct total amount of entries.

### Actual behavior

Instead the UI shows n/Y even if Y is smaller than n.
Screenshot_20220816-022947_Tachiyomi

### Crash logs

_No response_

### Tachiyomi version

App version: 0.13.6 (standard, 1f79444, 82, 2022-08-14T15:52Z)

### Android version

Android version: 11 (SDK 30) Android build ID: RP1A.200720.012.A125FXXS2BVE1 

### Device

Device brand: samsung Device manufacturer: samsung Device name: a12 Device model: SM-A125F Device product name: a12nseea

### Other details

_No response_

### Acknowledgements

stevenyomi:
Fixes #7770.

Now the app will toast "An update is already running" and swipe refresh indicators will be dismissed immediately.


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt
```diff
@@ -154,18 +154,15 @@ class LibraryUpdateService(
...
-                val intent = Intent(context, LibraryUpdateService::class.java).apply {
-                    putExtra(KEY_TARGET, target)
-                    category?.let { putExtra(KEY_CATEGORY, it.id) }
-                }
-                ContextCompat.startForegroundService(context, intent)
+            if (isRunning(context)) return false
 
-                true
-            } else {
-                instance?.addMangaToQueue(category?.id ?: -1)
```
arkon, line 166: Aren't we sort of losing this?

stevenyomi: This does not work in the first place.

#7770's cause is that `addMangaToQueue` actually replaces the `mangaToUpdate` list and changes the total count:
https://github.com/tachiyomiorg/tachiyomi/blob/1f9f9662bc5abe75f24dae30d48e843005d86394/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt#L286-L288

But that's not the root cause. Adding manga to queue does not really start updating them. The job is only started on the first trigger:
https://github.com/tachiyomiorg/tachiyomi/blob/1f9f9662bc5abe75f24dae30d48e843005d86394/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt#L240-L253

And will not change during running because it's already converted to some map:
https://github.com/tachiyomiorg/tachiyomi/blob/1f9f9662bc5abe75f24dae30d48e843005d86394/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateService.kt#L321-L323

The job is singleton so I'm afraid there's nothing much we can do. Moreover, appending manga to the existing job requires a lot of changes to avoid updating one manga multiple times or querying a source in parallel.

PR #7802, opened 2022-08-18
Lint changes

ghostbear:
- Fixes current lint issues
- Changes lint task to lintKotlin on PRs


build.gradle.kts
```diff
@@ -26,8 +26,11 @@ subprojects {
     kotlinter {
         experimentalRules = true
 
-        // Doesn't play well with Android Studio
-        disabledRules = arrayOf("experimental:argument-list-wrapping")
+
+        disabledRules = arrayOf(
+            "experimental:argument-list-wrapping", // Doesn't play well with Android Studio
+            "filename", // Often broken to give a more general name
```
ghostbear, line 32: We could also disable to rule in the individual file most of them were for extension function

arkon: It's annoying with Compose too so it's fine.

PR #7803, opened 2022-08-18
Add app metadata to the project repo

simplepad:
This PR adds app metadata to the project repo. It can be used for integration with Fastlane if so desired but also provides metadata that can be parsed by fdroid to enhance Tachiyomi's fdroid page.


arkon:
Thanks, but Tachiyomi on F-Droid isn't actually supported by us nor do we really recommend it since we already handle updates directly through the app.

PR #7824, opened 2022-08-20
Disable explicit nulls for JSON serializer

### LINKED ISSUE #7823: Disable explicit nulls for JSON serializer

https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#explicit-nulls

https://github.com/tachiyomiorg/tachiyomi/blob/3760b310df8701f929d4ca4acbe7facfa97a2220/app/src/main/java/eu/kanade/tachiyomi/AppModule.kt#L80
```diff
- Json { ignoreUnknownKeys = true }
+ Json { ignoreUnknownKeys = true; explicitNulls = false }
```

Disabling explicit nulls enables people to write `field: String?` instead of `field: String? = null` when creating DTO classes for extensions (without creating another instance of Json). Having to write the default `null` is kinda unnecessary in my opinion since people rarely create instances of those manually.

This can avoid unnecessary errors when people forget to write the default `null`. These cases can be hard to test because:
- Tested manga list pages or manga details might happen to contain all those fields. The specific entry that omitted these fields (that the extension developer has seen when testing the API) might be un-tested in the app.
- Some websites have different behavior (either supply `null` or omit field) for the same model across different API endpoints, or even different entries. This might fool people into writing `String?` confidently on seeing the API returning `"field": null` for one specific response.

However people should not depend on this change until the next extension-lib update.

### Other details

_No response_

### Acknowledgements

stevenyomi:
Closes #7823, see the discussion there.

Extension developers should not depend on this change until the app and other main forks have all released a new stable version to match this behavior. By then we should update the contributing guide in the extension repo.

This PR has conflict with #7815, so whichever is merged later might need to rebase.


app/src/main/java/eu/kanade/tachiyomi/AppModule.kt
```diff
@@ -77,7 +77,7 @@ class AppModule(val app: Application) : InjektModule {
 
         addSingletonFactory<DatabaseHandler> { AndroidDatabaseHandler(get(), get()) }
 
-        addSingletonFactory { Json { ignoreUnknownKeys = true } }
+        addSingletonFactory { Json { ignoreUnknownKeys = true; explicitNulls = false } }
```
arkon, line 80: Can you just format this to multiple lines?

stevenyomi: Done.

PR #7829, opened 2022-08-20
Tweak flag classes for Library

ghostbear:
- Add interface for Flag and Mask
- Merge Sort Type and Direction into one class
- Use custom serializers for preferences
  - Mainly to not break the old


app/src/main/java/eu/kanade/domain/category/interactor/ResetCategoryFlags.kt
```diff
@@ -14,12 +12,6 @@ class ResetCategoryFlags(
...
         val display = preferences.libraryDisplayMode().get()
         val sort = preferences.librarySortingMode().get()
-        val sortDirection = preferences.librarySortingAscending().get()
-
-        var flags = 0L
-        flags = flags and DisplayModeSetting.MASK.inv() or (display.flag and DisplayModeSetting.MASK)
-        flags = flags and SortModeSetting.MASK.inv() or (sort.flag and SortModeSetting.MASK)
-        flags = flags and SortDirectionSetting.MASK.inv() or (sortDirection.flag and SortDirectionSetting.MASK)
-        categoryRepository.updateAllFlags(flags)
+        categoryRepository.updateAllFlags(0L + display + sort.type + sort.direction)
```
arkon, line 15: Is the `0L` necessary?

app/src/main/java/eu/kanade/tachiyomi/Migrations.kt
```diff
@@ -266,15 +264,24 @@ object Migrations {
             if (oldVersion < 81) {
                 // Handle renamed enum values
                 @Suppress("DEPRECATION")
```
arkon, line 266: Is there still something deprecated being used?

app/src/main/java/eu/kanade/tachiyomi/ui/library/setting/LibraryDisplayMode.kt
```diff
@@ -0,0 +1,60 @@
...
+            return values
+                .find { mode -> mode.flag == flag and mode.mask }
+                ?: default
+        }
+    }
+}
+
+fun LibraryDisplayMode.serialize(): String {
+    return when (this) {
+        LibraryDisplayMode.ComfortableGrid -> "COMFORTABLE_GRID"
```
arkon, line 42: You could also just store these in the classes themselves.

ghostbear: Do you mean the function or string?

arkon: The strings, but doesn't matter too much.

PR #7837, opened 2022-08-21
Crayola Blue theme

1-minute-to-midnight:
Added a new modern looking theme with perfect contrast and readability

This is also in response to my own feature request 

I wasn't quite satisfied with the new Tidal Wave theme. Still kudos to @NahutabDevelop.

This is my first pull request (as well as my first time using android studio) so please guide me if there are any problems.

Dark Mode Screenshots: 

image

image

image

image

Light Mode Screenshots:

image

image

image

image


arkon:
I'd rather there not be too many similar themes. Can you and @NahutabDevelop hash out a mix of the two themes if appropriate?

1-minute-to-midnight:
they are not really that similar (they're about as similar as the other themes are with each other), just born from the same issue is all
for side by side comparison:
image 
image

I just think my theme has a very neat and modern look to it

arkon:
These two + default results in 3 themes that are primarily blue, which is too many IMO. I don't particularly mind tweaking the default theme too if that seems like the right option.

1-minute-to-midnight:
that would be great.....could you tweak the default blue theme with white bold text for the unread count and slightly stronger blue, enough to stand out from the manga covers?

1-minute-to-midnight:
OR
The more fun option would be to allow the community to vote between the themes

jferdous70:
I would prefer if Crayola replaced the default blue them, it looks much better imo

1-minute-to-midnight:
Hey @CrepeTF, What do you think about this theme?

CrepeTF:
I think it's too derivitive of some of the existing themes. Like the light version of this theme is basically the same as the default light theme. 
I also agree with @arkon that there's just too many blue themes now though I don't necessarily agree with this theme replacing the default theme as that is meant to be have a very standard Material Design colour scheme.

1-minute-to-midnight:
Ah okay, guess I'll have to compile my own apk every update to keep my theme.

PR #7889, opened 2022-08-29
Rewrite RateLimitInterceptor

### LINKED ISSUE #7817: Make rateLimit and rateLimitHost not limit cached responses

According to the first figure in [OkHttp's documentation](https://square.github.io/okhttp/features/interceptors/), the current rate limiting interceptors will also rate limit cached results. These should be made into network interceptors to not needlessly delay fetching from cache.

### Other details

Consider ignoring rate limits for redirects (301, 302, 307, 308) and partial network responses (304), this will likely require rewriting the interceptors to look at responses, and rate limit the subsequent requests instead.

### Acknowledgements

stevenyomi:
Rewrote the interceptor with `java.util.ArrayDeque` and slightly changed logic.

Now it removes cached responses from queue and notifies the waiting request. Resolves #7817.

Currently it only allows completely cached responses to do this. We can tweak it to allow conditional get requests. (change `response.networkResponse == null` to `response.cacheResponse != null`)


app/src/main/java/eu/kanade/tachiyomi/network/interceptor/SpecificHostRateLimitInterceptor.kt
```diff
@@ -28,58 +24,4 @@ fun OkHttpClient.Builder.rateLimitHost(
...
-                Thread.sleep(waitTime) // Sleep inside synchronized to pause queued requests
-            } else {
-                requestQueue.add(now)
-            }
-        }
-
-        return chain.proceed(chain.request())
-    }
-}
+) = addInterceptor(RateLimitInterceptor(httpUrl.host, permits, period, unit))
```
AntsyLich, line 27: Put them in same file?

stevenyomi: > Put them in same file?

Then existing extensions would be broken. This function lives under the class `SpecificHostRateLimitInterceptorKt`.

PR #7895, opened 2022-08-30
Add "multi" as a recognized source/ext lang

AntsyLich:
Related to tachiyomiorg/tachiyomi-extensions#12892


app/src/main/res/values/strings.xml
```diff
@@ -258,6 +258,7 @@
 
       <!-- Extension section -->
     <string name="all_lang">All</string>
+    <string name="multi_lang">Multiple</string>
```
vetleledaal, line 261: Multi is already a well-known term in scene releases.

AntsyLich: i'll wait for more response

arkon:
What's the difference between all and multi? Should the `all` code just be displayed as "Multi"?

AntsyLich:
it's kinda misleading since they don't contain all the languages so i was thinking it'll more appropriate to change the lang of the exts and sources too

alessandrojean:
Just a remember that changing the language code in the package will change the package and make the app detect the extensions as new ones, so an update will not appear to the users. Maybe then is it an special case that needs to be handled by the app too?

vetleledaal:
I agree with Alessandro, should just change the label so users aren't forced to manually update extensions.

Migrating package names can be considered when the majority of users are on a version that handles changing package names more seamlessly. Hopefully this will also work for extensions migrating to multisrc, maybe rely on each extensions `id`?

AntsyLich:
Hmm makes sense. Updated the pr and also shortened the string to "Multi"

Issue #7876, opened 2022-08-27
download delay between chapters.


RyabDev:
### Describe your suggested feature

To avoid the chances of bulk downloads possibly causing sites to block tachiyomi. I think it would be reasonable to at least have the option to be able to have a slight delay between chapter downloads.
Another benefit to this is being able to select to download an entire series without worry.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.13.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

arkon:
Not sure this is really desirable considering sources can already implement rate limiting.

PR #7897, opened 2022-08-30
Add delay between chapter downloads(#7876)

d-najd:
* The current implementation delays all chapter downloads including the first one which is probably not what the user wants.
* Added option for delay of 1, 5 and 30 seconds between chapter downloads located in More/Settings/Downloads/Download ahead

Partially resolves https://github.com/tachiyomiorg/tachiyomi/issues/7876


abhishekabk5 (approved):
LGTM

arkon:
Not sure this is really desirable considering sources can already implement rate limiting.

PR #7898, opened 2022-08-30
Rename source download folder on source rename

AntsyLich:


app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
```diff
@@ -89,7 +92,13 @@ class SourceManager(
     private fun registerStubSource(sourceData: SourceData) {
         scope.launch {
             val (id, lang, name) = sourceData
-            sourceRepository.upsertSourceData(id, lang, name)
+            val dbSourceData = sourceRepository.getSourceData(id)
+            if (dbSourceData != sourceData) {
```
ghostbear, line 96: Is this check needed? The other check could just check if the source from the database is not null and check the names, no?

app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
```diff
@@ -89,7 +92,13 @@ class SourceManager(
...
+                sourceRepository.upsertSourceData(id, lang, name)
+                if (dbSourceData != null) {
```
ghostbear, line 98:
```suggestion
                if (dbSourceData != null && dbSourceData.name != sourceData.name) {
```

app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
AntsyLich: I mean no need to upsert data in db if it's the same thing

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
```diff
@@ -343,6 +343,22 @@ class DownloadManager(
...
+    /**
+     * Renames source download folder
+     *
+     * @param oldSource the old source.
+     * @param newSource the new source.
+     */
+    fun renameSource(oldSource: Source, newSource: Source) {
+        val oldFolder = provider.findSourceDir(oldSource) ?: return
+        val newName = provider.getSourceDirName(newSource)
+        // To handle capitalization changes
```
arkon, line 355: You can check beforehand to avoid the unnecessary extra IO.

app/src/main/java/eu/kanade/tachiyomi/source/SourceManager.kt
ghostbear: You could invert the statement and return early if that is the case to avoid the nesting

```suggestion
            if (dbSourceData == sourceData) return
```

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
```diff
@@ -343,6 +343,30 @@ class DownloadManager(
...
+        val capitalizationChanged = oldFolder.name.equals(newName, ignoreCase = true)
+        if (capitalizationChanged) {
+            val tempName = newName + "_tmp"
+            if (oldFolder.renameTo(tempName).not()) {
+                logcat(LogPriority.ERROR) { "Could not rename source download folder: ${oldFolder.name}." }
+                return
+            }
+        }
+
+        if (oldFolder.renameTo(newName).not()) {
```
arkon, line 365: Is `oldFolder` pointing to the renamed tmp folder at this point?

AntsyLich: according to my testing it is

PR #7899, opened 2022-08-30
Don't update page progress and allow preload when on StencilPage

AntsyLich:
Also cleanup a little in WebtoonPageHolder


app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonPageHolder.kt
```diff
@@ -246,30 +246,28 @@ class WebtoonPageHolder(
...
+                        isAnimated,
+                        ReaderPageImageView.Config(
+                            zoomDuration = viewer.config.doubleTapAnimDuration,
+                            minimumScaleType = SubsamplingScaleImageView.SCALE_TYPE_FIT_WIDTH,
+                            cropBorders = viewer.config.imageCropBorders,
+                        ),
+                    )
+                }
             }
-            // Keep the Rx stream alive to close the input stream only when unsubscribed
```
arkon, line 270: I actually don't know why this was here, do you?

AntsyLich: Nope. While I was testing I didn't feel any difference. This probably is some old code that wasn't reviewed since pager viewer has similar code like this pr.

PR #7902, opened 2022-08-30
Add support to update strategy on global update

### LINKED ISSUE #6539: Add a `dontUpdate`/`isOneShot` flag to `SManga` to avoid updating single-chapter titles

alessandrojean:
This PR adds a new `update_strategy` field to `SManga`, which defaults to `UpdateStrategy.ALWAYS_UPDATE`. If an extension marks it as `UpdateStrategy.ONLY_FETCH_ONCE`, it will be skipped in the library update by default (it's not configurable, and it's forced by the source).

Still needs to be included in a stable release (and in forks too) so it can be added to the extensions' lib.

The new field is also included in new backup files, and works just fine for older backups since it has a default value of `UpdateStrategy.ALWAYS_UPDATE` set to it (I tested with a backup created in Preview without this change).

Implemented as an enum to allow possible more cases to be added if needed.

Closes #6539.


alessandrojean:
Tested with a fork of the extensions-lib, it's working as expected.

Important to notice, if a user have an updated extension with the field in an older version of the app, it throws an error.

stevenyomi:
How is it different from a "Completed" status?

alessandrojean:
A source could not mark the status as complete (like it being unknown) and it could still be single chapter. The change is most done thinking about hentai sources that always have single chapter entries.

Also some users tends to disable the complete restriction, so having an extra one just for this case could at least help to some users keep it on after recommendation. We could maybe make it uncheckable and always on thought.

alessandrojean:
Status could be an alternative, but I'm not very sure if "Single chapter", "One shot", or other similar descriptions would really fit into being an own status option when you compare to the other options available.

stevenyomi:
I think single chapter is only one of the many reasons to prevent auto-updating. Adding only this one as a boolean field is not "general" and might cause people to request adding other reasons. It's better to add `no_auto_updates = true` to prevent it out of any reason, or add a flag field `no_auto_updates_reason = 0b00010` to allow adding more reasons.

alessandrojean:
Updated the code to have a more generic enum now, reviews are welcome 👍🏻.

app/src/main/java/eu/kanade/tachiyomi/source/model/UpdateStrategy.kt
```diff
@@ -0,0 +1,6 @@
+package eu.kanade.tachiyomi.source.model
+
+enum class UpdateStrategy {
```
arkon, line 3: Can you add a JavaDoc to this with `@since extension-lib 1.4` and also bump `LIB_VERSION_MAX` to 1.4 in `ExtensionLoader`?

app/src/main/res/values/strings.xml
```diff
@@ -799,6 +799,7 @@
     <string name="skipped_reason_completed">Skipped because series is complete</string>
     <string name="skipped_reason_not_caught_up">Skipped because there are unread chapters</string>
     <string name="skipped_reason_not_started">Skipped because no chapters are read</string>
+    <string name="skipped_reason_not_always_update">Skipped because series is not marked as always update</string>
```
arkon, line 802:
```suggestion
    <string name="skipped_reason_not_always_update">Skipped because series does not require updates</string>
```

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/source/model/UpdateStrategy.kt
alessandrojean: Done. Does the JavaDoc also needs to be added to the extension-lib?

arkon: No, that's a bit easier to figure out based on GitHub showing tags on commits.

stevenyomi:
Isn't it easier to make the field an `Int` like `status`?

alessandrojean:
Integers could work but honestly an enum it's best as it's designed for it. Using an integer could possibly allow sources to set invalid values (as it can be done with the status field with the current implementation).

stevenyomi:
We can add a data verifier/cleaner to clamp the values from source. That's a lot more efficient than doing all those conversions of an Enum.

alessandrojean:
I'd really prefer to use a structure designed for that, as status ideally should also be one. The conversions are only done when saving and reading from the database once and they're not that time consuming as they only get the `name` while encoding and do a `valueOf` (iirc) when decoding. That can really be improved later if it's really a bad performatic issue.

alessandrojean:
On a second thought maybe it would be interesting to save it as integers in the database but still expose it as enum. A column adapter could be created to encode the enum as it's ordinal value and decoding by using the `values()[index]` of the enum, clamping it to the enum bounds when doing so.

I'm gonna change this later today.

app/src/main/java/eu/kanade/tachiyomi/source/model/UpdateStrategy.kt
```diff
@@ -0,0 +1,22 @@
+package eu.kanade.tachiyomi.source.model
+
+/**
+ * Define the update strategy for a single [SManga].
+ * The strategy used will only take effect on the library update.
+ *
+ * @since extensions-lib 1.4
+ */
+enum class UpdateStrategy {
```
arkon, line 9: I'm actually not sure if it'd be nicer to just stay consistent with the existing way we handle "enums" as int constants like status or to start introducing actual enums.

I'd be okay either way though; depends on if anyone has strong feelings about it.

arkon (commented):
Generally looks fine, but need to fix conflicts.

alessandrojean:
Will fix the conflicts this week 👍🏻.

app/src/main/java/eu/kanade/tachiyomi/source/model/UpdateStrategy.kt
alessandrojean: The status can be easily replaced with an enum in the future by using the database enum adapter, so I think it's worth the move.

alessandrojean:
Conflicts are fixed now, and the build is still running fine, so it's ready to merge again.

Issue #6623, opened 2022-02-10
HTTP 103 fails in asObservableSuccess


vetleledaal:
### Describe your suggested feature

Related to #6562, but that issue is closed for comments.

asObservableSuccess() does not properly handle HTTP 103 codes. These should be ignored instead of causing the whole request to fail.

### Other details

This is difficult to test as it seems like Cloudflare is doing A/B testing on this feature, but here are some repro steps:
1. Revert #6562
2. Open GManga v1.2.9 (you _may_ need to change the UA to bypass CF (setting it to any number is sufficient))
3. Open "National School Prince is a Girl"
4. If and only if the toast "HTTP error 103" is shown on manual refresh; continue
5. Open chapter 237 OR 236 (others are not affected)
 
I used this snippet for breakpoints
```kt
override fun fetchPageList(chapter: SChapter): Observable<List<Page>> {
    return client.newCall(pageListRequest(chapter))
        .asObservable()
        .map { response ->
            pageListParse(response)
        }
}
```
![studio64_DW2aY4a4fX](https://user-images.githubusercontent.com/13540478/153489431-17bdcf9e-bb99-423c-83db-b1d0d77381c5.png)

Used a VPN to get HTTP 103 codes, Surfshark (Germany - Berlin)

Related issue tachiyomiorg/tachiyomi-extensions#10786

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.13.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

vetleledaal:
this is still happening in a131e28b6. With the official GMANGA v1.2.9 it says "HTTP error 103" when opening the mentioned chapters, switching to .asObservable() makes he toast say "expected 0 bytes but received 3463"

vetleledaal:
The GManga extension was fixed in tachiyomiorg/tachiyomi-extensions#10934 by changing the domain to one that doesn't return 103 codes. The underlying issue is still present, but probably not as urgent as before.

hakudansho:
[...]

PR #7907, opened 2022-08-31
Manually handle HTTP 103 through WebView

AntsyLich:
Related to #6623


app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
```diff
@@ -0,0 +1,158 @@
...
+
+    internal class JsInterface(private val latch: CountDownLatch, var payload: String? = null) {
+        @JavascriptInterface
+        fun passPayload(passedPayload: String) {
+            payload = passedPayload
+            latch.countDown()
+        }
+    }
+
+    private fun http103Exception(msg: String) = Exception("Http103Interceptor - $msg")
```
arkon, line 82: Should be an `IOException` considering it's in an interceptor.

app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
```diff
@@ -0,0 +1,158 @@
...
+            return@lazy
+        }
+
+        WebSettings.getDefaultUserAgent(context)
+    }
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+        val response = chain.proceed(request)
+        if (response.code != 103) return response
```
arkon, line 56: I feel like we might want to just handle all 1xx responses in this interceptor. Thoughts?

app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
```diff
@@ -0,0 +1,158 @@
...
+    private val executor = ContextCompat.getMainExecutor(context)
+
+    private val networkHelper: NetworkHelper by injectLazy()
+
+    /**
+     * When this is called, it initializes the WebView if it wasn't already. We use this to avoid
+     * blocking the main thread too much. If used too often we could consider moving it to the
+     * Application class.
+     */
+    private val initWebView by lazy {
```
arkon, line 42: We can definitely dedupe some of this stuff, but I'll refactor that afterwards.

app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        exception?.let { throw it }
+
+        val payload = jsInterface.payload ?: throw http103Exception("Couldn't fetch site through webview")
+
+        return ogResponse.newBuilder()
+            .code(200)
+            .protocol(Protocol.HTTP_1_1)
+            .message("SUCCESS")
```
arkon, line 154: Is a message even required?

app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
AntsyLich: imo others doesn't impact tachi that much. 100 is handled by okhttp and 102 is mainly used by WebDAV, and 101 indicates switch protocol which should just be handled by the ext itself

app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
AntsyLich: not sure but no harm in keeping it

app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
AntsyLich: All exception are wrapped in `IOException` in intercept function

app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
arkon: 1xx responses should've had no effect at all but here we are. I guess we can address it again if they decide to add more codes.

app/src/main/java/eu/kanade/tachiyomi/network/interceptor/Http103Interceptor.kt
AntsyLich: > But these messages are something called HTTP status messages. In fact, the server always returns a message for every request. 

Guess that's that

PR #7909, opened 2022-08-31
Fix tmp cbzs are counted towards manga download count

AntsyLich:


arkon (commented):
Why not just make the tmp CBZs end with `Downloader.TMP_DIR_SUFFIX` too?

PR #7910, opened 2022-08-31
Cleanup chapter name when syncing chapter with source

### LINKED ISSUE #7351: Regression: Manga's title is no longer stripped from chapter names as of 0.13.4
### Steps to reproduce

Add a manga (in my case it's local) with the manga's title included in the chapter titles.

### Expected behavior

The name should be stripped, as it previously was (following the resolution of issue #6913).

### Actual behavior

There seems to have been a regression between 0.13.3 and 0.13.4. In 0.13.3, issue #6913 was resolved with commit 89decf3, returning the title-stripping behavior. (Thank you for that, by the way!) When I recently updated to 0.13.4, however, the previous behavior that duplicates the manga's title in several locations had returned.

In the following screenshots, 0.13.3 is on the left, and 0.13.4 is on the right.
Before. After.

The old behavior also returns in the reader:

### Crash logs

_No response_

### Tachiyomi version

0.13.4

### Android version

Android 12

### Device

Google Pixel 5a

### Other details

I haven't looked immensely deeply into it as I'm not familiar with the codebase, but it looks like the culprit may be commit 65a8b63, which moves the strip code for organizational reasons, but also removes the relevant `cleanChapterNames` call – likely unintentionally.

### Acknowledgements

AntsyLich:
Closes #7351


arkon (commented):
This doesn't work for local, does it?

arkon:
Yeah, but chapter names need to match local folder/files to work, and those local folder/file names are the source of truth for the local source. It needs to be a display-layer thing, not a database change.

AntsyLich:
isn't source of truth the url?

AntsyLich:
Did some testing and chapter name doesn't effect local source functionality

PR #7916, opened 2022-09-01
Lessen the use of GlobalScope `launchIO`

AntsyLich:


arkon:
A bunch of these jobs shouldn't actually be cancelled if their view/presenter scopes are cancelled.

AntsyLich:
do i wrap them with `NonCancelable` context?

arkon:
@ghostbear what should be preferred in this codebase?

ivaniskandar:
Ideally file operations should be moved to a separate job but for now using NonCancellable context should work

PR #7927, opened 2022-09-03
Fix download ahead missing out 1 chapter

### LINKED ISSUE #7917: Download ahead missing out 1 chapter.
### Steps to reproduce

Happens when all downloaded chapters are read

Set download ahead to X amount.
1. Get a manga with X+2 chapters. 
2. Download chapter 1 and 2
3. Read chapter 1 and 2

### Expected behavior

Upto Chapter X+2 should be in download queue

### Actual behavior

Only upto Chapter X+1 is in queue while Chapter X+2 is not added

### Crash logs

_No response_

### Tachiyomi version

r4845 

### Android version

N/A

### Device

N/A

### Other details

Should add a check for if it's the last downloaded chapter then instead of just next one it should add next 2 chapters

### Acknowledgements

nzoba:
Closes #7917


app/src/main/java/eu/kanade/tachiyomi/util/system/BooleanExtensions.kt
```diff
@@ -0,0 +1,3 @@
+package eu.kanade.tachiyomi.util.system
+
+fun Boolean.toInt() = if (this) 1 else 0
```
ghostbear, line 3: You could move this function into this file as well

https://github.com/tachiyomiorg/tachiyomi/blob/809da49301cceacd433f38354fb04fc616efcc5f/app/src/main/java/eu/kanade/data/DatabaseUtils.kt

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -452,12 +453,12 @@ class ReaderPresenter(
         val isNextChapterDownloaded =
             downloadManager.isChapterDownloaded(nextChapter.name, nextChapter.scanlator, manga.title, manga.source)
         if (isNextChapterDownloaded) {
-            downloadAutoNextChapters(chaptersNumberToDownload, nextChapter.id)
+            downloadAutoNextChapters(chaptersNumberToDownload, nextChapter.id, nextChapter.read)
```
AntsyLich, line 456: Won't this be invoked every time? Since next chapter always won't be read

nzoba: I think I misunderstood your issue. I thought you were talking about the rare case where the next chapter is already read and downloaded (and you don't have the setting `skip chapters marked read` enabled) then it would be included in the next unread chapters count even thought it's already read

I think your issue happens if you read too fast and the next chapter is not yet downloaded (still in the queue), so the last read chapter won't add a chapter to the queue (since one of the condition is that the next chapter should already be downloaded). Let me know if I understood correctly this time :sweat_smile: 
I can add a check whether the next one is already downloaded or in the queue to fix that

AntsyLich: yup that's the issue

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -449,15 +450,19 @@ class ReaderPresenter(
...
         if (chaptersNumberToDownload == 0 || !manga.favorite) return
-        val isNextChapterDownloaded =
-            downloadManager.isChapterDownloaded(nextChapter.name, nextChapter.scanlator, manga.title, manga.source)
-        if (isNextChapterDownloaded) {
-            downloadAutoNextChapters(chaptersNumberToDownload, nextChapter.id)
+        val isNextChapterDownloadedOrQueued = downloadManager.isChapterDownloaded(
+            nextChapter.name,
+            nextChapter.scanlator,
+            manga.title,
+            manga.source,
```
AntsyLich, line 457: do a `skipCache = true` check

nzoba: Is it really necessary in this case ? While you are reading you probably won't touch your downloads via the file system so the cache should be enough no ? And it's faster to check the cache

PR #7955, opened 2022-09-05
Cleanup logic of `onStripSplit` in WebtoonPageHolder

AntsyLich:


app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonPageHolder.kt
```diff
@@ -289,25 +289,24 @@ class WebtoonPageHolder(
...
         }
 
         return imageStream
     }
 
-    private fun onStripSplit(imageStream: BufferedInputStream): InputStream? {
-        val page = page ?: return null
-        val streamFn = page.stream ?: return null
+    private fun onStripSplit(imageStream: BufferedInputStream): InputStream {
+        // If we have reached this point [page] and it's stream shouldn't be null
```
arkon, line 300:
```suggestion
        // If we have reached this point [page] and its stream shouldn't be null
```

PR #7956, opened 2022-09-05
Flatten available extensions with its source

AntsyLich:
We should also do something similar for installed extensions

Supersedes #7200


ghostbear:
> We should also do something similar for installed extensions

Please god no

PR #7967, opened 2022-09-08
Prevent okhttp from caching covers and chapter images

stevenyomi:
The okhttp client that sources use has a 5 MiB disk cache created here:

https://github.com/tachiyomiorg/tachiyomi/blob/6b91f6545704ba6facace7b1d0cdd129ab132f71/app/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt#L19-L20

https://github.com/tachiyomiorg/tachiyomi/blob/6b91f6545704ba6facace7b1d0cdd129ab132f71/app/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt#L63

We already cache these responses elsewhere:
- covers - library ones in app's cover cache and others in Coil's disk cache
- chapter images - either in chapter cache or downloaded to storage

Previously they are also cached by okhttp, evicting cached web pages while not being read again.

By setting `Cache-Control` header to `no-store`, we keep them from the okhttp cache, preserving as many cached web pages as possible to increase cache hit rate. This also saves disk write.

I also cleaned up some code in `MangaCoverFetcher`.


arkon (approved):
LGTM. @ivaniskandar you want to review too?

app/src/main/java/eu/kanade/tachiyomi/data/coil/MangaCoverFetcher.kt
```diff
@@ -234,18 +228,9 @@ class MangaCoverFetcher(
     }
 
     private fun writeToDiskCache(
-        snapshot: DiskCache.Snapshot?,
```
ivaniskandar, line 237: I'm not feeling this tbh. Is the snapshot passed here always null?

stevenyomi: Yes.

PR #7969, opened 2022-09-09
DownloadController: Partial Compose conversion

ivaniskandar:
Item list is not changed as currently there is no fitting Compose component to replace the drag-drop behavior.

CI build most likely will fail to complete :/


ivaniskandar:
Interesting

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadHeaderItem.kt
```diff
@@ -35,16 +35,39 @@ data class DownloadHeaderItem(
...
     override fun hashCode(): Int {
-        return id.hashCode()
+        var result = id.hashCode()
+        result = 31 * result + name.hashCode()
+        result = 31 * result + size
+        result = 31 * result + subItems.hashCode()
+        return result
     }
 
+//    override fun equals(other: Any?): Boolean {
```
arkon, line 59: Remove these.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadService.kt
```diff
@@ -47,6 +49,13 @@ class DownloadService : Service() {
          */
         val runningRelay: BehaviorRelay<Boolean> = BehaviorRelay.create(false)
 
+        private val _isRunning = MutableStateFlow(false)
+
+        /**
+         * True if the service is running.
+         */
```
arkon, line 56: Is this really useful...?

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadController.kt
```diff
@@ -1,134 +1,318 @@
...
-        DownloadService.runningRelay
-            .observeOn(AndroidSchedulers.mainThread())
-            .subscribeUntilDestroy { onQueueStatusChange(it) }
+    @Composable
+    override fun ComposeContent() {
+        val context = LocalContext.current
+        val downloadList by presenter.state.collectAsState()
+
+        val scrollBehavior = TopAppBarDefaults.pinnedScrollBehavior(rememberTopAppBarState())
+        var fabExpanded by remember { mutableStateOf(true) }
```
arkon, line 111: Really ought to lift this to our `Scaffold` or something at some point.

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadController.kt
```diff
@@ -1,134 +1,318 @@
...
+                        Row(verticalAlignment = Alignment.CenterVertically) {
+                            Text(
+                                text = stringResource(R.string.label_download_queue),
+                                maxLines = 1,
+                                modifier = Modifier.weight(1f, false),
+                                overflow = TextOverflow.Ellipsis,
+                            )
+                            if (downloadList.isNotEmpty()) {
+                                val pillAlpha = if (isSystemInDarkTheme()) 0.12f else 0.08f
+                                Pill(
```
arkon, line 147: I feel like we could also just make `AppBar` accept something for a pill considering it's a pattern we have in the library too.

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadController.kt
```diff
@@ -1,134 +1,318 @@
...
+        ) { contentPadding ->
+            val density = LocalDensity.current
+            val layoutDirection = LocalLayoutDirection.current
+            val left = with(density) { contentPadding.calculateLeftPadding(layoutDirection).toPx().roundToInt() }
+            val top = with(density) { contentPadding.calculateTopPadding().toPx().roundToInt() }
+            val right = with(density) { contentPadding.calculateRightPadding(layoutDirection).toPx().roundToInt() }
+            val bottom = with(density) { contentPadding.calculateBottomPadding().toPx().roundToInt() }
+
+            Box(modifier = Modifier.nestedScroll(nestedScrollConnection)) {
+                if (downloadList.isEmpty()) {
```
arkon, line 275: Couldn't we do the empty state before the rest of it and short circuit?

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadService.kt
ivaniskandar: As in the value? It's used for start/pause FAB. But yeah with this, there's a three way to check for the service status. So it's kinda all over the place

arkon: I meant the Javadoc

ivaniskandar: Ah. Now that I look at it again no

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadController.kt
ivaniskandar: This one's specific for interop nested scroll. A simpler method by using LazyList state is available and IMO should be kept separate.

PR #7973, opened 2022-09-10
Add support for ComicInfo.xml files

### LINKED ISSUE #4965: [Feature Request] ComicInfo.xml Support for Local Manga
There is no standard for tagging comics/manga, but the ComicRack.xml by ComicRack is widely supported by self hosted services like Ubiqooquity, Komga and Calibre plugins. ComicTagger and MangaTagger use the ComicRack.xml format to automatically tag manga/comics using Anilist or ComicVine metadata, making it much faster and easier to tag things.
## What/Requirements
Tachiyomi would be able to read the xml file from a cbz/cbr file or from the chapter folder, and parse the information outlined in the wiki page
https://wiki.mobileread.com/wiki/ComicRack

Implemented in komga using kotlin like this
https://github.com/gotson/komga/commit/af01d25ede29feb5159e8a800b789f200e6e7e5c

Shamicen:
Closes #4965

Based on jobobby04@b1f243e3c and arkon@936655a 

I implemented reading `ComicInfo.xml` files from the chapter root and from chapter archives ( `zip`, `rar`, `cbr` and `cbz`)
For now the function only reads the title, writer, summary, and genre fields as well as the the 5 artist fields penciller, inker, colorist, letterer and coverArtist. I implemented the artist field in a way that removes duplicate entries since artists often fill more then one role.

I also changed the xml factory settings in `AppModule.kt` to add the `unknownChildHandler` to parse xml files with extra fields not specified in ComicInfo.kt and made all vals in `ComicInfo.kt` nullable to allow xml files with less fields then defined in `ComicInfo.kt`. Adding the unkownChildHandler causes the warning: _This XML Serialization code is not stable and may be subject to binary and source incompatible changes_ but it seems to be the only way to allow parsing files with extra fields as of now. 

I left the Status field empty since there is no equivalent to Tachiyomis Status in the ComicInfo.xml standard. We could add a Tachiyomi exclusive custom field for Status, but there would be a small risk that some apps out there may not be able to handle xml files with extra fields. We could also make a request on the anansi-project github to add a Publishing Status field to the official ComicInfo.xml standard. 

The function first scans the chapter root and only if it finds nothing will it scan the archive files. If it finds a `ComicInfo.xml` file in one of the archives, then the function will copy it to the chapter root to save resources on the next scan. If neither the chapter root or archive files have a ComicInfo.xml file, a `.noxml` file will be generated in the chapter root so that the chapter archives will not be scanned again in the future. 

My implementation causes some `Inappropriate blocking method call` compiler warnings, but I think it should be alright since I call every function that causes these warnings with:  `withContext(Dispatchers.IO)` 
If someone knows of a better approach that does not cause these warnings please let me know.

I'm pretty new to programming and appreciate all feedback, so please just tell me if I should change anything.


app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +161,114 @@ class LocalSource(
                 }
             }
 
+        // get Metadata from ComicInfo.xml Files
+        val xmlFile = getMangaDirsFiles(manga.url, baseDirsFile)
+            .firstOrNull { it.extension.equals("xml", ignoreCase = true) }
+
+        val noXmlFile = getMangaDirsFiles(manga.url, baseDirsFile)
+            .firstOrNull { it.extension.equals("noxml", ignoreCase = true) }
```
ghostbear, line 169: How about searching for both files at the same time? Then instead of checking for null in the if statement below check what the extension is as it will either be `xml` or `noxml`

```suggestion
        val metadataRegex = "(no)?xml".toRegex()
        val metadataFile = getMangaDirsFiles(manga.url, baseDirsFile)
            .firstOrNull { it.extension.matches(metadataRegex) } ?: return manga
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +161,118 @@ class LocalSource(
...
+                val folderPath = mangaDir?.absolutePath
+                val fileName = "ComicInfo.xml"
+
+                val chapterArchives = getMangaDirsFiles(manga.url, baseDirsFile)
+                    .filter { isSupportedFile(it.extension) && !it.extension.equals("epub", true) }
+                    .toList()
+
+                for (chapter in chapterArchives) {
+                    if (chapter.extension.equals("zip", true) || chapter.extension.equals("cbz", true)) {
+                        val zip = withContext(Dispatchers.IO) { ZipFile(chapter) }
```
ghostbear, line 190: We have a shortcut function for `withContext(Dispatchers.IO)` called `withIOContext`

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +161,118 @@ class LocalSource(
...
+                        val zipEntry = zip.getEntry("ComicInfo.xml")
+
+                        if (zipEntry != null) {
+                            val archiveInputStream = withContext(Dispatchers.IO) { zip.getInputStream(zipEntry) }
+                            val comicInfo = xml.decodeFromReader<ComicInfo>(AndroidXmlReader(archiveInputStream, encoding))
+                            if (comicInfo.title != null || comicInfo.writer != null || comicInfo.summary != null ||
+                                comicInfo.genre != null || comicInfo.penciller != null
+                            ) setMetadataXml(comicInfo, manga)
+                            withContext(Dispatchers.IO) { archiveInputStream.close() }
```
ghostbear, line 199: Instead of closing the stream manually use the function called `use` and it will handle it for you

```suggestion
                                withIOContext {
                                    zip.getInputStream(zipEntry).use { archiveInputStream ->
                                        val comicInfo = xml.decodeFromReader<ComicInfo>(AndroidXmlReader(archiveInputStream, encoding))
                                        if (comicInfo.title != null || comicInfo.writer != null || comicInfo.summary != null ||
                                            comicInfo.genre != null || comicInfo.penciller != null) {
                                            setMetadataXml(comicInfo, manga)
                                        }
                                    }
                                }
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +161,118 @@ class LocalSource(
...
+                val mangaDir = getMangaDir(manga.url, baseDirsFile)
+                val folderPath = mangaDir?.absolutePath
+                val fileName = "ComicInfo.xml"
+
+                val chapterArchives = getMangaDirsFiles(manga.url, baseDirsFile)
+                    .filter { isSupportedFile(it.extension) && !it.extension.equals("epub", true) }
+                    .toList()
+
+                for (chapter in chapterArchives) {
+                    if (chapter.extension.equals("zip", true) || chapter.extension.equals("cbz", true)) {
```
ghostbear, line 189: You're doing similar things in both of the condition statements. Try to extract the things related to `ComicInfo.xml` out into a separate function

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +161,114 @@ class LocalSource(
...
+        if (comicInfo.penciller != null) artist += comicInfo.penciller.value + ", "
+        if (comicInfo.inker != null) artist += comicInfo.inker.value + ", "
+        if (comicInfo.colorist != null) artist += comicInfo.colorist.value + ", "
+        if (comicInfo.letterer != null) artist += comicInfo.letterer.value + ", "
+        if (comicInfo.coverArtist != null) artist += comicInfo.coverArtist.value + ", "
+
+        if (artist.isNotBlank()) {
+            val artistList: List<String> = artist.split(",").map { it.trim() }.distinct()
+            manga.artist = artistList.joinToString(", ").dropLast(2)
+        }
```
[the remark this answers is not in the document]
Shamicen: The value Strings are separated with a comma that's why I did it that way. Example:
`<Penciller>Paco Medina, Sean Izaakse</Penciller>
  <Inker>Paco Medina, Sean Izaakse</Inker>
  <Colorist>Jesus Aburtov, Marcio Menyz</Colorist>`

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
Shamicen: OK I will change it

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
Shamicen: Will do thank you

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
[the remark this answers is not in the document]
ghostbear: Split the strings before calling distinct

```suggestion
        manga.artist = listOfNotNull(
            comicInfo.penciller?.value,
            comicInfo.inker?.value,
            comicInfo.colorist?.value,
            comicInfo.letterer?.value,
            comicInfo.coverArtist?.value
        )
            .flatMap { it.split(", ") }
            .distinct()
            .joinToString(", ") { it.trim() }
            .takeIf { it.isNotEmpty() }
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
Shamicen: I tried implementing this in my new commit hope it is OK

Shamicen:
What am I supposed to do about the build check failing? 

Do I need to update my branch or did I mess something up?

arkon:
You need to rebase against the tip of `master`.

Shamicen:
Does this work?

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +160,115 @@ class LocalSource(
...
+                            }
+                        }
+                        if (breaker) break
+                    } else if (chapter.extension.equals("rar", true) || chapter.extension.equals("cbr", true)) {
+                        withIOContext {
+                            Archive(chapter).use { archive: Archive ->
+                                var comicInfoFileHeader: FileHeader? = null
+                                archive.fileHeaders.map {
+                                    if (it.fileName == "ComicInfo.xml") comicInfoFileHeader = it
+                                }
```
ghostbear, line 209: Isn't this trying to find the first file called `ComicInfo.xml`

```suggestion
                                var comicInfoFileHeader = archive.fileHeaders.firstOrNull { it.fileName == "ComicInfo.xml" }
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +160,115 @@ class LocalSource(
...
+                                    createNoXmlFile = false
+                                    breaker = true
+                                }
+                            }
+                        }
+                        if (breaker) break
+                    }
+                }
+                if (createNoXmlFile) {
+                    File("$folderPath/.noxml").outputStream().use {}
```
ghostbear, line 226:
```suggestion
                    File("$folderPath/.noxml").createNewFile()
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +160,115 @@ class LocalSource(
...
+                            Archive(chapter).use { archive: Archive ->
+                                var comicInfoFileHeader: FileHeader? = null
+                                archive.fileHeaders.map {
+                                    if (it.fileName == "ComicInfo.xml") comicInfoFileHeader = it
+                                }
+
+                                if (comicInfoFileHeader != null) {
+                                    archive.getInputStream(comicInfoFileHeader).use { archiveInputStream ->
+                                        val comicInfo = xml.decodeFromReader<ComicInfo>(AndroidXmlReader(archiveInputStream, encoding))
+                                        val archiveInputStreamOut = archive.getInputStream(comicInfoFileHeader)
```
ghostbear, line 214: We are opening the same input stream twice (`archive.getInputStream(comicInfoFileHeader)`). Maybe we should use the functions `mark` and `reset` to reset the first one.

Shamicen: Did that in my new commit

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +159,114 @@ class LocalSource(
...
+            comicInfo.genre != null || comicInfo.penciller != null
+        ) setMetadataXml(comicInfo, manga)
+
+        val comicInfoFile = File("$folderPath/ComicInfo.xml")
+
+        withIOContext {
+            comicInfoFile.outputStream().use { comicInfoOut ->
+                archiveInputStreamOut.copyTo(comicInfoOut)
+            }
+            archiveInputStreamOut.close()
```
jobobby04, line 246: `.use` is preferred.
```suggestion
            comicInfoFile.outputStream().use { comicInfoOut ->
                archiveInputStreamOut.use { it.copyTo(comicInfoOut) }
            }
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +159,114 @@ class LocalSource(
...
 
+        // get Metadata from ComicInfo.xml Files
+        val metadataRegex = "(no)?xml".toRegex()
+        val metadataFile = getMangaDirsFiles(manga.url, baseDirsFile)
+            .firstOrNull { it.extension.matches(metadataRegex) }
+
+        try {
+            val encoding = "UTF-8"
+            if (metadataFile?.extension == "xml") {
+                val comicInfo = xml.decodeFromReader<ComicInfo>(AndroidXmlReader(metadataFile.inputStream(), encoding))
```
jobobby04, line 170: Need to close the reader/stream, if this doesnt work use `.use` on the `inputStream()` instead
```suggestion
                val comicInfo = AndroidXmlReader(metadataFile.inputStream(), encoding).use { xml.decodeFromReader<ComicInfo>(it) }
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +159,113 @@ class LocalSource(
                 }
             }
 
+        // get Metadata from ComicInfo.xml Files
+        val metadataRegex = "(no)?xml".toRegex()
+        val metadataFile = getMangaDirsFiles(manga.url, baseDirsFile)
+            .firstOrNull { it.extension.matches(metadataRegex) }
+
+        try {
+            val encoding = "UTF-8"
```
ghostbear, line 168:
```suggestion
            val encoding = Charsets.UTF_8.name()
```

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
Shamicen: I had to change this code a little a while back since the way it was implemented here caused it to exit the function if **no** `.noxml` file was found instead of exiting if it finds a `.noxml` file. 

I now noticed that by continuing to use the Regex `metadataFile` will always be the `.noxml` file which in turn means that if there is both a `.noxml` and a `ComicInfo.xml` file at the chapter root it won't read the `ComicInfo`. 

Now my Question should we keep it that way? 

On the plus side users could create a .noxml file to stop Tachiyomi from parsing ComicInfo files which could maybe save some resources when opening Local Source. But some users may get confused why their ComicInfo files aren't parsed especially since we create .noxml files automatically.

ghostbear: Could perhaps use `lastOrNull` if they always come in the order of [`noxml`, `xml`]

Shamicen: yeah that sounds good

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +159,109 @@ class LocalSource(
                 }
             }
 
+        // get Metadata from ComicInfo.xml Files
+        val metadataRegex = "(no)?xml".toRegex()
+        val metadataFile = getMangaDirsFiles(manga.url, baseDirsFile)
+            .lastOrNull { it.extension.matches(metadataRegex) }
```
arkon, line 165: It'd be safer to just check for the exact filenames. There's a chance that some stray `.xml` file exists but isn't actually a `ComicInfo.xml` file.

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +159,109 @@ class LocalSource(
...
+
+                                if (zipEntry != null) {
+                                    zip.getInputStream(zipEntry).buffered().use { archiveInputStream ->
+                                        archiveInputStream.mark(0)
+                                        val comicInfo = xml.decodeFromReader<ComicInfo>(AndroidXmlReader(archiveInputStream, encoding))
+                                        archiveInputStream.reset()
+                                        setAndCopyMetadata(comicInfo, archiveInputStream, manga, folderPath)
+                                    }
+                                    createNoXmlFile = false
+                                    breaker = true
```
arkon, line 196: You could avoid this sort of breaking if you just pull out this logic into a separate function that'd `return` when you find/create a file.

app/src/main/java/eu/kanade/domain/manga/model/ComicInfo.kt
```diff
@@ -0,0 +1,126 @@
...
+    val genre: ComicInfoGenre?,
+    val web: ComicInfoWeb?,
+    val languageISO: ComicInfoLanguageISO?,
+    val translator: ComicInfoTranslator?,
+    val ageRating: ComicInfoAgeRating?,
+    val manga: ComicInfoManga?,
+    val characters: ComicInfoCharacters?,
+    val pageCount: ComicInfoPageCount?,
+    val pages: ComicInfoPages?,
+
```
arkon, line 30: Nitpick: remove empty line

app/src/main/java/eu/kanade/domain/manga/model/ComicInfo.kt
```diff
@@ -0,0 +1,126 @@
...
+    val title: ComicInfoTitle?,
+    val series: ComicInfoSeries?,
+    val summary: ComicInfoSummary?,
+    val writer: ComicInfoWriter?,
+    val penciller: ComicInfoPenciller?,
+    val inker: ComicInfoInker?,
+    val colorist: ComicInfoColorist?,
+    val letterer: ComicInfoLetterer?,
+    val coverArtist: ComicInfoCoverArtist?,
+    val editor: ComicInfoEditor?,
```
arkon, line 19: Let's just leave out all of the various fields/classes that aren't being used. We can add them later if we need them.

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
Shamicen: I implemented the requested changes in my new commits

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -154,9 +159,122 @@ class LocalSource(
...
+                            val comicInfoFileHeader = archive.fileHeaders.firstOrNull { it.fileName.equals("ComicInfo.xml", true) }
+
+                            if (comicInfoFileHeader != null) {
+                                comicInfoFound = true
+                                archive.getInputStream(comicInfoFileHeader).buffered().use { archiveInputStream ->
+                                    archiveInputStream.mark(0)
+                                    val comicInfo = xml.decodeFromReader<ComicInfo>(AndroidXmlReader(archiveInputStream, encoding))
+                                    archiveInputStream.reset()
+                                    setAndCopyMetadata(comicInfo, archiveInputStream, manga, folderPath)
+                                }
```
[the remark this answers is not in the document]
Shamicen: We can't return anything in line 230 since it's in  withIOContext and use

But I'll change it to hasComicFile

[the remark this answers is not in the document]
Shamicen: I Changed to has ComicFile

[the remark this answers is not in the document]
Shamicen: Will resolve this for now since I don't think there is a way to return from `withContext{}` directly

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
Shamicen: I pulled the logic in a separate function like requested to replace break with return, but like I already wrote in the review from @ghostbear I'm unable to get rid of the Boolean since this part is inside `withIOContext{}` and I don't think there is a way to return directly from it.

Shamicen: I cleaned it up now by changing the couroutine scope of the whole function. I hope this does not go against best practices will revert to the clunky Boolean method if it does.

 `private suspend fun getComicInfoFromArchive(chapterArchives: List<File>, manga: SManga, encoding: String, folderPath: String?): Boolean = withIOContext {`

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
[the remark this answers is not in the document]
Shamicen: I cleaned it up now by changing the Coroutine scope of the whole function. I hope this does not go against best practices will revert to the clunky Boolean method if it does.

`private suspend fun getComicInfoFromArchive(chapterArchives: List<File>, manga: SManga, encoding: String, folderPath: String?): Boolean = withIOContext {`

Shamicen:
I implemented functionality to create ComicInfo.xml files on chapter download Shamicen@cbd35d6.

I decided against just pushing them to this branch since @arkon   said:
>It'd probably make sense to only generate them on-demand, i.e. via a "move to local" sort of feature (https://github.com/tachiyomiorg/tachiyomi/issues/2289) or if it's to be used as the data storage for in-app manga detail edits

on my last commit that did something similar with Details.json files.

I do think that it would be a good Idea to generate ComicInfo files at this point since we would not have to write ComicInfo.xml files to all `CBZ` archives in the source that has to be migrated, after they have already been created. We also have convenient access to chapter specific Information at this point namely the chapter Name and Scanlator. 
We could also add the upload date at this point but it doesn't fit 100% with the Anansi project documentation: 
`Usually contains the release date of the book.` 
But we could just add a note that it is not the publishing but upload date to the Note field and use it anyway.

It also makes using chapters created by Tachiyomi in 3rd party apps that support the ComicInfo standard a lot more seamless.

Now a few Questions to the details of the Implementation if it is OK to go ahead with creating ComicInfo files at this point.

Settings: should the users be able to choose
1. nothing ComicInfo files are always generated at the chapter root and in `CBZ` archives.
2. Choose if they will be generated at the chapter root but always generate them for `CBZ` archives.
3. Choose between the 3 following options via a dropdown menu after choosing if they should be created at all: `create ComicInfo files at the chapter root and CBZ archives`, `create ComicInfo files in CBZ archives`, `create ComicInfo files at the chapter root`.
(Performance shouldn't be an Issue as creating the ComicInfo files is basically instantaneous.)

Should we add a custom field against the spec: Should we go against the Schema and use a custom PublishingStatusTachiyomi field:
There are 2 things that should be considered regarding this question:

1. I found this comment on the Anansi project Github regarding custom fields: 
> Just to point out, you can still include it in your ComicInfo.xml files. No consumer to my knowledge validates against the spec formally... https://github.com/anansi-project/comicinfo/issues/31#issuecomment-1183319783

2. There is already an open Issue about implementing an official Status field https://github.com/anansi-project/comicinfo/issues/5#issue-1109397095 and the proposed field specs appear to satisfy all of Tachiyomis needs. But the Issue has been open since January and the opinions on implementing it seem to be split.

We could implement a custom field now and just switch to the spec one if it gets merged. 

below is an example output were I changed the content to be more descriptive. I also tested the created file with a online xml validator against the Anansi project 2.1 Schema file and it passed the validation. 

```
<?xml version='1.0' encoding='UTF-8' ?>
<ComicInfo xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <Title>example chapter name</Title>
    <Series>example series name (Manga Title)</Series>
    <Summary>example description</Summary>
    <Writer>Author</Writer>
    <Translator>example Scanlator (chapter.scanlator from Downloader.Kt)</Translator>
    <Genre>Genre 1, Genre 2, Genre 3, Genre 4, Genre 5, Genre 6</Genre>
    <Web>example Tracker link(track.remoteUrl)</Web>
</ComicInfo>
```

I don't really know that much about xml and this may go a bit against the standard since it calls for `xmlns:xs` instead of `xmlns:xsd` But this is the exact header all ComicInfo.xml files I could find on Github use.

I did not use the prefix and namespace field of the elements provided by our dependency since this resulted in  

```
<?xml version='1.0' encoding='UTF-8' ?>
<xsd:ComicInfo>
</xsd:ComicInfo>

(written from memory so could be a bit different)
```
like output which did not pass the validation and produced an error if opened in Android Studio.

I can open a new pull request with this Code and close this PR if everything is OK so far.

arkon:
I'll take a look at this tomorrow or on Saturday.

Shamicen:
2 more Ideas for using ComicInfo.xml files in the future:

1. Add some more fields to the trackers track variable like Description and Tags and use those as well.
2. Scan ComicInfo.xml files for all Manga (just chapter root for convenience) and allow users to customize their libraries that way. (I admit that I have no Idea how complex implementing this would be though)

This would also allow to more or less automatically use tracker Information instead of the Information provided by the sources if both are implemented.

arkon:
I'd like to keep this PR focused on the first part in supporting the file so that it's more manageable, i.e. being able to read them in for local source manga.

For further ideas, we should move discussion to a separate issue to figure out the details.

Based on what you've said thus far:
- I generally think the app needs to have less settings, so the most sensible way of creating the files would be at the chapter root (for usage by the app) and within individual archives (for external usage)
- If we want to add anything custom to the files, we should probably prefix them to avoid future collisions

As for this PR, it looks like a decent implementation to start with. I'll merge it into a branch and clean it up a bit before merging back into `master`.

Shamicen:
OK thank you very much will do.

Also thank you @ghostbear for the Code reviews and improvements.

PR #7975, opened 2022-09-10
Extract translation from app module

ghostbear:
Probably best to merge #7953 before this one and lock Weblate before merging this

- Prep for further extraction to modules in the nearish future


.gitignore
```diff
@@ -13,7 +13,7 @@
 app/**/output.json
 
 # Hebrew assets are copied on build
-app/src/main/res/values-iw/
+i18n/src/main/res/values-iw/
 
 # Generated
 locales_config.xml
```
arkon, line 19: Probably makes sense to move things into per-module `.gitignore`s at some point.

PR #8014, opened 2022-09-15
Extract source api from app module

ghostbear:


core/src/main/java/eu/kanade/tachiyomi/network/NetworkHelper.kt
```diff
@@ -1,20 +1,20 @@
...
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
-import uy.kohesive.injekt.injectLazy
 import java.io.File
 import java.util.concurrent.TimeUnit
 
 class NetworkHelper(context: Context) {
 
-    private val preferences: PreferencesHelper by injectLazy()
+    // TODO: Abstract preferences similar to 1.x
```
arkon, line 16: Are you planning on doing that soon? Not a fan of losing the type safety here...

ghostbear: PreferenceHelper is too tightly coupled to things, imo, so I will probably have to look at it very soon

PR #8016, opened 2022-09-16
Upgrade AGP 7.3.0

ivaniskandar:
Removed the pinned R8 version as it doesn't work with this AGP. So this may cause some issues with extensions.


arkon:
Did you test with R8?

ivaniskandar:
Yeah I've used this on my personal build for a while

app/proguard-rules.pro
```diff
@@ -13,6 +12,14 @@
 -keep,allowoptimization class app.cash.quickjs.** { public protected *; }
 -keep,allowoptimization class uy.kohesive.injekt.** { public protected *; }
 
+# From extensions-lib
```
arkon, line 15: Might be more convenient to just annotate things with `@Keep`

source-api/consumer-proguard.pro
```diff
@@ -0,0 +1,5 @@
+-keep,allowoptimization class eu.kanade.tachiyomi.source.model.** { public protected *; }
```
arkon, line 1: Wouldn't we realistically want to keep everything in this module?

ivaniskandar: Yeah but the `eu.kanade.tachiyomi.source` package also contains some stuff from main app module

app/proguard-rules.pro
ivaniskandar: Honestly I prefer keeping the rules centralized but I can try using it.

arkon: I'll merge this one first to keep the changes somewhat isolated.

PR #8020, opened 2022-09-17
Use 1.x preference abstraction

ghostbear:
- Uses SharedPreferences compared to 1.x impl which uses DataStore but it breaks all settings screens currently
- Move PreferencesHelper to new PreferenceStore
  - PreferencesHelper should be split into smaller preference stores and be in core or domain
- Remove flow preferences as new PreferenceStore handles changes for us


arkon:
A few issues I noticed when trying this changeset:

- Sources/extensions/migrate screens don't load
- Downloaded only / Incognito mode toggles in More don't have an immediate visual effect

ghostbear:
> A few issues I noticed when trying this changeset:
> 
> * Sources/extensions/migrate screens don't load
> * Downloaded only / Incognito mode toggles in More don't have an immediate visual effect

Should be fixed now

PR #8025, opened 2022-09-17
Read metadata from ComicInfo.xml files in Local source

### LINKED ISSUE #4965: [Feature Request] ComicInfo.xml Support for Local Manga

arkon:
Closes #4965

Followup to #7973 with some minor cleanup.

Schema docs: https://anansi-project.github.io/docs/comicinfo/schemas/v2.0

Kudos to @Shamicen for most of the legwork.


app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -134,27 +140,129 @@ class LocalSource(
...
-                    description?.let { manga.description = it }
-                    genre?.let { manga.genre = it.joinToString() }
-                    status?.let { manga.status = it }
+        // Augment manga details based on metadata files
+        val mangaDirFiles = getMangaDirsFiles(manga.url, baseDirsFile).toList()
+        val comicInfoMetadata = mangaDirFiles
+            .firstOrNull { it.name == COMIC_INFO_FILE || it.name == ".noxml" }
+        try {
+            // Top level ComicInfo.xml
+            if (comicInfoMetadata?.name == COMIC_INFO_FILE) {
```
ghostbear, line 156: when statement instead of if statement?

PR #8026, opened 2022-09-18
Only reset filterSheet to source settings instead of changing state

### LINKED ISSUE #8015: Filter Reset button doesn't work
### Steps to reproduce

1. Open the filter in any source. 
2. Change any of the options
3. Hit the reset button to reset filter options

### Expected behavior

The selected filter options should revert back to the default / first options. 

### Actual behavior

Any selected filters remain selected. 
Interestingly, the actual results page behind the filter select tray in the gets refreshed. 

### Crash logs

_No response_

### Tachiyomi version

Preview r4904

### Android version

Android 13

### Device

Pixel 6a

### Other details

_No response_

### Acknowledgements

geospiza-fortis:
Fixes #8015

This fixes the behavior found in #8015, where the filter button doesn't do anything.

The current behavior is caused by `resetFilter`:

https://github.com/tachiyomiorg/tachiyomi/blob/0086743a5311c22fb8c07f596ab5de384862a68a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourcePresenter.kt#L158-L161

When we first go into the filtering modal, the filter type of `presenter.currentFilter` is still set to `Filter.Popular` (or `Filter.Latest`). This means that the user input is completed ignored in the `resetFilter` call, and nothing changes in the filtering modal. Alternatively, the catalogue is changed from underneath if the previous filter was a user filter, before even hitting the filter button.

This change just sets the filterSheet to the source's default filters, without actually making state changes. I've done some manual testing, and this seems to be the correct behavior (at least from what I expect).

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
@@ -124,8 +124,7 @@ open class BrowseSourceController(bundle: Bundle) :
                 presenter.setSourceFilter(presenter.filters)
             },
             onResetClicked = {
-                presenter.resetFilter()
-                filterSheet?.setFilters(presenter.filterItems)
+                filterSheet?.setFilters(presenter.source!!.getFilterList().toItems())
             },
         )
 


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
```diff
@@ -124,8 +124,7 @@ open class BrowseSourceController(bundle: Bundle) :
                 presenter.setSourceFilter(presenter.filters)
             },
             onResetClicked = {
-                presenter.resetFilter()
```
arkon, line 127: @ghostbear Was `resetFilter` meant just for resetting the text search query?

ghostbear: `resetFilter` was just meant to reset filters. But we can make it reset the text search query as well, as there is no obvious way to do that currently.

happywillow0:
I was just testing out this change. While this does result in my expected behavior of the filters tray, after pressing the reset button, new filter selections are not being applied / pulled up. 

Example: Comic.fun
1. Filter sort by "Most Views" (working)
2. Filter sort by "Latest updated" (working)
3. Hit Reset, selections get reset. 
4. Filter sort by "Most Views" (pulls up Latest updated)

arkon:
Superseded by #8040, but thanks!

PR #7794, opened 2022-08-17
Create local source readable details.json file on chapter download

Shamicen:
If the option "Create details file on download" is checked, a local source readable details.json file will be created on chapter download. 

This means that if a manga gets copied from it's source folder to Tachiyomi/local all information like Title, Author, Artist, Description Genre and Status will be preserved. 

Could also be implemented without the toggle in Settings, since the details.json files are normally only a few hundred bytes in size and creating them had no noticeable impact on performance. 

This is my first ever Github commit I would be very grateful for all feedback and I apologize if my code does not meet the quality standards of Tachiyomi.

PR #8033, opened 2022-09-18
Create ComicInfo Metadata files on chapter download

Shamicen:
Supersedes #7794
I implemented the  capability to auto generate ComicInfo Metadata files on chapter download 

The output looks like this:
```
<?xml version='1.0' encoding='UTF-8' ?>
<ComicInfo xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    <Title>chapter.name"</Title>
    <Series>manga.title</Series>
    <Summary>manga.description</Summary>
    <Writer>manga.author</Writer>
    <Penciller>manga.artist</Penciller>
    <Translator>chapter.scanlator</Translator>
    <Genre>manga.genre</Genre>
    <Web>track?.remoteUrl</Web>
    <ty:PublishingStatusTachiyomi xmlns:ty="http://www.w3.org/2001/XMLSchema">manga.status converted to it's meaning for example: Ongoing</ty:PublishingStatusTachiyomi>
</ComicInfo>
```

I'm not the biggest expert on xml but it does pass online validators and our parser can parse it.
 
the header: 
```
<?xml version='1.0' encoding='UTF-8' ?>
<ComicInfo xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
```
is also the exact header used by all ComicInfo files I could find on Github

The files will be generated inside `CBZ` archives, in normal manga chapter folders and at the chapter root.


app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -567,6 +591,59 @@ class Downloader(
...
+            title = ComicInfoTitle(chapter.name),
+            series = ComicInfoSeries(manga.title),
+            summary = manga.description?.let { ComicInfoSummary(it) },
+            writer = manga.author?.let { ComicInfoWriter(it) },
+            penciller = manga.artist?.let { ComicInfoPenciller(it) },
+            translator = chapter.scanlator?.let { ComicInfoTranslator(it) },
+            genre = manga.genre?.let { ComicInfoGenre(it) },
+            web = track?.remoteUrl?.let { ComicInfoWeb(it) },
+            publishingStatusTachiyomi = when (manga.status) {
+                1 -> ComicInfoPublishingStatusTachiyomi("Ongoing")
```
ghostbear, line 625: Use the constant values in SManga instead of the magic numbers

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -266,6 +266,16 @@ class LocalSource(
...
+
+        when (comicInfo.publishingStatusTachiyomi?.value) {
+            "Ongoing" -> manga.status = 1
+            "Completed" -> manga.status = 2
+            "Licensed" -> manga.status = 3
+            "Publishing finished" -> manga.status = 4
+            "Cancelled" -> manga.status = 5
+            "On hiatus" -> manga.status = 6
+            else -> { manga.status = 0 }
+        }
```
ghostbear, line 278: Same here use the constant value in SManga

Also you can move out the assignment

```suggestion
        manga.status = when (comicInfo.publishingStatusTachiyomi?.value) {
            "Ongoing" ->  1
            "Completed" -> 2
            "Licensed" -> 3
            "Publishing finished" -> 4
            "Cancelled" -> 5
            "On hiatus" -> 6
            else -> 0
        }
```

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -519,29 +537,35 @@ class Downloader(
...
+
             Download.State.DOWNLOADED
         } else {
             Download.State.ERROR
         }
     }
 
     /**
-     * Archive the chapter pages as a CBZ.
+     Archive the chapter pages as a CBZ and create a ComicInfo.xml file inside the archive.
```
ghostbear, line 557: Feels like this function has an unnecessary responsibility to create the ComicInfo.xml file. Make the creation of the ComicInfo.xml file a separate function.

You can probably use the `createComicInfoAtChapterRoot` as a base as the creation of the ComicInfo.xml seems to be a bit DRY

```kt
private fun createComicInfoAt(dir: UniFile, downloadManga: Manga, chapter: Chapter) {
        File("${dir.filePath}/ComicInfo.xml").outputStream().use { it.write(getComicInfo(downloadManga, chapter)) }
}
```

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -567,6 +591,59 @@ class Downloader(
...
+                6 -> ComicInfoPublishingStatusTachiyomi("On hiatus")
+                else -> { ComicInfoPublishingStatusTachiyomi("Unknown") }
+            },
+            inker = null,
+            colorist = null,
+            letterer = null,
+            coverArtist = null,
+            tags = null,
+        )
+        val format = XML {
```
ghostbear, line 639:
```kt
XML {
  xmlDeclMode = XmlDeclMode.Charset
  indent = 4
  xmlVersion = XmlVersion.XML10
}
```
This should be added to a DI module and injected into this file as we might want to use it elsewhere

Shamicen: First of all thank you for the review.

This is already in AppModule.kt:
```
addSingletonFactory {
            XML {
                unknownChildHandler = UnknownChildHandler { _, _, _, _, _ -> emptyList() }
                autoPolymorphic = true
            }
        }
```
should I add it there? Sorry I have never worked with DI before.

ghostbear: Move `xmlDeclMode`, `indent`, and `xmlVersion` to the already defined XML instance in AppModule, as they only seem to affect the writer and not the reader. Then inject the XML instance into the Downloader either in the constructor with `Injekt.get()` or define a `val` with `by injektLazy()`

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -567,6 +596,54 @@ class Downloader(
...
+    /**
+     * returns a ByteArray containing the Manga Metadata of the chapter to download in ComicInfo.xml format
+     *
+     * @param downloadManga the manga of the chapter to download.
+     * @param chapter the name of the chapter to download
+     */
+    private fun getComicInfo(downloadManga: Manga, chapter: Chapter): ByteArray {
+        val track: Track? = runBlocking { (getTracks.await(downloadManga.id)).firstOrNull() }
+
+        val manga = downloadManga.toSManga()
```
ghostbear, line 619: I don't think the conversion is necessary as `Manga` should hold everything necessary

ghostbear (commented):
Side note can you use `Co-authored-by` instead of mentioning me in commits. After a few commits, I start getting spammed mentioned. Thank you for understanding.


You can use https://coauthoredby.netlify.app/ to generate the `Co-authored-by` bit

```
Your commit message

Co-authored-by: name <name@example.com>
```

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -519,29 +542,35 @@ class Downloader(
         download.status = if (downloadedImages.size == download.pages!!.size) {
             // Only rename the directory if it's downloaded.
             if (preferences.saveChaptersAsCBZ().get()) {
-                archiveChapter(mangaDir, dirname, tmpDir)
+                archiveChapter(mangaDir, dirname, tmpDir, download.manga, download.chapter)
```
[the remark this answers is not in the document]
Shamicen: Do services normally scan for ComicInfo files if they aren't in `CBR`/`CBZ ` archives?

We already create them at the chapter root for Tachiyomi itself.

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -567,6 +596,54 @@ class Downloader(
...
 
+    /**
+     * Creates a ComicInfo.xml file at the chapter root
+     *
+     * @param mangaDir the manga directory of the download.
+     * @param downloadManga the manga of the chapter to download.
+     * @param chapter the chapter to download
+     */
+    private fun createComicInfoFile(mangaDir: UniFile, downloadManga: Manga, chapter: Chapter) {
+        File("${mangaDir.filePath}/ComicInfo.xml").outputStream().use { it.write(getComicInfo(downloadManga, chapter)) }
```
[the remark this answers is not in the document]
Shamicen: I made it that way to update them once a new download starts. 
Shouldn't be that expensive since creating the ComicInfo files is basically instantaneous (0,026s per `createComicInfo()` call) compared to the time it takes to download just one image.

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
Shamicen: Isn't it kinda weird to use `SManga.ONGOING.toLong` if manga isn't `SManga` itself?

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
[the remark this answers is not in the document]
ghostbear: Should probably flush the file then as if not done it can cause corruption if the content size changes

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
Shamicen: Actually I'll just change it now and change it back if this goes against good style

ghostbear: `Manga` and `SManga` share the statuses so it's okay for you to use `toLong`

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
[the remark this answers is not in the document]
Shamicen: OK will change it

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
Shamicen: > `Manga` and `SManga` share the statuses so it's okay for you to use `toLong`

OK thank you

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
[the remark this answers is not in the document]
Shamicen:
```
File("${mangaDir.filePath}/ComicInfo.xml").outputStream().use { stream ->
            stream.flush()
            stream.write(getComicInfo(manga, chapter))
        }
```
Would this be OK?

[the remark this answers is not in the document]
ghostbear: You can do it like how the BackupManager does it 

https://github.com/tachiyomiorg/tachiyomi/blob/4ea512f6c22b76d1d0e3f777a7b15fd675f2eb97/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt#L121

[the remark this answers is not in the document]
Shamicen: Implemented the changes in the new commit

PR #8065, opened 2022-09-24
support reading and writing password protected CBZ/Zip archives

Shamicen:
Added support for reading and writing password protected `Zip`/`CBZ` archives using `zip4j`.

This implementation still uses the normal java `ZipFile` in `ZipPageLoader.kt` if the archive is not password protected, because the `splitTallImages` function does not work with InputStreams from zip4j `ZipFiles` for some reason. I would add more checks to make sure that `splitTallImages` does not get called with a zip4j ZipFile if this can't be fixed.

Their are some more details that need to be changed (especially regarding user input) but I wanted to get a general opinion before putting any more work into this PR.


arkon:
General consensus from a couple of contributors is that this doesn't belong in the main app. It may be applicable to one of the 18+ focused forks.

Shamicen:
> General consensus from a couple of contributors is that this doesn't belong in the main app. It may be applicable to one of the 18+ focused forks.

OK

PR #8074, opened 2022-09-25
Split the rest of the preferences in PreferencesHelper

ghostbear:


app/src/main/java/eu/kanade/domain/ui/model/ThemeMode.kt
```diff
@@ -0,0 +1,15 @@
+package eu.kanade.domain.ui.model
+
+/* ktlint-disable experimental:enum-entry-name-case */
+
+// Keys are lowercase to match legacy string values
+/* ktlint-disable enum-entry-name-case */
+enum class ThemeMode {
```
AntsyLich, line 7: Migrate the values to upper case?

PR #8097, opened 2022-09-27
MangaScreen: Adjust insets

ivaniskandar:
I should have done this when Scaffold was rebased


app/src/main/java/eu/kanade/presentation/components/MangaBottomActionMenu.kt
```diff
@@ -86,7 +91,7 @@ fun MangaBottomActionMenu(
             }
             Row(
                 modifier = Modifier
-                    .navigationBarsPadding()
+                    .padding(WindowInsets.navigationBars.only(WindowInsetsSides.Bottom).asPaddingValues())
```
arkon, line 94: Won't this be weird with the 3 button nav bar on the side on phones?

ivaniskandar: Yeah I've made some changes related to this

PR #8106, opened 2022-09-28
Migrate Accompanist SwipeRefresh to Compose PullRefresh

stevenyomi:
~~Not bumping Compose version because of Scaffold + AnimatedVisibility issue in beta, see https://issuetracker.google.com/issues/247605039.~~

~~Draft because indicator sometimes stays on the screen when idle or glitches when refreshing, see https://issuetracker.google.com/issues/248274004.~~

Fixes are extracted to #8169.

Issue with A12 stretch overscroll is documented in #8168.


stevenyomi:
For the stretch overscroll issue, we can wrap it with the following code as a workaround for A12+, but this disables the stretch VFX at the bottom of the list (and top of the list on fling). See https://issuetracker.google.com/issues/233515751.

```kt
CompositionLocalProvider(LocalOverscrollConfiguration provides null) {
    // ...
}
```

PR #8107, opened 2022-09-28
Requests: add `GET(HttpUrl)` overload

stevenyomi:
Many extensions use `HttpUrl.Builder` to build a search request. This overload allows them to do `GET(urlBuilder.build(), headers)` instead of `GET(urlBuilder.toString(), headers)`, saving an unnecessary URL parse.

OkHttp's `Request.Builder.url(String)` silently converts WebSocket URLs to HTTP ones, but we don't actually need that, so doing `url.toHttpUrl()` should be safe.


AntsyLich:
Open a PR in [tachiyomiorg/extensions-lib](https://github.com/tachiyomiorg/extensions-lib) too

stevenyomi:
Done: https://github.com/tachiyomiorg/extensions-lib/pull/7.

arkon:
Any reason you're not also adding it for the other methods?

stevenyomi:
> Any reason you're not also adding it for the other methods?

In most cases, people need to manually build the URL (path, queries, fragments) with proper escaping only for GET requests. You usually POST, PUT, etc, to static URLs - string concatenations should be sufficient.

PR #8109, opened 2022-09-28
Improve DelayedTrackerStore

AntsyLich:
- Add to delayed tracking store when it fails to update tracker
- Only remove item from store when it successfully updates


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -843,13 +844,22 @@ class ReaderPresenter(
...
+                                    if (context.isOnline()) {
+                                        service.update(updatedTrack.toDbTrack(), true)
+                                        insertTrack.await(updatedTrack)
+                                        success = true
+                                    }
+                                } catch (e: Exception) {
+                                    exception = e
+                                }
+                                if (success.not()) {
                                     delayedTrackingStore.addItem(updatedTrack)
```
ghostbear, line 859: Can't this be done in the catch?

AntsyLich: if i throw a exception when device is not online.

ghostbear: Do it

PR #8111, opened 2022-09-29
gives ComicInfo.xml files priority over noxml files if both are at th…

Shamicen:
This gives `ComicInfo` files priority over `noxml` files. If both a `.noxml` files and a `ComicInfo.xml` file are inside the chapter root Tachiyomi will now use the ComicInfo file to set Metadata from it.


app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
```diff
@@ -151,7 +151,7 @@ class LocalSource(
         try {
             val mangaDirFiles = getMangaDirsFiles(manga.url, baseDirsFile).toList()
             val comicInfoMetadata = mangaDirFiles
-                .firstOrNull { it.name == COMIC_INFO_FILE || it.name == ".noxml" }
+                .lastOrNull { it.name == COMIC_INFO_FILE || it.name == ".noxml" }
```
arkon, line 154: Considering we already fetch all the files as `mangaDirFiles`, we should just check for the 2 files individually. If a `ComicInfo.xml` file exists, we should delete the `.noxml` file.

arkon:
Does this fix #8112?

Shamicen:
> Does this fix #8112?

No

app/src/main/java/eu/kanade/tachiyomi/source/LocalSource.kt
Shamicen: OK will do

Shamicen: I implemented it in the new commit

PR #8114, opened 2022-09-29
Add methods to get manga and chapter url

AntsyLich:
Some fork supports directly opening chapters with webview hence `getChapterUrl`


source-api/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt
```diff
@@ -356,6 +356,26 @@ abstract class HttpSource : CatalogueSource {
...
     }
 
+    /**
+     * Returns the url of the provided manga
+     *
+     * @param manga the manga
+     * @return url of the manga
+     */
+    open fun getMangaUrl(manga: SManga): String {
+        return "$baseUrl${manga.url}"
```
alessandrojean, line 366: Maybe make it return the `mangaDetailsRequest(manga).url.toString()` instead?

So for existing sources (that doesn't rely on an API) and have changed it, the new method will already use the correct URL without needing a new update to the extension.

source-api/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt
```diff
@@ -356,6 +356,26 @@ abstract class HttpSource : CatalogueSource {
...
+    }
+
+    /**
+     * Returns the url of the provided chapter
+     *
+     * @param chapter the chapter
+     * @return url of the chapter
+     */
+    open fun getChapterUrl(chapter: SChapter): String {
+        return "$baseUrl${chapter.url}"
```
alessandrojean, line 376: Probably same here?

source-api/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt
```diff
@@ -356,6 +356,26 @@ abstract class HttpSource : CatalogueSource {
         }
     }
 
+    /**
```
arkon, line 359: Can you add `@since extensions-lib 1.4` to these?

AntsyLich: Done

PR #8116, opened 2022-09-29
Reduce some usages of `toDbManga`

AntsyLich:


app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -99,6 +99,38 @@ data class Manga(
...
+        }
+
+        manga = manga.copy(status = other.status.toLong())
+
+        manga = manga.copy(updateStrategy = other.update_strategy)
+
+        if (!initialized) {
+            manga = manga.copy(initialized = other.initialized)
+        }
+        return manga
```
stevenyomi, line 131:
```suggestion
        val author = other.author ?: author
        val artist = other.artist ?: artist
        val description = other.description ?: description
        val genres =  if (other.genre != null) {
            other.getGenres()
        } else {
            genres
        }
        val thumbnailUrl = other.thumbnail_url ?: thumbnailUrl
        return manga.copy(
            author = author,
            artist = artist,
            description = description,
            genres = genres,
            thumbnailUrl = thumbnailUrl,
            status = other.status.toLong(),
            updateStrategy = other.update_strategy,
            initialized = other.initialized && initialized
        )
```
Do we really need to create so many copies of this? I think this is more efficient.

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -99,6 +99,28 @@ data class Manga(
         it.initialized = initialized
     }
 
+    fun copyFrom(other: SManga): Manga {
+        val author = other.author ?: author
+        val artist = other.artist ?: artist
+        val description = other.description ?: description
+        val genres =  if (other.genre != null) {
```
stevenyomi, line 106:
```suggestion
        val genres = if (other.genre != null) {
```
I typed an extra space here, sorry.

PR #8121, opened 2022-10-01
Tweak behavior on Browse screen

### LINKED ISSUE #8088: Tapping on Source Name while Browsing Manga and going back exits Browsing
### Steps to reproduce

1. Install latest preview (r4963, as of writing this)
2. go to Browse, allow perms, see that Multi, English are enabled (shouldn't matter, I think)
3. open a source in browse. Eg: MangaDex
4. open whichever manga. Eg: open Meiruko-chan
5. tap on the Source Name in Manga info (MangaDex (EN); its on the right side of Ongoing Status)
6. Now try pressing/doing Back action to go back to the previous screen

### Expected behavior

A browse search is triggered that searches the source with the source name itself
and we are able to go back to browsing manga if we want to

### Actual behavior

In step 5, A browse search is triggered that searches the source with the source name itself
In step 6, the Back action completely exits the Browse section and we're brought back to the Sources screen. So, now I have to tap on Source page and scroll to whichever manga I was seeing again.

### Crash logs

NA

### Tachiyomi version

r4963

### Android version

12

### Device

SM-N770F

### Other details

in step 5, the source search is triggered. If the manga were in library, then it searches the library for other manga from the same source. 
But if the manga isn't added to library, is there a need to tap and search by source? most sources dont have an entry which is the source itself (Exception: MangaDex contains Bocchi-sensei teach me MangaDex - (an outdated) introductory guide to using MangaDex; but you can search that directly or much better: see it on their website)

### Acknowledgements

ghostbear:
- Fix system back popping controller (fixes #8088)
- Add behavior to system back to cancel search toolbar before clearing current search or popping controller


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceController.kt
```diff
@@ -114,11 +110,32 @@ open class BrowseSourceController(bundle: Bundle) :
...
+    // Let Compose view handle this
+    override fun handleBack(): Boolean {
+        val dispatcher = (activity as? OnBackPressedDispatcherOwner)?.onBackPressedDispatcher ?: return false
+        return if (dispatcher.hasEnabledCallbacks()) {
+            dispatcher.onBackPressed()
+            true
+        } else {
+            false
+        }
+    }
```
ivaniskandar, line 129: Should be able to put this inside `FullComposeController` instead

PR #8122, opened 2022-10-01
Downloader: Don't trust index provided by source

AntsyLich:


CarlosEsco:
Maybe @arkon disagrees but I don't think we should hiding issues with extensions that are broken and need fixed.

PR #8134, opened 2022-10-02
Implementation of pdf loader

sagix:
I saw that issues were reported for pdf: https://github.com/tachiyomiorg/tachiyomi/issues/1751
I hope is implementation can answer to it :) 

Android lets us read a pdf though the object PdfRenderer.
This renderer transforms pdf pages to Bitmaps.

With that, two PageLoader were created:
1. PdfPageLoader: that streams pdf pages to pngs
2. CachedPdfPageLoader: that writes pdf pages to pngs in the cache (like HttpPageLoader)

Rendering pdf can be slow, so having the page cached gives a better experience.
PdfRenderer can only render one page at a time.

Also, the pdf pages are encoded a pngs, because [image-decoder](https://github.com/tachiyomiorg/image-decoder) cannot read bitmaps directly.
So maybe if it could, encoded and decoded pngs could be skipped and the streaming of pdf could be faster.

I'm open to feedbacks and to do further development on this feature :)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/cache/ChapterCache.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/cache/ChapterCache.kt
@@ -13,9 +13,11 @@ import kotlinx.serialization.json.Json
 import okhttp3.Response
 import okio.buffer
 import okio.sink
+import okio.source
 import uy.kohesive.injekt.injectLazy
 import java.io.File
 import java.io.IOException
+import java.io.InputStream
 
 /**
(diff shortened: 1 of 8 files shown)


arkon:
I feel like this is more of a "should" rather than "could" -- PDF is a pretty convoluted file format and people would be better off extracting images from one (and optionally repackaging it a ZIP or something), both from a portability and performance perspective. I'm already on the fence on whether the app should support EPUBs, but at least that's a "normal" e-book format.

PR #8141, opened 2022-10-03
Update german translations to fix typos

marcelbrode:
Update german translations to fix typos


stevenyomi:
You should update translation on Weblate instead of opening a separate PR: https://hosted.weblate.org/projects/tachiyomi/strings/de/

PR #8150, opened 2022-10-06
Fix #8149 - Chapter download state not updating

### LINKED ISSUE #8149: Chapter download state not updated when deleted when marked as read
### Steps to reproduce

1. Enable more > settings > downloads > delete chapters > after manually marked as read
2. Download a chapter of a manga that has a chapter above it
3. Hold down to select the chapter above the downloaded chapter
4. Press the "mark previous as read" button

### Expected behavior

The chapter should be deleted.

### Actual behavior

The chapter remains, unmodified.

### Crash logs

_No response_

### Tachiyomi version

App version: 0.13.6-5003 (standard, 6be9cccc7, 86, 2022-10-01T21:11Z)

### Android version

Android version: 12 (SDK 31)

### Device

Device model: SM-G781V Device product name: r8qsqw | "Samsung Galaxy S20 FE 5G" 

### Other details

[Video Example](https://user-images.githubusercontent.com/54383546/194221031-a6dfc18b-f449-4776-bfea-659421b2b41e.mp4) 

Please note that, as is demonstrated in the video, this function still works when chapters are individually selected. It is strictly the "mark previous as read" that is nonfunctional. 

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.

alessandrotedd:
Fixed chapter download state icon not updating after setting a chapter as read (issue #8149).
The chapter was indeed deleted but the UI wasn't being updated.


arkon:
I'm not sure this is really fixing it, it's just lucky enough to trigger a recomposition of the chapter items because the read status change occurs later. The existing logic probably makes more sense, i.e. it prioritizes actually marking something as read over any sort of file operations (although those occur asynchronously, which contributes to the problem).

The state change is actually [faked](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt#L686) when you explicitly delete something since the download state is normally consumed via a provider rather than an actual change on the model itself due to the (external) way that the download states are tracked.

I think we'd need to either fake the status change here too, or figure out a way to have the "real" download states actually be part of the state objects being composed.

PR #8158, opened 2022-10-08
MoreScreen: Remove title

ivaniskandar:
~~Also removes the state banners as it's redundant~~

image


arkon:
The banners are intentional to make users toggling them be more aware of their effect (although there's still people who don't remember how they enabled them...).

ivaniskandar:
Alright I added them back

PR #8160, opened 2022-10-08
 Hiding the keyboard

d-najd:
Closes https://github.com/tachiyomiorg/tachiyomi/issues/8157

This addresses https://github.com/tachiyomiorg/tachiyomi/issues/8157 by disabling the keyboard when a state change is detected in BaseController.kt or any of its subclasses

The reason why the code changes are located in BaseController.kt and not in something like LibraryController.kt is because while testing I had realised that this bug was affecting more classes than just LibraryController.kt, and more importantly to prevent the same bug to appear in future classes that extend BaseController.kt.

the "keyboard bug" can be triggered in the following classes:
- LibraryController.kt by searching manga from the "Library" tab
- MangaController.kt by searching manga from the "Browse" tab and choosing an extension
- SettingsController.kt by searching for specific setting from "More" tab and then Settings 
- LoginDialogPreference.kt not confirmed since I don't know how to trigger the activity at the moment, but if the activity gets triggered while keyboard is opened it will stay open without a doubt

**Thing to note**

The keyboard may be unintentionally closed on some activities that depend on this behaviour, but the logcat log should help with debugging that


app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
```diff
@@ -69,6 +73,12 @@ abstract class BaseController<VB : ViewBinding>(bundle: Bundle? = null) : Contro
         super.onChangeStarted(handler, type)
     }
 
+    fun View.hideKeyboard() {
```
arkon, line 76: Can you move this to `ViewExtensions.kt`?

It's also repeated in `TrackSearchDialog`, so could replace that too.

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
```diff
@@ -69,6 +73,12 @@ abstract class BaseController<VB : ViewBinding>(bundle: Bundle? = null) : Contro
...
+        val imm = context.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+        imm.hideSoftInputFromWindow(windowToken, 0)
+        logcat { "Attempting to hide the keyboard" }
```
arkon, line 79: Probably don't need this, it should be relatively obvious when debugging issues.

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/ui/base/controller/BaseController.kt
d-najd: sure

d-najd:
I accidentally closed the branch....

d-najd:
I was unable to get any to trigger any bug with the keyboard in TrackSearchDialog.kt so I put code to close the keyboard when attaching and closing the view, that should solve the problem.

and also I am unable to render track_search_dialog.xml for some reason in android studio, is it bug on my end?

app/src/main/java/eu/kanade/tachiyomi/util/view/ViewExtensions.kt
```diff
@@ -197,6 +198,11 @@ inline fun <reified T : Drawable> T.copy(context: Context): T? {
     }
 }
 
+inline fun View.hideKeyboard() {
+    val imm = context.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
```
AntsyLich, line 202: Move this to ContextExtensions.kt

d-najd: arkon had requested me to move this to ViewExtensions.kt so that's why I placed it here, should I still move it there and tell him?

app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
```diff
@@ -103,6 +104,11 @@ fun Context.notification(channelId: String, block: (NotificationCompat.Builder.(
     return builder.build()
 }
 
+fun View.hideKeyboard() {
+    val imm = context.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
```
AntsyLich, line 108: I meant to only move `Context.getSystemService` not the whole function

d-najd: I think the commit solves the issue, I am sorry for any inconvinience caused.

app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
```diff
@@ -185,6 +186,10 @@ val Context.powerManager: PowerManager
 val Context.keyguardManager: KeyguardManager
     get() = getSystemService()!!
 
+fun Context.getInputMethodManager(): InputMethodManager {
+    return getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
```
AntsyLich, line 190: Should make this similar to other extensions imo

d-najd: Does the new commit look good?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSearchDialog.kt
```diff
@@ -147,6 +148,9 @@ class TrackSearchDialog : DialogController {
 
     override fun onAttach(view: View) {
         super.onAttach(view)
+
+        view.hideKeyboard()
```
arkon, line 152: I meant replacing [this](https://github.com/tachiyomiorg/tachiyomi/blob/4b4be58d0d2cac6636858d57119e0657b1b56f59/app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackSearchDialog.kt#L106).

d-najd: ok, replaced it and there are no other places where the keyboard is closed using code, I think we can finally mark this issue  as closed?

PR #8162, opened 2022-10-08
Overwrite saved images instead of creating new ones

Platiplus:
## ISSUE
Closes [6988](https://github.com/tachiyomiorg/tachiyomi/issues/6988)

## Description
User AXVin opened a feature request to be able to overwrite images that are already in the device when hitting the "save button".

## Previous Behavior
When clicking the save button on an image that was already in the device, the app tried to save the image again, making Android save the same image again with a number at the end of the file name.

## Actual Behavior
This behavior is now fixed. The app tries to find out if the image is already in the device and if it is, it just updates the file with the "new version" instead of duplicating the file.


app/src/main/java/eu/kanade/tachiyomi/data/saver/ImageSaver.kt
```diff
@@ -85,6 +91,36 @@ class ImageSaver(
...
+        val projection = arrayOf(
+            MediaStore.MediaColumns._ID,
+            MediaStore.MediaColumns.DISPLAY_NAME,
+            MediaStore.Images.Media.MIME_TYPE,
+            MediaStore.MediaColumns.RELATIVE_PATH,
+            MediaStore.MediaColumns.DATE_MODIFIED,
+        )
+
+        val selection = "${MediaStore.MediaColumns.RELATIVE_PATH}='$relativePath' AND " +
+            "${MediaStore.MediaColumns.DISPLAY_NAME}='$imagePath.jpg'"
```
ghostbear, line 105: Should probably use the file extension, instead of `jpg`, which is in the `Image` data class

Platiplus: Sorry, totally missed the fact that we already had the filename with the extension on a variable up there hahaha

app/src/main/java/eu/kanade/tachiyomi/data/saver/ImageSaver.kt
```diff
@@ -85,6 +91,36 @@ class ImageSaver(
 
         return destFile.getUriCompat(context)
     }
+
+    private fun getExistingImageUriOrNullQ(relativePath: String, imagePath: String): Uri? {
```
ghostbear, line 95:
```suggestion
    private fun getExistingImageUriOrNull(relativePath: String, imagePath: String): Uri? {
```

app/src/main/java/eu/kanade/tachiyomi/data/saver/ImageSaver.kt
```diff
@@ -85,6 +91,36 @@ class ImageSaver(
...
+        val selection = "${MediaStore.MediaColumns.RELATIVE_PATH}='$relativePath' AND " +
+            "${MediaStore.MediaColumns.DISPLAY_NAME}='$imagePath'"
+
+        context.contentResolver.query(
+            MediaStore.Images.Media.EXTERNAL_CONTENT_URI,
+            projection,
+            selection,
+            null,
+            null,
+        ).use { c ->
```
ghostbear, line 113: I would suggest using a more descriptive name
```suggestion
        ).use { cursor ->
```

app/src/main/java/eu/kanade/tachiyomi/data/saver/ImageSaver.kt
ghostbear, line 95: You could make this function take in a lambda which could produce the default value 

```diff
-private fun getExistingImageUriOrNullQ(relativePath: String, imagePath: String): Uri? {
+private fun findUriOrDefault(relativePath: String, imagePath: String, default: () -> Uri): Uri {
    // Your code
-   return null
+   return default()
}
```

```kt
val picture = findUriOrDefault(folderRelativePath, ("$imageLocation$filename")) {
    context.contentResolver.insert(
        pictureDir,
        contentValues,
    ) ?: throw IOException(context.getString(R.string.error_saving_picture))
}
```

ghostbear, line 95: You can decorate the function with `@RequiresApi(Build.VERSION_CODES.Q)` to silence the warnings in this function. The caller already checks if Android Q APIs are available

arkon (approved):

ghostbear (approved):

Issue #8168, opened 2022-10-09
SwipeRefresh/PullRefresh issue with A12 stretch overscroll


stevenyomi:
### Describe your suggested feature

This is extracted from #8106.

Accompanist SwipeRefresh and Compose PullRefresh both have issue with A12 stretch overscroll effect, see https://github.com/google/accompanist/issues/1059 and video below.

We might be using the following code as a workaround for A12+, but this disables the stretch VFX at the bottom of the list (and top of the list on fling). See https://issuetracker.google.com/issues/233515751.

```kt
CompositionLocalProvider(LocalOverscrollConfiguration provides null) {
    // ...
}
```

~~When the issue is fixed, we should remove the workaround.~~ This workaround is reverted in https://github.com/tachiyomiorg/tachiyomi/commit/b4f2da12ea58b2f386f4d49356074c0a38351083 and currently not in use.

### Other details

https://user-images.githubusercontent.com/95685115/192684393-56865ea0-8f67-4fb3-8f6d-46efc1c3a984.mp4

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.13.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

stevenyomi:
https://issuetracker.google.com/issues/266874741

PR #8169, opened 2022-10-09
Refactor SwipeRefresh and fix some issues

stevenyomi:
- This is extracted from #8106
- Refactored SwipeRefresh to have the same API as PullRefresh so we can swap easily when that one is stable
- Forced an `enabled` parameter to fix refreshing being allowed in selection mode
- Disabled stretch overscroll effect for A12+ to avoid glitches, see #8168


arkon:
Merge conflict, btw.

stevenyomi:
Fixed. FYI the app is currently using different `TextButton`'s from both M3 and app components.

stevenyomi:
Overscroll reverted in https://github.com/tachiyomiorg/tachiyomi/commit/b4f2da12ea58b2f386f4d49356074c0a38351083.

PR #8173, opened 2022-10-09
Zoom wide image only when view is completely visible

zbue:
because in some cases, `Rect.intersects()` returns true when view is not completely visible at all


app/src/main/java/eu/kanade/tachiyomi/util/view/ViewExtensions.kt
```diff
@@ -198,17 +198,20 @@ inline fun <reified T : Drawable> T.copy(context: Context): T? {
...
-fun View?.isVisibleOnScreen(): Boolean {
+fun View?.isCompletelyVisibleOnScreen(): Boolean {
     if (this == null) {
         return false
     }
     if (!this.isShown) {
         return false
     }
     val actualPosition = Rect()
+    val isGlobalVisible = getGlobalVisibleRect(actualPosition)
```
arkon, line 209: What's the difference between this and L212?

PR #8181, opened 2022-10-10
Fix Scrollbar when the list contains sticky header

ivaniskandar:
Honestly I don't really like this solution I came up with, but currently there's no public way of knowing a sticky header item.


app/src/main/java/eu/kanade/presentation/browse/MigrateSourceScreen.kt
```diff
@@ -85,7 +85,7 @@ private fun MigrateSourceList(
     ScrollbarLazyColumn(
         contentPadding = contentPadding + topPaddingValues,
     ) {
-        stickyHeader(key = "header") {
+        stickyHeader(key = "stickyHeader") {
```
arkon, line 88: Use the constant?

PR #8182, opened 2022-10-10
`Date added` as extra information in migration manga list view

zbue:
image


AntsyLich:
This kinda seems useless imo. Total chapter count or last update date might be more useful

arkon:
I'm also not really sure this is useful. Are people trying to only migrate new or old additions or something?

zbue:
It's just there's no way a user can see when they did exactly added the manga to library so I put it there😅 Feel free to close🙂

PR #8198, opened 2022-10-14
Add bookmark filter

### LINKED ISSUE #8187: Add filter by bookmarks to the library

When viewing a manga, there is a filter option for bookmarked chapters, adding this to the general library filter would make it significantly easier to remember where the bookmarked chapters are, and return to them without having to remember the specific manga.

### Other details

_No response_

### Acknowledgements

SleepySwords:
## Objective
Resolves #8187

Adds a filter in the library for Manga that contains a bookmarked chapter.


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -218,6 +219,19 @@ class LibraryPresenter(
             }
         }
 
+        val filterFnBookmarked: (LibraryItem) -> Boolean = bookmarked@{ item ->
+            if (filterBookmarked == State.IGNORE.value) return@bookmarked true
+            return@bookmarked runBlocking {
```
SleepySwords, line 224: Not really sure if this is the way it is supposed to be done, or if there is a better method.

arkon: I think ideally there'd just be an actual SQL query (and corresponding repository and interactor implementations) that's used to just check if a particular manga has any bookmarked chapters.

SleepySwords: wdym

arkon: Rather than ultimately using [this query](https://github.com/tachiyomiorg/tachiyomi/blob/147455f99c6ebcfbd09ff91cbc4cb52ffbbae0ba/app/src/main/sqldelight/data/chapters.sq#L26) and sifting through all the data here to figure out if there's bookmarks:

1. Have a query that gets bookmarked chapters for a given manga
2. Have an interactor that calls that query and returns whether or not it's an empty list (i.e. whether there's bookmarks in a manga) instead of [the one currently being used](GetChapterByMangaId).

SleepySwords: ah ok

PR #8200, opened 2022-10-14
BrowseSourceToolbar: Match display mode dropdown to stable and change toolbar icon based on display mode

zbue:


app/src/main/java/eu/kanade/presentation/browse/components/BrowseSourceToolbar.kt
```diff
@@ -126,7 +128,12 @@ fun BrowseSourceRegularToolbar(
...
-                                imageVector = Icons.Outlined.Check,
+                                imageVector = Icons.Outlined.RadioButtonChecked,
+                                contentDescription = "",
+                            )
+                        } else {
+                            Icon(
+                                imageVector = Icons.Outlined.RadioButtonUnchecked,
                                 contentDescription = "",
                             )
                         }
```
ghostbear, line 139: Extract this to a Composable function called RadioIcon with a parameter if it's checked or not

app/src/main/java/eu/kanade/presentation/components/DropdownMenu.kt
```diff
@@ -26,3 +31,28 @@ fun DropdownMenu(
         content = content,
     )
 }
+
+@Composable
+fun RadioIcon(
```
ghostbear, line 36: I meant just extract the icon part but this work as well just the naming should be a bit different imo
```suggestion
fun RadioButton(
```

ghostbear (commented):
The build is failing due to unused imports in BrowseSourceToolbar

PR #8205, opened 2022-10-15
Retain bookmark of readded chapters

AntsyLich:


app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
```diff
@@ -136,11 +136,11 @@ class SyncChaptersWithSource(
 
         val deletedChapterNumbers = TreeSet<Float>()
         val deletedReadChapterNumbers = TreeSet<Float>()
+        val deletedBookmaredChapterNumbers = TreeSet<Float>()
```
arkon, line 139: Bookmarked

PR #8206, opened 2022-10-15
DeleteChaptersDialog: Add ability to include bookmarked chapters for deletion

zbue:


app/src/main/java/eu/kanade/presentation/manga/components/MangaDialogs.kt
```diff
@@ -1,18 +1,68 @@
...
 import eu.kanade.tachiyomi.R
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
 
 @Composable
 fun DeleteChaptersDialog(
+    showCheckbox: Boolean,
     onDismissRequest: () -> Unit,
     onConfirm: () -> Unit,
+    preferences: DownloadPreferences = Injekt.get(),
```
ghostbear, line 31: Hoist the value rather than using `Injekt.get()` as `DownloadPreferences` is available in the presenter

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -157,9 +157,14 @@ class MangaController : FullComposeController<MangaPresenter> {
...
             is Dialog.DeleteChapters -> {
                 DeleteChaptersDialog(
+                    checked = presenter.removeBookmarkedChapters.get(),
+                    showCheckbox = dialog.chapters.any { it.bookmark },
                     onDismissRequest = onDismissRequest,
-                    onConfirm = {
+                    onConfirm = { toggle ->
                         presenter.toggleAllSelection(false)
+                        if (dialog.chapters.any { it.bookmark }) {
+                            presenter.removeBookmarkedChapters.set(toggle)
```
arkon, line 166: Why is it just disabling the setting? I thought it'd be more of a "force" remove them even if it's set.

zbue: Uhhm... no??? Its a user-based preference so I thought that its not needed to force them

arkon: "Include bookmarked chapters" doesn't imply "Turn off the setting permanently until you go and turn it back on" though.

arkon:
Closing this for now, but please reopen it if it's updated to be a one-time force instead of just turning off the setting.

PR #8218, opened 2022-10-16
Add "Play" button on manga in library

### LINKED ISSUE #3071: "Play" button on manga in library

As a user, I'd like the ability to more easily and quickly resume a manga from the manga list. The process of resuming a manga today is as follows:

1. Click the manga in the list
2. Navigate to the "Chapters" tab
3. Click the "Play" (resume) button at the bottom right.

While not really inconvenient as it is today, it does make it a little tedious when you're catching up on 5+ different manga that each only have 1 or 2 updated chapters. You actually can spend quite a bit of time just clicking through UI to do a simple task.

### What/Requirements

To simplify the user experience of resuming manga, I suggest adding a "Play" button (similar to the one you see in the Chapters tab for a manga, except without the "Resume" text) on top of the manga image for each manga in the manga list. The button would be small enough that it doesn't interfere with the user's ability to simply click the manga image, which would still take them to the "Detail" view for that manga. Clicking the play button (which should be anchored to the bottom-right corner of each manga entry in the list) will act as if the user clicked the "Resume" button on the Chapters tab for that manga. It simply removes steps 1 and 2 in the steps given earlier.

d-najd:
Resolves #3071

- Adds "play" button to each of the items in the library.
- The "play" button has the same functionality as resume/start button in the fab in `MangaScreen`.
- Option to turn on/off the feature, off by default.
- The design is the same as compact form of the play button and follows theme colors.

~~- Does not include List Display Mode since I am not able to come up with good design -_-~~
~~- Downloaded chapter filter is **not** applied, more info provided in the review by AntsyLich~~

Some photos, (the design is [specterflare](https://github.com/specterflare)'s idea)

| Compact  | Comfortable | Cover-only | List | Display Settings |
| ------------- | ------------- |  ------------- | ------------- | ------------- |
| Screenshot_2022-10-19-20-36-15-756_eu kanade tachiyomi debug  | Screenshot_2022-10-20-19-47-33-049_eu kanade tachiyomi debug | Screenshot_2022-10-20-19-47-37-815_eu kanade tachiyomi debug |  | Screenshot_2022-10-20-19-47-46-103_eu kanade tachiyomi debug |


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
```diff
@@ -471,6 +474,11 @@ class LibraryPresenter(
             .reduce { set1, set2 -> set1.intersect(set2).toMutableList() }
     }
 
+    suspend fun getNextUnreadChapter(mangaId: Long): Chapter? {
+        val chapters = getMangaWithChapters.awaitChapters(mangaId)
+        return chapters.findLast { !it.read }
```
ghostbear, line 479: This should take user chapter sort into consideration. You can probably tweak the existing logic in `GetNextChapter`

app/src/main/java/eu/kanade/presentation/library/components/LibraryGridItemContinueReadingButton.kt
```diff
@@ -0,0 +1,47 @@
...
+    ) {
+        Icon(
+            painter = painterResource(R.drawable.ic_continue_reading_24dp),
+            contentDescription = null,
+            modifier = Modifier
+                .offset(y = (-.75).dp) // make the icon more "centered"
+                .size(24.dp),
+        )
+    }
+}
```
d-najd, line 47: Placing this in `LibraryGridCover` by adding it to the `content` parameter before passing it to `MangaGridCover`, this would allow removal code duplication in `LibraryComfortableGrid`, `LibraryCompactGrid`, `LibraryCoverOnlyGrid` but I am not able to figure out the right syntax.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
d-najd: `MangaPresenter` had code for sorting but I wasn't sure what it was sorting so I decided not to include it since it didn't seem to cause any problems this way, here is the code

```
    fun getNextUnreadChapter(): DomainChapter? {
        val successState = successState ?: return null
        return successState.processedChapters.map { it.chapter }.let { chapters ->
            if (successState.manga.sortDescending()) {
                chapters.findLast { !it.read }
            } else {
                chapters.find { !it.read }
            }
        }
    }
```

should I copy the logic from this instead?

Soitora:
Should be an option, off by default
And surely there should be better designs for this, it's not very aesthetic

d-najd:
> Should be an option, off by default And surely there should be better designs for this, it's not very aesthetic

good point, got any recommendations for better design?

also if there are no unread chapters left it will appear as clickable but will actually do nothing, I planned to mention it in the description and implement that feature sometime in the near future but forgot to do so.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
ghostbear: > also are you talking about GetNextChapter from HistoryRepositoryImpl?

Yes

> should I copy the logic from this instead?

Copy-pasting code isn't ideal they should share an interactor/use case because it will make it easier to change the behavior without finding all the code scattered in presenters. I think the `getNextChapter` function in `HistoryRepositoryImpl` can be split into `getNextChapter(mangaId: Long)` and `getNextChapter(mangaId: Long, chapterId: Long)` as the `chapterId` is only use in the first part of `getNextChapter(mangaId: Long, chapterId: Long)`

Soitora:
> good point, but imo I think it should be on by default given how many people requested this feature in the past.

The amount of people who has requested this is not high enough to warrant such a drastic change be on by default

I could maybe be neutral on this point if it didn't look so out of place 

> got any ideas how to improve the design?

Even back when it was added to forks years ago, I disliked the design, even now I am unable to think of something that looks natural

If I had any better looking examples to go by I would post it

d-najd:
Now that you mention it, it does look out of place, I will try to change the colors a bit till someone comes up with better design, I will also disable it by default and let arkon (or whoever will does the merge) decide if it should be enabled by default.

Thanks for the feedback.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
d-najd: ok I will do that, and after that I will create a separate request to replace code for next chapter in `MangaController` and other places where I can find code duplication.

specterflare:
Why don't you just reuse the compact form of the play button from the chapter list? It would better match every user's individual theme, and you could grey it out when there are no unread chapters. Maybe something like this:

thing

specterflare:
Why don't you just reuse the compact form of the play button from the chapter list? It would better match every user's individual theme, and you could grey it out (or just get rid of it) when there are no unread chapters.

Maybe something like this:

finaldfffff

Soitora:
> Maybe something like this:
> 
> finaldfffff
> 
> 

This looks way better

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaController.kt
```diff
@@ -343,8 +346,12 @@ class MangaController : FullComposeController<MangaPresenter> {
     // Chapters list - start
 
     private fun continueReading() {
-        val chapter = presenter.getNextUnreadChapter()
-        if (chapter != null) openChapter(chapter)
+        val presenterScope: CoroutineScope = MainScope()
```
AntsyLich, line 349: Use provided viewScope

app/src/main/java/eu/kanade/data/history/HistoryRepositoryImpl.kt
```diff
@@ -28,6 +28,22 @@ class HistoryRepositoryImpl(
...
+        val manga = handler.awaitOne { mangasQueries.getMangaById(mangaId, mangaMapper) }
+
+        val sortFunction: (Chapter, Chapter) -> Int = when (manga.sorting) {
+            Manga.CHAPTER_SORTING_SOURCE -> { c1, c2 -> c2.sourceOrder.compareTo(c1.sourceOrder) }
+            Manga.CHAPTER_SORTING_NUMBER -> { c1, c2 -> c1.chapterNumber.compareTo(c2.chapterNumber) }
+            Manga.CHAPTER_SORTING_UPLOAD_DATE -> { c1, c2 -> c1.dateUpload.compareTo(c2.dateUpload) }
+            else -> throw NotImplementedError("Unknown sorting method")
+        }
+
+        val chapters = handler.awaitList { chaptersQueries.getChaptersByMangaId(mangaId, chapterMapper) }
```
AntsyLich, line 41: You are ignoring filters that user might have set

d-najd: I think that the issue is resolved but I will do some testing just in case and close the issue after.

d-najd:
> > Maybe something like this:
> > finaldfffff
> 
> This looks way better

what about this?
Screenshot_2022-10-19-20-36-15-756_eu kanade tachiyomi debug

app/src/main/java/eu/kanade/presentation/library/components/LibraryGridItemContinueReadingButton.kt
d-najd: I will mark this as resolved since I don't think what I am asking is possible and nobody has come up with a solution.

arkon:
I'm holding off on merging this until after a stable release so let me know if this was for Hacktoberfest so I can label it as needed.

app/src/main/res/drawable/ic_continue_reading_24dp.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:autoMirrored="true" android:height="24dp"
```
AntsyLich, line 1: Is this needed since you changed to a button?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
```diff
@@ -579,15 +581,9 @@ class MangaPresenter(
...
+    suspend fun getNextUnreadChapter(): DomainChapter? {
         val successState = successState ?: return null
-        return successState.processedChapters.map { it.chapter }.let { chapters ->
-            if (successState.manga.sortDescending()) {
-                chapters.findLast { !it.read }
-            } else {
-                chapters.find { !it.read }
-            }
-        }
+        return getNextChapter.await(successState.manga.id)
```
AntsyLich, line 586: We should probably avoid a db call here since we already have everything needed

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibrarySettingsSheet.kt
```diff
@@ -282,12 +282,14 @@ class LibrarySettingsSheet(
...
         private val tabsGroup: TabsGroup
+        private val otherGroup: OtherGroup
 
         init {
             displayGroup = DisplayGroup()
             badgeGroup = BadgeGroup()
             tabsGroup = TabsGroup()
-            setGroups(listOf(displayGroup, badgeGroup, tabsGroup))
+            otherGroup = OtherGroup()
+            setGroups(listOf(displayGroup, badgeGroup, tabsGroup, otherGroup))
```
AntsyLich, line 292: Should be ok to put the option at top of badge group. @arkon what do you think?

app/src/main/java/eu/kanade/domain/library/service/LibraryPreferences.kt
```diff
@@ -81,6 +81,12 @@ class LibraryPreferences(
 
     // endregion
 
+    // region Other
+
+    fun showContinueReadingButton() = preferenceStore.getBoolean("display_continue_reading_button", false)
```
AntsyLich, line 86: Change the location depending on the other comments output

app/src/main/java/eu/kanade/data/history/HistoryRepositoryImpl.kt
```diff
@@ -28,6 +28,43 @@ class HistoryRepositoryImpl(
         }
     }
 
+    override suspend fun getNextChapter(mangaId: Long): Chapter? {
```
AntsyLich, line 31: This still doesn't handle user set filter lol

d-najd:
> I'm holding off on merging this until after a stable release so let me know if this was for Hacktoberfest so I can label it as needed.

Nope, just me wanting to help out tachiyomi 🙂

app/src/main/res/drawable/ic_continue_reading_24dp.xml
ghostbear: @AntsyLich having use cases in the presenter will just lead to different implementations for each get next chapter use case. And the database operation shouldn't take more than a few millisecond.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
ghostbear: @AntsyLich having use cases in the presenter will just lead to different implementations for each get next chapter use case. And the database operation shouldn't take more than a few millisecond.

app/src/main/res/drawable/ic_continue_reading_24dp.xml
d-najd: Nope my bad.

app/src/main/java/eu/kanade/data/history/HistoryRepositoryImpl.kt
d-najd: I misread the comment as "sorting" last time I read it for some reason 😑

I managed to add filters for bookmark and read but can't figure out for the life of me what to do with the download filter, the implementation in `MangaPresenter` uses `DownloadManager` as well as every other implementation and I don't think it is an option to use dependency injection here.

perhaps tag for downloaded can be added like bookmark and read for each chapter? Or I could just get the code from my first commit and let some code duplication happen....

oh and I removed the return method cases for when the manga is sorted like the method below since it altered the way that the original continue reading button worked and it was sometimes opening read chapters for some reason

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
d-najd: I agree with ghostbear here and if it takes noticable time we can always change it, but if someone modifies one of the buttons they will have different behaviour and it may become difficult to transfer the behaviour (like the problem I have right now with the downloaded filter)

app/src/main/java/eu/kanade/data/history/HistoryRepositoryImpl.kt
d-najd: I will implement the filtering outside of this `HistoryRepositoryImpl` since I have no idea how to go about the download filter

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryPresenter.kt
d-najd: @ghostbear getting next chapter with filters applied (we need filters or the behaviour will be altered) is impossible without adding `downloaded` tag to each chapter (like `bookmarked` or `read`), I will instead get list of all chapters and apply the filters after (there is existing code for that so I will refactor some code as well).

if adding `downloaded` tag is more desireable I can try to implement it that way or if you, ants or someone else has other idea I will be glad to hear it.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaPresenter.kt
d-najd: the latest commit avoids db call and different implementations

d-najd:
I will resolve the conflicts when the code gets reviewed.

app/src/main/java/eu/kanade/presentation/components/CommonMangaItem.kt
```diff
@@ -101,9 +118,10 @@ private fun BoxScope.CoverTextOverlay(title: String) {
             .fillMaxWidth()
             .align(Alignment.BottomCenter),
     )
+    val endPadding = if (showContinueReadingButton) 38.dp else 8.dp
```
arkon, line 121: This can be a bit less magic-number-y if you expose a constant from `CommonMangaItem` that you also use for the size there (i.e., take a size constant and add a bit here for padding).

app/src/main/java/eu/kanade/presentation/components/CommonMangaItem.kt
```diff
@@ -315,3 +344,30 @@ fun MangaListItem(
...
+            .align(Alignment.BottomEnd),
+        shape = MaterialTheme.shapes.small,
+        colors = IconButtonDefaults.filledIconButtonColors(
+            containerColor = MaterialTheme.colorScheme.primaryContainer,
+            contentColor = contentColorFor(MaterialTheme.colorScheme.primaryContainer),
+        ),
+    ) {
+        Icon(
+            imageVector = Icons.Filled.PlayArrow,
+            contentDescription = null,
```
arkon, line 368: Needs a content description.

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterFilter.kt
```diff
@@ -0,0 +1,82 @@
+package eu.kanade.tachiyomi.util.chapter
```
arkon, line 1: This stuff probably belongs in the domain module, but this is fine for now.

arkon (commented):
Could just put it at the end for the list item view?

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterFilter.kt
d-najd: might as well move it while I am at it then, should I move just the methods to `Chapter.kt` or the whole file?

ghostbear: Move the whole file that way it says more about what the file contains

PR #8221, opened 2022-10-17
Fix cache renew interval logic

AntsyLich:
The current logic will return false and renew cache until we actually hit the interval time period (which we don't since the last renew is overwritten later on.) Making it renew the download cache every single time.

This seems to be also have been the case in the previous logic


app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
```diff
@@ -195,7 +195,7 @@ class DownloadCache(
      */
     @Synchronized
     private fun renewCache() {
-        if (lastRenew + renewInterval >= System.currentTimeMillis()) {
+        if (lastRenew + renewInterval < System.currentTimeMillis()) {
```
arkon, line 198: This is never going to not be true if the later logic never gets executed (it'll always be `0 + 60000 < current time`). You'd need to also check if `lastRenew` is still 0.

arkon:
Wait, does your change even make sense?

It'll initially be `0 + 60000`, which is less than the current time, so it'll the initial cache renewal.

It'll then later check `some time + 60000` and skip it if it hasn't been reached yet, or do a cache renewal once we've surpassed it.

AntsyLich:
I feel so stupid

PR #8232, opened 2022-10-18
Add ability to open random manga

### LINKED ISSUE #1248: Open random manga from library
In #964, there was a mention of a random manga from each catalogue. I'm not requesting that, but rather one that will open a random manga that I've already added to my library.

Essentially, there will be a "Random" button at the top of the my library page that'll open a chapter from one of the manga that we have in the library. In the best case scenario, the random button would also have a set of options to choose from like: 

- Selecting from downloaded
- Selecting from unread
- Selecting the first unread chapter instead of the latest

zbue:
Closes #1248 
image


AntsyLich:
Add the option to disable it

THElegend5:
Probably have an option to only open downloaded chapter if there isn't already an option for it.

THElegend5:
Also maybe have an option to open random for that particular category? As per category makes more sense on large libraries rather than opening a random from the whole library.

jferdous70:
Also, please turn it off by default if it's not already

stevenyomi:
You should not make it a FAB since it's not a primary action.

Probably put it into the overflow menu, together with the global update button.

jferdous70:
it also closes #1248

zbue:
> Add the option to disable it

@AntsyLich Once I resolve all the things in this PR. 🙂

> Probably have an option to only open downloaded chapter if there isn't already an option for it.

@THElegend5 FYI, it works with filter and/or search in library 🙂

> Also maybe have an option to open random for that particular category? As per category makes more sense on large libraries rather than opening a random from the whole library.

This implementation already opens a random manga that is in library and only from current active category.
> You should not make it a FAB since it's not a primary action.

@stevenyomi See https://github.com/tachiyomiorg/tachiyomi/issues/1248#issuecomment-800747877

pikatwinky:
@zbue You are God, thanks for this PR. I think this is such an important feature which has been missing for years

zbue:
> I'm okay with it just being in the overflow menu, which would also remove the need for a setting for it.

Done.
Btw, user search query (e.g. searching genre) + opening random manga won't be possible.

app/src/main/java/eu/kanade/presentation/library/components/LibraryToolbar.kt
```diff
@@ -127,8 +135,34 @@ fun LibraryRegularToolbar(
             IconButton(onClick = onClickFilter) {
                 Icon(Icons.Outlined.FilterList, contentDescription = stringResource(R.string.action_filter), tint = filterTint)
             }
-            IconButton(onClick = onClickRefresh) {
-                Icon(Icons.Outlined.Refresh, contentDescription = stringResource(R.string.pref_category_library_update))
+            Box {
+                val (moreExpanded, onMoreExpanded) = remember { mutableStateOf(false) }
```
AntsyLich, line 139: Use `by` instead of `=` that way you'll get a mutable value instead of an immutable value and a function

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
```diff
@@ -42,6 +42,8 @@ class LibraryController(
     @Composable
     override fun ComposeContent() {
         val context = LocalContext.current
+        val getMangaForCategory = presenter.getMangaForCategory(page = presenter.activeCategory)
```
AntsyLich, line 45: Move this inside the lambda block

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
```diff
@@ -59,6 +61,15 @@ class LibraryController(
                 context.toast(if (started) R.string.updating_category else R.string.update_already_running)
                 started
             },
+            onClickOpenRandomManga = {
+                val items = getMangaForCategory.map { it.libraryManga.manga.id }
+                if (getMangaForCategory.isNotEmpty()) {
+                    val item = items.random()
+                    openManga(item)
```
AntsyLich, line 68:
```suggestion
                    openManga(items.random())
```
And remove the val

zbue:
> User search query (e.g. searching genre) + opening random manga won't be possible.

Perhaps after the app has the option to group by tags, etc., like its forks, this case will be resolved.

i18n/src/main/res/values/strings.xml
```diff
@@ -846,6 +847,7 @@
     <string name="information_no_recent">No recent updates</string>
     <string name="information_no_recent_manga">Nothing read recently</string>
     <string name="information_empty_library">Your library is empty</string>
+    <string name="information_nothing_in_here">Nothing in here</string>
```
AntsyLich, line 850:
```suggestion
    <string name="information_failed_to_open_random_manga">Couldn\'t open a random title</string>
```

i18n/src/main/res/values/strings.xml
```diff
@@ -72,6 +72,7 @@
     <string name="action_disable_all">Disable all</string>
     <string name="action_edit">Edit</string>
     <string name="action_add">Add</string>
+    <string name="action_open_random_manga">Open random manga</string>
```
AntsyLich, line 75: Though internally we specify everything as "manga" on the UI side we mention them as "titles".
```suggestion
    <string name="action_open_random_manga">Open random title</string>
```

app/src/main/java/eu/kanade/presentation/library/components/LibraryToolbar.kt
zbue: I am getting `Initializer required for destructuring declaration` here after applying what you suggested

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryController.kt
zbue: `@Composable invocations can only happen from the context of a @Composable function` if this is inside lambda block

i18n/src/main/res/values/strings.xml
zbue: `Nothing in here` indicates that the current category is empty, which is better I you ask me, rather than giving the user a more general message

app/src/main/java/eu/kanade/presentation/library/components/LibraryToolbar.kt
```diff
@@ -127,8 +135,34 @@ fun LibraryRegularToolbar(
...
+                IconButton(onClick = { onMoreExpanded(!moreExpanded) }) {
+                    Icon(
+                        imageVector = Icons.Default.MoreVert,
+                        contentDescription = stringResource(R.string.abc_action_menu_overflow_description),
+                    )
+                }
+                val onDismissRequest = { onMoreExpanded(false) }
+                DropdownMenu(
+                    expanded = moreExpanded,
```
AntsyLich, line 148:
```suggestion
                val expanded by remember { mutableStateOf(false) }
                IconButton(onClick = { expanded = !expanded }) {
                    Icon(
                        imageVector = Icons.Default.MoreVert,
                        contentDescription = stringResource(R.string.abc_action_menu_overflow_description),
                    )
                }
                val onDismissRequest = { expanded = false }
                DropdownMenu(
                    expanded = expanded,
```

app/src/main/java/eu/kanade/presentation/library/components/LibraryToolbar.kt
arkon: You'd get rid of the destructuring so that it's

```
var moreExpanded by remember { ... }
```

i18n/src/main/res/values/strings.xml
```diff
@@ -72,6 +72,7 @@
     <string name="action_disable_all">Disable all</string>
     <string name="action_edit">Edit</string>
     <string name="action_add">Add</string>
+    <string name="action_open_random_title">Open random title</string>
```
arkon, line 75: I actually recently reworded things like "manga" to "entry" so let's go with that instead of "title".

i18n/src/main/res/values/strings.xml
arkon: Can we make it a bit more specific so it's "No entries found in this category"?

arkon (changes requested):

PR #8262, opened 2022-10-21
Tweak tablet UI mode setting

stevenyomi:
- Tweak auto tablet UI mode (previously enabled when smallest width ≥ 720dp)
  - Portrait: smallest width ≥ 700dp, for [tablets with width of 711dp = 1600px ÷ 2.25](https://www.gsmarena.com/xiaomi_pad_5-11042.php)
  - Landscape: smallest width ≥ 600dp, for smaller tablets, also ensures nav rail icons fit
    - (what if there's a square device with 600dp width...?)
- Move tablet UI setting to "Appearance" screen, before the nav rail option
- Show nav rail option when tablet UI is available instead of when it's enabled
  - so the option will always be shown if tablet UI is only enabled in landscape orientation either manually or automatically
- Remove dependency on M3 Window Size Class (useless)

I think we should move the tablet UI helper functions from `ContextExtensions.kt` to somewhere else, but idk where's appropriate...


app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
```diff
@@ -263,28 +261,46 @@ fun Context.createFileInCacheDir(name: String): File {
...
- * We consider anything with a width of >= 720dp as a tablet, i.e. with layouts in layout-sw720dp.
- */
-fun Context.isTablet(): Boolean {
-    return resources.configuration.smallestScreenWidthDp >= TABLET_UI_MIN_SCREEN_WIDTH_DP
+private const val TABLET_UI_REQUIRED_SCREEN_WIDTH_DP = 720
+
+// some tablets have screen width like 711dp = 1600px / 2.25
+private const val TABLET_UI_MIN_SCREEN_WIDTH_PORTRAIT_DP = 700
+
+// make sure icons on the nav rail fit
```
arkon, line 269: There ought to be a height requirement too, but I guess Material doesn't really check that either.

Issue #3465, opened 2020-07-13
[Bug] Category name letter casing change not allowed when renaming


bbtora:
### Device information
* Tachiyomi version: Preview r2511
* Android version: 10

## Steps to reproduce
1. Have a category, for example: Reading/reading/READING
2. Change the category name uppercase to lowercase or visa-versa

### Expected behavior
Category name should be changed, for example: reading/READING/reading

### Actual behavior
Renaming not allowed. Gives message: `A category with this name already exist!`

PR #8274, opened 2022-10-22
Category name letter casing change not allowed when renaming

d-najd:
For more info #3465.

### NOTE

The actual behaviour altered is in `RenameCategory` but because of the renaming it is not included in the commit.


app/src/main/java/eu/kanade/domain/category/model/Category.kt
```diff
@@ -18,5 +18,9 @@ data class Category(
 }
 
 internal fun List<Category>.anyWithName(name: String): Boolean {
```
arkon, line 20: Could just add an argument (`ignoreCase = false`)

d-najd: Sure the change are up!

stevenyomi:
> including gradle.properites.swp

Why?

d-najd:
> > including gradle.properites.swp
> 
> Why?

if you don't include it the PR build check would fail, and I read on stackoverflow that you should include it so

stevenyomi:
> if you don't include it the PR build check would fail, and I read on stackoverflow that you should include it so

Never heard of that. None of your build checks failed because of that. If it's true, that file would have been added to the repo in the beginning. Just remove it.

app/src/main/java/eu/kanade/domain/category/model/Category.kt
```diff
@@ -18,5 +18,5 @@ data class Category(
...
-    return any { name.equals(it.name, ignoreCase = true) }
+    return any { name.equals(it.name, ignoreCase = false) }
```
AntsyLich, line 21:
```suggestion
    return any { name == it.name }
```

curche:
> > > including gradle.properites.swp
> > 
> > 
> > Why?
> 
> if you don't include it the PR build check would fail, and I read on stackoverflow that you should include it so

that should be gradle.properties
gradle.properties.swp sounds like a swap/temporary file (of gradle.properties) which usually something like Vim creates when you have a file open
its kinda rare for it to show up and probably why it isn't there in `.gitignore` or just have Vim [create swp files in separate directory](https://vi.stackexchange.com/a/16047)

d-najd:
> > > > including gradle.properites.swp
> > > 
> > > 
> > > Why?
> > 
> > 
> > if you don't include it the PR build check would fail, and I read on stackoverflow that you should include it so
> 
> that should be gradle.properties gradle.properties.swp sounds like a swap/temporary file (of gradle.properties) which usually something like Vim creates when you have a file open its kinda rare for it to show up and probably why it isn't there in `.gitignore` or just have Vim [create swp files in separate directory](https://vi.stackexchange.com/a/16047)

I guess you are correct, looking again at the stackoverflow answer it is refering to gradle.properties, my bad, but I have one question did https://github.com/tachiyomiorg/tachiyomi/actions/runs/3304018563/jobs/5452632515, get canceled because I got a review requesting changes?

stevenyomi:
It was cancelled because you pushed a new commit. Only one PR build check can be run at the same time.

PR #8279, opened 2022-10-22
Fixed "0 mins ago" in "Library last updated" string

Dexroneum:
With help of Syer10


app/src/main/java/eu/kanade/presentation/updates/UpdatesUiItem.kt
```diff
@@ -54,7 +54,12 @@ fun LazyListScope.updatesLastUpdatedItem(
 ) {
     item(key = "updates-lastUpdated") {
         val time = remember(lastUpdated) {
-            DateUtils.getRelativeTimeSpanString(lastUpdated, Date().time, DateUtils.MINUTE_IN_MILLIS)
+            val now = Date().time
+            if (now - lastUpdated < 60000) {
```
arkon, line 58: Could do `1.minutes.inWholeMilliseconds`

app/src/main/java/eu/kanade/presentation/updates/UpdatesUiItem.kt
```diff
@@ -54,7 +54,12 @@ fun LazyListScope.updatesLastUpdatedItem(
...
+                ""
```
arkon, line 59: I feel like `null` is more appropriate here.

i18n/src/main/res/values/strings.xml
```diff
@@ -750,7 +750,8 @@
     <string name="updating_library">Updating library</string>
     <string name="update_already_running">An update is already running</string>
     <string name="cant_open_last_read_chapter">Unable to open last read chapter</string>
-    <string name="updates_last_update_info">Library last updated: %1$s</string>
+    <string name="updates_last_update_info">Library last updated: %s</string>
+    <string name="updates_last_update_info_extra">Just now</string>
```
arkon, line 754: Make the name a bit more descriptive (e.g., `updates_last_update_info_just_now`)

app/src/main/java/eu/kanade/presentation/updates/UpdatesUiItem.kt
arkon: Should be, yeah.

PR #8281, opened 2022-10-24
WebView for chapter link

### LINKED ISSUE #4770: [Feature Request] WebView for chapter link
i will lot easier to check, because most of the issues would need to check chapter in webview
like page issue

## What/Requirements
hold tap chapter, if chapter selected only 1, show webview icon to open

d-najd:
closes #4770 

- Will add button inside reader topbar, the button will open the chapter in webview

- Adds button to the bottom navbar when manga is opened and only a single chapter is selected.
## Problem (until ext-lib 1.4 starts being used)
It seems that some extensions use api to access the chapter data so `getChapterUrl` in `HttpSource` in most extension leads to the correct url but sometimes leads to the api which only shows json data, example (Manga Plus Chainsaw Man chapter 1):

Correct url
[https://mangaplus.shueisha.co.jp/viewer/1001249](https://mangaplus.shueisha.co.jp/viewer/1001249?fbclid=IwAR3GtqpVLN6rZkzUu-b7QV34P4n4H-DxavvFLNjiTRSZ0YZgwuiCv82PFP4)

Actual url from `getChapterUrl`
[https://jumpg-webapi.tokyo-cdn.com/api/manga_viewer?chapter_id=1001249&split=yes&img_quality=super_high&format=json](https://jumpg-webapi.tokyo-cdn.com/api/manga_viewer?chapter_id=1001249&split=yes&img_quality=super_high&format=json&fbclid=IwAR02t-pmEmTwbszXvEmGs07sFqRFNQ9y3nunT4xUVTCyE7Vjn3BfKRgH_gM)

I will mark request as open in case I forget to do so (if I leave it as draft) after ext-lib 1.4 gets implemented.

## Picture

~~My opinion on what should be done
#4770 as well as this branch should be closed since to implement this small feature changes to extensions like MangaPlus and MangaDex will have to be done to support the feature adding more breaking points which is not worth it for in my opinion.
Or an alternative could be to add new request in `HttpSource` for example `getChapterWebviewUrl` and show a alertDialog saying not supported for extensions which have not implemented the behaviour.~~

~~I didn't want to spend hours altering extensions and for it to turn out that the feature is not worth it so that's why I made the pull request now.~~


AntsyLich:
`getChapterUrl` was added to be used in ext-lib 1.4 (currently all ext is using ext-lib 1.3) so this pr won't work for many ext until after a week or so of stable release when ext-lib of extensions will be bumped to 1.4

stevenyomi:
These functions already have a default implementation and will work as expected. J2K (iirc) already has this feature via the default `pageListRequest()`.

Not sure if extlib 1.4 will be merged any time soon as there are few new features. Also we haven't decided on how to let 1.2 users update.

AntsyLich:
I didn't check the code first but move the option inside reader topbar

d-najd:
> These functions already have a default implementation and will work as expected. J2K (iirc) already has this feature via the default `pageListRequest()`.
> 
> Not sure if extlib 1.4 will be merged any time soon as there are few new features. Also we haven't decided on how to let 1.2 users update.

I checked j2k implementation and it seems kinda hacky
```
    /** Helper method to handle guya-like sources */
    private fun HttpSource.fullChapterUrl(mangaUrl: String, chapterUrl: String, chapter: Chapter): String {
        val lowerUrl = baseUrl.lowercase()
        return when {
            chapter.url.startsWith("http") -> {
                chapter.url
            }
            lowerUrl.contains("guya") || lowerUrl.contains("danke") ||
                lowerUrl.contains("hachirumi") || lowerUrl.contains("mahoushoujobu") ||
                (lowerUrl.contains("cubari") && !mangaUrl.contains("imgur")) -> {
                // cubari links would have double / without the trim end
                mangaUrl.trimEnd('/') + "/" + chapter.chapter_number.fmt().replace(".", "-")
            }
            else -> baseUrl + chapterUrl
        }
    }
```
If something like this is acceptable I will borrow the code from here. TachiyomiSY seems to only open the manga in the browser not the chapter.

CarlosEsco:
> > These functions already have a default implementation and will work as expected. J2K (iirc) already has this feature via the default `pageListRequest()`.
> > Not sure if extlib 1.4 will be merged any time soon as there are few new features. Also we haven't decided on how to let 1.2 users update.
> 
> I checked j2k implementation and it seems kinda hacky
> 
> ```
>     /** Helper method to handle guya-like sources */
>     private fun HttpSource.fullChapterUrl(mangaUrl: String, chapterUrl: String, chapter: Chapter): String {
>         val lowerUrl = baseUrl.lowercase()
>         return when {
>             chapter.url.startsWith("http") -> {
>                 chapter.url
>             }
>             lowerUrl.contains("guya") || lowerUrl.contains("danke") ||
>                 lowerUrl.contains("hachirumi") || lowerUrl.contains("mahoushoujobu") ||
>                 (lowerUrl.contains("cubari") && !mangaUrl.contains("imgur")) -> {
>                 // cubari links would have double / without the trim end
>                 mangaUrl.trimEnd('/') + "/" + chapter.chapter_number.fmt().replace(".", "-")
>             }
>             else -> baseUrl + chapterUrl
>         }
>     }
> ```
> 
> If something like this is acceptable I will borrow the code from here. TachiyomiSY seems to only open the manga in the browser not the chapter.

it's not which is why its not in main

d-najd:
> > I checked j2k implementation and it seems kinda hacky
> 
> That's not what I meant. I mean you can just use the simple `getChapterUrl()` and ignore everything else.
> 
> > It seems that some extensions use api to access the chapter data so `getChapterUrl` in `HttpSource` in most extension leads to the correct url but sometimes leads to the api which only shows json data, example (Manga Plus Chainsaw Man chapter 1):
> 
> Yeah just ignore that. Some sources have their manga URLs pointing to API too. Just implement the feature and wait for extensions to adapt to it later.

The code is already using getChapterUrl() so everything should be good to go.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderActivity.kt
```diff
@@ -275,6 +276,10 @@ class ReaderActivity : BaseRxActivity<ReaderPresenter>() {
      */
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         when (item.itemId) {
+            R.id.action_open_in_web_view -> {
+                openChapterInWebview()
+                invalidateOptionsMenu()
```
AntsyLich, line 281: This is not necessary. [invalidateOptionsMenu] is used when we need to recreate the topbar.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderPresenter.kt
```diff
@@ -606,6 +607,15 @@ class ReaderPresenter(
         return viewerChaptersRelay.value?.currChapter
     }
 
+    fun getSource() = manga?.source?.let { sourceManager.getOrStub(it) } as? HttpSource
+
+    fun getChapterUrl(): String? {
+        val sChapter = getCurrentChapter()?.chapter?.toDomainChapter()?.toSChapter() ?: return null
```
AntsyLich, line 613:
```suggestion
        val sChapter = getCurrentChapter()?.chapter ?: return null
```
DbChapter inherits SChapter

AntsyLich (commented):
Put WebView at right so it gets hidden when there's not enough space

PR #8292, opened 2022-10-26
Use `Channel` in Download cache

AntsyLich:
Co-Authored-By: Andreas <6576096+ghostbear@users.noreply.github.com>


app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
```diff
@@ -41,10 +43,8 @@ class DownloadCache(
     private val downloadPreferences: DownloadPreferences = Injekt.get(),
 ) {
 
-    // This is just a mechanism of notifying consumers of updates to the cache, the value itself
-    // is meaningless.
-    private val _state: MutableStateFlow<Long> = MutableStateFlow(0L)
-    val changes = _state.asStateFlow()
+    private val _changes: Channel<Unit> = Channel(Int.MAX_VALUE)
```
stevenyomi, line 46:
```suggestion
    private val _changes: Channel<Unit> = Channel(Channel.Factory.UNLIMITED)
```

ghostbear: Should be able to do `Channel(Channel.UNLIMITED)` instead of `Channel(Channel.Factory.UNLIMITED)`

PR #8303, opened 2022-10-28
Prompt Extension update if ext-lib is updated

AntsyLich:


app/src/main/java/eu/kanade/tachiyomi/extension/api/ExtensionGithubApi.kt
```diff
@@ -109,6 +109,7 @@ internal class ExtensionGithubApi {
                     pkgName = it.pkg,
                     versionName = it.version,
                     versionCode = it.code,
+                    libVersion = it.version.substringBeforeLast('.').toDouble(),
```
arkon, line 112: Should dedupe this logic from `ExtensionLoader`

arkon:
Superseded by e1ab1fdb6

PR #8326, opened 2022-10-29
Clean up strings and icons

stevenyomi:
- Reduce usage of `android.R.string`
- Fixed misused strings
- Used Outlined variant icons where possible (dedupe) and specify Filled explicitly instead of Default
  - Changed add/remove bookmark icon in chapter toolbar to outlined versions to match the others
- Added and fixed icon content description


i18n/src/main/res/values/strings.xml
```diff
@@ -110,7 +103,6 @@
     <string name="action_disable">Disable</string>
     <string name="action_pin">Pin</string>
     <string name="action_unpin">Unpin</string>
-    <string name="action_cancel">Cancel</string>
```
arkon, line 113: We'd probably want to to keep this if we ever want multiplatform, considering `android.R` isn't going to be a thing on other platforms.

app/src/main/java/eu/kanade/presentation/category/components/CategoryFloatingActionButton.kt
```diff
@@ -21,7 +21,7 @@ fun CategoryFloatingActionButton(
 ) {
     ExtendedFloatingActionButton(
         text = { Text(text = stringResource(R.string.action_add)) },
-        icon = { Icon(imageVector = Icons.Outlined.Add, contentDescription = "") },
+        icon = { Icon(imageVector = Icons.Outlined.Add, contentDescription = stringResource(R.string.action_add)) },
```
arkon, line 24: Should just `null` this, the text provides better a11y context than the icon.

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -264,16 +264,16 @@ private fun MangaScreenSmallImpl(
...
                     text = {
-                        val id = if (chapters.any { it.chapter.read }) {
-                            R.string.action_resume
-                        } else {
-                            R.string.action_start
-                        }
                         Text(text = stringResource(id))
                     },
-                    icon = { Icon(imageVector = Icons.Default.PlayArrow, contentDescription = null) },
+                    icon = { Icon(imageVector = Icons.Filled.PlayArrow, contentDescription = stringResource(id)) },
```
arkon, line 276: Likewise, rely on the text for a11y context rather than the icon.

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -477,16 +477,16 @@ fun MangaScreenLargeImpl(
...
                         text = {
-                            val id = if (chapters.any { it.chapter.read }) {
-                                R.string.action_resume
-                            } else {
-                                R.string.action_start
-                            }
                             Text(text = stringResource(id))
                         },
-                        icon = { Icon(imageVector = Icons.Default.PlayArrow, contentDescription = null) },
+                        icon = { Icon(imageVector = Icons.Filled.PlayArrow, contentDescription = stringResource(id)) },
```
arkon, line 489: Same here.

app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadController.kt
```diff
@@ -223,22 +223,22 @@ class DownloadController :
...
                         icon = {
                             val icon = if (isRunning) {
-                                Icons.Default.Pause
+                                Icons.Outlined.Pause
                             } else {
-                                Icons.Default.PlayArrow
+                                Icons.Filled.PlayArrow
                             }
-                            Icon(imageVector = icon, contentDescription = null)
+                            Icon(imageVector = icon, contentDescription = stringResource(id))
```
arkon, line 241: Same here

arkon (changes requested):

i18n/src/main/res/values/strings.xml
stevenyomi: Should I remove all other uses of `android.R.string`? This will include the `OK` button in dialogs.

app/src/main/java/eu/kanade/presentation/category/components/CategoryFloatingActionButton.kt
stevenyomi: I only added these to collapsible FABs as they might collapse and remove the text completely.

i18n/src/main/res/values/strings.xml
arkon: Doesn't have to be in this PR, but we should eventually yes.

app/src/main/java/eu/kanade/presentation/category/components/CategoryFloatingActionButton.kt
arkon: Visual collapsing doesn't really matter for visual impaired users anyway if they're relying on Talkback to read out the node contents.

stevenyomi: I mean when the FAB is collapsed, the text is gone, so Talkback can't read anything if the icon has no content description.

arkon: I see, I didn't actually realize that it wasn't smart enough to read the invisible text (which I just confirmed is the case).

PR #8350, opened 2022-10-30
Disable `ChapterHeader` & `ChapterDownloadIndicator` click when in selection mode

zbue:


app/src/main/java/eu/kanade/presentation/components/ChapterDownloadIndicator.kt
```diff
@@ -43,33 +43,49 @@ enum class ChapterDownloadAction {
...
 private fun NotDownloadedIndicator(
+    enabled: Boolean,
     modifier: Modifier = Modifier,
     onClick: (ChapterDownloadAction) -> Unit,
 ) {
     Box(
         modifier = modifier
             .size(IconButtonTokens.StateLayerSize)
             .commonClickable(
+                enabled = enabled,
```
arkon, line 88: I feel like it's visually confusing when they look the same but can no longer be clicked.

It _might_ make sense to just hide this indicator but keep the other status ones? Open to other opinions on that.

zbue: I actually consider hiding the indicator, the negative is that, they won't be able to see which is downloaded or which is not in case where they plan to delete some of it.

arkon: Yeah I guess it's fine since the action menu has the download action anyhow.

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -351,6 +351,7 @@ private fun MangaScreenSmallImpl(
                         contentType = MangaScreenItem.CHAPTER_HEADER,
                     ) {
                         ChapterHeader(
+                            enabled = chapters.none { it.selected },
```
arkon, line 354: Should probably use something like `fastAny` rather than `none`, which an iterator internally.

arkon:
There's a merge conflict btw.

app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
```diff
@@ -128,12 +132,18 @@ fun MangaChapterListItem(
 
         // Download view
         if (onDownloadClick != null) {
-            ChapterDownloadIndicator(
-                modifier = Modifier.padding(start = 4.dp),
-                downloadStateProvider = downloadStateProvider,
-                downloadProgressProvider = downloadProgressProvider,
-                onClick = onDownloadClick,
-            )
+            AnimatedVisibility(
```
arkon, line 135: Wait, I was fine with the original implementation. I agree that it's useful to see the statuses while in selection mode.

zbue: Oopsies. I misunderstood.

PR #8365, opened 2022-10-30
Tabs: Don't explicitly set text color in the text

ivaniskandar:
The container already provides color option for both states


app/src/main/java/eu/kanade/presentation/library/components/LibraryTabs.kt
```diff
@@ -1,56 +1,51 @@
...
-                    remember { mutableStateOf<Int?>(null) }
-                }
                 Tab(
-                    selected = state.currentPage == index,
-                    onClick = { scope.launch { state.animateScrollToPage(index) } },
+                    selected = currentPageIndex == index,
+                    onClick = { onTabItemClick(index) },
                     text = {
-                        TabText(category.visualName, count, state.currentPage == index)
+                        val count by getNumberOfMangaForCategory(category.id)
```
arkon, line 42: Wouldn't it be better to simply not call this if `showMangaCount` is false?

PR #8369, opened 2022-10-30
AppBar filter action: Use `onSurfaceVariant` to match with the rest of action bar

zbue:
Old  | New
------------- | -------------
image | image


arkon:
`IconButton` and `Icon` use `LocalContentColor.current` by default too. Why switch to `onSurfaceVariant` explicitly?

zbue:
> `IconButton` and `Icon` use `LocalContentColor.current` by default too. Why switch to `onSurfaceVariant` explicitly?

Top app bar uses its own [color token](https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-main/compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/tokens/TopAppBarSmallTokens.kt#34) and not the `Icon` nor `IconButton`.

arkon:
Right, and that's used as the `LocalContentColor` for the `actions`, so why not inherit it properly with `LocalContentColor.current`?

zbue:
> Right, and that's used as the `LocalContentColor` for the `actions`, so why not inherit it properly with `LocalContentColor.current`?

It defaults to black. You can clearly see the difference in the picture above.

arkon:
Not gonna lie, I couldn't actually see that difference.

Thanks for pointing it out though, proper fix is in 781888540

PR #8370, opened 2022-10-30
Added Interceptor to fetch a new url on Forbidden

brettmcginnis:


arkon:
But why? The source should just set a valid URL?

brettmcginnis:
@arkon  Sorry this was meant to be a draft in my project.

Anyways the reason is time. It was valid at the time the url was made but invalid when re-used. Aka export a backup from one device, and then restore on another and you'll have mostly blank series MangaCover images

PR #8402, opened 2022-11-01
Add per-category update frequency

### LINKED ISSUE #822: Per-category update frequency
I follow a lot of manga using the app, and library updates can take a very long time. However, a lot of the manga are weekly or monthly, and therefore don't need to be updated that frequently. 

An option to change the update frequency based on category (either accessible from the three dot menu in the upper right corner in the "My library" screen or under "General" in the settings where "Library Update Frequency" is) would reduce the number of checks needed, while still allowing users to keep up to date with more frequently-updated manga. 

Version: v0.5.2
(Android 7.0 on HTC 10)

d-najd:
closes #822

### Short version of how the feature works
The per-category update works by assigning the global update interval to each category as default update interval, categories can then change this interval and they won't follow the global update interval unless the user changes the setting back

### Long version of how the feature works
When category is created it is assigned the global update interval by default and will follow whatever the setting is set to.
If the user then goes to the New Dialog and changes "Default" value (global update interval) to something else, then that category will follow the update interval set in there.

for ex: if we have `Global update interval = 24h`, `Category A update interval = 12h`, `Category B update interval = Default (global update interval)`, `Category C update interval = 48h`, then
- Category A will update every 12h
- Category B will update every 24h (or whatever the global update interval is set to)
- Category C will update every 48h

### How the code works

First it checks for the lowest update interval from each category and the global update, with the example above it will be 12h and sets it as the update interval for the `LibraryUpdateJob`

when the update gets called in the `LibraryUpdateService`, first it checks if the category was ever updated, 
- if it wasn't it adds it to the update list 

If it was updated before it checks if the update interval is set to global update interval
- if it is, it runs a check to see if enough time has passed since the last update, in the example above that would be 24h, and if so adds it to the update list
- it it wasn't, it checks what the custom interval was (in the example above that would be 12h or 48h) and if enough time has passed, if it did the category will be added to update list

items that were not added to the update list will be filtered later and they won't be updated (ex excluding categories) but the last update interval will still be set to the time that the checks were ran.

### Note
I did not include off option in the update interval because there is already option to exclude categories from update and I didn't want users to have 1 more thing to check and annoy discord staff.

### Some photos

| Categories screen | New Dialog |
| --- | --- |
|  |  |


Soitora:
Nice method to further increase your own librarys update time and also easier for people to find a setting for changing it

d-najd:
I will resolve the merge conflicts once the code gets reviewed.

arkon:
Closing this as #5734 is more convenient.

PR #8418, opened 2022-11-02
Add back haptic feedback long tap to fav

Riztard:
- add back haptic when long tap on manga to add to library


app/src/main/java/eu/kanade/presentation/browse/components/BrowseSourceComfortableGrid.kt
```diff
@@ -80,7 +84,10 @@ fun BrowseSourceComfortableGridItem(
                 Badge(text = stringResource(R.string.in_library))
             }
         },
-        onLongClick = onLongClick,
+        onLongClick = {
+            onLongClick()
+            haptic.performHapticFeedback(HapticFeedbackType.LongPress)
```
arkon, line 89: You can lift this up a few levels, `onLongClick` is the same thing for all 3 of these composables.

Riztard: like this?

PR #8425, opened 2022-11-02
Ignore gradle.properties.swp

d-najd:
The file is from intellij vim plugin if I remember correctly, 
I accidentally included it once (and nearly did again 5min ago) with a commit and I remember seeing another commit from arkon removing it after (sorry about that).


arkon:
Could just ignore all swap files.

stevenyomi:
You can add it to `.git/info/exclude` to make Git ignore it locally.

d-najd:
> You can add it to `.git/info/exclude` to make Git ignore it locally.

that will work but if some other contributor is using vim plugin it will still generate the files and he probably wont notice, plus there are no drawbacks so why not.

d-najd:
> Could just ignore all swap files.

sure give me a min

stevenyomi:
> that will work but if some other contributor is using vim plugin it still might generate the file and he probably wont notice, plus there are no drawbacks so why not.

Yeah but if some other contributor is using _(insert plugin name here)_ plugin it still might generate the file and he probably wont notice, so why not add every kind of temp file into our gitignore? 😏

stevenyomi:
Imo gitignore file should only cover commonly used temp files and individual tool generated ones should be ignored locally, otherwise the gitignore file will be filled with all kinds of files.

d-najd:
> > that will work but if some other contributor is using vim plugin it still might generate the file and he probably wont notice, plus there are no drawbacks so why not.
> 
> Yeah but if some other contributor is using _(insert plugin name here)_ plugin it still might generate the file and he probably wont notice, so why not add every kind of temp file into our gitignore? smirk

might as well open a request to ignore _every_ possible plugin while we are at that and mark it as top priority

jokes aside I still think that we can add files to gitignore as we notice them, the decision is up to you all to make 🙂

PR #8427, opened 2022-11-02
Revert "Tweak how `getChapterUrl` works (#8392)"

stevenyomi:
As some sources store keys instead of URL/paths.


AntsyLich:
We should tweak the ext to store urls and paths instead of keys

stevenyomi:
> We should tweak the ext to store urls and paths instead of keys

Not gonna happen. There's no way to list them, and changing the behavior will lead to complicated code, need for migration or duplicate entries.

PR #8446, opened 2022-11-04
Show missing chapter warning when resuming

maxim380:
Added a warning if chapters are skipped when resuming.

Based on issue tachiyomiorg#7843
Screenshot_20221104-142932_Tachiyomi


Riztard:
maybe also tell which chapter? just like in the chapter transition

TheKingTermux:
Well it will be very helpful if the reader knows how many chapters are missing

foxscore:
I think it might be better to move the check of unread/missing chapters into the openChapter function, so that not only the `Resume` button triggers it, but also selecting the chapter in general *(to prevent missclicks 'n such)*.

Furthermore, you should probably check if the source is actually missing chapters, or if you just haven't read them yet *(these should result in different messages)*. Both cases are widely different, but currently being treated the same.

But otherwise, good idea.

PR #8489, opened 2022-11-09
add Suwayomi tracker

AriaMoradi:


app/src/main/java/eu/kanade/tachiyomi/data/track/tachidesk/Tachidesk.kt
```diff
@@ -0,0 +1,110 @@
...
+import okhttp3.Dns
+import okhttp3.OkHttpClient
+import eu.kanade.domain.manga.model.Manga as DomainManga
+import eu.kanade.domain.track.model.Track as DomainTrack
+
+class Tachidesk(private val context: Context, id: Long) : TrackService(id), NoLoginTrackService, EnhancedTrackService {
+    override val client: OkHttpClient =
+        networkService.client.newBuilder()
+            .dns(Dns.SYSTEM) // don't use DNS over HTTPS as it breaks IP addressing
+            .build()
```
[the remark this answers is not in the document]
jobobby04: Nvm I see its going to be handled in the API class

arkon:
Is there still an initiative to rename Tachidesk? This seems like it'd make users think that it's an official desktop version of Tachiyomi, which it isn't.

AriaMoradi:
> Is there still an initiative to rename Tachidesk? This seems like it'd make users think that it's an official desktop version of Tachiyomi, which it isn't.

We had a vote last year to rename it, Suwayomi came out as the second winner and we decided to not rename the app but name the org Suwayomi. Also I have never advertised Tachidesk as an official Tachiyomi app and have never seen anyone treat it that way in out discord server not sure how anyone could come to that conclusion.

arkon:
> I'm not sure how anyone could come to that conclusion

I've seen people on both Reddit and Discord that were under that impression, and putting it in the app itself will only further perpetuate that false association. I'm a bit confused as to how you can possibly think no one would come to that conclusion given the naming.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsTrackingScreen.kt
```diff
@@ -198,7 +198,7 @@ class SettingsTrackingScreen : SearchableSettings {
                             if (hasValidSourceInstalled) {
                                 trackManager.suwayomi.loginNoop()
                             } else {
-                                context.toast(context.getString(R.string.enhanced_tracking_warning, context.getString(trackManager.suwayomi.nameRes())), Toast.LENGTH_LONG)
+                                context.toast(context.getString(R.string.enhanced_tracking_warning, R.string.tracker_tachidesk), Toast.LENGTH_LONG)
```
Soitora, line 201: Why did you add back Tachidesk naming?

AriaMoradi: Referring to Extension?

AriaMoradi: It is referring to the extension and it's named Tachidesk.

PR #8491, opened 2022-11-10
Add option to invalidate download cache

AntsyLich:


app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsAdvancedScreen.kt
```diff
@@ -157,6 +158,7 @@ class SettingsAdvancedScreen : SearchableSettings {
         val context = LocalContext.current
         val navigator = LocalNavigator.currentOrThrow
         val libraryPreferences = remember { Injekt.get<LibraryPreferences>() }
+        val downloadCache = remember { Injekt.get<DownloadCache>() }
```
stevenyomi, line 161: No need to remember this, just get it inside the onclick lambda.

PR #8512, opened 2022-11-12
Changed Web element from using manga.url back to using tracker remote url (ComicInfo.xml files)

Shamicen:
I changed the web element back to using  the tracker remote url since manga.url does not contain a valid hyperlink and instead contains something like this:
```
/manga/05e36dbe-ddf2-45e4-8d12-1f45d8397817
```

this also fits the Schema documentation of this element:
```
A URL pointing to a reference website for the book.
```
We could also use the source url instead but like I previously said that is not actually contained in manga.url


arkon:
Superseded by 262f8449b4369fde42cbd15d2a768e54461e84af

PR #8528, opened 2022-11-13
Fix related to cancelling queued chapters

AntsyLich:
Tachi removes the downloaded chapter (if it exists) when you just cancelled a download from queue. Which shouldn't happen imo so changed that to only remove from queue.

Also removes redundant return


arkon:
Why is something in the queue if it's already downloaded?

AntsyLich:
Not really sure but after download ahead was added I had some instances where downloaded chapters would disappear while reading and exiting from reader will add the chapter to the top of queue indicating that the chapter was in queue while it was still downloaded

PR #8559, opened 2022-11-18
Rework on the wheel picker

### LINKED ISSUE #8800: Picker takes too long to select and highlight a different value after scrolling to a new value.
### Steps to reproduce

1. Go to anywhere in the app with a picker (ex. Item per row count, rating/chapter count/finish date/start in the tracker sheet)
2. Swipe to change the value of the picker but do not hit the max value or min value.

### Expected behavior

The picker should select and highlight the new value as soon as the picker stops moving. 

### Actual behavior

The value is only selected and highlighted a few moments after the picker stops moving and requires the user to wait for the value to be highlighted before closing the prompt.

### Crash logs

_No response_

### Tachiyomi version

Preview r5369

### Android version

Android 9

### Device

Amazon Fire HD 10+

### Other details

_No response_

### Acknowledgements

ivaniskandar:
doesn't need to wait for the animation stop to change the value (closes #8800)
now have min-max selection date
etc


.github/renovate.json
```diff
@@ -7,6 +7,5 @@
     "androidx.core:core-splashscreen",
     "com.android.tools:r8",
     "com.google.guava:guava",
```
arkon, line 9: Not entirely sure if Renovate accepts trailing commas.

app/src/main/java/eu/kanade/presentation/components/WheelPicker.kt
```diff
@@ -0,0 +1,310 @@
...
+    }
+}
+
+private fun calculateSnappedItemIndex(lazyListState: LazyListState): Int {
+    return lazyListState.layoutInfo.visibleItemsInfo
+        .maxBy { calculateAnimatedAlpha(lazyListState, it.index) }
+        .index
+}
+
+private fun calculateDayOfMonths(localDate: LocalDate): List<Int> {
```
arkon, line 265: https://developer.android.com/reference/java/time/YearMonth#lengthOfMonth() ?

app/src/main/java/eu/kanade/presentation/components/WheelPicker.kt
```diff
@@ -0,0 +1,310 @@
...
+                onSelectionChanged = { index ->
+                    val newMonth = months[index].first
+                    internalOnSelectionChange(internalSelection.withMonth(newMonth))
+                },
+            )
+
+            // Year
+            val years = remember(minDate, maxDate) {
+                val minYear = minDate?.year?.coerceAtLeast(1900) ?: 1900
+                val maxYear = maxDate?.year?.coerceAtMost(2100) ?: 2100
```
arkon, line 213: We could probably limit it to current year or current year + 1.

app/src/main/java/eu/kanade/presentation/components/WheelPicker.kt
Two-Ai: or [`LocalDate.lengthOfMonth()`](https://developer.android.com/reference/java/time/LocalDate#lengthOfMonth()) directly

app/src/main/java/eu/kanade/presentation/components/WheelPicker.kt
ivaniskandar: You sure about this? IMO the date picker as a component shouldn't have major restrictions, but if you want to set the restrictions specifically to start/finish reading date we can do that

arkon: Well those are already set to current date, so that's fine.

I'm not entirely sure why this is limited to 2100, although I guess it's fine for now.

ivaniskandar: I'm just following the others https://github.com/androidx/androidx/blob/androidx-main/compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/DatePicker.kt#L478

PR #8643, opened 2022-11-29
Fix Stub Source migration screen broken

AntsyLich:
Also remove name shadowing


app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/manga/MigrationMangaScreenModel.kt
```diff
@@ -32,21 +32,19 @@ class MigrationMangaScreenModel(
     init {
         coroutineScope.launch {
             mutableState.update { state ->
-                val source = sourceManager.get(sourceId)
-                if (source == null) {
-                    _events.send(MigrationMangaEvent.FailedGettingSource)
```
ghostbear, line 37: Doesn't this remove the screen pop when uninstalling

AntsyLich: This is a source not a extension

PR #8651, opened 2022-12-01
Use Voyager on Source Preference screen

ivaniskandar:
~~Maps View-based Preferences to existing Compose PreferencesItem
Unknown Preference class will be skipped silently but logged for easy debug.~~

Wraps a PreferenceFragmentCompat.


app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/details/SourcePreferencesScreen.kt
```diff
@@ -0,0 +1,150 @@
...
+import eu.kanade.tachiyomi.source.getPreferenceKey
+import eu.kanade.tachiyomi.util.system.logcat
+import kotlinx.coroutines.launch
+import logcat.LogPriority
+import nl.adaptivity.xmlutil.core.impl.multiplatform.name
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+import androidx.preference.Preference as AndroidXPreference
+
+class SourcePreferencesScreen(val sourceId: Long) : Screen {
```
arkon, line 39: So when are we choosing classes vs data classes vs objects?

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/details/SourcePreferencesController.kt
```diff
@@ -1,179 +0,0 @@
...
-                pref.order = Int.MAX_VALUE // reset to default order
-
-                // Apply incognito IME for EditTextPreference
-                if (pref is EditTextPreference) {
-                    val setListener = pref.getOnBindEditTextListener()
-                    pref.setOnBindEditTextListener {
-                        setListener?.onBindEditText(it)
-                        it.setIncognito(viewScope)
-                    }
-                }
```
arkon, line 128: Are we still handling this?

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/details/SourcePreferencesScreen.kt
ivaniskandar: object generally for screen that will only be showed once (e.g nav tabs)
class when a param is needed
data class same as the above but usually for easy recreate with different param using `copy` method (which actually never needed so far)

app/src/main/java/eu/kanade/tachiyomi/ui/browse/extension/details/SourcePreferencesController.kt
ivaniskandar: sadly no just like other TextFields in other screens https://issuetracker.google.com/issues/211567805

arkon: What about handling things like `inputType = InputType.TYPE_CLASS_TEXT or InputType.TYPE_TEXT_VARIATION_PASSWORD`?

ivaniskandar: Yeah worry not I'm scrapping this and going with different approach

PR #8659, opened 2022-12-03
Change Updates icon badge to show new updates count

AntsyLich:


app/src/main/java/eu/kanade/tachiyomi/ui/updates/UpdatesTab.kt
```diff
@@ -110,5 +110,8 @@ object UpdatesTab : Tab {
                 (context as? MainActivity)?.ready = true
             }
         }
+        LaunchedEffect(Unit) {
+            screenModel.resetNewUpdatesCount()
```
arkon, line 114: Maybe it should reset after we navigate away (so I guess when this gets disposed)?

i18n/src/main/res/values/strings.xml
```diff
@@ -254,7 +254,7 @@
     <string name="pref_update_only_completely_read">With unread chapter(s)</string>
     <string name="pref_update_only_non_completed">With \"Completed\" status</string>
     <string name="pref_update_only_started">That haven\'t been started</string>
-    <string name="pref_library_update_show_tab_badge">Show unread count on Updates icon</string>
+    <string name="pref_library_update_show_tab_badge">Show new updates count on Updates icon</string>
```
arkon, line 257: If it gets cleared, we might as well just kill the option to hide it.

app/src/main/java/eu/kanade/tachiyomi/ui/updates/UpdatesTab.kt
AntsyLich: Resetting it at both enter and exit

PR #8719, opened 2022-12-11
Fix SetMangaViewerFlags

### LINKED ISSUE #8450: Per manga reading mode and rotation type resets to default if the other is changed
### Steps to reproduce

1. Open manga chapter. 
2. Change reading mode to webtoon. 
3. Exit chapter and close app. 
4. Reopen app and the same chapter. 
5. Webtoon setting is saved.
6. Change rotation type to fixed portrait.
7. Exit chapter and close app.
8. Reopen app and the same chapter.
9. Rotation type is saved, but reading mode resets to default.
10. Change reading mode to webtoon.
11. Exit chapter and app.
12. Reopen app and same chapter.
13. Reading mode stays as webtoon, but now the rotation type resets to default. 

### Expected behavior

Both reading mode and rotation type should stay the same as previously set. 

### Actual behavior

Either reading mode or rotation type resets to default if the other is changed. 

### Crash logs

_No response_

### Tachiyomi version

App version: 0.14.2 (standard, 3222247, 91, 2022-10-31T21:42Z) 

### Android version

12 (SDK 31)

### Device

Samsung Galaxy S10E (SM-G970U1) 

### Other details

_No response_

### Acknowledgements

Two-Ai:
Fixes #8450

Stop clearing old viewer flags when setting a flag

Alternative to #8718
#8718 relies on the `Manga` object staying in sync with the DB, but this costs an extra DB request.


arkon:
I suspect this one would be the more resilient of the two?

Two-Ai:
> I suspect this one would be the more resilient of the two?

Yes, though only by a bit.

On the other hand, it turns out [`SetMangaChapterFlags`](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/domain/manga/interactor/SetMangaChapterFlags.kt) already uses the pattern from https://github.com/tachiyomiorg/tachiyomi/pull/8718.

ghostbear:
The pattern from this PR is used in https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/domain/category/interactor/SetDisplayModeForCategory.kt

So both patterns exists in the codebase

PR #8746, opened 2022-12-15
App state banner tweaks

ivaniskandar:
* Move download indexing notification to this banner group
* Animate state changes


app/src/main/java/eu/kanade/tachiyomi/data/notification/Notifications.kt
```diff
@@ -40,8 +40,6 @@ object Notifications {
     const val ID_DOWNLOAD_CHAPTER_PROGRESS = -201
     const val CHANNEL_DOWNLOADER_ERROR = "downloader_error_channel"
     const val ID_DOWNLOAD_CHAPTER_ERROR = -202
-    const val CHANNEL_DOWNLOADER_CACHE = "downloader_cache_renewal"
```
arkon, line 43: Add it to `deprecatedChannels` so it gets deleted.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
```diff
@@ -66,6 +64,8 @@ class DownloadCache(
      */
     private var lastRenew = 0L
     private var renewalJob: Job? = null
+    val isRenewing: Boolean
```
arkon, line 67: Why not just make this a flow? Feels weird to check this based on every cache change in the activity.

app/src/main/java/eu/kanade/tachiyomi/data/notification/Notifications.kt
ivaniskandar: Done

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
ivaniskandar: Done. Still piggybacking off of the `changes` flow but added `distinctUntilChanged()`

PR #8763, opened 2022-12-17
Pass listing query to BrowseSourceScreen

stevenyomi:
~~Mark some classes as Serializable to avoid crashes because Voyager's Screen needs to be Serializable.~~

Pass string to the Screen as per https://voyager.adriel.cafe/state-restoration. Reverted some changes in #8762.

This would cause inconvenience when we eventually introduce custom listing - maybe just rewrite them all lol.

Also fixes a TextButton typed as Button.


jobobby04:
Are the Filters in the filter list serializable?

stevenyomi:
> Are the Filters in the filter list serializable?

I'm not sure. And I'm not sure if that even matters. But I'll make the field a string anyway to avoid introducing too much mess.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreen.kt
```diff
@@ -125,7 +125,7 @@ data class BrowseSourceScreen(
                         horizontalArrangement = Arrangement.spacedBy(8.dp),
                     ) {
                         FilterChip(
-                            selected = state.listing == Listing.Popular,
+                            selected = state.listing === Listing.Popular,
```
arkon, line 128: Does it really matter if they're the exact same instances or not?

stevenyomi: No but that's a singleton so referential equality should also work. I can revert this if you don't like.

arkon: Been bitten in the past by this so might as well avoid it.

stevenyomi: Done.

PR #8772, opened 2022-12-18
Update WorkManager

ivaniskandar:
Replaced deprecated `ExistingPeriodicWorkPolicy.REPLACE` mostly with `ExistingPeriodicWorkPolicy.UPDATE`
Also backup creator will attempt to run in the context of foreground service.

Tested on android 8.1, 12, 13.


app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupCreatorJob.kt
```diff
@@ -25,14 +26,21 @@ import java.util.concurrent.TimeUnit
...
         val backupPreferences = Injekt.get<BackupPreferences>()
-        val notifier = BackupNotifier(context)
         val uri = inputData.getString(LOCATION_URI_KEY)?.toUri()
             ?: backupPreferences.backupsDirectory().get().toUri()
         val flags = inputData.getInt(BACKUP_FLAGS_KEY, BackupConst.BACKUP_ALL)
         val isAutoBackup = inputData.getBoolean(IS_AUTO_BACKUP_KEY, true)
 
+        try {
+            setForeground(getForegroundInfo())
+        } catch (e: IllegalStateException) {
```
arkon, line 40: This is really the solution for the crashes we originally got? :|

ivaniskandar: i can't see the logs from 5364 but yeah ig so

PR #8803, opened 2022-12-25
Implement scanlator filter

### LINKED ISSUE #969: Filter chapters list by scanlator

AntsyLich:
Closes #969 (3 digit *phew*)

And @max-niederman i'm very sorry if you were working on this since i didn't notice your comment until now

| Chapter settings sheet | Filter dialog |
| ------- | ------- |
| Screenshot_1698829844 | Screenshot_1698829852 |


CarlosEsco:
As I stated on discord.  100% should be a filter to hide scanlator you don't want to see. 

Main benefit is you don't need extra logic if a new scanlator group picks up a series.  If you do a show filter, you need extra logic when a group is added or else it silently hides the group.

Also you have no logic for Library updates. Which is needed since right now a user is going to be notified there are new chapters even if it isn't in their selected group.

jobobby04:
Why is it an extra table that just contains a reference to the manga and the filtered scanlators? It makes much more sense to have it as a column.

CarlosEsco:
Two things missing

1. Still notifying a user when a chapter is of an excluded group. When it should be ignored.

2.  Not backing up the excluded scanlators (I believe neko and j2k use the same proto number so use that one)

specterflare:
Just wanted to say thank you to AntsyLitch for working on this! This has been my #1 most anticipated feature for years now. Can't wait to see it finally merged!

AntsyLich:
Update: Due to my exams and job I'm unable to work on this currently (feel free to work on this if you want). I'll start working again from October.

AntsyLich:
Ready to review.
Notes: 
1. I didn't implement backup for excluded scanlator filter (someone else do it please thanks)
2. From my testing the new library view query didn't slow down library loading any bit but we need to do more testing
3. Do we want the filter icon to change to "activate" color when scanlator filter is active? Currently it does not and only the icon inside the settings sheet is changed to active (see screenshot)

app/src/main/java/eu/kanade/presentation/manga/ChapterSettingsDialog.kt
```diff
@@ -132,6 +148,41 @@ private fun ColumnScope.FilterPage(
         state = bookmarkedFilter,
         onClick = onBookmarkedFilterChanged,
     )
+    if (onScanlatorFilterClicked != null) {
```
arkon, line 151: It might be less confusing to just always show this but show in the dialog that there's no scanlators?

app/src/main/java/eu/kanade/presentation/manga/components/ScanlatorFilterDialog.kt
```diff
@@ -0,0 +1,118 @@
...
+            usePlatformDefaultWidth = true,
+        ),
+        confirmButton = {
+            Row {
+                TextButton(onClick = mutableExcludedScanlators::clear) {
+                    Text(text = stringResource(R.string.action_reset))
+                }
+                Spacer(modifier = Modifier.weight(1f))
+                TextButton(onClick = onDismissRequest) {
+                    Text(text = stringResource(android.R.string.cancel))
```
arkon, line 105:
```suggestion
                    Text(text = stringResource(R.string.action_cancel))
```

app/src/main/java/eu/kanade/presentation/manga/components/ScanlatorFilterDialog.kt
```diff
@@ -0,0 +1,118 @@
...
+                }
+                TextButton(
+                    onClick = {
+                        onConfirm(mutableExcludedScanlators)
+                        onDismissRequest()
+                    },
+                ) {
+                    Text(text = stringResource(android.R.string.ok))
```
arkon, line 113:
```suggestion
                    Text(text = stringResource(R.string.action_ok))
```

app/src/main/java/eu/kanade/presentation/manga/components/ScanlatorFilterDialog.kt
```diff
@@ -0,0 +1,118 @@
...
+import androidx.compose.ui.window.DialogProperties
+import eu.kanade.tachiyomi.R
+import tachiyomi.presentation.core.components.material.TextButton
+import tachiyomi.presentation.core.components.material.padding
+import tachiyomi.presentation.core.util.isScrolledToEnd
+import tachiyomi.presentation.core.util.isScrolledToStart
+
+@Composable
+fun ScanlatorFilterDialog(
+    availableScanlator: List<String>,
```
arkon, line 38:
```suggestion
    availableScanlators: List<String>,
```

Also maybe we should be using `Set`s in more places?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreen.kt
```diff
@@ -219,6 +227,15 @@ class MangaScreen(
                 )
             }
         }
+
+        if (showScanlatorsDialog) {
+            ScanlatorFilterDialog(
+                availableScanlator = successState.availableScanlators,
+                excludedScanlators = successState.excludedScanlators,
```
arkon, line 234: Feels like we could get these on demand rather than subscribing to them.

data/src/main/sqldelight/tachiyomi/data/chapters.sq
```diff
@@ -38,6 +38,16 @@ WHERE _id = :id;
 getChaptersByMangaId:
 SELECT *
 FROM chapters
+WHERE manga_id = :mangaId
+AND scanlator NOT IN (
```
arkon, line 42: Wouldn't a left join be more efficient?

data/src/main/sqldelight/tachiyomi/migrations/26.sqm
```diff
@@ -0,0 +1,46 @@
...
+        count(*) AS total,
+        sum(read) AS readCount,
+        coalesce(max(chapters.date_upload), 0) AS latestUpload,
+        coalesce(max(history.last_read), 0) AS lastRead,
+        coalesce(max(chapters.date_fetch), 0) AS fetchedAt,
+        sum(chapters.bookmark) AS bookmarkCount
+    FROM chapters
+    LEFT JOIN history
+    ON chapters._id = history.chapter_id
+    WHERE NOT EXISTS (
```
arkon, line 35: Same here re:  left join.

data/src/main/sqldelight/tachiyomi/view/libraryView.sq
```diff
@@ -21,6 +21,12 @@ LEFT JOIN(
     FROM chapters
     LEFT JOIN history
     ON chapters._id = history.chapter_id
+    WHERE NOT EXISTS (
```
arkon, line 24: And here.

app/src/main/java/eu/kanade/presentation/manga/components/ScanlatorFilterDialog.kt
```diff
@@ -0,0 +1,133 @@
...
+                }
+                if (!state.isScrolledToStart()) HorizontalDivider(modifier = Modifier.align(Alignment.TopCenter))
+                if (!state.isScrolledToEnd()) HorizontalDivider(modifier = Modifier.align(Alignment.BottomCenter))
+            }
+        },
+        properties = DialogProperties(
+            usePlatformDefaultWidth = true,
+        ),
+        confirmButton = {
+            Row {
```
arkon, line 108: Use a `FlowRow` to deal with smaller windows/longer translations.

PR #8818, opened 2022-12-28
Handle 1000+ pages properly in the downloader

jobobby04:
Closes https://github.com/tachiyomiorg/tachiyomi/issues/1561

Had an idea on how to handle this while keeping the same behavior, decided to just PR it instead of add it to my fork first since its small.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
@@ -376,7 +376,9 @@ class Downloader(
             return Observable.just(page)
         }
 
-        val filename = String.format("%03d", page.number)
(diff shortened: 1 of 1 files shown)


arkon:
I'm not sure I understand what this is doing.

Don't `String.format("%03d", 1000)` and `String.format("%04d", 1000)` both just output `"1000"`?

[This sorting](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt#L180) doesn't take that into account still either.

jobobby04:
Yes, but for pages under 1000 like 900, it would result in `String.format("%03d", 1000) = 900` and `String.format("%04d", 1000) = 0900`, which fixes the order which the reader displays them

arkon:
Ah right, it's based on the total number of pages. That works.

PR #8847, opened 2023-01-07
Minor cleanup dialog

zbue:
- Enable `confirmButton` when it only meets the need to do so
- Added [ConfirmActionDialog] to every bottom action menu button (except for change category, download action) since they cannot undo the action once it is performed
- Fixed [WheelPicker] background height changing when the value is changed from initial value
- Others

This PR does not contain strings cleanup (user agent error, category exists error)


arkon:
Please make separate PRs.

PR #8848, opened 2023-01-07
Enable `confirmButton` only when needed to respond to user input

zbue:
This PR does not contain strings cleanup (user agent error)


app/src/main/java/eu/kanade/presentation/category/components/CategoryDialogs.kt
```diff
@@ -23,17 +24,21 @@ import kotlin.time.Duration.Companion.seconds
...
 
     AlertDialog(
         onDismissRequest = onDismissRequest,
         confirmButton = {
-            TextButton(onClick = {
-                onCreate(name)
-                onDismissRequest()
-            },) {
+            TextButton(
+                enabled = !categories.anyWithName(name) && name.isNotEmpty(),
```
arkon, line 36: This doesn't actually tell the user why it's disabled, whereas the flow before showed the error as a toast.

zbue:
Maybe this will give a better ux rather than closing the dialog and showing a toast(?)
| When opened | Name already exists | Name do no exist |
| ------------- | ------------- | ------------- |
| image  |    image | image |
| image _initial value_  | image _value has changed_ | image |

arkon:
https://m3.material.io/components/text-fields/guidelines#9ad90554-a793-4506-9075-6812fd7b381a

PR #8849, opened 2023-01-07
Make center zoom start vertically centered too

### LINKED ISSUE #8747: "true center" Zoom start position

Currently "Center" zoom start position only horizontally center the page. It would be helpful if there;s  an option for "true center", where zoom start position would be horizontally and vertically centered.

This would make devices with a shorter screen ratio (e.g. 4:3, 5:3) to be able to display many manga in fullscreen (with auto crop border enabled), without needing to scroll down. Because many manga did not place important elements near the top and bottom edges.

this feature request, "true center" zoom's usefulness, would diminish on taller devices. So it would be pretty much useless for most modern phones nowadays (21:9). as i don't think there's any manga with panels that tall.

it basically would help eink tablets. as most of them have shorter aspect ratio. Just like the current zoom start positions helps tall phones to view manga in fullscreen.

### Other details

_No response_

### Acknowledgements

RS156:
"true center" Zoom start position 
Changed zoom position to get true center value.

Closes #8747


arkon:
Not entirely sure if people would dislike this and whether it should be a separate option, but I guess we'll see from the feedback.

PR #8857, opened 2023-01-07
Fix reader settings sheet's mode section not updated

stevenyomi:
Fix changing reading mode not updating the corresponding section.


arkon:
Don't need the same thing for orientation?

stevenyomi:
> Don't need the same thing for orientation?

Orientation doesn't really affect other parts of the menu. Reading mode affects the "Paged/Webtoon" part.

PR #8884, opened 2023-01-10
Allow partially read chapters to be marked as unread in updates screen

zbue:


app/src/main/java/eu/kanade/tachiyomi/ui/updates/UpdatesScreenModel.kt
```diff
@@ -110,24 +112,32 @@ class UpdatesScreenModel(
...
                 downloadProgressProvider = { activeDownload?.progress ?: 0 },
-                selected = it.chapterId in selectedChapterIds,
+                selected = update.chapterId in selectedChapterIds,
+                readProgressString = update.lastPageRead
+                    .takeIf { !update.read && it > 0 }
+                    ?.let {
+                        context.getString(
+                            R.string.chapter_progress,
+                            it + 1,
+                        )
```
arkon, line 139: This should be happening in the composable, not the screen model.

zbue: Should I do it for manga screen as well?

arkon: Sure, but I don't mind having that change seperate

app/src/main/java/eu/kanade/presentation/updates/UpdatesUiItem.kt
```diff
@@ -113,6 +117,14 @@ fun LazyListScope.updatesUiItems(
                     modifier = Modifier.animateItemPlacement(),
                     update = updatesItem.update,
                     selected = updatesItem.selected,
+                    readProgress = updatesItem.update.lastPageRead
+                        .takeIf { !updatesItem.update.read && it > 0L }
+                        ?.let {
+                            context.getString(
```
arkon, line 123: Can just use `stringResource` in composables.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreenModel.kt
```diff
@@ -507,12 +507,14 @@ class MangaInfoScreenModel(
                             dateFormat,
                         )
                     },
-                readProgressString = chapter.lastPageRead.takeIf { !chapter.read && it > 0 }?.let {
-                    context.getString(
-                        R.string.chapter_progress,
-                        it + 1,
-                    )
-                },
+                readProgressString = chapter.lastPageRead
```
arkon, line 510: Set the number and do the string part in the composable.

arkon (changes requested):

PR #8895, opened 2023-01-12
yeet `ChapterHeader` click

zbue:


arkon:
It's an intended shortcut.

PR #8897, opened 2023-01-12
Fix downloaded filter unmatched state in manga screen

zbue:
Since the checking is already done in [downloadedFilter](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt#L64)


app/src/main/java/eu/kanade/presentation/manga/ChapterSettingsDialog.kt
```diff
@@ -76,14 +75,9 @@ fun ChapterSettingsDialog(
         ) {
             when (page) {
                 0 -> {
-                    val forceDownloaded = manga?.forceDownloaded() == true
                     FilterPage(
-                        downloadFilter = if (forceDownloaded) {
-                            TriStateFilter.ENABLED_NOT
```
Two-Ai, line 82: This is the reverse of the checking in [`Manga.downloadedFilter`](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt#L66):
```kotlin
            if (forceDownloaded()) return TriStateFilter.ENABLED_IS
```
Is the change in behavior intended?

zbue: Yes. This fixes the visual bug in downloaded filter state.

zbue: To reproduce:
1. Open any manga with few downloaded chapters and turn on downloaded filter in chapter settings sheet
2. Toggle downloaded only switch on
3. Open the entry you've filtered from step 1

You'll see the downloaded filter state is "❌" when it should be "✅"

Issue #6985, opened 2022-04-23
Skipped manag log 


iAdoly:
### Describe your suggested feature

Like chapter error to see what is skipped

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.13.4](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

SeriousMatters:
I think the Skipped manga log can go into the Updates section. Perhaps in a separate tab/category

Noriller:
I'm in the middle of migrating a lot of entries that have been skipped because of source errors.
But I must either know which is which or check the logs manually.

Instead of adding a new issue, I'm adding a suggestion here:

Add a "with error" badge to entries, on each global update it would refresh them (or maybe, you could even add those with errors to be skipped).

This way, it would be possible to know with a glance which entries have errors and filter by them.

Issue #6654, opened 2022-02-15
Toggle sources from being used in migration 


Onering20:
### Describe your suggested feature

When there are one off extensions (XKCD), comic readers, NSFW sources, or a source that doesn't have searches properly implemented; to be able to toggle them from being in the migration pool.

I think it would save time and processing effort, while making efforts to try to consolidate sources easier.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.13.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

vengiare:
While `Settings > Browse > Only include pinned sources` kinda does this functionality already, yeah it would be better if there's a separate setting for migrate search.

arkon:
Related from #6315: should allow quickly selecting just the pinned sources.

zaghdaneh:
Hi, I'm working on this and need some clarifications:
- Where should this "toggle" be ? In the sources ? (accessible by clicking on the top right icon in the browse tab)
and when clicking it it would enable all the sources that are pinned ?

arkon:
Global search currently just opens a new screen with a search text field in the top app bar, so we could probably put some chips to control settings like sources between the search field and the results list underneath.

PR #8928, opened 2023-01-16
Exclude disabled sources in global update

zbue:
It could be useful if:
- Source is down
- Source hasn't been fixed yet
- Source has been removed and user hasn't yet found another source to migrate to.

As a result, users won't be notified when an update fails if the source disabled.
#
They can still manually refresh the entry, or pull to update category in the library to check the source.


arkon:
It seems relatively common to hide sources from your list even though you've added something from it. I do wonder if doing this is more inline with the concept of "disabling" a source though?

AntsyLich:
I don't know about other but i have some source that i only use for 1 or 2 entry and does it have it disabled so it doesn't appear in Sources

stevenyomi:
At least:
- Don't skip disabled languages, only disabled sources
- Mark those entries as skipped and implement #6985 (should be easy, just need to print the logs or just yoink code from J2K)

arkon:
I think #6654 would be a prerequisite to this too.

PR #8951, opened 2023-01-19
Move shared configuration to subprojects in root Gradle file

ghostbear:


build.gradle.kts
```diff
@@ -11,6 +16,64 @@ plugins {
     alias(kotlinx.plugins.serialization) apply false
 }
 
+subprojects {
+    tasks.withType<KotlinJvmCompile> {
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8.toString()
+            // See https://kotlinlang.org/docs/reference/experimental.html#experimental-status-of-experimental-api(-markers)
+            freeCompilerArgs += listOf(
```
arkon, line 24: It'd probably make sense to keep these to the relevant module(s). A lot of these are specific to the UI stuff that should eventually be split off into a separate module anyway.

build.gradle.kts
```diff
@@ -11,6 +16,64 @@ plugins {
...
+
+    plugins.withType<BasePlugin> {
+        plugins.apply("tachiyomi.lint")
+        configure<BaseExtension> {
+            compileSdkVersion(AndroidConfig.compileSdk)
+            defaultConfig {
+                minSdk = AndroidConfig.minSdk
+                targetSdk = AndroidConfig.targetSdk
+                versionCode = 94
+                versionName = "0.14.3"
```
arkon, line 58: This also seems appropriate to keep in `app`.

PR #8955, opened 2023-01-21
Replace RxJava in HttpPageLoader downloader

Two-Ai:
- Convert downloader `Observable` to `flow`
  - Uses [`runInterruptible`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/run-interruptible.html) to turn the blocking call to `queue.take()` into a cancellable call.
  - Flow collection is ended by cancelling the scope in `recycle`. (This means the `HttpPageLoader` can't be reused after calling `recycle`, but this was true with the `Observable` as well.)
- Convert and combine `fetchImageFromCacheThenNet`/`getImageUrl`/`getCachedImage`/`cacheImage` into a single suspending function (`loadPage`)
  - Inlining the observables allows for some simplification of the error handling.
  - Behavior should be otherwise identical.
- Convert cleanup `Completable` to coroutine
  - Uses global `launchIO`, not ideal but similar to previous behavior. Can't be scheduled on the local `scope` as this runs after `scope` is cancelled.

RxJava is still used in `getPage`. `getPage` interfaces with the downloader via the `PriorityBlockingQueue`, so the two can be refactored independently.


arkon:
Pages don't seem to load with some sources after this btw, but I haven't deep dived into it.

PR #5759, opened 2021-08-20
Long Strip Split for Webtoon

AntsyLich:
The splitting is on by default to know how it may effect different devices

Continues #4791 

Thanks @ghostbear for helping out time to time.

Issue #8433, opened 2022-11-03
Split tall images (BETA) bugged


TsumiHokiro:
### Steps to reproduce

1 - Long strip comic with a sufficient number of pages (10+);
2 - Scroll down the chapter as pages load;
3 - Notice the discontinuity of the comic.

### Expected behavior

No page skipping or transparent notification/display of page not downloaded.

### Actual behavior

Pages which might have been downloaded or are yet to be are skipped/not displayed as such.

### Tachiyomi version

Tachiyomi Preview486

### Android version

Android 10

### Device

SM-G9600

### Other details

I have noticed this happening at least since r478. A few things I've noticed are that:
1 - The longer the pages are, the more likely they are to be hidden/cropped/not displayed.
2 - In some cases, images that were downloaded and I managed to see a part of  soon were hidden, and the next one was displayed [page skip].
3 - A few ways to try to make the app display the images: Scroll Down and then all the way up again (Works best with chapters with more pages); restart reading the chapter; close the app and try it all again.
4 - The app may or not later re-display images that were skipped without user intervention.
5 - The middle of the image more often than not gets displayed, however, either/both upper or/and may be hidden.
6 - The Issue is gone once Split Tall Images (BETA) is disabled in Reader Settings. [Edited for English]
7 - This issue is similar to, but not exactly the same as #7943 and has been reported to jobobby04/TachiyomiSY#711 who recommended to report it here.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] I have updated the app to version **[0.14.2](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

haganbmj:
r5202 here, same behavior. 

Applies to both Webtoon and Continuous Vertical and only when Split Tall Images is enabled. 

Continuous Vertical makes it more obvious when it happens as it makes the image splitting more obvious. 

Behavior appears to be that one of the resulting split images is skipped. 

[...]

PR #8959, opened 2023-01-21
Change default 32-bit color setting to on

specterflare:
Now that Long Strip Split (#5759) has been added, I can't see any reason to keep the default color bit setting to 16. The performance difference is minimal in nearly all cases, but the banding is very much noticeable. At this point, I feel that 32-bit color is the better default experience.


arkon:
Reader long split strip still has bugs (#8433) and isn't in stable, so this wouldn't really be mergable until that's resolved.

arkon:
Tracking this as a TODO (8d7a7919a9bc4b3e2242486c16d0c8b947e8ab64).

PR #8965, opened 2023-01-22
Prevent changing tab when in selection mode

zbue:
Hiding navigation rail does not look good tho.

Can be closed if anyone can think of better solution


arkon:
What's the motivation behind this? Sort of matching what happens on phones vs. tablet UI?

PR #9006, opened 2023-01-31
Show exception class in snackbar message

stevenyomi:
Show simple name for exceptions that are not `HttpException`, `IOException` or plain `Exception`.

#8759 → #8900 → this


alessandrojean (approved):

alessandrojean:
Probably just double check if the No chapters found simpleName is omitted as well.

PR #9007, opened 2023-01-31
Fully utilize WorkManager for library updates

ivaniskandar:
No more trampolining, and stuff.

It's pretty much straight copy-paste from the service, with some changes related to cancellation handling. Manual updates will also runs with workman job so auto update work scheduling need some adjustments too.

Bumped version code to re-enqueue auto update job with the new spec.


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
```diff
@@ -1,44 +1,551 @@
...
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
+import java.io.File
+import java.util.Date
+import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
 
 class LibraryUpdateJob(private val context: Context, workerParams: WorkerParameters) :
```
arkon, line 87: I feel like you could've still left the logic in the separate class rather than merging the two into a bigger class.

PR #9016, opened 2023-02-03
Replace RxJava in DownloadQueue

Two-Ai:
- Misc cleanup
  - Replace `!List.isEmpty` with `List.isNotEmpty`
  - Remove redundant case in `MoreScreenModel`
  - Drop no-op `StateFlow.catch`
    - From [lint warning](https://github.com/Kotlin/kotlinx.coroutines/blob/32848d967722caa476b538e30c8e265638311ae5/kotlinx-coroutines-core/common/src/flow/operators/Lint.kt#L101-L109): "SharedFlow never completes, so this operator typically has not effect, it can only catch exceptions from 'onSubscribe' operator"
- Convert `DownloadQueue` queue to `MutableStateFlow`
  - Replace delegation to a `MutableList` with an internal `MutableStateFlow<List>`.
  - In order to avoid modifying every usage of the queue as a list, add passthrough functions for the currently used list functions. This should be later refactored, possibly by inlining `DownloadQueue` into `Downloader`.
  - `DownloadQueue.updates` was a `SharedFlow` which updated every time a change was made to the queue. This is now equivalent to the queue `StateFlow`.
  - Simultaneous assignments to `_state.value` could cause concurrency issues. To avoid this, always modify the queue using [`_state.update`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/update.html).
-  Add `Download.statusFlow`/`progressFlow`
   - `progressFlow` is based on the [`DownloadQueueScreenModel` implementation](https://github.com/tachiyomiorg/tachiyomi/blob/7559c133c05cff0a41d11757ec31c86da7bf5c74/app/src/main/java/eu/kanade/tachiyomi/ui/download/DownloadQueueScreenModel.kt#L207-L219) rather than the [`DownloadQueue` implementation](https://github.com/tachiyomiorg/tachiyomi/blob/7559c133c05cff0a41d11757ec31c86da7bf5c74/app/src/main/java/eu/kanade/tachiyomi/data/download/model/DownloadQueue.kt#L108-L113).
-  Reimplement `DownloadQueue.statusFlow`/`progressFlow`
   - Use [`StateFlow<List<T>>.flatMapLatest()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/flat-map-latest.html) and [`List<Flow<T>>.merge()`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/merge.html) to replicate the effect of `PublishSubject`.
   - Use `drop(1)` to avoid re-emitting the state of each download each time the merged flow is recreated.


app/src/main/java/eu/kanade/tachiyomi/data/download/model/DownloadQueue.kt
```diff
@@ -1,131 +1,99 @@
...
+        _state.value.filter { download -> download.status == Download.State.DOWNLOADING }.asFlow()
+
+    fun count(predicate: (Download) -> Boolean) = _state.value.count(predicate)
+    fun filter(predicate: (Download) -> Boolean) = _state.value.filter(predicate)
+    fun find(predicate: (Download) -> Boolean) = _state.value.find(predicate)
+    fun <K> groupBy(keySelector: (Download) -> K) = _state.value.groupBy(keySelector)
+    fun isEmpty() = _state.value.isEmpty()
+    fun isNotEmpty() = _state.value.isNotEmpty()
+    fun none(predicate: (Download) -> Boolean) = _state.value.none(predicate)
+    fun toMutableList() = _state.value.toMutableList()
```
arkon, line 98: Not entirely sure it's worth duplicating all of the list functions, but I guess we can revisit that separately.

Issue #8382, opened 2022-10-31
Copy manga tags to clipboard when long pressed


Satria41:
### Steps to reproduce

1: open manga
2: long press a tags

### Expected behavior

Tags should be copied to clipboard

### Actual behavior

No tags was Copied to clipboard

### Tachiyomi version

0.14.0 & 0.14.1

### Android version

Android 9 

### Device

Asus Zenfone Max Pro M2(on pixel experience)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] I have updated the app to version **[0.14.1](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

mayorJAY:
I will like to fix this bug @arkon

kingtie:
Don't know if this is already known, but the same thing happens for titles, sources and authors also.

PR #9058, opened 2023-02-08
Add action to copy tags to clipboard

0x7673:
Implements #8382 

Detecting long press on Compose chips doesn't look straightforward. [source](https://stackoverflow.com/questions/74119839/detecting-long-press-events-on-chips-in-jetpack-compose).

I suggest we add new item in menu to copy tags.


arkon:
I'm not really a fan of this. I could maybe see having an actual menu pop up when tapping the individual chips for:

1. Search in library
2. Global search
3. Copy to clipboard

Although open to other ideas.

PR #9063, opened 2023-02-10
Add copy tags to clipboard feature

0x7673:
Implements #8382 

Clicking on individual genre chip will open pop up menu which will allow user to search genre (global or library) and copy tags to clipboard.

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
@@ -72,6 +72,7 @@ import eu.kanade.tachiyomi.ui.manga.ChapterItem
 import eu.kanade.tachiyomi.ui.manga.MangaScreenState
 import eu.kanade.tachiyomi.ui.manga.chapterDecimalFormat
 import eu.kanade.tachiyomi.util.lang.toRelativeString
+import eu.kanade.tachiyomi.util.system.copyToClipboard
 import tachiyomi.domain.chapter.model.Chapter
 import tachiyomi.domain.manga.model.Manga
 import java.text.DateFormat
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -91,7 +92,10 @@ fun MangaScreen(
     onWebViewClicked: (() -> Unit)?,
     onWebViewLongClicked: (() -> Unit)?,
     onTrackingClicked: (() -> Unit)?,
-    onTagClicked: (String) -> Unit,
+
+    // For tags menu
+    onGenreSearch: (String) -> Unit,
```
arkon, line 97: Let's stick with the `Tag` terminology for now.

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -117,6 +121,15 @@ fun MangaScreen(
     onAllChapterSelected: (Boolean) -> Unit,
     onInvertSelection: () -> Unit,
 ) {
+    val context = LocalContext.current
+    val onCopyTagsToClipboard: () -> Unit = {
```
arkon, line 125: It'd make more sense to copy the individual tag rather than all of them.

i18n/src/main/res/values/strings.xml
```diff
@@ -96,6 +96,7 @@
     <string name="action_resume">Resume</string>
     <string name="action_open_in_browser">Open in browser</string>
     <string name="action_show_manga">Show entry</string>
+    <string name="action_copy_tags_to_clipboard">Copy tags to clipboard</string>
```
arkon, line 99: Let's just make this more generic like "Copy to clipboard" so we can reuse it elsewhere.

arkon (changes requested):

PR #9103, opened 2023-02-18
Remove RxJava in PageHolder

Two-Ai:
Inline `readImageHeaderSubscription` in `PagerPageHolder` and `WebtoonPageHolder` by converting `setImage()` into a suspend function.
- The image processing runs in the `loadPageAndProcessStatus()` continuation.
- `WebtoonPageHolder`: Use `suspendCancellableCoroutine` as a substitute for `doOnUnsubscribe`.
  - Closing `openStream` after the `frame.setImage()` but before the `WebtoonPageHolder` is recycled causes the page display to fail for reasons that are not currently understood. (See also https://github.com/tachiyomiorg/tachiyomi/commit/8ef200861cbc9c0332fedf4622ee7cf9394230a8)
- Remove subscription handling from WebtoonViewer/WebtoonBaseHolder as it is no longer used.


app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/WebtoonPageHolder.kt
```diff
@@ -121,7 +114,6 @@ class WebtoonPageHolder(
     override fun recycle() {
         loadJob?.cancel()
         loadJob = null
-        unsubscribeReadImageHeader()
```
arkon, line 124: I guess this would've technically cancelled the work to read the header / setting the image if it was getting recycled. Might still be useful to do something like that?

PR #9143, opened 2023-02-25
Fix banners-related issues

ivaniskandar:
This is most likely Compose issue so these changes will be reevaluated when new Compose ver is out.


app/src/main/java/eu/kanade/presentation/components/Banners.kt
```diff
@@ -96,7 +98,12 @@ fun AppStateBanners(
         }.fastMap { it.measure(constraints) }
         val downloadedOnlyHeight = downloadedOnlyPlaceable.fastMaxBy { it.height }?.height ?: 0
 
-        val incognitoPlaceable = subcompose(2) {
+        val incognitoId = when {
+            indexing && downloadedOnlyMode -> 3
+            indexing || downloadedOnlyMode -> 2
+            else -> 1
+        }
```
stevenyomi, line 105: Can you just make a variable and increment it?

ivaniskandar: Feel free to suggest I can't think of anything simpler than this

PR #9152, opened 2023-02-26
Move Local Source to separate module

ghostbear:
- Move Local Source to a separate module
- Move ComicInfo and MangaDetails (details.json) to a separate module
- Break down Local Source into smaller single-responsibility classes
- Create extension functions for extracting metadata from EpubFile (removes the dependency of SManga in it)
- Move RequestStoragePermission from DiskUtil as it concerns the presentation layer only


source-local/src/main/java/tachiyomi/source/local/io/AndroidLocalSourceFileSystem.kt
```diff
@@ -0,0 +1,39 @@
...
+            .asSequence()
+    }
+
+    override fun getFilesInBaseDirectories(): Sequence<File> {
+        return getBaseDirectories()
+            // Get all the files inside all baseDir
+            .flatMap { it.listFiles().orEmpty().toList() }
+    }
+
+    override fun getMangaDirectories(mangaUrl: String): File? {
```
arkon, line 26: `mangaUrl` seems like an odd name considering it's supposed to be the name.

The function name is also confusing. It sounds like it gets multiple directories, but it only ever returns 0 or 1 directory.

source-local/src/main/java/tachiyomi/source/local/io/AndroidLocalSourceFileSystem.kt
```diff
@@ -0,0 +1,39 @@
...
+        return getFilesInBaseDirectories()
+            // Get the first mangaDir or null
+            .firstOrNull { it.isDirectory && it.name == mangaUrl }
+    }
+
+    override fun getFilesMangaDirectories(mangaUrl: String): Sequence<File> {
```
arkon, line 32: Same re: `mangaUrl`

source-local/src/main/java/tachiyomi/source/local/metadata/EpubFile.kt
```diff
@@ -0,0 +1,60 @@
+package tachiyomi.source.local.metadata
+
+import eu.kanade.tachiyomi.source.model.SChapter
+import eu.kanade.tachiyomi.source.model.SManga
+import eu.kanade.tachiyomi.util.storage.EpubFile
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.*
```
arkon, line 8: Is the linting not working? I thought we don't allow wildcards.

source-local/src/main/java/tachiyomi/source/local/image/AndroidLocalCoverManager.kt
```diff
@@ -0,0 +1,55 @@
...
+import tachiyomi.source.local.io.LocalSourceFileSystem
+import java.io.File
+import java.io.InputStream
+
+private const val DEFAULT_COVER_NAME = "cover.jpg"
+
+class AndroidLocalCoverManager(
+    private val context: Context,
+    private val fileSystem: LocalSourceFileSystem,
+) : LocalCoverManager {
```
arkon, line 17: Did you want to set up the module as multiplatform from the get-go? We could just have a single Android source set.

source-local/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
+plugins {
+    id("com.android.library")
+    id("org.jetbrains.kotlin.android")
```
arkon, line 3:
```suggestion
    kotlin("android")
```

source-local/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+    implementation(project(":domain"))
+
+    implementation(kotlinx.bundles.serialization)
+
+    implementation(libs.unifile)
+    implementation(libs.junrar)
+
+    testImplementation(libs.junit)
+    androidTestImplementation(androidx.test.ext)
+    androidTestImplementation(androidx.test.espresso.core)
```
arkon, line 32: There's no tests yet, could emit the test deps/config.

source-local/src/main/java/tachiyomi/source/local/metadata/EpubFile.kt
ghostbear: It did work for import order so it's doing at least running

source-local/src/main/java/tachiyomi/source/local/image/AndroidLocalCoverManager.kt
ghostbear: I was thinking more about making it possible when coding, but we could set it up.  It will probably be worth making it a separate PR as it will likely involve more rewriting

PR #9179, opened 2023-03-06
Initialize download index disk cache

ivaniskandar:


stevenyomi:
Should we consider using ProtoBuf instead of JSON for better performance and smaller file size?

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
```diff
@@ -352,28 +392,65 @@ class DownloadCache(
...
 
 /**
  * Class to store the files under a manga directory.
  */
+@Serializable
 private class MangaDirectory(
+    @Serializable(with = UniFileAsStringSerializer::class)
     val dir: UniFile,
     var chapterDirs: MutableSet<String> = mutableSetOf(),
 )
```
stevenyomi, line 445: Assign `@ProtoNumber`s?

ivaniskandar: doesn't need one imo

stevenyomi: Just a reminder not to reorder the fields.

ivaniskandar: should anything happen the disk cache will be regen'd automatically

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
```diff
@@ -352,28 +392,65 @@ class DownloadCache(
...
 }
 
 /**
  * Class to store the files under the root downloads directory.
  */
+@Serializable
 private class RootDirectory(
+    @Serializable(with = UniFileAsStringSerializer::class)
     val dir: UniFile,
-    var sourceDirs: ConcurrentHashMap<Long, SourceDirectory> = ConcurrentHashMap(),
```
arkon, line 362: Using non-concurrent data structures resulted in crashes before.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
```diff
@@ -180,6 +213,7 @@ class DownloadCache(
         // Save the chapter directory
         mangaDir.chapterDirs += chapterDirName
 
+        updateDiskCache()
         notifyChanges()
```
arkon, line 217: Just do `updateDiskCache` within `notifyChanges`?

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
ivaniskandar: lmk how to replicate the crashes too so I can check. ConcurrentHashMap is not kotlin serializable so uhh we may need to go the other way to avoid the crash

ivaniskandar: I opted to use mutex lock instead, gonna test this setup for the next few days

ivaniskandar: So far so good.

PR #9184, opened 2023-03-08
Added missing chapters count in MangaInfoHeader

foxscore:
# Summary
Added a `Missing X chapters` chip to the chapter header. Will not be shown if there are none missing *(duh)*.
Alternatively, a `Might be missing chapters` message will appear if the source doesn't assign chapters properly.

# Why tho?
Because nobody wants to start reading a manga, only to find out after a few hours that half the chapters are just not there.
This way, the reader might be inclined to look up the manga from a different source beforehand.

# Images
<details>
  <summary>Example: Missing 260 chapters</summary>
  

Screenshot_20230317-144851
	
</details>

<details>
	<summary>Example: Might be missing chapters</summary>
	

Screenshot_20230317-144802

</details>


app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -77,6 +77,7 @@ import tachiyomi.presentation.core.util.isScrolledToEnd
 import tachiyomi.presentation.core.util.isScrollingUp
 import java.text.DateFormat
 import java.util.Date
+import eu.kanade.tachiyomi.util.getMissingChapters as getMissingChapters
```
arkon, line 80: Unnecessary alias?

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
```diff
@@ -0,0 +1,36 @@
+package eu.kanade.tachiyomi.util
+
+import eu.kanade.tachiyomi.ui.manga.ChapterItem
+
+public fun getMissingChapters(chaptersInput: List<ChapterItem>): List<Int>? {
```
arkon, line 5: Could just return an `Int`, we never use the actual items.

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
```diff
@@ -0,0 +1,36 @@
...
+    // Get and evaluate most recent chapter
+    var prevChapter = chapters[0].chapter.chapterNumber
+
+    // Evaluate chapters from most to least recent
+    for (i in 1 until chapters.count()) {
+        // Get chapter
+        currentChapter = chapters[i].chapter.chapterNumber
+
+        // Evaluate
+        if (currentChapter == -1f) {
```
arkon, line 22: Could probably just filter out everything without a recognized chapter number first.

app/src/main/java/eu/kanade/presentation/manga/components/MangaInfoHeader.kt
```diff
@@ -371,6 +377,55 @@ private fun MangaAndSourceTitlesLarge(
                 textAlign = TextAlign.Center,
             )
         }
+        if (missingChapters == null) {
```
arkon, line 380: It'd probably make more sense to put this as a subtitle in `ChapterHeader`.

arkon (changes requested):

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
foxscore: Was added automatically by IDE 🤷‍♂️ No idea why it did that

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
foxscore: Solved in 9cf5454b1d33720d517ecda37a8ad1207ec314e5

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
foxscore: Solved in 9cf5454b1d33720d517ecda37a8ad1207ec314e5

app/src/main/java/eu/kanade/presentation/manga/components/MangaInfoHeader.kt
foxscore: — Solved in 4637c241f0ad502d1d2b25c0613e4f3d68bae25b —
I also changed the style so that it fits better with the new location.

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
```diff
@@ -0,0 +1,47 @@
+package eu.kanade.tachiyomi.util
```
arkon, line 1: Put this in the `domain` module under `tachiyomi.domain.chapter.service`. It'd be ideal to have tests too.

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import eu.kanade.tachiyomi.ui.manga.ChapterItem
+import kotlin.math.ceil
+
+public fun countMissingChapters(chaptersInput: List<ChapterItem>): Int? {
```
arkon, line 6: `public` is redundant on Kotlin.

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
```diff
@@ -0,0 +1,47 @@
...
+    if (chaptersInput.isEmpty()) {
+        return 0
+    }
+
+    // If any chapters[i].chapter.chapterNumber is -1, return null
```
arkon, line 11: I'd probably just filter out the `-1`s and continue with the rest of it in cases that you have a list like `[-1, -1, 1, 2, 3, 4, 5, ...]`.

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
```diff
@@ -0,0 +1,47 @@
...
+    if (chaptersInput.any { it.chapter.chapterNumber == -1f }) {
+        return null
+    }
+
+    var count = 0
+    val chapters = chaptersInput.sortedBy { it.chapter.chapterNumber }.reversed()
```
arkon, line 17: Doing it in reverse seems kind of unnecessary?

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
```diff
@@ -0,0 +1,47 @@
...
+    var currentChapter = 0f
+
+    // If the first chapter is equal to or greater than 2, there are missing chapters
+    if (chapters.last().chapter.chapterNumber >= 2) {
+        // The chapter number might be, for example, 16.5, (with 16 being the previous one)
```
arkon, line 22: When you do the initial sorting, you can also convert them to `Int`s and `distinct` the list to skip this sort of check.

app/src/main/java/eu/kanade/presentation/manga/components/ChapterHeader.kt
```diff
@@ -40,5 +45,42 @@ fun ChapterHeader(
...
+                    id = R.plurals.missing_chapters,
+                    count = missingChapters,
+                    missingChapters,
+                ),
+            )
+        }
     }
 }
+
+@Composable
```
arkon, line 66: I'm not sure I'm really a fan of this UI, but I guess we can iterate on it separately based on feedback.

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/util/MangaMissingChapters.kt
foxscore: Should there be a check for `if all are -1`, or should these *(likely very rare)* cases be ignored?

arkon: It'd sort of end up depend on how the check after works, but it's also not uncommon for that to happen (even if just basic cases like a single `[-1]` for a doujin). That shouldn't be considered as missing things though.

app/src/main/java/eu/kanade/presentation/manga/components/ChapterHeader.kt
foxscore: I mainly chose it because it stands out enough to catch the users attention, but I don't like that it makes the whole row take up more empty space on the top & bottom

arkon (approved):

PR #9217, opened 2023-03-16
BrowseSource: do networkToLocal and initializeManga inside flow

akozlins:
during browsing of a source, after finishing the query for list of mangas for a given page
the networkToLocalManga is essentially called one by one (serially),
and only after that getManga.subscribe creates flow for each manga

the problem is that networkToLocalManga is relatively slow when inserting into database,
consequently only after all entries are committed into DB (slow) and flows are created (fast)
the new page can be requested

with this change the flow for each manga is created right after the list is loaded,
such that networkToLocalManga, filtering and initializeManga happen inside individual flows,
this allows much faster response to user actions (e.g. when reaching the end of current page,
new page is requested much faster, also each flow can be cancelled when out of view of user)

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreenModel.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreenModel.kt
@@ -31,8 +31,10 @@ import eu.kanade.tachiyomi.util.removeCovers
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.emptyFlow
+import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.firstOrNull
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreenModel.kt
ivaniskandar, line 1: I'm quite sure L77 can be removed now

arkon: Wouldn't that actually still be desired to receive updates from after the manga is opened and potentially updated via the details screen?

akozlins: @arkon i think the behavior should still be the same: before this change one could still go into manga detail page while the manga is not yet committed to DB (as the networkToLocal happens one by one),

- though I do agree that this seems a bit dangerous as subscribe does try to get manga from local storage
- I did run with the above changes for past week on 3 devices and did not have any crashes or problems accessing detail pages during page loads

arkon: I think what I was describing is probably what `initializeManga` is doing so it's potentially fine.

PR #9220, opened 2023-03-16
Fallback chapter name if it ends up as blank

### LINKED ISSUE #8810: Chapter renamed by number, blank or (invalid)
### Steps to reproduce

1. Download comic
2. Chapter name is from number
3. Or chapter folder named (invalid) or blank

### Expected behavior

Chapter have "Chapter" attribute on first
Like "Chapter 01"

### Actual behavior

Chapter renamed by number and in the storage became (invalid) if not have a name or blank

### Crash logs

_No response_

### Tachiyomi version

0.14.2

### Android version

Android 11

### Device

Realme 7i

### Other details

Screenshot_2022-12-27-21-48-38-95_afd66b7916d19a60815f56273408b73e
Screenshot_2022-12-27-21-49-07-46_83ab3b20d3fba5c9c905d050db1f8de3

### Acknowledgements

Hipp0x:
fix #8810 

Summary :
I changed the directory name when we download a manga which chapters name only composed of numbers or spaces.
It will appear like "Chapter 5" instead of "5" in the download directory.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
@@ -138,12 +138,33 @@ class DownloadProvider(
      * @param chapterScanlator scanlator of the chapter to query
      */
     fun getChapterDirName(chapterName: String, chapterScanlator: String?): String {
+        val newChapterName = getNewChapterName(chapterName)
         return DiskUtil.buildValidFilename(
             when {
-                chapterScanlator.isNullOrBlank().not() -> "${chapterScanlator}_$chapterName"
-                else -> chapterName
+                chapterScanlator.isNullOrBlank().not() -> "${chapterScanlator}_$newChapterName"
+                else -> newChapterName
             },
         )
+    }
+
+    /**
+     * Return the new name for the chapter (in case it's only a number or blanck)
+     *
+     * @param chapterName the name of the chapter
+     */
+    private fun getNewChapterName(chapterName: String): String {
+        var newChapterName = chapterName
+        try {
+            val chapterToInt = Integer.parseInt(chapterName)
+            if (chapterToInt.toString() == chapterName) {
+                newChapterName = "${R.string.chapter} $chapterName"
+            }
+        } catch (e: NumberFormatException) {
+            if (chapterName.isBlank()) {
+                newChapterName = "${R.string.chapter}"
+            }
+        }
+        return newChapterName
     }
 
     fun isChapterDirNameChanged(oldChapter: Chapter, newChapter: Chapter): Boolean {
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
@@ -138,7 +138,7 @@ class DownloadProvider(
      * @param chapterScanlator scanlator of the chapter to query
      */
     fun getChapterDirName(chapterName: String, chapterScanlator: String?): String {
-        val newChapterName = getNewChapterName(chapterName)
+        val newChapterName = sanitizeChapterName(chapterName)
         return DiskUtil.buildValidFilename(
             when {
                 chapterScanlator.isNullOrBlank().not() -> "${chapterScanlator}_$newChapterName"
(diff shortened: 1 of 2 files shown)


Hipp0x:
Can someone review my PR ?

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
```diff
@@ -138,14 +138,35 @@ class DownloadProvider(
...
+     * Return the new name for the chapter (in case it's only a number or blanck)
+     *
+     * @param chapterName the name of the chapter
+     */
+    private fun getNewChapterName(chapterName: String): String {
+        var newChapterName = chapterName
+        try {
+            val chapterToInt = Integer.parseInt(chapterName)
+            if (chapterToInt.toString() == chapterName) {
+                newChapterName = "${R.string.chapter} $chapterName"
```
arkon, line 160: `R.string.chapter` resolves to some number, not the string. You'd need to do `context.getString(R.string.chapter)`.

That said, there's no point in localizing this, it'd break downloads if you switched the app language of the translation changes for some reason.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
```diff
@@ -138,14 +138,35 @@ class DownloadProvider(
...
 
+    /**
+     * Return the new name for the chapter (in case it's only a number or blanck)
+     *
+     * @param chapterName the name of the chapter
+     */
+    private fun getNewChapterName(chapterName: String): String {
+        var newChapterName = chapterName
+        try {
+            val chapterToInt = Integer.parseInt(chapterName)
```
arkon, line 158: There's a `String.toIntOrNull()` function that'd be useful in these kind of scenarios.

However, I don't think this is useful to do. If the chapter name is already just a number, then the downloaded chapter would already be fine?

Hipp0x: should we just close the issue and this PR if it's not useful?

arkon: The blank name is still applicable.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
Hipp0x: Do you think it's okay if I set a default value like "Chapter", or should I just find another way ?

arkon: That's fine.

Hipp0x: I've update the code

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
```diff
@@ -138,14 +138,28 @@ class DownloadProvider(
...
         )
     }
 
+    /**
+     * Return the new name for the chapter (in case it's empty or blank)
+     *
+     * @param chapterName the name of the chapter
+     */
+    private fun getNewChapterName(chapterName: String): String {
+        return if (chapterName.isEmpty() || chapterName.isBlank()) {
```
arkon, line 156: `isBlank` is a superset of `isEmpty`

i18n/src/main/res/values/strings.xml
```diff
@@ -446,6 +446,7 @@
     <string name="save_chapter_as_cbz">Save as CBZ archive</string>
     <string name="split_tall_images">Split tall images</string>
     <string name="split_tall_images_summary">Improves reader performance</string>
+    <string name="chapter">Chapter</string>
```
arkon, line 449: Remove this.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
```diff
@@ -138,14 +138,28 @@ class DownloadProvider(
...
             },
         )
     }
 
+    /**
+     * Return the new name for the chapter (in case it's empty or blank)
+     *
+     * @param chapterName the name of the chapter
+     */
+    private fun getNewChapterName(chapterName: String): String {
```
arkon, line 155: `sanitizeChapterName` maybe

arkon (changes requested):

arkon:
I'll merge this on the basis that it's a _little_ bit better. It'll only realistically handle the case where there's _one_ chapter that's affected since you'll end up with conflicting names if there's multiple chapters that end up with a blank name normally.

Realistically, that should be fixed "properly" at the source/extension level.

Otherwise, the better solution would be #9238.

PR #9242, opened 2023-03-21
Fix spanish latam is missing in in-app language selection option

AntsyLich:


arkon (changes requested):
Should update `registerLocalesConfigTask` to output the correct locale instead.

arkon:
Superseded by 290efb0283145d81290972991047064c1d905c9c

PR #9289, opened 2023-04-02
Rephrase the wording on Crash Screen

curche:
mainly to avoid users sharing screenshots instead of the file directly
feedback appreciated: if you feel like there's a better rewording, please do share

maybe need to mention the file shows up in notifications somehow? 
or closing the request because it feels unnecessary works too

or could make the Dump Crashlogs Button *also* open up Android Share screen from which you can save to Files/share to Discord?


i18n/src/main/res/values/strings.xml
```diff
@@ -794,7 +794,7 @@
 
     <!-- Crash screen -->
     <string name="crash_screen_title">Whoops!</string>
-    <string name="crash_screen_description">%s ran into an unexpected error. We suggest you screenshot this message, dump the crash logs, and then share it in our support channel on Discord.</string>
+    <string name="crash_screen_description">%s ran into an unexpected error. We recommend you use the button to dumb a crashlog file, and share it to our support channel on Discord.</string>
```
ghostbear, line 797:
```suggestion
    <string name="crash_screen_description">%s ran into an unexpected error. We recommend you dump a crashlog file, and share it in our support channel on Discord.</string>
```

arkon:
We should probably just make the dump buttons both on the crash screen and advanced settings just directly open up a share sheet. There probably isn't much value in having it in a notification that people consistently miss or don't understand how to expand.

curche:
Yea I agree. I'll close this and create an issue instead

PR #9299, opened 2023-04-06
GlobalSearchScreen: Skip result screen when using search intent

ivaniskandar:


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/globalsearch/GlobalSearchScreen.kt
```diff
@@ -27,21 +34,49 @@ class GlobalSearchScreen(
...
+                        if (manga != null) {
+                            navigator.replace(MangaScreen(manga.id, true))
+                        } else {
+                            // Backoff to result screen
+                            showSingleLoadingScreen = false
+                        }
+                    }
+                    is SearchItemResult.Error -> {
+                        context.toast(result.throwable.message)
+                        navigator.pop()
```
arkon, line 59: Should maybe just fall back to the global search here too?

PR #9304, opened 2023-04-07
Add swipe actions for chapters 

d-najd:
Closes: https://github.com/tachiyomiorg/tachiyomi/issues/3630

# Features
   https://github.com/tachiyomiorg/tachiyomi/pull/5739#issuecomment-901511902 was used as a guide for implementing this feature but here is more detailed info
* Configurable swipe actions and swipe threshold, available actions are:
  * Mark as read/unread
  * Bookmark / remove bookmark
  * Download chapter / Cancel Download / Remove chapter 
* Swipe background colors adapt to the theme
* Icons and snackbar messages adapt to the swipe action 
* Appear and drag animations
* Animate elevation on drag

## Possible improvements
* currently the swipe action for downloaded chapter removes the chapter, this leads to user having to redownload the chapter if the user presses the undo button, creating temporary cache for the chapter while the snackbar is being displayed will fix this ~~issue~~ feature

## Notes
* surfaceVariant and primary theme colors are used and have been tested with dark, light and pure black dark mode
* I wasn't sure if I should put the swipe action settings in library or create a new settings category (it isn't really a library feature and creating a new category will take time and it will only contain config for one feature), I asked for opinion/help on discord but didn't receive an answer 😔

# Video Preview
## With reworked corner animation

https://user-images.githubusercontent.com/59766732/230716727-06266f4f-af7d-41b0-ad3a-043ffebc416a.mp4

### With the old animation (not really visible but shows other parts of the feature better)

https://user-images.githubusercontent.com/59766732/230670912-a3b4cff0-4dc7-47c3-96fb-364dd01031a2.mp4 

~~Reason the for draft~~

~~I just realised that material's DismissState (the one that I used to develop the feature) provides ability to see fraction more accurately than material 3  thus allowing me to animate the corner's better (material 3's implementation started accounting for swipe fraction when the item was swiped more than 14.5% which was not accurate enough for what I had in mind)~~


app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
```diff
@@ -135,6 +142,9 @@ fun MangaScreen(
             snackbarHostState = snackbarHostState,
             dateRelativeTime = dateRelativeTime,
             dateFormat = dateFormat,
+            chapterSwipeRightAction = chapterSwipeRightAction,
+            chapterSwipeLeftAction = chapterSwipeLeftAction,
```
arkon, line 146: Consider start/end instead of left/right for LTR/RTL ambiguity.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsLibraryScreen.kt
```diff
@@ -266,6 +267,53 @@ object SettingsLibraryScreen : SearchableSettings {
...
+                    pref = chapterSwipeLeftActionPref,
+                    title = stringResource(R.string.pref_chapter_swipe_left),
+                    entries = mapOf(
+                        LibraryPreferences.ChapterSwipeAction.Bookmark to stringResource(R.string.pref_chapter_swipe_action_bookmark),
+                        LibraryPreferences.ChapterSwipeAction.MarkAsRead to stringResource(R.string.pref_chapter_swipe_action_mark_read),
+                        LibraryPreferences.ChapterSwipeAction.Download to stringResource(R.string.pref_chapter_swipe_action_download_chapter),
+                    ),
+                ),
+                Preference.PreferenceItem.ListPreference(
+                    pref = chapterSwipeThresholdPref,
```
arkon, line 300: Should choose a sensible threshold. Having this be configurable seems overkill.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsLibraryScreen.kt
```diff
@@ -266,6 +267,53 @@ object SettingsLibraryScreen : SearchableSettings {
...
+        val chapterSwipeLeftActionPref = libraryPreferences.swipeLeftAction()
+        val chapterSwipeThresholdPref = libraryPreferences.swipeThreshold()
+
+        return Preference.PreferenceGroup(
+            title = stringResource(R.string.pref_chapter_swipe),
+            preferenceItems = listOf(
+                Preference.PreferenceItem.ListPreference(
+                    pref = chapterSwipeRightActionPref,
+                    title = stringResource(R.string.pref_chapter_swipe_right),
+                    entries = mapOf(
```
arkon, line 284: Should add a disabled option for both the left/right.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreenModel.kt
```diff
@@ -523,6 +528,87 @@ class MangaInfoScreenModel(
...
+                // updated chapter is needed because the state of the chapter item does not get updated
+                val updatedChapterItem = allChapters?.find { it.chapter.id == chapterItem.chapter.id }
+                    ?: throw NullPointerException("Chapter with id ${chapterItem.chapter.id} not found")
+
+                executeChapterSwipeAction(updatedChapterItem, swipeAction)
+            }
+        }
+    }
+
+    private fun chapterUndoSwipeSnackbarMessage(
```
arkon, line 550: Bit of a code smell if the screen model has display logic in it. The composables should be the ones with the strings.

i18n/src/main/res/values/strings.xml
```diff
@@ -273,6 +273,22 @@
     <string name="include">Include: %s</string>
     <string name="exclude">Exclude: %s</string>
 
+    <string name="pref_chapter_swipe">Chapter swipe</string>
```
arkon, line 276:
```suggestion
    <string name="pref_chapter_swipe">Chapter swipe actions</string>
```

i18n/src/main/res/values/strings.xml
```diff
@@ -677,6 +693,14 @@
...
     <string name="no_chapters_error">No chapters found</string>
     <string name="are_you_sure">Are you sure?</string>
+    <string name="snack_chapter_marked_as_read">Marked as read</string>
+    <string name="snack_chapter_marked_as_unread">Marked as unread</string>
+    <string name="snack_chapter_bookmarked">Bookmarked</string>
+    <string name="snack_chapter_bookmark_removed">Removed bookmark</string>
+    <string name="snack_chapter_downloading">Downloading</string>
+    <string name="snack_chapter_deleted">Deleted</string>
+    <string name="snack_chapter_canceled_download">Cancelled download</string>
+    <string name="snack_chapter_download_error">Download Error</string>
```
arkon, line 703: Aren't all of these already obvious from UI changes in the chapter list items?

domain/src/main/java/tachiyomi/domain/library/service/LibraryPreferences.kt
```diff
@@ -118,6 +118,22 @@ class LibraryPreferences(
...
+    fun swipeRightAction() = preferenceStore.getEnum("pref_chapter_swipe_right_action", ChapterSwipeAction.Bookmark)
+
+    fun swipeLeftAction() = preferenceStore.getEnum("pref_chapter_swipe_left_action", ChapterSwipeAction.MarkAsRead)
+
+    fun swipeThreshold() = preferenceStore.getFloat("pref_chapter_swipe_threshold", .2f)
+
+    // endregion
+
+    enum class ChapterSwipeAction {
+        MarkAsRead,
```
arkon, line 132:
```suggestion
        ToggleRead,
```

domain/src/main/java/tachiyomi/domain/library/service/LibraryPreferences.kt
```diff
@@ -118,6 +118,22 @@ class LibraryPreferences(
...
+        Bookmark,
```
arkon, line 133:
```suggestion
        ToggleBookmark,
```

app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
```diff
@@ -47,103 +72,236 @@ fun MangaChapterListItem(
...
-                    ),
-                ) {
-                    if (date != null) {
-                        Text(
-                            text = date,
-                            maxLines = 1,
-                            overflow = TextOverflow.Ellipsis,
+        },
+        dismissContent = {
+            val animateCornerRato = if (dismissState.offset.value != 0f) min(dismissState.progress.fraction / .075f, 1f) else 0f
```
arkon, line 155: Ratio

app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
```diff
@@ -47,103 +72,236 @@ fun MangaChapterListItem(
...
+            }
+            DismissValue.Default -> { }
+        }
+    }
+    SwipeToDismiss(
+        state = dismissState,
+        dismissThresholds = { FractionalThreshold(chapterSwipeThreshold) },
+        background = {
+            var backgroundColor by remember { mutableStateOf(Color.Unspecified) }
+            backgroundColor = when (dismissState.dismissDirection) {
```
arkon, line 115: Not sure I'm really feeling the need for different colors. Would be simpler to just stick with one.

app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
```diff
@@ -47,103 +72,236 @@ fun MangaChapterListItem(
...
+            }
+        }
+        LibraryPreferences.ChapterSwipeAction.Bookmark -> {
+            if (!bookmark) {
+                Icons.Default.Bookmark
+            } else {
+                Icons.Default.BookmarkRemove
+            }
+        }
+        LibraryPreferences.ChapterSwipeAction.Download -> {
```
arkon, line 290: It's not terribly clear what some of these icons actually mean in terms of what it'll do if you swipe. Download/delete are fine, but is `FileDownloadOff` supposed to cancel? I don't know what `ErrorOutline` would do.

app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
d-najd: > Would be simpler to just stick with one.

If you are refering to code complexity the difference will be 1 less line of code so not much.
> Not sure I'm really feeling the need for different colors.

In that case I will go with ```MaterialTheme.colorScheme.primary``` since it looks a little bit more consistent with the content on screen

app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
d-najd: Yes ```FileDownloadOff``` is supposed to be cancel the download, do you have suggestion for better icon to use instead? Oh and I think we will also we will have to replace ```FileDownload`` for consistency if we replace ```FileDownloadOff```.

ErrorOutline is supposed to mean that there was error with the previous download attempt and will display ```Download Error``` on the snackbar but do nothing when the swipe or undo action is called.
On second thought replacing it with download action seems more appropriate, I will do that if its okay with you.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreenModel.kt
d-najd: Oh that is nice to know, I will move it to ```MangaChapterListItem``` then.

That reminds, maybe we should consider removing the context dependency from the screen models because of this.
```
ViewModels shouldn't hold a reference to any Lifecycle-related type.
Don't pass Activity, Fragment, Context or Resources as a dependency.
If something needs a Context in the ViewModel, you should strongly evaluate if that is in the right layer.
```

Source: https://developer.android.com/topic/architecture/recommendations

arkon: Yes, the code smell definitely already exists, it'd be ideal to move that stuff out.

app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
arkon: I think that makes sense. It should just (try to) download if it's not downloaded, cancel if it's downloading, and delete if it's downloaded.

i18n/src/main/res/values/strings.xml
d-najd: > Aren't all of these already obvious from UI changes in the chapter list items?

You are right about that. But there is one problem, since in the current implementation I am displaying a snackbar with undo button we have to display some text as the snackbar's message. 
So we could either replace the snackbar with something else, or use single text for all actions something along the lines of ```Action Executed``` but imo that is a bad idea.

arkon: The download deletion is probably the only one that really needs an undo since the other actions are easily reversible by just doing the action again, but that also applies outside of the context of the swipe actions. It may be more ideal to just leave that out for now and do a followup to have an undo snackbar for deletes separately.

d-najd: Sure thing, by followup you mean a followup pull request?

arkon: Yeah, I just mean as a separate isolated change.

app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
d-najd: Resolved with commit: https://github.com/tachiyomiorg/tachiyomi/pull/9304/commits/5b618073b644609ebc36a3d5be6462a7aff3a98b
Also tested with RTL/LTR layouts and it seems to work fine

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsLibraryScreen.kt
d-najd: The default threshold seems to work fine, I will go with that.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsLibraryScreen.kt
d-najd: The option has been added but the default implementation has "overflow" by default when I disable the item (as shown in the video below).
I could "disable" this feature if you want by offsetting the SwipeToDismiss component by the amount that it has been "overflown" by.

https://user-images.githubusercontent.com/59766732/233198086-e45de918-7e9c-49f4-93c6-4f58aa6ce8ae.mp4

If you are happy with the current implementation feel free to mark the conversation as resolved

arkon: It's not controlled by what `directions` you pass in?

d-najd: kind of, if a direction is not passed in it wont allow you to swipe but the overflow is still kept in.

The api is really weird overall, material3's implementation of the swipe progress value starts to update the swipe progress after it goes over 14.5%, this forced me to use material's implementation.
The progress goes automatically to 100% when you overswipe (even if the direction is disabled like above) which is one of the reasons I disable the background when the direction is disabled the other being it looks weirder that way.

And worst of all if you manage to swipe the item while the snap behaviour behaviour is being executed (which is fairly easy if you know how) the item gets stuck between dismissed and default state for some reason (I tried printing log and it prints progress 0% then 100% then 0% and so on). On screen this looks like the item is stuck dismissed, from what I understand this is caused because the snapTo method is suspend method and this is the reason why the pull request is not on hold (I have a working solution but it can be improved).

I suggest that I make a separate SwipeToDismiss class which accounts for these inconsistencies so that it could be reused or use an alternate library if anyone knows of one.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsLibraryScreen.kt
```diff
@@ -216,4 +217,46 @@ object SettingsLibraryScreen : SearchableSettings {
...
+    ): Preference.PreferenceGroup {
+        val chapterSwipeEndEnabledPref = libraryPreferences.swipeEndEnabled()
+        val chapterSwipeStartEnabledPref = libraryPreferences.swipeStartEnabled()
+        val chapterSwipeEndActionPref = libraryPreferences.swipeEndAction()
+        val chapterSwipeStartActionPref = libraryPreferences.swipeStartAction()
+
+        return Preference.PreferenceGroup(
+            title = stringResource(R.string.pref_chapter_swipe),
+            preferenceItems = listOf(
+                Preference.PreferenceItem.SwitchPreference(
```
arkon, line 233: I think it'd be simpler if you just had an option in the action lists as "Disabled" so you'd end up with 2 prefs instead of 4.

d-najd: I am dumb....

i18n/src/main/res/values/strings.xml
```diff
@@ -273,6 +273,16 @@
...
 
+    <string name="pref_chapter_swipe">Chapter swipe</string>
+    <string name="pref_chapter_swipe_end_enabled">Swipe to end</string>
+    <string name="pref_chapter_swipe_start_enabled">Swipe to start</string>
+    <string name="pref_chapter_swipe_end">Swipe to end action</string>
+    <string name="pref_chapter_swipe_start">Swipe to start action</string>
+    <string name="pref_chapter_swipe_action_bookmark">Bookmark</string>
+    <string name="pref_chapter_swipe_action_mark_read">Mark as read</string>
+    <string name="pref_chapter_swipe_action_download_chapter">Download chapter</string>
+    <string name="pref_chapter_swipe_action_disabled">Disabled</string>
```
arkon, line 284: There's already a generic string `disabled`

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsLibraryScreen.kt
```diff
@@ -216,4 +217,38 @@ object SettingsLibraryScreen : SearchableSettings {
...
+            title = stringResource(R.string.pref_chapter_swipe),
+            preferenceItems = listOf(
+                Preference.PreferenceItem.ListPreference(
+                    pref = chapterSwipeEndActionPref,
+                    title = stringResource(R.string.pref_chapter_swipe_end),
+                    entries = mapOf(
+                        LibraryPreferences.ChapterSwipeAction.ToggleBookmark to stringResource(R.string.pref_chapter_swipe_action_bookmark),
+                        LibraryPreferences.ChapterSwipeAction.ToggleRead to stringResource(R.string.pref_chapter_swipe_action_mark_read),
+                        LibraryPreferences.ChapterSwipeAction.Download to stringResource(R.string.pref_chapter_swipe_action_download_chapter),
+                        LibraryPreferences.ChapterSwipeAction.Disabled to stringResource(R.string.pref_chapter_swipe_action_disabled),
```
arkon, line 238: Put this as the first option, similar to things like the global update/download schedule settings.

i18n/src/main/res/values/strings.xml
arkon: Yeah let's just use the existing ones even if they're longer. They're used elsewhere too so if we do change it, we'll change it everywhere.

PR #9326, opened 2023-04-11
Add Tachiyomi Cross Device Sync Feature

KaiserBh:
Hello Tachiyomi team,

I've added a new sync feature to Tachiyomi that allows users to sync their manga library, reading history, tracking, categories, and chapters across devices. The server-side implementation for this feature is still a work in progress and can be found on the [develop branch](https://github.com/SyncYomi/SyncYomi/tree/develop) of the SyncYomi repository.

The syncing functionality is operational and ready for testing. The server is designed to be self-hosted, allowing users to set it up for their friends and family. An OpenAPI documentation is available for generating tokens/API keys, which act as unique user identifiers.

Please note that the data is not encrypted, so server administrators can access the stored information. The current implementation supports localhost and reverse proxy configurations, with plans to add HTTPS support in the future. The web UI is under active development and will eventually allow token/API key generation.

As I'm relatively new to Kotlin and Android development, I welcome any feedback and suggestions for improvement. I would also appreciate pointers on handling local state changes, specifically for updating the read status of individual chapters.

The current implementation should prevent conflicts during syncing as it uses Unix epoch timestamps in milliseconds for tracking changes.

I look forward to hearing your thoughts and collaborating on this feature. Thank you for considering my contribution!

Best regards,
KaiserBh

edit: I forgot about the images.
sync_settings1

sync_settings2

sync_settings_3

Also demo below:
[Device 1 marks manga as read](https://cdn.ayo.icu/demo_device_1.mp4)
[Device 2 requests sync from server then updates manga as unread](https://cdn.ayo.icu/demo_device_2.mp4)
[Device 1 requests sync from server (marks berserk as unread and restore missing mangas)](https://cdn.ayo.icu/demo_device_3.mp4)
[Device 2 requests sync from server (fresh install)](https://cdn.ayo.icu/demo_device_4.mp4)


AntsyLich:
Hello! Great work. I noticed that you are Using `java.time.Instant` in some places. But the `Instant` class was added In Android 8 while the app supports Android 6.

ghostbear:
@AntsyLich we ha the desugaring library so Instant should be fine on older Android versions

KaiserBh:
> Hello! Great work. I noticed that you are Using `java.time.Instant` in some places. But the `Instant` class was added In Android 8 while the app supports Android 6.

@AntsyLich Thank you for pointing this out! I wasn't aware that the `java.time.Instant` class was introduced in Android 8, while the app supports Android 6.

ghostbear mentioned that the desugaring library might resolve this issue, so let me know if I need to change anything.

ghostbear:
> ghostbear mentioned that the desugaring library might resolve this issue, so let me know if I need to change anything.

It should already be enabled in all submodules

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncDataJob.kt
```diff
@@ -0,0 +1,88 @@
...
+            SyncManager(context).syncData()
+            notifier.showSyncComplete()
+            Result.success()
+        } catch (e: Exception) {
+            logcat(LogPriority.ERROR, e)
+            notifier.showSyncError(e.message)
+            Result.failure()
+        } finally {
+            context.cancelNotification(Notifications.ID_BACKUP_PROGRESS)
+        }
```
d-najd, line 44: From what I understand if "finally" is not called here the notification "ID_BACKUP_PROGRESS" will not be canceled and will stay up indefinetely, if that is the case then you may consider replacing it with try-with-resources, or if that is not possible cleaner.

Reasons being: There is no guarantee that finalizers will run at all, they are deprecated (as of java 9), are pain to debug (since they don't throw exceptions) among other things.

Source: Effective java Third Edition item 8-9.
And here website containing summary (I won't provide link of the full book for obvious reasons) scroll to item 8 and 9: https://ahdak.github.io/blog/effective-java-part-1/

KaiserBh: Hi 👋 

Thank you for reviewing my code and providing valuable feedback. I understand your concern about the use of "finally" and the potential issues with the notification not being cancelled.

I appreciate your suggestion of using try-with-resources or a cleaner as an alternative approach. However, I noticed that in other parts of the codebase, a similar pattern has been followed. Here are two examples:

[Backup restore job](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupRestoreJob.kt#L36-L51)

[Backup create job](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupCreateJob.kt#L44-L54)

I followed the existing practice in the codebase to maintain consistency. However, I also acknowledge that we should strive for improvement and consider adopting better practices if they provide better guarantees and are less prone to issues.

Please let me know your thoughts on this, and I look forward to discussing it further.

ghostbear: `try-with-resources` only works when the resource implements AutoClosable

`Object.finalize()` is deprecated `try-finally` is not

`finally` is always executed (for full detail of how it works read [this](https://docs.oracle.com/javase/specs/jls/se20/html/jls-14.html#jls-14.20.2)). However, it can create resource leaks if you need to close multiple resources and one of those throws an exception, for example

```kt
val input: FileInputStream    = null;
val output: FileOutputStream  = null;    
try {
  input  = FileInputStream(file1);
  output = FileOutputStream(file2);
  // If completes normally or abruptly it will go to the finally block
} catch (e: Exception) {
  // If completes normally, abruptly, or doesn't cover all exception it will go to the finally block
} finally {
    if (output != null) output.close(); // If this throw an exception
    if (input  != null) input.close(); // Then this is a resource leak
}
```

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    @Composable
+    fun getSyncHostPref(syncPreferences: SyncPreferences): Preference.PreferenceItem.EditTextPreference {
+        return Preference.PreferenceItem.EditTextPreference(
+            title = stringResource(R.string.pref_sync_host),
+            subtitle = stringResource(R.string.pref_sync_host_summ),
+            onValueChanged = { newValue ->
+                syncPreferences.syncHost().set(newValue)
+                true
+            },
```
arkon, line 51: I don't think this is actually necessary, the `pref` should already be set internally.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+import eu.kanade.tachiyomi.data.sync.SyncDataJob
+import eu.kanade.tachiyomi.util.system.toast
+import kotlinx.coroutines.launch
+import tachiyomi.domain.sync.SyncPreferences
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+
+object SettingsSyncScreen : SearchableSettings {
```
arkon, line 25: I'd consider putting this in the same screen as manual backups/restores, but probably with a new name (or Sync encompasses both?).

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+    fun getSyncHostPref(syncPreferences: SyncPreferences): Preference.PreferenceItem.EditTextPreference {
+        return Preference.PreferenceItem.EditTextPreference(
+            title = stringResource(R.string.pref_sync_host),
+            subtitle = stringResource(R.string.pref_sync_host_summ),
+            onValueChanged = { newValue ->
+                syncPreferences.syncHost().set(newValue)
+                true
+            },
+            enabled = true,
+            icon = null,
```
arkon, line 53: Can omit the defaults.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+
+        return listOf(
+            getSyncHostPref(syncPreferences = syncPreferences),
+            getSyncAPIKeyPref(syncPreferences = syncPreferences),
+            getSyncNowPref(syncPreferences = syncPreferences),
+        )
+    }
+
+    @Composable
+    fun getSyncHostPref(syncPreferences: SyncPreferences): Preference.PreferenceItem.EditTextPreference {
```
arkon, line 44: Don't really need to split these basic preferences up into smaller functions, it'd be a bit more straightforward to just list them out within `getPreferences`.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+                true
+            },
+            enabled = true,
+            icon = null,
+            pref = syncPreferences.syncAPIKey(),
+        )
+    }
+
+    @Composable
+    fun getSyncNowPref(syncPreferences: SyncPreferences): Preference.PreferenceGroup {
```
arkon, line 74: Make sure to make these `private` if applicable.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+        val scope = rememberCoroutineScope()
+        val showDialog = remember { mutableStateOf(false) }
+        val context = LocalContext.current
+        val lastSync = syncPreferences.syncLastSync().get()
+        val formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss").withZone(ZoneId.systemDefault())
```
arkon, line 79: Consider using `DateUtils.getRelativeTimeSpanString` like we do elsewhere in the app.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+            preferenceItems = listOf(
+                Preference.PreferenceItem.TextPreference(
+                    title = stringResource(R.string.pref_sync_now),
+                    subtitle = stringResource(R.string.pref_sync_now_subtitle),
+                    onClick = {
+                        showDialog.value = true
+                    },
+                ),
+                Preference.PreferenceItem.InfoPreference("Last sync at: $formattedLastSync"),
+                Preference.PreferenceItem.InfoPreference("Last local change at: $formattedLastLocalChange"),
```
arkon, line 109: Could combine the two, and make sure they're localized.

I haven't looked at the server implementation yet, but what happens when there's a significant gap between these and there's local changes on the other device too?

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+            pref = syncPreferences.syncAPIKey(),
+        )
+    }
+
+    @Composable
+    fun getSyncNowPref(syncPreferences: SyncPreferences): Preference.PreferenceGroup {
+        val scope = rememberCoroutineScope()
+        val showDialog = remember { mutableStateOf(false) }
+        val context = LocalContext.current
+        val lastSync = syncPreferences.syncLastSync().get()
```
arkon, line 78: Does this actually get visually updated when this changes? Should we be collecting this as state?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncDataJob.kt
```diff
@@ -0,0 +1,88 @@
...
+        return try {
+            SyncManager(context).syncData()
+            notifier.showSyncComplete()
+            Result.success()
+        } catch (e: Exception) {
+            logcat(LogPriority.ERROR, e)
+            notifier.showSyncError(e.message)
+            Result.failure()
+        } finally {
+            context.cancelNotification(Notifications.ID_BACKUP_PROGRESS)
```
arkon, line 43: `ID_SYNC_PROGRESS`?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncDataJob.kt
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    override suspend fun getForegroundInfo(): ForegroundInfo {
+        return ForegroundInfo(
+            Notifications.ID_BACKUP_PROGRESS,
```
arkon, line 49: `ID_SYNC_PROGRESS`?

app/src/main/java/eu/kanade/tachiyomi/data/sync/models/Sync.kt
```diff
@@ -0,0 +1,241 @@
...
+import kotlinx.serialization.Serializable
+import tachiyomi.domain.category.model.Category
+import tachiyomi.domain.chapter.model.Chapter
+import tachiyomi.domain.manga.model.Manga
+import tachiyomi.domain.track.model.Track
+
+@Serializable
+data class SyncStatus(
+    @SerialName("last_synced") val lastSynced: String? = null,
+    val last_synced_epoch: Long? = null,
```
arkon, line 15: Use consistent casing in the API?

app/src/main/java/eu/kanade/tachiyomi/data/sync/models/Sync.kt
```diff
@@ -0,0 +1,241 @@
...
+    val status: String? = null,
+)
+
+@Serializable
+data class SyncManga(
+    val source: Long? = null,
+    val url: String? = null,
+    @SerialName("favourite") val favorite: Boolean? = null,
```
arkon, line 23: Just use American English spelling in the API for consistency?

app/src/main/java/eu/kanade/tachiyomi/data/sync/models/Sync.kt
```diff
@@ -0,0 +1,241 @@
...
+
+@Serializable
+data class SyncStatus(
+    @SerialName("last_synced") val lastSynced: String? = null,
+    val last_synced_epoch: Long? = null,
+    val status: String? = null,
+)
+
+@Serializable
+data class SyncManga(
```
arkon, line 20: It bothers me that we're adding yet another model for the same data. Could we reuse the existing backup models?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -459,6 +462,8 @@ class LibraryScreenModel(
                     read = read,
                 )
             }
+            // Update last local changes time to prevent conflicts for syncing
+            syncPreferences.syncLastLocalUpdate().set(Instant.now())
```
arkon, line 466: I wonder if it'd be better to just track auto-incremented (not sure if that's the right word, but I mean automatically set the current time on modification) `last_modified` timestamps in the relevant tables and query those. Might help with dealing with conflicts too.

Also prevents changes needing to made in all of the relevant screen models (although even with this current implementation, I'd point you to doing it in the interactors instead).

domain/src/main/java/tachiyomi/domain/sync/SyncPreferences.kt
```diff
@@ -0,0 +1,19 @@
...
+    private val preferenceStore: PreferenceStore,
+) {
+    fun syncHost() = preferenceStore.getString("sync_host", "https://sync.tachiyomi.org")
+    fun syncAPIKey() = preferenceStore.getString("sync_api_key", "")
+    fun syncLastLocalUpdate() = preferenceStore.getInstant("sync_last_local_update", Instant.EPOCH)
+    fun syncLastSync() = preferenceStore.getInstant("sync_last_sync", Instant.EPOCH)
+
+    fun syncInterval() = preferenceStore.getInt("sync_interval", 0)
+
+    fun deviceName() = preferenceStore.getString("device_name", android.os.Build.MANUFACTURER + android.os.Build.PRODUCT)
```
arkon, line 16: Would probably need to make this customizable in case someone ends up using multiple of the same model.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+import tachiyomi.domain.manga.interactor.GetFavorites
+import tachiyomi.domain.manga.model.Manga
+import tachiyomi.domain.source.service.SourceManager
+import tachiyomi.domain.sync.SyncPreferences
+import tachiyomi.domain.track.model.Track
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+import java.lang.StrictMath.max
+import java.time.Instant
+import java.util.*
```
arkon, line 54: Prefer explicit imports.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+ */
+class SyncManager(
+    private val context: Context,
+) {
+    private val handler: DatabaseHandler = Injekt.get()
+    private val sourceManager: SourceManager = Injekt.get()
+    private val syncPreferences: SyncPreferences = Injekt.get()
+    private val libraryPreferences: LibraryPreferences = Injekt.get()
+    private val getCategories: GetCategories = Injekt.get()
+    private val getFavorites: GetFavorites = Injekt.get()
```
arkon, line 70: Can stick these in the constructor.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+     * The data is sent to the server by calling `sendSyncRequest()`.
+     *
+     * This function should be called when you want to synchronize local library
+     * data with a remote server. It is designed to be used with a coroutine
+     * as it is marked as a suspend function.
+     */
+    suspend fun syncData() {
+        val favorites = getFavorites.await()
+
+        Log.i("SyncManager", "Mangas to sync: $favorites")
```
arkon, line 89: Use `logcat(LogPriority.DEBUG) { ... }`

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+
+        // Manga data
```
arkon, line 91: These comments are kind of redundant.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+     * SyncManga objects with the retrieved data. The SyncManga objects are then added
+     * to a mutable list called syncMangaList, which is returned at the end.
+     *
+     * @param mangaList A list of Manga objects to be converted into SyncManga objects.
+     * @return A list of SyncManga objects created from the given mangaList.
+     */
+    private suspend fun mangaList(mangaList: List<Manga>): List<SyncManga> {
+        val syncMangaList = mutableListOf<SyncManga>()
+
+        for (m in mangaList) {
```
arkon, line 121: It'd probably be ideal if we can extract the logic that's happening for this and backups (i.e. something that just produces a data object with all of the relevant data), which realistically shouldn't differ.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+            .filterNot(Category::isSystemCategory)
+            .map(syncCategoryMapper)
+    }
+
+    /**
+     * Sends a sync request to the server containing manga, categories, and extensions data.
+     *
+     * This function constructs the required objects (SyncStatus, Data, SyncDevice, and SData)
+     * using the given input parameters, converts the SData object to a JSON string,
+     * and sends the JSON data to the server using the provided URL and API key.
```
arkon, line 223: Should probably try to compress that data before sending it off, it can be potentially really big.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+     * and retrieves the device name and update device ID if is 0 from the preferences.
+     *
+     *
+     * @param url The server URL to send the sync data to.
+     * @param apiKey The API key for authentication.
+     * @param jsonData The JSON string containing the sync data.
+     */
+    private suspend fun sendSyncData(url: String, apiKey: String, jsonData: String) {
+        val client = OkHttpClient()
+        val json = Json { ignoreUnknownKeys = true }
```
arkon, line 281: Inject these?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+
+        val mediaType = "application/json".toMediaTypeOrNull()
+        val body =
+            jsonData.toRequestBody(
+                mediaType,
+            )
+        val request = Request.Builder()
```
arkon, line 288: There's an existing `POST` convenience method.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+            .addHeader("X-API-Token", apiKey)
+            .build()
+
+        client.newCall(request).execute().use { response ->
+            val responseBody = response.body.string()
+
+            if (response.isSuccessful) {
+                val syncDataResponse: SData = json.decodeFromString(responseBody)
+                logcat(
+                    LogPriority.INFO,
```
arkon, line 301: Should probably be `DEBUG`

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+                    val chapters = mangaList.flatMap { it.chapters ?: emptyList() }
+
+                    // Restore system categories first.
+                    syncDataResponse.data?.categories?.let { restoreSyncCategories(it) }
+
+                    // restore manga / everything.
+                    restoreManga(mangaList, history, tracking, syncCategories, chapters)
+
+                    // Trigger a global update after restoring everything: this should fix the UI prev/next chapter buttons.
+                    LibraryUpdateJob.startNow(context)
```
arkon, line 321: I'm kind of iffy on this, since it'll increase traffic.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+
+                    syncPreferences.syncLastSync().set(Instant.now())
+                    // if device id is 0, update it
+                    if (syncPreferences.deviceID().get() == 0) {
+                        syncDataResponse.device?.id?.let { syncPreferences.deviceID().set(it) }
+                    }
+                } else {
+                    // no update is required as local device is up to date
+                    if (syncDataResponse.update_required == false) {
```
arkon, line 330: Is this not a given if it's already in this `else` block?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,846 @@
...
+
+    /**
+     * Restores the categories from a list of SyncCategory objects. If a category with the same
+     * name already exists in the database, the existing category's ID is assigned to the
+     * SyncCategory's corresponding category. Otherwise, a new category is created and added to
+     * the database.
+     *
+     * @param syncCategories A list of SyncCategory objects containing the category data to sync.
+     */
+    private suspend fun restoreSyncCategories(syncCategories: List<SyncCategory>) {
```
arkon, line 445: A lot of this restore logic seems similar to or the same as the backup restore logic.

arkon (commented):
General thoughts:

- This is an impressive effort! Thanks to tackling it.
- The implementation could probably be a bit more generic to not be tied to the hosted service but rather just something that'll return a set of data and the necessary timestamps (such that the same data could go to some place like Google Drive, for example, considering self-hosting something is going to be a non-starter for many users)
- A lot of the actual backup/restore logic seems very similar to the existing logic, so we should try to dedupe that

KaiserBh:
Hi @arkon Thanks for reviewing the code, I will go through it tomorrow and see what I need to change or improve, I did skim through it.

In the meantime, if you or anyone else would like to test it, I have created a working server/frontend that can be used to generate keys and try out the syncing feature.

[SyncYomi](https://github.com/SyncYomi/SyncYomi/releases)

Setting it up is quite straightforward. Simply download the binary compatible with your OS and run it. For those who prefer, there is also a Docker version available.

**Note:** If you are running the binary without changing the config host to 0.0.0.0, it will not be available on the local network. Make sure to update this setting. The Docker version does this by default. (A config.toml file is generated once the binary is run/started for the first time.)

Once the server is up, navigate to localhost:8282, and you should be greeted with an onboarding process. Create an account and then go to Settings -> API Keys -> Add API Key.

Copy the API key and add it to the Tachiyomi app. From there, everything should be straightforward. If you have two or more devices, try syncing.

image

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
KaiserBh: The server implementation is designed to handle the synchronization process based on the last local change timestamp provided by the device. The server checks whether the last local change on the device is newer than the last synced data on the server, and if it is, the server updates its data accordingly. Conversely, if the device's last local change is older, the server sends the latest data back to the client to update their local storage.

Regarding the last sync timestamp, it is primarily used for user information purposes. It helps the user identify which device has outdated data so they can take appropriate action if needed.

In the case of a significant gap between the last sync and last local change, the server would rely on the last local change timestamp to decide whether to update its data or send the latest data back to the client. This approach helps ensure that the most recent changes are always synced across devices, even if there are local changes on another device.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
KaiserBh: > Does this actually get visually updated when this changes? Should we be collecting this as state?

It does get updated but only when the user navigate back and then go back into the sync screen, as am not sure how am meant to update the state async so if you have a pointer for me I can update it accordingly.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
KaiserBh: hmm yea that would be great actually, but again if you could guide me or direct me to it then I will update them,
I do agree that it's better if we can just put this in one place rather than in random places. That being said when we get the data from last_modified what exactly update those? because we don't want the user to accidentally update the "last local update" just because they go to setting and update those etc, I was thinking at the moment since it's library sync:

Update on read
Update on new manga added to favourites
Update on category change
Update on manga removed from favourites

That's what I can think of at the moment but do let me know what do you think?

domain/src/main/java/tachiyomi/domain/sync/SyncPreferences.kt
KaiserBh: so would you suggest we let the user set their own device name? since the server have notification feature currently (discord, telegram and notifiarr) supported and I plan to use their device name to show sync status and when was last synced etc. If so we can just add it in the pref settings.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
KaiserBh: If you have a better solution on fixing the prev and next button I am all ears, we could omit that and let the user just update it manually (refresh each manga they are reading manually drag top to bottom) but yea otherwise there will be issue with next/prev button, it will get flipped at the same time it won't know that there is actually previous chapter etc.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
arkon: You can look at examples like

```kt
val defaultCategory by libraryPreferences.defaultCategory().collectAsState()
```

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
arkon: That explanation doesn't really convince me that gaps between syncs are handled nicely, but maybe it's realistically an edge case if someone doesn't read on multiple devices sort-of simultaneously.

It might be worth thinking more about though. I don't think treating timestamps for the entire piece of data is going to be scalable for handling more complex conflicts.

domain/src/main/java/tachiyomi/domain/sync/SyncPreferences.kt
arkon: It'd probably be a better user experience to allow it be user customizable, yes.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
arkon: It's an existing issue when backups are restored too. I'd prefer if we tackle that separately.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
arkon: > hmm yea that would be great actually, but again if you could guide me or direct me to it then I will update them

I'm actually not super sure how it'd work either. Potentially a SQL trigger on alter?

> That being said when we get the data from last_modified what exactly update those?

Presumably some form of automatic trigger, but if not then something akin to what you have but centralized in the specific interactors instead.

> because we don't want the user to accidentally update the "last local update" just because they go to setting and update those etc

I don't understand what you mean by this.

>  I was thinking at the moment since it's library sync

Should probably include any change in read status (including last read page), bookmark status, or favorite status. That should implicitly cover history changes.

Otherwise, category changes and library updates too?

KaiserBh: No worries, I will have a look at it and try to centralized it, maybe like you said SQL trigger on alter might do the job,

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
KaiserBh: This is done! gzip the data before sending it now.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
KaiserBh: > It's an existing issue when backups are restored too. I'd prefer if we tackle that separately.

hmm okay I didn't know that, I guess the issue is with prev and next button logic? because it really shouldn't break have anyone had a crack at it? or maybe found the issue? assuming not since it's still an issue.

rickymohk:
Hello, just a random guy coming across this great PR and want to drop some idea. Do you think it is possible to move the database logic to the app side, and connect to any FTP server without needing to run a dedicated server program?

KaiserBh:
> Hello, just a random guy coming across this great PR and want to drop some idea. Do you think it is possible to move the database logic to the app side, and connect to any FTP server without needing to run a dedicated server program?

Hello, random guy! Yes, it is possible to move the database logic to the app side and connect to an FTP server instead of running a dedicated server program. However, there are a few trade-offs to consider, such as the potential for increased complexity.

At the moment, I am focusing on ironing out any issues with the existing implementation and syncing with a server and Google Drive. Expanding the functionality to support FTP server connections and moving database logic to the app side can be considered for future development once the current priorities have been addressed.

Hello @arkon I've made some changes to the syncing implementation, and now it accounts for granular sync. Granular sync allows updating only specific data elements that have changed, which is particularly helpful in situations where there's a significant gap between sync times and local changes on both devices.

As a side note, I could use your help with finding the appropriate place in the code for this snippet:
```kotlin
 // Update last local changes time to prevent conflicts for syncing
syncPreferences.syncLastLocalUpdate().set(Instant.now())
```

This should trigger on events we discussed earlier, such as updating last read page, bookmark status, favorite status, adding new manga to favorites, category changes, and manga removal from favorites.

Could you please point me in the right direction for where to place this code in the app? Your guidance would be greatly appreciated.

KaiserBh:
https://user-images.githubusercontent.com/41852205/233789834-a4ce0bbd-b2b3-4926-971d-1f7bb225033f.mp4

Gdrive added, works fine but I broke self host (server side code issue) will fix it tomorrow.

arkon:
Import to note, we don't want to use the GMS libraries because of their licensing. We'd want to actually implement Google Drive integration using their REST API.

KaiserBh:
> Import to note, we don't want to use the GMS libraries because of their licensing. We'd want to actually implement Google Drive integration using their REST API.

No worries, the logic will be similar or same mostly, I will fix it tomorrow.

zhongfly:
Can this feature be considered to support webdav?
Some file hosting services provide webdav.
And it's easy for people to create their own webdav-enabled network drive.

arkon:
I don't know if I'm just out of the loop, but I've only ever seen Chinese users ask for WebDAV and I've never seen it used here. I'm not sure if it's worth supporting a niche like that if most things don't support it.

KaiserBh:
Ok I think it's done, categories for Gdrive kinda funky need some tests to be sure but the ordering can be messed up. So if anyone re-ordering the categories it will just change the "name" of the category basically, my advice for now is just rename their categories instead of re-ordering but not sure at the moment I will take a look at it tomorrow (maybe).

@arkon let me know if this is better, I removed the gms so it should be okay now, Added browser because google doesn't allow/blocks webview user-agent to sign in. Feel free to review it and let me know any changes that needs to be done.

mutsumi63:
> I don't know if I'm just out of the loop, but I've only ever seen Chinese users ask for WebDAV and I've never seen it used here. I'm not sure if it's worth supporting a niche like that if most things don't support it.

' only ever seen Chinese users ask for WebDAV '
Because we can't connect to gd, we need some self-host services, or use services like webdav or S3. 

SyncYomi is a good solution,
But as you said, self-hosting service is too hard for many users.

Webdav or S3 is actually a good solution for Sync, and many open source software use it as an alternative.

KokaKiwi:
> SyncYomi is a good solution, But as you said, self-hosting service is too hard for many users.
> 
> Webdav or S3 is actually a good solution for Sync, and many open source software use it as an alternative.

Also there's many Nextcloud instances out there to use easily as WebDAV backends

Mohamed-Ali-K:
Hey, Congratulations on adding the new sync feature to Tachiyomi, this will certainly be a useful addition to the community. I wanted to share my solution for Cross Device Sync which is using [Syncthing](https://syncthing.net/). [Syncthing](https://syncthing.net/). is a free and open-source tool that synchronizes files between two or more computers or smartphones, and it works on any platform in real-time. You can  [download](https://syncthing.net/downloads/) it from their website and read about it more.

Personally, I keep a fresh backup of my Tachiyomi library across all my devices using [Syncthing](https://syncthing.net/). It's a one-time setup and I don't have to worry about files being in sync anymore. Syncthing begins the synchronization on file change, so it's effortless.
example :
my galaxy Note : 
- Not20 Ulta 5G
my google pixel : 
google Pixle

However, I still have to go to Tachiyomi and restore the backup every time I change devices. With Syncthing, we can get an idea of how to cross-device sync without depending on a third-party server or cloud provider oy any google API token or whatever . It's a peer-to-peer file sync solution, and you can refer to this [YouTube](https://www.youtube.com/watch?v=PSx-BkMOPF4) video to know more about it, I find it very handy.

OncePunchedMan:
That's one way to do it, sure. However you run into the same limitation as for the server the author created: self-hosting.

binyaminyblatt:
I'm really excited for this feature

KaiserBh:
Hi @arkon,

I made the changes and dedup as much as I could.  These changes aim to address the issues and feedback that were previously brought up. I believe that the PR is now essentially ready for merging, but I would like to hear your thoughts on whether there is anything further that needs improvement or clarification.

Also if anyone want to test it for gdrive feel free to dm me on discord and send me your email address so I can add you as a test user. [discord](https://discord.gg/PW5h5GKn)

reitenth:
is this close to be implemented? thanks.

mystic-man-rhino:
> is this close to be implemented? thanks.

If interested in testing cross device sync you can join unofficial fork SyncYomi https://discord.gg/bMjGCyw6

erpbridge:
A review has been requested from @arkon and @ghostbear. Review is currently pending, and a workflow is awaiting approval.

app/src/debug/assets/client_secrets.json
```diff
@@ -0,0 +1 @@
+{"web":{"client_id":"1046609911130-dfclj7n32mt4ar3kifuva7oubgh1al5n.apps.googleusercontent.com","project_id":"tachiyomi","auth_uri":"https://accounts.google.com/o/oauth2/auth","token_uri":"https://oauth2.googleapis.com/token","auth_provider_x509_cert_url":"https://www.googleapis.com/oauth2/v1/certs","client_secret":"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","redirect_uris":["http://127.0.0.1:53682/auth"]}}
```
arkon, line 1: This file seems the same as the next one?

gradle/libs.versions.toml
```diff
@@ -90,6 +90,11 @@ voyager-transitions = { module = "cafe.adriel.voyager:voyager-transitions", vers
 
 kotlinter = "org.jmailen.gradle:kotlinter-gradle:3.13.0"
 
+google-api-services-drive = "com.google.apis:google-api-services-drive:v3-rev197-1.25.0"
+google-api-client-oauth = "com.google.oauth-client:google-oauth-client:1.34.1"
+google-api-client-oauth-jetty = "com.google.oauth-client:google-oauth-client-jetty:1.34.1"
+browser = "androidx.browser:browser:1.5.0"
```
arkon, line 96: We actually removed use of custom tabs for tracker OAuth auth in favor of just normal browser usage.

gradle/libs.versions.toml
```diff
@@ -90,6 +90,11 @@ voyager-transitions = { module = "cafe.adriel.voyager:voyager-transitions", vers
 
 kotlinter = "org.jmailen.gradle:kotlinter-gradle:3.13.0"
 
+google-api-services-drive = "com.google.apis:google-api-services-drive:v3-rev197-1.25.0"
+google-api-client-oauth = "com.google.oauth-client:google-oauth-client:1.34.1"
+google-api-client-oauth-jetty = "com.google.oauth-client:google-oauth-client-jetty:1.34.1"
```
arkon, line 95: Is this actually being used?

app/src/main/java/eu/kanade/tachiyomi/data/notification/Notifications.kt
```diff
@@ -59,6 +59,16 @@ object Notifications {
     const val ID_BACKUP_COMPLETE = -502
     const val ID_RESTORE_COMPLETE = -504
 
+    /**
+     * Notification channel and ids used by the sync system.
```
arkon, line 63: I'd consider grouping this/renaming the backup/restore group, or just relying on the existing channels entirely.

app/src/main/assets/client_secrets.json
```diff
@@ -0,0 +1 @@
+{"web":{"client_id":"1046609911130-dfclj7n32mt4ar3kifuva7oubgh1al5n.apps.googleusercontent.com","project_id":"tachiyomi","auth_uri":"https://accounts.google.com/o/oauth2/auth","token_uri":"https://oauth2.googleapis.com/token","auth_provider_x509_cert_url":"https://www.googleapis.com/oauth2/v1/certs","client_secret":"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","redirect_uris":["http://127.0.0.1:53682/auth"]}}
```
arkon, line 1: Would it be problematic if this is reused by other forks?

app/src/main/java/eu/kanade/tachiyomi/data/sync/OAuthCallbackServer.kt
```diff
@@ -0,0 +1,138 @@
...
+import java.io.BufferedWriter
+import java.io.IOException
+import java.io.InputStreamReader
+import java.io.OutputStreamWriter
+import java.net.ServerSocket
+import java.net.Socket
+import java.net.URI
+
+/**
+ * OAuthCallbackServer is a lightweight HTTP server that listens for an authorization code from Google OAuth.
```
arkon, line 15: This is a bit weird. Any reason why we couldn't just have an activity like the tracker OAuth flows?

app/src/debug/assets/client_secrets.json
KaiserBh: yea one of them is for standard release and the other one is debug release, both have same value but just different location.

gradle/libs.versions.toml
KaiserBh: Oh do you want me to just use normal browser usage? I can change that and remove the custom tab.

gradle/libs.versions.toml
KaiserBh: Jetty is not being used, I have removed it.

app/src/main/java/eu/kanade/tachiyomi/data/notification/Notifications.kt
KaiserBh: No worries, using existing channel now.

app/src/main/assets/client_secrets.json
KaiserBh: That shouldn't be the case, it's basically the same thing as what [rclone ](https://github.com/rclone/rclone/blob/4f8dab8bcec40f34f02074549747437ac17b0bef/backend/drive/drive.go#L59) uses, which they encrypt a very good practice (I suggest you do the same I plan to revoke the client_secrets once it's merged), otherwise it could allow someone to impersonate your application and potentially gain unauthorized access to user data. So that being said I suggest other fork use their own client_secrets, otherwise it will just show Tachiyomi, instead of TachiyomiJ2K or TachiyomiSY. 

Everything else works the same as long we get access to the users gdrive.

app/src/main/java/eu/kanade/tachiyomi/data/sync/OAuthCallbackServer.kt
KaiserBh: I don't think it's possible, I haven't seen how the Tracker OAuth flow work. But essentially we need authorization code that google will send, hence why using local http server. In the past we could use OOB method [oob-migration](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration) but it has been deprecated.

gradle/libs.versions.toml
arkon: Yes, that'd be ideal. It doesn't seem that useful to bring it another dependency.

app/src/debug/assets/client_secrets.json
arkon: It should just be using the other one if this doesn't exist, shouldn't it? I don't see why we'd need the duplicate file.

app/src/main/java/eu/kanade/tachiyomi/data/sync/OAuthCallbackServer.kt
arkon: You can refer to how it's done for MAL for example:

- [Activity](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/tachiyomi/ui/setting/track/MyAnimeListLoginActivity.kt)
- [Manifest entry](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/AndroidManifest.xml#L141)

It notably uses a URI like `tachiyomi://myanimelist-auth`, which is registered as the callback. Logging in then redirects to `tachiyomi://myanimelist-auth?code=...`, which redirects to that activity, extracts the `code`, does something with it, then goes back to the `MainActivity`.

KaiserBh: Ohh great, I will have a look at that and do that instead.

gradle/libs.versions.toml
KaiserBh: No worries, will change this as well.

app/src/debug/assets/client_secrets.json
KaiserBh: Not sure exactly, when I tried using the debug one it crashes because it couldn't find the file, it's somewhere in google drive sync, or we can just hard code it? something like what rclone does encrypt it and hard code?

app/src/main/java/eu/kanade/tachiyomi/data/sync/OAuthCallbackServer.kt
KaiserBh: Cool I think I know why we need local http server to receive the code, it's because google doesn't allow redirect to a custom URI @arkon.

image

AntsyLich: You are setting custom redirect uri for a "web" client type. You need to setup a "android" client type and set redirect uri for that

app/src/debug/assets/client_secrets.json
AntsyLich: Put it under `app/src/main/assets/client_secrets.json`

app/src/main/java/eu/kanade/tachiyomi/data/sync/OAuthCallbackServer.kt
KaiserBh: > You are setting custom redirect uri for a "web" client type. You need to setup a "android" client type and set redirect uri for that. More info [here](https://developers.google.com/identity/protocols/oauth2#1.-obtain-oauth-2.0-credentials-from-the-dynamic_data.setvar.console_name-.).

Okay thank you, fixed and I didn't know that google actually uses the SHA1 and package name for the redirect uri. 
So when creating the client in the future be sure to match the package name with the activity.

In this case I have set it to:

```
  <data android:scheme="eu.kanade.google.oauth" />
```

More information on redirect uri can be found [here](https://developers.google.com/identity/protocols/oauth2/native-app#request-parameter-redirect_uri).

app/src/debug/assets/client_secrets.json
KaiserBh: > Put it under `app/src/main/assets/client_secrets.json`

Done, it seems to be working now.

playful1510:
Am I correct in understanding that this has not been implemented yet? Just curious because I'm interested in trying out the sync functionality, but the debug/preview APKs are gone from a previous post.

KaiserBh:
> Am I correct in understanding that this has not been implemented yet? Just curious because I'm interested in trying out the sync functionality, but the debug/preview APKs are gone from a previous post.

Hi, they are old build. I wouldn't recommend using those. Anyway, if you are interested in testing the syncing, feel free to join 
[Discord server](https://discord.gg/X966brhFgN) The new build is in the announcement channel.

KaiserBh:
We are reworking parts of the code in my [downstream fork](https://github.com/KaiserBh/tachiyomi) and splitting it into parts, closing this one in the time being.

Issue #9328, opened 2023-04-12
Option to skip duplicate chapters when downloading chapters


TraceLosu:
### Describe your suggested feature

When the "Skip Duplicate Chapters" option is enabled (or another separate option if need be), the downloader should also skip duplicate chapters. This should only apply to when downloading multiple chapters at once.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.14.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

PR #9329, opened 2023-04-12
Allow option to skip duplicate chapters when downloading multiple chapters

TraceLosu:
Closes: #9328 

### Feature Description
Adds a new option in settings that skips duplicate chapters when downloading multiple chapters. For example, if this setting is enabled, when downloading the next 5 chapters, it will download the next 5 unique chapters, and skips over duplicate chapters if there is any.

The logic for this is the same as the current "Skip Duplicate Chapter" option for the reader. It grabs all chapters that are valid for downloading (unless the chapter list is already picked by the user) and groups them by chapter number. Then we loop through each chapter number and grab the first chapter with the same scanlator group as the first chapter in the list (or the first chapter, if none is found).

Demo: https://streamable.com/zelz6h


arkon:
I'm not entirely sure just selecting the first scanlator will be desirable. I wonder if we should prioritize choosing scanlators (#8803) instead?

Issue #8912, opened 2023-01-14
Skip duplicate chapters in download ahead feature


zbue:
### Describe your suggested feature

_Why?_
Because download chapter ahead stops working when duplicate chapter is skipped

### Other details

Note that it only happens when `Skip duplicate chapters` is enabled in reader settings

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.14.3](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

PR #9334, opened 2023-04-14
fix: skip duplicate chapters on download ahead if option to skip duplicates is enabled

TraceLosu:
Closes: #8912

### Issue Description
In `ReaderViewModel#downloadNextChapters()`, `getNextChapters` is used to get the next unread chapters from the current chapter that's being read. However, `getNextChapters` returns *all* next unread chapters, including duplicates. This causes issues when the reader is set to skip duplicate chapters.

When the downloader receives the chapters to download from `getNextChapters`, said chapters could contain duplicates. The chapters that are duplicates will be skipped over by the reader (but still be downloaded). Because a chapter was skipped, the download ahead feature might stop working, because the next chapter might not have been downloaded by the feature. 

### Solution
To fix this, the resulting list from `getNextChapters` is filtered, removing duplicates the same way how the reader does it.


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderViewModel.kt
```diff
@@ -456,8 +456,19 @@ class ReaderViewModel(
             )
             if (!isNextChapterDownloaded) return@launchIO
 
-            val chaptersToDownload = getNextChapters.await(manga.id, nextChapter.id!!)
-                .take(amount)
+            val chaptersToDownload = getNextChapters.await(manga.id, nextChapter.id!!).run {
+                if (readerPreferences.skipDupe().get()) {
+                    this.groupBy { it.chapterNumber }
```
arkon, line 461: Considering this logic is duplicated with what's in the initialization of `chapterList`, consider making a function to do this so the logic doesn't drift in the future.

PR #9362, opened 2023-04-18
Download video in different regulation and add local subtitle button

Imon24:
The beta version (r-6145) is pretty good. But sometimes it crashes. I think it will be fixed in the upcoming stable version.🥰
   However, there are no options available to download an anime episode in different regulations. It always download the episode in 1080p.. But sometimes it is not possible to download an episode in such a high regulation for people like me who don't have a Wi-Fi connection. So please add this option. And most of the website (Almost all) don’t allow us to watch dubbed anime with subtitle so please add an option to add local subtitle from our mobile storage. Lastly, thanks from the core of my heart for making an app like aniyomi, it's like a heaven for anime and manga lover like me. ❤️


alessandrojean:
This isn't the Aniyomi repository, you're opening your requests in the wrong format (should be issues) and in the wrong repository.

Issue #9216, opened 2023-03-16
Move chapter scanlator on chapter transition to a new line


noaione:
### Describe your suggested feature

It certainly nice having the scanlator name when moving inbetween chapters, but the scanlator name being on the same line and separated by dot is confusing to read as shown in the image below.

I suggest moving it to a separate line making it less confusing to read.

### Other details

![original](https://user-images.githubusercontent.com/34302902/225634007-c28ee1fa-eef2-49b4-b658-0bd6bd5addce.png)

How it looked originally, the right image just makes me think that the scanlator name is the chapter title.

![modified](https://user-images.githubusercontent.com/34302902/225634232-4d8a22a9-5c0c-46ef-9518-a7c419344fab.png)

Mocked image, less confusing compared to the current implementation.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.14.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

Dalekin9:
I'm going to be working on this

PR #9371, opened 2023-04-20
Changed the string so that the scanlator appears on a new line

Dalekin9:
Changed both transition page's strings so that scanlators are on another line

 _close_ #9216 

| Screenshot_20230420_135139_eu kanade tachiyomi |Screenshot_20230420_135218_eu kanade tachiyomiDlkn debug |
| ------- | ------- |


AntsyLich:
The output is no different than when a chapter's name is too big to fit in a line. How are user supposed to differentiate the next line is the scanlator and not just part of the chapter's name

Dalekin9:
Then having both the dot seperator and a new line would be the best solution ? ( without remodeling the UI )

specterflare:
Personally, I'd prefer italics
233419028-4d45ed7c-a8e7-4b27-90e9-3c53554a3374

arkon:
I'll just have a differently styled line in #9373.

PR #9384, opened 2023-04-24
Double tap zoom toggle

### LINKED ISSUE #4145: [Feature Request] Ability to disable zooming
Currently, the slightest accidental gesture seems to keep causing unwanted zoom in/out which can get very annoying. Just a simple way to disable zoom gestures for people who don't want to use zoom can be very helpful. 

## What/Requirements

An on/off switch for zoom related gestures in the preferences>reader>navigation section, which can be used to disable pinch and double taps. 

Any other simple way of disabling zoom would also work. For example the way the Neko app allows disabling zoom for webtoons only)
IMG_20201217_154028

Flat:
Implements a toggle that allows users to disable double tap zoom including QuickScaling for webtoons. Partially resolves #4145


i18n/src/main/res/values/strings.xml
```diff
@@ -317,6 +317,7 @@
     <string name="pref_page_rotate">Rotate wide pages to fit</string>
     <string name="pref_page_rotate_invert">Flip orientation of rotated wide pages</string>
     <string name="pref_long_strip_split">Split tall images (BETA)</string>
+    <string name="pref_double_tap_zoom">Enable double tap zoom</string>
```
arkon, line 320:
```suggestion
    <string name="pref_double_tap_zoom">Double tap to zoom</string>
```

PR #9398, opened 2023-04-26
Add interval data layer

quangkieu:
- Part of https://github.com/tachiyomiorg/tachiyomi/pull/5734


data/src/main/sqldelight/tachiyomi/migrations/24.sqm
```diff
@@ -0,0 +1 @@
+ALTER TABLE mangas ADD COLUMN calculate_interval INTEGER DEFAULT 0 NOT NULL;
```
AntsyLich, line 1: Add a new line here

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt
```diff
@@ -526,6 +527,8 @@ class BackupManager(
                 thumbnailUrl = manga.thumbnailUrl,
                 favorite = manga.favorite.toLong(),
                 lastUpdate = manga.lastUpdate,
+                nextUpdate = 0L,
+                calculateInterval = 0L,
```
AntsyLich, line 531:
```suggestion
                nextUpdate = null,
                calculateInterval = null,
```
So it doesn't overwrite db value

AntsyLich (changes requested):

data/src/main/sqldelight/tachiyomi/data/mangas.sq
```diff
@@ -16,6 +16,7 @@ CREATE TABLE mangas(
     favorite INTEGER AS Boolean NOT NULL,
     last_update INTEGER AS Long,
     next_update INTEGER AS Long,
+    calculate_interval INTEGER DEFAULT 0 NOT NULL ,
```
arkon, line 19:
```suggestion
    calculate_interval INTEGER DEFAULT 0 NOT NULL,
```

data/src/main/java/tachiyomi/data/manga/MangaMapper.kt
```diff
@@ -4,13 +4,15 @@ import eu.kanade.tachiyomi.source.model.UpdateStrategy
...
-    { id, source, url, artist, author, description, genre, title, status, thumbnailUrl, favorite, lastUpdate, _, initialized, viewerFlags, chapterFlags, coverLastModified, dateAdded, updateStrategy ->
+val mangaMapper: (Long, Long, String, String?, String?, String?, List<String>?, String, Long, String?, Boolean, Long?, Long?, Long, Boolean, Long, Long, Long, Long, UpdateStrategy) -> Manga =
+    { id, source, url, artist, author, description, genre, title, status, thumbnailUrl, favorite, lastUpdate, nextUpdate, calculateInterval, initialized, viewerFlags, chapterFlags, coverLastModified, dateAdded, updateStrategy ->
         Manga(
             id = id,
             source = source,
             favorite = favorite,
             lastUpdate = lastUpdate ?: 0,
+            nextUpdate = nextUpdate ?: 0,
+            calculateInterval = calculateInterval.toInt(),
```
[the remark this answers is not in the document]
arkon: I'm guessing the intention is that the interval will never be that big, but it's probably simpler to just leave it as a `Long`.

arkon (commented):
Make sure to bump up the `versionCode` in `app/build.gradle.kts` too to prevent downgrades.

data/src/main/java/tachiyomi/data/manga/MangaMapper.kt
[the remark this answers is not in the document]
quangkieu: Yes, I did not want to change column type so I just use Long for Inverval

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt
quangkieu: Updated

data/src/main/sqldelight/tachiyomi/migrations/24.sqm
quangkieu: Updated

data/src/main/sqldelight/tachiyomi/data/mangas.sq
quangkieu: Updated

quangkieu:
> Make sure to bump up the `versionCode` in `app/build.gradle.kts` too to prevent downgrades.

I would bump before the whole feature merge

arkon:
We can do multiple version bumps as things get merged. Attempting to downgrade the database if you install an older version leads to more headaches than it's worth.

domain/src/main/java/tachiyomi/domain/manga/model/Manga.kt
```diff
@@ -8,6 +8,8 @@ data class Manga(
     val source: Long,
     val favorite: Boolean,
     val lastUpdate: Long,
+    val nextUpdate: Long,
+    val calculateInterval: Int,
```
AntsyLich, line 12: Tell me if I'm wrong but I think @arkon meant to change this to Long

quangkieu: Sure I could convert to Long but I see it would hit the front ui more to convert to Int for pluralStringResource and some calculations now use Long operation instead of Int.
Right now only the read/write to db need 1 time covert to Int

quangkieu:
Thanks arkon

PR #9422, opened 2023-04-29
Move GitHub Release/App Update logic to data

ghostbear:


domain/src/main/java/tachiyomi/domain/release/interactor/GetApplicationRelease.kt
```diff
@@ -0,0 +1,79 @@
...
+        // Check if latest version is different from current version
+        val isNewVersion = isNewVersion(arguments.isPreview, arguments.commitCount, arguments.versionName, release.version)
+        return when {
+            isNewVersion && arguments.isThirdParty -> Result.ThirdPartyInstallation
+            isNewVersion -> Result.NewUpdate(release)
+            else -> Result.NoNewUpdate
+        }
+    }
+
+    private fun isNewVersion(isPreview: Boolean, commitCount: Int, versionName: String, versionTag: String): Boolean {
```
arkon, line 40: Might be a good time to slap in some unit tests for this.

domain/build.gradle.kts
```diff
@@ -22,4 +22,5 @@ dependencies {
     api(libs.sqldelight.android.paging)
 
     testImplementation(libs.bundles.test)
+    testImplementation(libs.mockk)
```
arkon, line 25: Just add it to the `test` bundle?

domain/src/test/java/tachiyomi/domain/release/interactor/GetApplicationReleaseTest.kt
```diff
@@ -0,0 +1,169 @@
...
+            GetApplicationRelease.Arguments(
+                isPreview = false,
+                isThirdParty = true,
+                commitCount = 0,
+                versionName = "v1.0.0",
+                repository = "test",
+            ),
+        )
+
+        assertEquals(GetApplicationRelease.Result.ThirdPartyInstallation, result)
```
arkon, line 55: Use kotest assertions, e.g. `result shouldBe GetApplicationRelease.Result.ThirdPartyInstallation`

domain/src/test/java/tachiyomi/domain/release/interactor/GetApplicationReleaseTest.kt
```diff
@@ -0,0 +1,169 @@
...
+        val preferenceStore = mockk<PreferenceStore>()
+        preference = mockk()
+        every { preferenceStore.getLong(any(), any()) } returns preference
+        releaseService = mockk()
+
+        getApplicationRelease = GetApplicationRelease(releaseService, preferenceStore)
+    }
+
+    @Test
+    fun `When has update but is third party expect third party installation`() = runBlocking {
```
arkon, line 34: Use `runTest`?

domain/src/test/java/tachiyomi/domain/release/interactor/GetApplicationReleaseTest.kt
```diff
@@ -0,0 +1,169 @@
...
+            GetApplicationRelease.Arguments(
+                isPreview = true,
+                isThirdParty = false,
+                commitCount = 1000,
+                versionName = "",
+                repository = "test",
+            ),
+        )
+
+        assertDoesNotThrow { result as GetApplicationRelease.Result.NewUpdate }
```
arkon, line 82: Could just assert that is that. The test would fail if it throws anyway.

domain/src/test/java/tachiyomi/domain/release/interactor/GetApplicationReleaseTest.kt
```diff
@@ -0,0 +1,169 @@
...
+                repository = "test",
+            ),
+        )
+
+        assertDoesNotThrow { result as GetApplicationRelease.Result.NoNewUpdate }
+        assertEquals(GetApplicationRelease.Result.NoNewUpdate, result)
+    }
+
+    @Test
+    fun `When has lasted check is before three months`() = runBlocking {
```
arkon, line 143: I'm not sure I understand this test based on the name

ghostbear: Covers this case
```
if (arguments.forceCheck.not() && now.isBefore(Instant.ofEpochMilli(lastChecked.get()).plus(3, ChronoUnit.DAYS))) {
   return Result.NoNewUpdate
}
```

arkon: Is 3 months supposed to be some arbitrary amount larger than 3 days? Would also make sense to check that `releaseService` is never called then.

ghostbear: Ah yeah no it should be if now is before 3 days, not 3 months

PR #9430, opened 2023-04-30
Metadata filter for LocalSource

### LINKED ISSUE #5892: Filter local manga for details in . json

For your local manga, there should be a proper filter function for the details, where you can search for any combination of author, tags, status, etc. saved in the .json.

This Filter function should also include the ability to search for the pages- / chapter-length

### Other details

Atm, it's only possible to use the top-bar search function for local manga. This doesn't allow to search for anything that is not the filename.

### Acknowledgements

Shamicen:
partially resolves #5892

Include manga title, manga genre, author, artist and status from metadata files in LocalSource search

status can be searched by either searching for the `string value` (e.g. `ongoing`) or `status:SMangaValue` (e.g. `status:1` for ongoing)

The other fields can be searched by searching for the `exact string` with the exception of title, which only requires a `substring`


Shamicen:
I changed the `fetchSearchManga` function to automatically load Metadata when opening `LocalSource`, because searching metadata only used to work for manga which have been opened once before, which in my opinion kind of defeats the purpose.

I had to lift the part of the code that fetched chapters out of `forEach` into `SManga.create` since the manga metadata changed in `forEach` was lost after exiting the block. 

I also parallelized the creation of manga which includes fetching/creating their covers and metadata. 

This resulted in a performance improvement of ≈ 25% from opening LocalSource to Manga being displayed.

I ran 3 test each here are the results: 
non parallel: 117 seconds, 120 seconds, 111 seconds   `average: 118 seconds`
parallel: 97 seconds, 90 seconds, 87 seconds   `average: 91 seconds `

I used Java's `parallelStream` method for this but their may be a way to use Kotlins `launch` function instead

Shamicen:
Changed implementation from search to using filters.

**Other changes**

- Local Source Manga will only be loaded once and stay constant until app restart, instead of loading all manga from disk every time Local Source is opened  (already existing Local Source manga can still be updated individually)
- Local Source manga that already have populated metadata fields in the database will be populated from the database instead of re-loading metadata from disk

**Other things to note**
- the `OrderBy.Latest` filter is still broken since it's implementation requires to either add it twice or choose between `OrderBy.Latest` and `OrderBy.Popular`

Shamicen:
This is what the filters look like:

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreen.kt
```diff
@@ -295,7 +296,7 @@ class MangaScreen(
         }
 
         val previousController = navigator.items[navigator.size - 2]
-        if (previousController is BrowseSourceScreen && source is HttpSource) {
+        if (previousController is BrowseSourceScreen && (source is HttpSource || source is LocalSource)) {
```
arkon, line 299: Can just change the check to `CatalogueSource`

i18n/src/main/res/values/strings.xml
```diff
@@ -636,6 +636,13 @@
     <string name="chapter_not_found">Chapter not found</string>
     <string name="local_invalid_format">Invalid chapter format</string>
     <string name="local_filter_order_by">Order by</string>
+    <string name="local_filter_genres">Genres</string>
+    <string name="local_filter_authors">Authors</string>
+    <string name="local_filter_artists">Artists</string>
+    <string name="local_filter_status">Status</string>
```
arkon, line 642: Use more general names for these, they'll probably get reused later for other things.

source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/model/SManga.kt
```diff
@@ -72,6 +74,19 @@ interface SManga : Serializable {
         it.initialized = initialized
     }
 
+    fun getStatusString(): String {
```
arkon, line 77: Should refactor `ComicInfoPublishingStatus`.

source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
```diff
@@ -282,7 +606,65 @@ actual class LocalSource(
     }
 
     // Filters
```
arkon, line 608: Better off moving this stuff into a separate file.

source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
```diff
@@ -282,7 +606,65 @@ actual class LocalSource(
...
+        val authors = localManga.mapNotNull { it.author?.split(",") }
+            .flatMap { it.map { author -> author.trim() } }.toSet()
+
+        val artists = localManga.mapNotNull { it.artist?.split(",") }
+            .flatMap { it.map { artist -> artist.trim() } }.toSet()
+
+        val filters = try {
+            mutableListOf<Filter<*>>(
+                OrderBy.Popular(context),
+                LoadRandomMangaPage(context),
```
arkon, line 637: Not a fan of this, feels like feature bloat.

source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
```diff
@@ -282,7 +606,65 @@ actual class LocalSource(
...
+                        context.getString(R.string.licensed),
+                        context.getString(R.string.publishing_finished),
+                        context.getString(R.string.cancelled),
+                        context.getString(R.string.on_hiatus),
+                        context.getString(R.string.unknown),
+                    ).map { StatusFilter(it) },
+                ),
+                Separator(),
+                TextSearchHeader(context),
+                GenreTextSearch(context),
```
arkon, line 656: I don't see the point in having groups _and_ text search. Not having the groups also seems like it'd avoid needing to load metadata upfront.

source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
```diff
@@ -65,40 +85,196 @@ actual class LocalSource(
...
-            // Filter out files that are hidden and is not a folder
-            .filter { it.isDirectory && !it.name.startsWith('.') }
-            .distinctBy { it.name }
-            .filter { // Filter by query or last modified
-                if (lastModifiedLimit == 0L) {
-                    it.name.contains(query, ignoreCase = true)
-                } else {
-                    it.lastModified() >= lastModifiedLimit
-                }
+        loadMangaForPage(page)
```
arkon, line 165: We should ideally avoid metadata loading if we're not actually searching (i.e. just loading popular/latest, which currently just call this).

Shamicen: I personally think that we should load metadata for manga when we first load them for the following reasons:

1. Users with large local libraries  would need to wait a very long time when filtering for the first time, as metadata would need to be fetched for each page that has already loaded. Instead adding a little bit of load time to the comparatively short time it takes to load a single manga page would be the better user experience in my opinion, especially since the metadata load time will shorten significantly after the metadata files have been copied (first ever load) and will be even shorter if the user does not clear their `LocalSource Database` as the metadata will be loaded from the database instead of being loaded from `ComicInfo` files.  
2. Loading metadata directly means that manga show up with their correct titles instead of just their directory names allowing users more freedom when organizing their libraries / spares them the hassle of having to rename all of their manga directories if their existing library does not use manga titles as directory names
3. It could be confusing to some users if their manga randomly change names after they press the filter button.

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreen.kt
Shamicen: I changed it

i18n/src/main/res/values/strings.xml
Shamicen: I changed the names to genres, authors etc.

source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/model/SManga.kt
Shamicen: Changed it to use `ComicInfoPublishingStatus`

source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
Shamicen: Moved the classes to designated `Filter` file

source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
Shamicen: Removed `LoadRandomMangaPage`

source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
Shamicen: I think we should keep the groups because some users

1. may not know what genre options they actually have without them
2. may not know the correct spelling of every author/artist name

PR #9472, opened 2023-05-07
Don't copy ComicInfo files to temp dir

Shamicen:
Filter out `ComicInfo.xml` when copying image files to temp directory in `ZipPageLoader`


arkon:
Irrelevant after revert in 6f59c6c6bb12f1ef123abd418001c7a86d43ef7a

PR #9511, opened 2023-05-15
Minor Downloader cleanup

Two-Ai:
* Inline `completeDownload` into the top-level Downloader Subscription (`initializeSubscription`)
* Inline post-download steps in `isDownloadSuccessful` into `downloadChapter`
* Consolidate queueState updates in `removeFromQueue` overloads
  * Avoids unnecessary sequential updates to the queue `MutableStateFlow`


app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -665,14 +656,26 @@ class Downloader(
         }
     }
 
-    fun removeFromQueue(chapters: List<Chapter>) {
-        chapters.forEach { chapter ->
-            queueState.value.find { it.chapter.id == chapter.id }?.let { removeFromQueue(it) }
+    private inline fun removeFromQueueByPredicate(predicate: (Download) -> Boolean) {
```
arkon, line 659: `removeFromQueueIf` would probably sound better, but you can lump that in with a followup PR.

PR #9514, opened 2023-05-18
Populate the ComicInfo Number field with chapter numbers

Shamicen:
Populate the Number field of ComicInfo files with chapter numbers.  

According to the [Official ComicInfo documentation](https://anansi-project.github.io/docs/comicinfo/documentation), [Kavita's documentation](https://wiki.kavitareader.com/en/archive/guides/metadata) and [Komga's documentation](https://komga.org/guides/scan-analysis-refresh.html#import-metadata-for-cbr-cbz-containing-a-comicinfo-xml-file) (only tested with Komga)

### DIFF
--- a/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
+++ b/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
@@ -98,6 +98,7 @@ fun Manga.hasCustomCover(coverCache: CoverCache = Injekt.get()): Boolean {
 fun getComicInfo(manga: Manga, chapter: Chapter, chapterUrl: String) = ComicInfo(
     title = ComicInfo.Title(chapter.name),
     series = ComicInfo.Series(manga.title),
+    number = ComicInfo.Number(chapter.chapterNumber.toString()),
     web = ComicInfo.Web(chapterUrl),
     summary = manga.description?.let { ComicInfo.Summary(it) },
     writer = manga.author?.let { ComicInfo.Writer(it) },
--- a/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
+++ b/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
@@ -98,7 +98,7 @@ fun Manga.hasCustomCover(coverCache: CoverCache = Injekt.get()): Boolean {
 fun getComicInfo(manga: Manga, chapter: Chapter, chapterUrl: String) = ComicInfo(
     title = ComicInfo.Title(chapter.name),
     series = ComicInfo.Series(manga.title),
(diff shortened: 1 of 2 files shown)


ThePromidius:
I'll only note that in Kavita this can lead to wrong results.
Because in Kavita, volumes are handled in a different way and we wanted 2 way tracking, the number for volumes is kind of encoded and won't match.
```kt
                    // Is a single-file volume
                    // We encode the chapter number to support tracking
                    name = "Volume ${volume.number}"
                    chapter_number = volume.number.toFloat() / 10000
```
I'm just pointing this out.
I once asked if comicinfo could be handled by extensions way back when it was being implemented and got a no iirc. So i'm not really sure what can be done in this case.

Shamicen:
> I once asked if comicinfo could be handled by extensions way back when it was being implemented and got a no iirc. So i'm not really sure what can be done in this case.

This shouldn't be an Issue since Tachiyomi itself never reads the Number field and even if it did in the future then it would probably only be relevant to LocalSource. This change is just to correctly populate the ComicInfo.xml file, when a chapter is downloaded from any source, with as much information currently available as possible. I only linked the Kavitas and Komga documentations to show that this is the way the field is used by consumers in the real world. 

The only way this could cause problems for Kavitas is if you use Tachiyomi as a download client for Kavitas and copy the chapter archive files from your Tachiyomi download folder to your server and they somehow could not deal with a file that supplies a chapter number without a Volume number, which is completely in spec for ComicInfo files btw. But that would be an Issue that should be raised to Kavitas.

Please correct me if I misunderstood anything I don't really know how Kavitas works.

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -98,6 +98,7 @@ fun Manga.hasCustomCover(coverCache: CoverCache = Injekt.get()): Boolean {
 fun getComicInfo(manga: Manga, chapter: Chapter, chapterUrl: String) = ComicInfo(
     title = ComicInfo.Title(chapter.name),
     series = ComicInfo.Series(manga.title),
+    number = ComicInfo.Number(chapter.chapterNumber.toString()),
```
arkon, line 101: Might want to account for it being `-1` sometimes?

ThePromidius:
I was just saying.
I don't think it'll case much of an issue since people that download from Kavita will mostly just want the offline read experience and to not actually move those files around. Either way I decided it'd be good to bring it out just in case.

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
Shamicen: I added a check for negative chapter numbers

Issue #6260, opened 2021-11-22
Ability to remove manga from trackers


THElegend5:
### Describe your suggested feature

Sometimes you might track the wrong manga and you have to go all the way to the site and remove it manually ( also sometimes you aren't even able to tell which one you tracked untill some time later ) , it would be nice if there would be something like a a pop up saying do you want to also remove this from mal/anilist when you remove the tracking for that title.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.12.3](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

zaghdaneh:
I started working on this so if possible please assign to me.
Should the dialog be like in the linked picture ? or something less obstructive?
![image](https://github.com/tachiyomiorg/tachiyomi/assets/46049558/ae8fdb5a-3a63-4756-81a9-3dee210b6d08)

AntsyLich:
We don't do assignments. Also it'll better if it's in the middle of the screen like an normal dialog

arkon:
That looks fine, it's consistent with the date removal.

PR #9535, opened 2023-05-22
Remove manga from trackers

zaghdaneh:
Closes : [#6260](https://github.com/tachiyomiorg/tachiyomi/issues/6260)

### Feature description
When tracking is removed for a manga, user is asked if they want to also remove tracking from the tracker service.  If the accept, a delete request is sent to the respective service.

- The api request was implemented for Myanimelist/Anilist/Kitsu/Shikimori/Mangaupdates.
- The api request was not implemented for Bangumi as the corresponding api request wasn't found, instead a flag was added to not show the dialogue for services that do not support the delete request.

Draft pull request for review and while double-checking if it really is not possible to send a delete request to bangumi.

### Images

--> The delete confirmation dialogue.


scb261:
Imo it's a bit confusing and the user may think (from quick glance w/o properly reading) that it's a confirmation to remove it locally.

Maybe insted make it a tick 'Also remove tracking from AniList'.

Could also rename the 'Cancel' button to something like 'Remove only locally', but I'm not sure if it's too long.

arkon:
Bangumi indeed seems to be missing an API for it: https://github.com/bangumi/api/issues/15

app/src/main/java/eu/kanade/tachiyomi/data/track/TrackService.kt
```diff
@@ -74,6 +77,7 @@ abstract class TrackService(val id: Long) {
 
     abstract suspend fun update(track: Track, didReadChapter: Boolean = false): Track
 
+    abstract suspend fun delete(track: Track): Track
```
arkon, line 80: Probably makes more sense to move this to a separate interface rather than using the flag.

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
```diff
@@ -167,6 +167,16 @@ class Anilist(id: Long) : TrackService(id) {
         return api.updateLibManga(track)
     }
 
+    override suspend fun delete(track: Track): Track {
+        if (track.library_id == null || track.library_id!! == 0L) {
+            val libManga = api.findLibManga(track, getUsername().toInt())
+                ?: throw Exception("$track not found on user library")
```
arkon, line 173: Probably safe to just swallow an error in this case.

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt
```diff
@@ -110,6 +111,39 @@ class AnilistApi(val client: OkHttpClient, interceptor: AnilistInterceptor) {
         }
     }
 
+    suspend fun deleteLibManga(track: Track): Track {
+        logcat { "sending request ! deleting : ${track.id} with mangaid ${track.manga_id}" }
```
arkon, line 115: Remember to clean up these logs.

app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/BangumiApi.kt
```diff
@@ -70,6 +70,18 @@ class BangumiApi(private val client: OkHttpClient, interceptor: BangumiIntercept
         }
     }
 
+    suspend fun deleteLibManga(track: Track): Track {
```
arkon, line 73: Bangumi doesn't seem to have an open API for it, so we can just remove this.

app/src/main/java/eu/kanade/tachiyomi/data/track/kitsu/KitsuApi.kt
```diff
@@ -123,6 +123,26 @@ class KitsuApi(private val client: OkHttpClient, interceptor: KitsuInterceptor)
...
+            authClient.newCall(
+                Request.Builder()
+                    .url("${baseUrl}library-entries/${track.media_id}")
+                    .headers(
+                        headersOf(
+                            "Content-Type",
+                            "application/vnd.api+json",
+                        ),
+                    )
+                    .delete()
```
arkon, line 137: Use the `DELETE` helper instead.

app/src/main/java/eu/kanade/tachiyomi/data/track/myanimelist/MyAnimeListApi.kt
```diff
@@ -158,6 +158,20 @@ class MyAnimeListApi(private val client: OkHttpClient, interceptor: MyAnimeListI
         }
     }
 
+    suspend fun deleteItem(track: Track): Track {
+        return withIOContext {
+            val request = Request.Builder()
+                .url(mangaUrl(track.media_id).toString())
+                .delete()
```
arkon, line 165: Use the `DELETE` helper instead.

app/src/main/java/eu/kanade/tachiyomi/data/track/shikimori/ShikimoriApi.kt
```diff
@@ -35,28 +36,45 @@ class ShikimoriApi(private val client: OkHttpClient, interceptor: ShikimoriInter
...
                 }
+                authClient.newCall(
+                    POST(
+                        "$apiUrl/v2/user_rates",
+                        body = payload.toString().toRequestBody(jsonMime),
+                    ),
+                ).awaitSuccess()
+                    .parseAs<JsonObject>()
+                    .let {
+                        track.library_id = it["id"]!!.jsonPrimitive.long // save id of the entry for possible future delete request
```
arkon, line 58: I guess this would be an issue for existing entries?

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackInfoDialog.kt
```diff
@@ -692,3 +703,77 @@ data class TrackServiceSearchScreen(
...
+            modifier = Modifier.windowInsetsPadding(WindowInsets.systemBars),
+            icon = {
+                Icon(
+                    imageVector = Icons.Default.Delete,
+                    contentDescription = null,
+                )
+            },
+            title = {
+                Text(
+                    text = "Remove ${stringResource(sm.getServiceNameRes())} tracking",
```
arkon, line 731: All UI strings should be localized (i.e add it to the `values/strings.xml` file).

app/src/main/java/eu/kanade/tachiyomi/data/track/shikimori/ShikimoriApi.kt
zaghdaneh: you mean for already existing entries before the feature is implemented? I can address it by retrieving the manga ID if it's missing before the delete request.

arkon:
The alternative UI seems a bit less confusing.

I'd also reword:
- "Do you really want to remove ... tracking?" with "This will remove the tracking locally." 
- "Also remove manga entry from ..." with "Also remove from ..."
- "Yes" with "OK"
- "No" with "Cancel"

zaghdaneh:
Thanks for the feedback. I updated it as recommended.
For bangumi, I simply hide the checkbox since it doesn't have a delete request.

## Image

i18n/src/main/res/values/strings.xml
```diff
@@ -754,6 +755,9 @@
     <string name="track_remove_date_conf_title">Remove date?</string>
     <string name="track_remove_start_date_conf_text">This will remove your previously selected start date from %s</string>
     <string name="track_remove_finish_date_conf_text">This will remove your previously selected finish date from %s</string>
+    <string name="track_delete_title">Remove %s tracking</string>
```
stevenyomi, line 758:
```suggestion
    <string name="track_delete_title">Remove %s tracking?</string>
```

i18n/src/main/res/values/strings.xml
```diff
@@ -754,6 +755,9 @@
     <string name="track_remove_date_conf_title">Remove date?</string>
     <string name="track_remove_start_date_conf_text">This will remove your previously selected start date from %s</string>
     <string name="track_remove_finish_date_conf_text">This will remove your previously selected finish date from %s</string>
+    <string name="track_delete_title">Remove %s tracking?</string>
+    <string name="track_delete_text">This will remove the tracking locally.</string>
+    <string name="track_delete_remote_text">Also remove from %s.</string>
```
arkon, line 760:
```suggestion
    <string name="track_delete_remote_text">Also remove from %s</string>
```

PR #9568, opened 2023-06-02
Filter out non-downloaded chapters in reader when Downloaded Only is enabled

### LINKED ISSUE #5463: Navigate only between the downloaded chapters while in "Downloaded only" filtered library
### Acknowledgements

- [X] I have updated the app to version **[0.11.1](https://github.com/tachiyomiorg/tachiyomi/releases/tag/v0.11.1)**.

### Tachiyomi version

Stable 0.11.1

### Android version

Android 6.0

### Device

Energy Phone Max 2+

### Steps to reproduce

1. Enable _Downloaded only_ mode in the menu _More_
2. Open a downloaded chapter with the next/previous  downloaded chapter separated by at least two numbers
3. Reach the end/start of the document and tap/scroll to open the next/previous chapter.

### Expected behavior

In the _Downloaded only_ mode all the non-downloaded chapters are hidden in the library, I therefore expected to be able to navigate only between them.

### Actual behavior

Since the next or previous chapter is not downloaded the app tries to load it if there is a network connection active, failing otherwise.

### Crash logs

_No response_

### Other details

_No response_

HerrGeist:
Closes #5463 by filtering the List which is used in all the widgets. I orientated myself on the code used to remove duplicated chapters.

I tested the changes and scrolling to the next and previous chapters works with 3 different sources and the skip buttons in both directions work on different sources except MangaDex. I cant seem to to find the difference between mangadex and any other Source in this situation.


app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterRemoveNotDownloaded.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Returns a copy of the list with not downloaded chapters removed
+ */
+fun List<Chapter>.removeNotDownloaded(manga: Manga): List<Chapter> {
+    val downloadProvider: DownloadProvider = Injekt.get()
+    val sourceManager: SourceManager = Injekt.get()
+    val source: Source = sourceManager.get(manga.source) ?: return ArrayList()
+
+    return filter { downloadProvider.findChapterDir(it.name, it.scanlator, manga.title, source) != null }
```
arkon, line 19: Use `isChapterDownloaded` from `DownloadCache` instead.

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterRemoveNotDownloaded.kt
```diff
@@ -0,0 +1,20 @@
...
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+/**
+ * Returns a copy of the list with not downloaded chapters removed
+ */
+fun List<Chapter>.removeNotDownloaded(manga: Manga): List<Chapter> {
+    val downloadCache: DownloadCache = Injekt.get()
+    val sourceManager: SourceManager = Injekt.get()
+    val source: Source = sourceManager.get(manga.source) ?: return ArrayList()
```
arkon, line 17: Fetching the source isn't really necessary here, `manga.source` should already be the source ID.

arkon (changes requested):

app/src/main/java/eu/kanade/tachiyomi/util/chapter/ChapterRemoveNotDownloaded.kt
```diff
@@ -0,0 +1,20 @@
...
+import tachiyomi.domain.chapter.model.Chapter
+import tachiyomi.domain.manga.model.Manga
+import tachiyomi.domain.source.service.SourceManager
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
+
+/**
+ * Returns a copy of the list with not downloaded chapters removed
+ */
+fun List<Chapter>.removeNotDownloaded(manga: Manga): List<Chapter> {
```
AntsyLich, line 14:
```suggestion
fun List<Chapter>.filterDownloaded(manga: Manga): List<Chapter> {
```

PR #9582, opened 2023-06-04
Cleanup MangaScreenModel successState usage

Two-Ai:
- Refactor `updateSuccessState`
   - Convert to `inline` function
   - Use `when` for type safety if we add other `MangaScreenState` types
- Replace equivalent expressions with `updateSuccessState`
- Replace safe cast in `MangaScreen`


arkon:
@Two-Ai build failed btw.

PR #9590, opened 2023-06-06
added windows file path for WSA file sharing support

ArthurKun21:
Added the windows file path for the WSA file sharing support https://github.com/tachiyomiorg/tachiyomi/issues/9589

`/storage/emulated/0/Windows/Tachiyomi/local`

The commit have changed the mapping of local Tachiyomi to local Tachiyomi in both android and windows


ArthurKun21:
oh thanks for this, hopefully it will come out soon.

In the mean time, the patch that I did kinda works but still be better for the new implementation to be build

PR #9603, opened 2023-06-11
Automatically convert Details.json to ComicInfo.xml

Shamicen:
I noticed that there was an open todo for converting `Details.json` to `ComicInfo.xml`

We don't really have to delete the `Details.json` file since they won't be scanned as long as a `ComicInfo` file exists, but it could be confusing to users when changes they make to their `Details.json` files are not reflected by the app. Please tell me if they should not be deleted.

### DIFF
--- a/source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
+++ b/source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
@@ -1,6 +1,7 @@
 package tachiyomi.source.local
 
 import android.content.Context
+import com.hippo.unifile.UniFile
 import eu.kanade.tachiyomi.source.CatalogueSource
 import eu.kanade.tachiyomi.source.Source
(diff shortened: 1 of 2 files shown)


source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
```diff
@@ -170,7 +173,6 @@ actual class LocalSource(
                     setMangaDetailsFromComicInfoFile(comicInfoFile.inputStream(), manga)
                 }
 
-                // TODO: automatically convert these to ComicInfo.xml
                 legacyJsonDetailsFile != null -> {
```
arkon, line 176: I was thinking more of a job that you can run from the advanced settings to do it so we could drop this reading logic entirely.

Shamicen: ah ok

arkon:
Ended up using this in 79b37df64

PR #9604, opened 2023-06-11
Better handle decimal chapter numbers and add categories in ComicInfo.xml files

Shamicen:
1. Changed the serialized chapter numbers to not use decimal points if possible.

2. Add the library categories to the Manga genres with 🔰 as a prefix. 

Example output:

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
@@ -631,7 +631,7 @@ class Downloader(
         source: HttpSource,
     ) {
         val chapterUrl = source.getChapterUrl(chapter.toSChapter())
-        val categories = runBlocking { getCategories.await(manga.id) }.map { "$CATAGORY_SYMBOL ${it.name.trim()}" }
+        val categories = runBlocking { getCategories.await(manga.id) }.map { it.name.trim() }.takeUnless { it.isEmpty() }
         val comicInfo = getComicInfo(manga, chapter, chapterUrl, categories)
         // Remove the old file
         dir.findFile(COMIC_INFO_FILE)?.delete()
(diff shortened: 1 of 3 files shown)


arkon:
I'm iffy on putting categories as genres, they're not really genres depending on what they are (like a "Reading" category doesn't make sense as a genre). If we really want to include that, it should be a separate custom field like what's happening with publishing status.

Shamicen:
>  If we really want to include that, it should be a separate custom field like what's happening with publishing status.

I added a new field for categories.

core-metadata/src/main/java/tachiyomi/core/metadata/comicinfo/ComicInfo.kt
```diff
@@ -7,6 +7,7 @@ import nl.adaptivity.xmlutil.serialization.XmlSerialName
 import nl.adaptivity.xmlutil.serialization.XmlValue
 
 const val COMIC_INFO_FILE = "ComicInfo.xml"
+private const val CATAGORY_SYMBOL = "\uD83D\uDD30"
```
arkon, line 10: No need for the symbol.

core-metadata/src/main/java/tachiyomi/core/metadata/comicinfo/ComicInfo.kt
```diff
@@ -18,6 +19,12 @@ fun SManga.copyFromComicInfo(comicInfo: ComicInfo) {
         comicInfo.tags?.value,
     )
         .flatMap { it.split(", ") }
+        .plus(
```
arkon, line 22: Shouldn't do this.

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -628,7 +631,8 @@ class Downloader(
         source: HttpSource,
     ) {
         val chapterUrl = source.getChapterUrl(chapter.toSChapter())
-        val comicInfo = getComicInfo(manga, chapter, chapterUrl)
+        val categories = runBlocking { getCategories.await(manga.id) }.map { it.name.trim() }.takeUnless { it.isEmpty() }
```
arkon, line 634: Make `createComicInfoFile` a suspend function instead of using `runBlocking`.

core-metadata/src/main/java/tachiyomi/core/metadata/comicinfo/ComicInfo.kt
Shamicen: I made the requested changes

PR #9621, opened 2023-06-20
Add function to delete downloaded chapters when migrating manga

### LINKED ISSUE #9153: Button to delete downloaded chapters when migrating

When migrating to another source, the downloaded chapters currently remain downloaded, but only for the previous source. Therefore, they occupy storage space without the user noticing or being able to easily fix it. There needs to be a way to manage downloaded chapters for titles that are not in the library and/or a prompt to delete downloaded chapters when a title with downloaded chapters is removed from the library.

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.14.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

semenvav:
I made it possible to delete downloaded manga during migration. In the course of work, I found an error: the value calculation in the getFlagsFromPositions method may not work correctly if not all checkboxes are visible.
For example: a manga does not have a track but has a custom cover and the user wants to migrate all available information i.e. chapters, categories and custom cover. In this case, the getFlagsFromPositions method will return 111 instead of 1011, and then there will be an attempt to migrate the track instead of the selected cover.
This was a problem for me, because in this case my rabolat function is only correct if the manga has a track and a custom cover.
In order to fix this problem, I reworked this function
Should close #9153

| **Before** | **After** |
| -- | -- |
| image | image |


arkon:
Some thoughts:

- Would it be more ideal to do a best attempt to "move" downloads over (i.e. move/rename chapters if they're the same recognized chapter number)?
- I feel like we could rewrite it to use something other than flags, but that's beyond the scope of this PR

semenvav:
1) The idea of moving chapters is good, but I think it's a different function. Moving chapters seems very handy when migrating between 2 sources with the same translation. I was guided by the standard migration scenario for me: I read all the translated chapters in Russian -> migrated to an English-language source. In such a scenario, I want to exactly delete the chapters. Maybe we should make a flag for deletion, and then making a separate move function? For example, if the user wants to replace the manga and has not selected the delete flag, then all chapters are automatically moved.
2) Yes, the flag system that exists now has problems with the extension, but I did not want to change the system much in this pr, just fix the problem I found with minimal changes.

PR #9624, opened 2023-06-22
Trackers empty scores start starts from middle

### LINKED ISSUE #9412: Make trackers empty scores start with 50 or 5

IMG_20230428_002623
When the score of an added tracker has yet to be specified, the initial score showed is 0. The suggestion is about making it start at 50 (for Anilist) and 5 for Myanimelist. Since most of the time users give score from 6 (or 60) or above in this way one doesn't have to scoll too much

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.14.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

semenvav:
close #9412

When the score of an added tracker has yet to be specified, the initial value was 0, and now the middle of the scale.
I think this is a good idea, because most of the ratings are > 5/50 and this will greatly reduce the time to scroll, if the tracker allows you to put ratings from 1 to 100


ivaniskandar:
This will also affect non-empty score selection

semenvav:
Yes, that was very stupid of me. Hope everything is ok now

app/src/main/java/eu/kanade/presentation/track/TrackInfoDialogSelector.kt
```diff
@@ -123,7 +123,7 @@ fun TrackScoreSelector(
         content = {
             WheelTextPicker(
                 modifier = Modifier.align(Alignment.Center),
-                startIndex = selections.indexOf(selection).coerceAtLeast(0),
+                startIndex = if (selections.indexOf(selection) > 0) selections.indexOf(selection) else (selections.size / 2),
```
arkon, line 126: Could probably do something like

```suggestion
                startIndex = selections.indexOf(selection).takeIf { it >= 0 } ?: selections.size / 2,
```

(Notably you need `>=`)

arkon (changes requested):

app/src/main/java/eu/kanade/presentation/track/TrackInfoDialogSelector.kt
semenvav: Are you sure I need to use >=. In this case? It will only work for Anilist with the Smiley system. In the bind method, if remoteTrack = zero, score = 0F is written. Everywhere except Anilist with the Smiley system 0 is included in the ScoreList and accordingly selections.indexOf(selection) will return 0 if no score is set. Also 0 in all these systems is not an score and means no score.

arkon: That's a good call, but the fact that it won't handle the smiley system is also a problem.

arkon: Sure. Can you just update it with the takeIf then?

semenvav: Of course, as soon as I get to the computer

PR #9641, opened 2023-06-25
Show a progress indicator while checking for updates in the about screen

alessandrojean:
Shows a small circular progress indicator in the "Check for updates" item to make it more explicit to the user that something is being done. The check action is also now only called once if the user happens to click the item multiple times while a check is still being done.

I don't know if it's worth adding a `delay` in the check action to make the progress indicator to show a little longer if the update check is fast enough to make the indicator to be hidden suddenly.

Probably the initial toast can be removed as well, but I let as it is for now.

https://github.com/tachiyomiorg/tachiyomi/assets/14254807/5ee37ab7-0dad-48b7-9bcb-f9aaff71a220

Took the inspiration from the [Moshidon app](https://github.com/LucasGGamerM/moshidon).


app/src/main/java/eu/kanade/presentation/more/settings/screen/AboutScreen.kt
```diff
@@ -186,7 +213,11 @@ object AboutScreen : Screen() {
     /**
      * Checks version and shows a user prompt if an update is available.
      */
-    private suspend fun checkVersion(context: Context, onAvailableUpdate: (GetApplicationRelease.Result.NewUpdate) -> Unit) {
+    private suspend fun checkVersion(
+        context: Context,
+        onAvailableUpdate: (GetApplicationRelease.Result.NewUpdate) -> Unit,
+        onCheckingFinished: () -> Unit,
```
arkon, line 219: `onFinish`?

app/src/main/java/eu/kanade/presentation/more/settings/screen/AboutScreen.kt
```diff
@@ -186,7 +213,11 @@ object AboutScreen : Screen() {
...
+    ) {
         val updateChecker = AppUpdateChecker()
         withUIContext {
             context.toast(R.string.update_check_look_for_updates)
```
arkon, line 223: Let's remove this (including the string).

PR #9662, opened 2023-07-02
Automatically mark new chapters as read if a duplicate chapter exists and has been read

TraceLosu:
Adds a toggle-able option to automatically mark new duplicate chapters (a new chapter with the same chapter number) as read if it has been read by the user before. This removes the need to manually mark new duplicate chapters as read to avoid skipping the manga when updating.


scb261:
Does it consider chapters that don't have numbers? What happens if there is a read chapter without a number and a new different unread chapter without a number? Iirc, they are stored as `-1`.

TraceLosu:
> Does it consider chapters that don't have numbers? What happens if there is a read chapter without a number and a new different unread chapter without a number? Iirc, they are stored as `-1`.

I completely forgot about chapters with no numbers. Will fix that.

TraceLosu:
Actually, I just realized the logic for removing duplicate chapters in the reader considers all `-1` numbered chapters the same. So I think following the same logic is better here as well.

Kyrela:
Would love to see that feature integrated.

Issue #9046, opened 2023-02-07
Global search bug 


Alxhe:
### Steps to reproduce

1. Search any MMM on the global search
2.  List of all your extension will shown
3. The search will start 
4. The extension that dont have the Manga for example will not move and will just stay
5. You will swipe down ward so you can check if the manga that you finding is thier

### Expected behavior

this will be happen if you search in the global search the manga that you searching will surpace with its extension and all of the extension that dont have the specific manga that you searching will go to the lower part of the search bar

### Actual behavior

this will be happen if you search in the global search the manga that you searching will surpace with its extension and all of the extension that dont have the specific manga that you searching will go to the lower part of the search bar

### Crash logs

![inbound3702624304794073395](https://user-images.githubusercontent.com/124679890/217232580-1f6f3c20-50ff-499f-96cb-988a6acd4b46.jpg)
![inbound5700911253248256412](https://user-images.githubusercontent.com/124679890/217232591-94a4df3f-50df-49cd-b180-f72306e3c83a.jpg)
![inbound65004658144585858](https://user-images.githubusercontent.com/124679890/217232604-5070adbb-135d-45f4-bebc-ffffb920697e.jpg)

### Tachiyomi version

0.14.4

### Android version

android 8.1.0

### Device

Vivo y91

### Other details

N/A

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have tried the [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] I have updated the app to version **[0.14.4](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

arkon:
This was intentional requested behavior.

Alxhe:
Sorry firstime in github and I just conclude that this problem is a bug, because on the old version the version before the latest the global search is not like this in the video.
![inbound912064333711718562](https://user-images.githubusercontent.com/124679890/217263114-d3e3ca27-149b-429d-88b3-4cf189cdbea9.jpg)

Alxhe:
[...]

PR #9670, opened 2023-07-04
Fix global search scroll position

zaghdaneh:
### Summary

Related issue : #9046 

Fixed the issue by rescrolling to the top of the page whenever the search progress increases.
Potentially can annoy users scrolling down while search is still in progress and getting sent back to the top of the page (adressable by making the scrollToTop conditions more specific.


AntsyLich:
This is an annoyance rather than fix imo.

arkon:
Yeah, this is not great.

PR #9675, opened 2023-07-06
Add Tachiyomi Cross Device Sync Feature

AriaMoradi:
Continuing work from #9326

Splitted the PR to make reviews more straight forward.

This PR only contains the bare bones needed for sync to work.

There will be subsequent PRs adding more features to sync. 

Reviewer note: The code is compatible with [SyncYomi 0.2.4](https://github.com/SyncYomi/SyncYomi/releases/tag/v0.2.4)


AntsyLich (changes requested):
Extra opnions:
- The backup related stuff that are now also being used by sync now should be renamed to something else to indicate it's used by both
- Imo we should implement google drive sync first since more people are likely to use that instead of a self hosted server. I personally won't be able to host syncyomi any time soon and won't be able to test either.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
```diff
@@ -0,0 +1,185 @@
...
+                    title = stringResource(R.string.pref_sync_now),
+                    subtitle = stringResource(R.string.pref_sync_now_subtitle),
+                    onClick = {
+                        showDialog.value = true
+                    },
+                    icon = Icons.Outlined.Sync,
+                ),
+            ),
+        )
+    }
```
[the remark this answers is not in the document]
AriaMoradi: errors out saying remember delegate doesn't have a `setValue()`

AriaMoradi:
> Imo we should implement google drive sync first since more people are likely to use that instead of a self hosted server. I personally won't be able to host syncyomi any time soon and won't be able to test either.

Coming up in the next PR, if I add it back we'll get what was in the original PR

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsSyncScreen.kt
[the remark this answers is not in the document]
jobobby04: You have to import setValue, `import androidx.compose.runtime.setValue`

domain/src/main/java/tachiyomi/domain/sync/SyncPreferences.kt
```diff
@@ -0,0 +1,20 @@
...
+    fun syncAPIKey() = preferenceStore.getString("sync_api_key", "")
+    fun syncLastSync() = preferenceStore.getInstant("sync_last_sync", Instant.EPOCH)
+
+    fun syncInterval() = preferenceStore.getInt("sync_interval", 0)
+
+    fun deviceName() = preferenceStore.getString("device_name", android.os.Build.MANUFACTURER + android.os.Build.PRODUCT)
+
+    fun deviceID() = preferenceStore.getInt("device_id", 0)
+
+    fun syncService() = preferenceStore.getInt("sync_service", 0)
```
AntsyLich, line 19: Use enum preference

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,209 @@
...
+                null
+            }
+        }
+
+        val remoteBackup = syncService?.doSync(syncData)
+
+        if (remoteBackup != null) {
+            val (filteredFavorites, nonFavorites) = filterFavoritesAndNonFavorites(remoteBackup)
+            updateNonFavorites(nonFavorites)
+            SyncHolder.backup = backup.copy(backupManga = filteredFavorites)
```
KaiserBh, line 118: This is wrong by the way, It won't restore the categories and just place all the manga in the default categories instead.

I have updated it in the master branch and fixed it @AriaMoradi but basically we want to copy the rest of the data as well.

```suggestion
            SyncHolder.backup = backup.copy(
                backupManga = filteredFavorites,
                backupCategories = remoteBackup.backupCategories,
                backupSources = remoteBackup.backupSources,
                backupBrokenSources = remoteBackup.backupBrokenSources,
            )
```

domain/src/main/java/tachiyomi/domain/sync/SyncPreferences.kt
```diff
@@ -0,0 +1,20 @@
...
+) {
+    fun syncHost() = preferenceStore.getString("sync_host", "https://sync.tachiyomi.org")
+    fun syncAPIKey() = preferenceStore.getString("sync_api_key", "")
+    fun syncLastSync() = preferenceStore.getInstant("sync_last_sync", Instant.EPOCH)
+
+    fun syncInterval() = preferenceStore.getInt("sync_interval", 0)
+
+    fun deviceName() = preferenceStore.getString("device_name", android.os.Build.MANUFACTURER + android.os.Build.PRODUCT)
+
+    fun deviceID() = preferenceStore.getInt("device_id", 0)
```
KaiserBh, line 17: Can remove this now I am not using it in syncyomi server anymore.
```suggestion
```

app/src/main/java/eu/kanade/tachiyomi/data/sync/service/SyncYomiSyncService.kt
KaiserBh, line 1: This whole file is changed and the api endpoint changed as well check master branch for the latest one, might need to refactor the merging function at the moment I just dump it there but maybe another class that manage them maybe? Since it will be used by all services? Such as GoogleDrive, SyncYomi, WebDav, and any other cloud services maybe. 

@AriaMoradi

KaiserBh (commented):
Need to change some stuff right now it's broken when I last tested this (fixes in master branch), Also remember what we talked about, since right now we are getting the whole manga from database instead of favorites only it will be a lot of unnecessary data, and one of your suggestion was that we add lastFavoritedAt, so might as well implement that with this PR since it's part of the base.

arkon:
If you're going to split into smaller PRs, do the database-related changes in 1 PR (and _only_ those changes, including an app version increase).

AriaMoradi:
Merged backup and sync settings screens

image
image
image

KaiserBh:
> If you're going to split into smaller PRs, do the database-related changes in 1 PR (and _only_ those changes, including an app version increase).

#9683 opened the PR.

app/src/main/java/eu/kanade/tachiyomi/data/sync/service/SyncService.kt
```diff
@@ -23,4 +43,189 @@ abstract class SyncService(
...
+
+    /**
+     * Download sync data from the remote storage
+     */
+    abstract suspend fun pushSyncData(): SyncData?
+
+    /**
+     * Upload sync data to the remote storage
+     */
+    abstract suspend fun pullSyncData(syncData: SyncData)
```
KaiserBh, line 60: I think you mixed the comments for the two @AriaMoradi

app/src/main/java/eu/kanade/tachiyomi/data/sync/service/SyncYomiSyncService.kt
```diff
@@ -81,7 +49,7 @@ class SyncYomiSyncService(
         }
     }
 
-    suspend fun uploadSyncData(body: RequestBody) {
+    override suspend fun pullSyncData(syncData: SyncData) {
```
KaiserBh, line 52: This too should it be pull for uploading/updating the data? @AriaMoradi

arkon:
Next PR should ideally just be any changes to the existing backup/restore stuff.

KaiserBh:
> Next PR should ideally just be any changes to the existing backup/restore stuff.

#9699

KaiserBh:
This PR should be ready once #9699 is merged.

quangkieu:
My question would be for the conflict comic? So Machine 1 sync comic A read to chapter 10 but Machine 2 said comic A is at chapter 20? or chapter the fetch time, chapter update time, I know this one would change as some source just said New without date for like 2 day before put a update date, or the cover image as user could customize?

KaiserBh:
> My question would be for the conflict comic? So Machine 1 sync comic A read to chapter 10 but Machine 2 said comic A is at chapter 20? or chapter the fetch time, chapter update time, I know this one would change as some source just said New without date for like 2 day before put a update date, or the cover image as user could customize? If we have a dialog to show the conflict and last edit day per manga/comic would be greater to resolve that I think the last edit of each comic would be good to identify which version to keep. But what about when user restore from a backup, the edit time would be always newer from the sync copy. I would suggest to disable sync when restoring from backup and have feature to force update or restore from the cloud snapshot

Hi, I believe we have addressed the issue with conflicts you can read more on the last PR from [here](https://github.com/tachiyomiorg/tachiyomi/pull/9326#discussion_r1167644619) when arkon first mentioned it, and [this](https://github.com/tachiyomiorg/tachiyomi/pull/9326#issuecomment-1516704933)

Riztard:
included in #9790

PR #9683, opened 2023-07-08
Database changes to support library syncing

KaiserBh:
Database Changes only as requested.

I believe this will be all that's needed in terms of database changes, let me know if there is anything I need to change.


data/src/main/sqldelight/tachiyomi/migrations/25.sqm
```diff
@@ -0,0 +1,152 @@
+import kotlin.collections.List;
+import eu.kanade.tachiyomi.source.model.UpdateStrategy;
+
+-- Drop indices
+DROP INDEX IF EXISTS library_favorite_index;
+DROP INDEX IF EXISTS mangas_url_index;
+DROP INDEX IF EXISTS chapters_manga_id_index;
+DROP INDEX IF EXISTS chapters_unread_by_manga_index;
+
+-- Rename existing tables to temporary tables
```
arkon, line 10: Can't you just add columns to the existing tables? Why the effectively complete replacements?

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupChapter.kt
```diff
@@ -20,6 +20,8 @@ data class BackupChapter(
     // chapterNumber is called number is 1.x
     @ProtoNumber(9) var chapterNumber: Float = 0F,
     @ProtoNumber(10) var sourceOrder: Long = 0,
+    @ProtoNumber(11) var lastModifiedAt: Long? = null,
+    @ProtoNumber(12) var lastFavoritedAt: Long? = null,
```
AriaMoradi, line 24: We decided to name the field `favoriteModifiedAt`, why change it to `lastFavoritedAt`?

data/src/main/sqldelight/tachiyomi/migrations/25.sqm
KaiserBh: Not sure why I did it that way but fixed it.

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupChapter.kt
KaiserBh: does it matter? I just made it the same for consistency.

AriaMoradi: the meaning is very different

domain/src/main/java/tachiyomi/domain/manga/model/Manga.kt
```diff
@@ -24,6 +24,8 @@ data class Manga(
     val thumbnailUrl: String?,
     val updateStrategy: UpdateStrategy,
     val initialized: Boolean,
+    val lastModifiedAt: Long?,
+    val lastFavoritedAt: Long?,
```
AntsyLich, line 28: Looking at #9675 this isn't actually being shown anywhere on the ui so any need to add it to `Manga` object?

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupManga.kt
```diff
@@ -39,6 +39,8 @@ data class BackupManga(
     @ProtoNumber(103) var viewer_flags: Int? = null,
     @ProtoNumber(104) var history: List<BackupHistory> = emptyList(),
     @ProtoNumber(105) var updateStrategy: UpdateStrategy = UpdateStrategy.ALWAYS_UPDATE,
+    @ProtoNumber(106) var lastModifiedAt: Long? = 0,
+    @ProtoNumber(107) var lastFavoritedAt: Long? = 0,
```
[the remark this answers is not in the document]
KaiserBh: > If this field stores when a title was added to/removed from library it should be `favoriteModifiedAt`

It only accounts for adding to favorite not removing, I will change it to `favoriteModifiedAt`.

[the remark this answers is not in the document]
AntsyLich: Why not removing? If I remove the title from one device it should be removed from another too

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupChapter.kt
AriaMoradi: There is no concept of favoriting chapters but there is a bookmark feature and reading status.

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupManga.kt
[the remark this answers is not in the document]
AriaMoradi: > It only accounts for adding to favorite not removing

Correction: default value is null, value is set when user toggles favorite so both adding and removing.

domain/src/main/java/tachiyomi/domain/manga/model/Manga.kt
AriaMoradi: every manga has these fields associated with it, the alternative is adding a one to one foreign key which won't be useful much as it is needed most of the time imo.

app/build.gradle.kts
```diff
@@ -22,8 +22,8 @@ android {
 
     defaultConfig {
         applicationId = "eu.kanade.tachiyomi"
-
-        versionCode = 103
+
```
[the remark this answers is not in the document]
AriaMoradi: arkon asked to bump app version, are we doing it incorrectly?

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupManga.kt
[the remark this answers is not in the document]
KaiserBh: > > It only accounts for adding to favorite not removing
> 
> Correction: default value is null, value is set when user toggles favorite so both adding and removing.

Oh yea that's true, I thought it was just accounting for adding and not removing. Thanks for the correction.

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupChapter.kt
KaiserBh: Yea that's not meant to be there. Fixed in the latest commit.

app/build.gradle.kts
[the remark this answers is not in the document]
AntsyLich: No the version bump is ok but you were also reverting a change from another commit in the process. It's fixed now

domain/src/main/java/tachiyomi/domain/manga/model/Manga.kt
AntsyLich: I mean it's fine to keep it in db but is it needed to be in the manga object too?

KaiserBh: > I mean it's fine to keep it in db but is it needed to be in the manga object too?

If we store these values in the DB without exposing them in the manga object, they essentially become redundant. The BackupManga class copies data from the Manga object, and these fields play an essential role in our internal merging and syncing logic. You can refer to the implementation [here](https://github.com/tachiyomiorg/tachiyomi/pull/9675/files#diff-c058878952993eb2cd615c786376f0c38efc6e1af2663c41ceaa425a40362558R61-R231)

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupManga.kt
```diff
@@ -39,6 +39,8 @@ data class BackupManga(
     @ProtoNumber(103) var viewer_flags: Int? = null,
     @ProtoNumber(104) var history: List<BackupHistory> = emptyList(),
     @ProtoNumber(105) var updateStrategy: UpdateStrategy = UpdateStrategy.ALWAYS_UPDATE,
+    @ProtoNumber(106) var lastModifiedAt: Long? = 0,
+    @ProtoNumber(107) var favoriteModifiedAt: Long? = 0,
```
arkon, line 43: I imagine looking at part of the larger PR might answer this, but why do we need both of these timestamps?

data/src/main/sqldelight/tachiyomi/migrations/25.sqm
```diff
@@ -0,0 +1,64 @@
+ALTER TABLE mangas ADD COLUMN last_modified_at INTEGER AS Long;
+ALTER TABLE mangas ADD COLUMN favorite_modified_at INTEGER AS Long;
+ALTER TABLE mangas_categories ADD COLUMN last_modified_at INTEGER AS Long;
+ALTER TABLE chapters ADD COLUMN last_modified_at INTEGER AS Long;
```
arkon, line 4: Should we attempt to backfill these values?

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupManga.kt
KaiserBh: > I imagine looking at part of the larger PR might answer this, but why do we need both of these timestamps?

There was an issue when syncing manga favorites between devices, primarily when a manga was removed from favorites (Library). Initially, our sync system only tracked favorited manga. Here's how that led to issues:

Let's say we have two devices, Device A and Device B. Now, imagine a user removes a manga – let's use "One Piece" as an example – from their favorites on Device A and then syncs the device. The problem arose when Device B, which already had "One Piece" in its favorites, tried to sync. Instead of deleting "One Piece" from its favorites to mirror the change made on Device A, Device B would end up restoring "One Piece" back into the favorites. The unintended consequence was that if we synced Device A again, "One Piece" would incorrectly be restored to its favorites, even though the user had initially removed it.

To address this issue, I decided to keep a record of every manga on the device, not just the ones marked as favorites. This approach also includes manga that users might have searched for but didn't add to their library. However,  Aria mentioned that some users have thousands data in their phone that are not in favorite etc, therefore it will be a lot of unnecessary data being sent to server/services increase data usage etc.

In order to tackle this, @AriaMoradi proposed incorporating a favoriteModifiedAt timestamp. With this timestamp, we still have non-favorite manga data, but it is not as extensive as it would have been without the timestamp. We can now sync manga where favoriteModifiedAt != null, which effectively reduces the amount of irrelevant data that are being synced.

data/src/main/sqldelight/tachiyomi/migrations/25.sqm
KaiserBh: Oh yea we definitely should do that.

Will something like this do?

```sql UPDATE mangas SET last_modified_at = strftime('%s', 'now');
UPDATE mangas_categories SET last_modified_at = strftime('%s', 'now');
UPDATE chapters SET last_modified_at = strftime('%s', 'now');
UPDATE mangas SET favorite_modified_at = strftime('%s', 'now') WHERE favorite = 1;
```

AntsyLich: You missed `mangas` `last_modified_at`. Also might as well `last_modified_at` for all of them non-null

data/src/main/sqldelight/tachiyomi/data/mangas.sq
```diff
@@ -22,12 +22,40 @@ CREATE TABLE mangas(
...
+END;
+
+CREATE TRIGGER insert_last_modified_at_mangas
+AFTER INSERT ON mangas
+FOR EACH ROW
+BEGIN
+  UPDATE mangas
+  SET last_modified_at = strftime('%s', 'now')
+  WHERE _id = new._id;
+END;
```
AntsyLich, line 57: Redundant?

data/src/main/sqldelight/tachiyomi/data/mangas_categories.sq
```diff
@@ -2,15 +2,34 @@ CREATE TABLE mangas_categories(
...
+END;
+
+CREATE TRIGGER insert_last_modified_at_mangas_categories
+AFTER INSERT ON mangas_categories
+FOR EACH ROW
+BEGIN
+  UPDATE mangas_categories
+  SET last_modified_at = strftime('%s', 'now')
+  WHERE _id = new._id;
+END;
```
AntsyLich, line 28: This too

data/src/main/sqldelight/tachiyomi/migrations/25.sqm
```diff
@@ -0,0 +1,69 @@
...
+
+DROP TRIGGER IF EXISTS insert_last_modified_at_mangas;
+CREATE TRIGGER insert_last_modified_at_mangas
+AFTER INSERT ON mangas
+FOR EACH ROW
+BEGIN
+  UPDATE mangas
+  SET last_modified_at = strftime('%s', 'now')
+  WHERE _id = new._id;
+END;
```
AntsyLich, line 30: 3

data/src/main/sqldelight/tachiyomi/migrations/25.sqm
```diff
@@ -0,0 +1,69 @@
...
+
+DROP TRIGGER IF EXISTS insert_last_modified_at_mangas_categories;
+CREATE TRIGGER insert_last_modified_at_mangas_categories
+AFTER INSERT ON mangas_categories
+FOR EACH ROW
+BEGIN
+  UPDATE mangas_categories
+  SET last_modified_at = strftime('%s', 'now')
+  WHERE _id = new._id;
+END;
```
AntsyLich, line 69: 4

AntsyLich (changes requested):

domain/src/main/java/tachiyomi/domain/chapter/model/Chapter.kt
```diff
@@ -13,6 +13,7 @@ data class Chapter(
     val dateUpload: Long,
     val chapterNumber: Float,
     val scanlator: String?,
+    val lastModifiedAt: Long?,
```
AntsyLich, line 16:
```suggestion
    val lastModifiedAt: Long,
```

domain/src/main/java/tachiyomi/domain/chapter/model/Chapter.kt
```diff
@@ -31,6 +32,7 @@ data class Chapter(
             dateUpload = -1,
             chapterNumber = -1f,
             scanlator = null,
+            lastModifiedAt = null,
```
AntsyLich, line 35:
```suggestion
            lastModifiedAt = 0,
```

domain/src/main/java/tachiyomi/domain/manga/model/Manga.kt
```diff
@@ -24,6 +24,8 @@ data class Manga(
     val thumbnailUrl: String?,
     val updateStrategy: UpdateStrategy,
     val initialized: Boolean,
+    val lastModifiedAt: Long?,
```
AntsyLich, line 27:
```suggestion
    val lastModifiedAt: Long,
```

AntsyLich (changes requested):

arkon (approved):

PR #9691, opened 2023-07-09
Add filters to Global search

zaghdaneh:
### Summary
Related issue #6654

Added a few filter chips to the global search to quickly filter the results
- **All** : Shows all sources
- **Pinned**: Shows only pinned sources
- **available**: Hides sources that have an error (Failed to bypass cloudflare ...) 

### Implementation
For the implementation process I used Filterchips so I followed the way it was used in the `BrowseSourceScreen` class.
But I'm unsure if this is the best way to create the filters (using an enum)
Also for the 'available' icon I'm not sure what icon to use.

### Images
| All | Pinned | Available |
| ------- | ------- |  ------- |
| all | Pinned | Available |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/globalsearch/GlobalSearchScreenModel.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/globalsearch/GlobalSearchScreenModel.kt
@@ -29,13 +29,7 @@ class GlobalSearchScreenModel(
         .distinctUntilChanged()
         .map { (filter, items) ->
             items
-                .filterNot { (_, result) ->
-                    filter == GlobalSearchFilter.AvailableOnly &&
-                        result is SearchItemResult.Error
-                }
-                .filterNot { (source, _) ->
-                    filter == GlobalSearchFilter.PinnedOnly && "${source.id}" !in sourcePreferences.pinnedSources().get()
-                }
+                .filter { (source, result) -> isSourceVisible(filter, source, result) }
         }.stateIn(ioCoroutineScope, SharingStarted.Lazily, state.value.items)
 
     init {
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/globalsearch/GlobalSearchScreenModel.kt
```diff
@@ -38,6 +50,14 @@ class GlobalSearchScreenModel(
             .sortedWith(compareBy({ "${it.id}" !in pinnedSources }, { "${it.name.lowercase()} (${it.lang})" }))
     }
 
+    private fun isSourceVisible(filter: GlobalSearchFilter, source: CatalogueSource, result: SearchItemResult): Boolean {
+        return when (filter) {
+            GlobalSearchFilter.AvailableOnly -> result !is SearchItemResult.Error
```
arkon, line 55: I feel like it'd be more useful as an option to only show sources with >= 1 result (i.e. "Has results").

I feel like the UI will need to change to become more of a set of options before doing the search than the chips allow for, but that can happen separately.

app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/globalsearch/GlobalSearchScreenModel.kt
```diff
@@ -50,14 +70,22 @@ class GlobalSearchScreenModel(
...
 
+    fun setFilter(filter: GlobalSearchFilter) {
+        mutableState.update { it.copy(searchFilter = filter) }
+    }
+
     override fun getItems(): Map<CatalogueSource, SearchItemResult> {
         return mutableState.value.items
     }
 }
+enum class GlobalSearchFilter {
```
arkon, line 81: Nitpick: add an extra empty line above this

app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/globalsearch/GlobalSearchScreenModel.kt
zaghdaneh: For the second point, you mean the goal is to use the options to limit the sources to reduce the number of requests done by the search?

AntsyLich: Yes

PR #9699, opened 2023-07-11
Refactor backup and restore to support cross device sync.

KaiserBh:
Backup and restore refactor.


app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncHolder.kt
```diff
@@ -0,0 +1,7 @@
+package eu.kanade.tachiyomi.data.sync
+
+import eu.kanade.tachiyomi.data.backup.models.Backup
+
+object SyncHolder {
+    var backup: Backup? = null
```
AntsyLich, line 6: What is the purpose of `SyncHolder`? I can see that it's being used in other files but why is it needed

AntsyLich (changes requested):

THElegend5:
Btw does this also result a change in tachiyomi backup modals?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncHolder.kt
KaiserBh: I don't quite remember but from this [commit](https://github.com/tachiyomiorg/tachiyomi/commit/0a1137414efa05fc87e2669ed654bcda71c37ad9) if am not wrong I added it because I couldn't use the object directly in the worker or something I tried to convert the json into backup object but getting error that there is limit in worker etc. So I made sync holder which just holds the current value/data we get from the server that have been turned into object from json. Then again I think it can be refactored, once the last bit of the PR is added.

AntsyLich: Imo it would be better to save the `Backup` object as protobuf inside application data folder. This way we can just directly pass the uri to restore functions

KaiserBh: Yea that would probably be better and we can eliminate the boolean that is being used @AriaMoradi mentioned something similar that we should refactor it and not use the boolean/syncHolder. Can refactor it once everything is together as it will be easier to test and see if it's working as intended.

AntsyLich: Imo we should make the changes now and test them slowly rather than when the changes are more big

KaiserBh: > Imo we should make the changes now and test them slowly rather than when the changes are more big

Next PR will be this #9675 I will refactor it there then before we open the one after that which will be Gdrive implementation which is the biggest one.

AntsyLich: OK then the restore functions could be renamed to sync (e.g. `restoreBackup` -> `syncFromBackup`) since at the end they are syncing (backup file being a local source) and similarly to backup functions (e.g. `backupExtensionInfo` -> `prepExtensionInfoForSync`)

AntsyLich:
@arkon can we just replace the backup terminology to sync in the code?

arkon:
I think I'd still consider it a backup restore, but syncing could be considered an overwrite mode instead of an append mode.

i18n/src/main/res/values/strings.xml
```diff
@@ -516,6 +516,10 @@
     <string name="restoring_backup_canceled">Canceled restore</string>
     <string name="backup_info">You should keep copies of backups in other places as well.</string>
 
+    <!-- Sync section -->
+    <string name="syncing_library">Syncing library</string>
+    <string name="library_sync_complete">Library sync complete</string>
```
arkon, line 521: I actually feel like this doesn't need to be a notification, but we can maybe mull about that in the final PR.

PR #9710, opened 2023-07-13
Add private extension install method

ivaniskandar:
Private extensions are put inside private data directory of the running app, so this kind of extensions can only be used by the running app and not shared with other apps.

One limitation of private extension is the lack of deeplink handlers (if there's any) since the extension APK is not installed to the system. The other is this kind of extension is not persistent, meaning any private extension will be deleted when the user uninstalled or cleared data of the app.

When both kinds of extensions are installed with a same package name, shared extension (the one installed to the system) will be used unless the version codes are different. In that case the one with higher version code will be used.

~~With this install method enabled, installed extensions will be automatically updated when a newer version is available.~~


app/src/main/java/eu/kanade/tachiyomi/extension/util/ExtensionInstallReceiver.kt
```diff
@@ -121,4 +126,30 @@ internal class ExtensionInstallReceiver(private val listener: Listener) :
...
+
+        fun notifyRemoved(context: Context, pkgName: String) {
+            notify(context, pkgName, ACTION_EXTENSION_REMOVED)
+        }
+
+        private fun notify(context: Context, pkgName: String, action: String) {
+            Intent(action).apply {
+                data = Uri.parse("package:$pkgName")
+                `package` = context.packageName
+                context.sendBroadcast(this)
```
arkon, line 151: Do a local broadcast instead.

app/src/main/java/eu/kanade/tachiyomi/extension/util/ExtensionLoader.kt
```diff
@@ -41,12 +52,11 @@ internal object ExtensionLoader {
     const val LIB_VERSION_MIN = 1.3
     const val LIB_VERSION_MAX = 1.5
 
-    private val PACKAGE_FLAGS = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-        PackageManager.GET_CONFIGURATIONS or PackageManager.GET_SIGNING_CERTIFICATES
-    } else {
-        @Suppress("DEPRECATION")
-        PackageManager.GET_CONFIGURATIONS or PackageManager.GET_SIGNATURES
-    }
+    @Suppress("DEPRECATION")
```
arkon, line 55: Better hope they never do something to `GET_SIGNATURES`...

app/src/main/java/eu/kanade/tachiyomi/extension/util/ExtensionLoader.kt
```diff
@@ -220,12 +360,50 @@ internal object ExtensionLoader {
...
+            ?.map { Hash.sha256(it.toByteArray()) }
+            ?.toList()
+    }
+
+    private fun hasTrustedSignature(signatures: List<String>): Boolean {
+        return trustedSignatures.any { signatures.contains(it) }
+    }
+
+    private fun isOfficiallySigned(signatures: List<String>): Boolean {
+        return signatures.find { it != officialSignature } == null
```
arkon, line 389:
```suggestion
        return signatures.all { it == officialSignature }
```

PR #9736, opened 2023-07-18
Fix "Trigger new search on source filter change"

KokaKiwi:
`SearchScreenModel::sources` being lazy-init evaluated, it's not computed again when changing source filter thus leading to being at an incorrect value (the old one)

This PR propose fixing it by simply evaluate the sources to search on each time `search()` is called, which sould not be too heavy to compute anyway

Fixes c7f0a54a375380a8b6bd100c712acb2e01c11ac3


AntsyLich:
And that has been fixed by https://github.com/tachiyomiorg/tachiyomi/commit/c7f0a54a375380a8b6bd100c712acb2e01c11ac3 afaik

KokaKiwi:
Not really actually which is why i did mention this commit, because what it does is trigger the `search()` method whenever filters are changed, but there is still this `sources` property not "re-triggered" at the same time, leading it to retain the first initialized value (because of the `lazy` delegate), which make the `search()` method re-trigger search on "outdated" sources list

KokaKiwi:
The only problem i actually see with my PR is that whenever you only change the query, the sources list is recomputed, which could be fixed by invalidating the `sources` property only on filter change

arkon:
The identified problem is correct, but the fix is pretty overcomplicated. Superseded by #3f868c043.

PR #9737, opened 2023-07-18
Add download stats screen

### LINKED ISSUE #9567: Viewable Folder size of entry

It would be great if there is an option in the settings which shows the downloaded file sizes of each manga like this:
Screenshot_20230601_223906_Cx File Explorer
 Or another option is to display the total size of the downloaded files inside an entry somewhere in a corner:
Screenshot_20230601_223950_Aniyomi

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.14.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.

semenvav:
I made new screen which displays the uploaded file sizes for each manga, category and source

Screen accessible from settings screen

Screenshot_20230718_180017_Tachiyomi

Manga can be displayed separately or be grouped by category and source

### separate display of manga by default

image

|Group by category  | Group by source |
| ------- | ------- |
| Screenshot_20230718_175722_Tachiyomi | Screenshot_20230718_175606_Tachiyomi|

2 types of sort by alphabet and by size works on manga and group
| | |
| ------- | ------- |
| Screenshot_20230718_175733_Tachiyomi | Screenshot_20230718_175744_Tachiyomi|

Search works by manga name and source/category name in case of grouping
|Manga search  | Source search|
| ------- | ------- |
| Screenshot_20230718_175656_Tachiyomi |  Screenshot_20230718_175637_Tachiyomi|

Manga items have button to delete it and button for multiply deletion in selection mod
|Category selection  | Default selection |
| ------- | ------- |
| Screenshot_20230718_181136_Tachiyomi | Screenshot_20230718_181202_Tachiyomi|

Should close #9567


Osyx:
I feel like it would make more sense to have it as one statistics entry in settings, rather than two, and then group them somehow afterwards. Either via tabs or some other way of grouping. Would also be nice to have an overall stat of how many chapters you have dowloaded over time, their size, the current total etc.

semenvav:
Yes, I thought about somehow adding this to the existing statistics screen, but I didn’t figure out how to do it conveniently and beautifully and decided to make a separate screen.
In the future, I am going to refine this functionality, including giving out more information, this is rather not a fully developed function, but a base with the most basic and important functionality for me - memory management

AntsyLich:
What about chapters that have been deleted or entries that are not in library anymore

semenvav:
For now this is not taken into account. Only currently downloaded entries are displayed. For beginning, my goal was simply to make the ability to track memory usage and free it more convenient than through some kind of file explorer and that's all for now.

Programmer-0-0:
I've tested this out and I'm not sure if this is intentional or not but when you start downloading a series (and let's say you also pressed download on a wrong chapter, and it finished downloading). When you press pause and got to Download Stats to delete that chapter, it delete's that chapter but also clears the download queue of all the other chapter of the same series. Also, one suggestion I would like to make is to change "Alphabetically" to "A-Z" as on my phone it was wrapping to second line and it didn't look that nice. Otherwise, everything looked pretty good (at least from my limited testing)

AntsyLich:
Why is a BackHandler needed

Programmer-0-0:
@AntsyLich Because when I was in selection mode, when I press back instead of taking me out of selection mode it returns me back to the more screen. In the library section and updates section when you press back it takes you out of selection mode so maybe that's why I got used to it that way.

semenvav:
@Programmer-0-0 Yes, I agree, if on other screens the back button cancels the selection mode / search, then I need to do the same. The text on sorting also changed

Programmer-0-0:
@semenvav I was a bit curious and tried to see if I can save the expanded state on a configuration change (like rotating the screen for example). So I did this:
```
val expandedMapSaver = run {
        mapSaver(
            save = {expanded: MutableMap<String, Boolean> ->
                expanded.toMap()
            },
            restore = { restorationMap: Map<String, Any?> ->
                restorationMap.mapValues { it.value as Boolean }.toMutableMap()
            },
        )
    }
    val expanded = rememberSaveable(saver = expandedMapSaver) {
        mutableStateMapOf(*categoryMap.keys.toList().map { it to false }.toTypedArray())
    }
```
 There is one problem though. Initially when I press on any header for it to expand and contract it works perfectly fine but then when I rotate the screen (configuration change) the expanded sate of the list gets saved but at this point if I press on any header for it to expand or collapse the UI doesn't update (or recompose) until I rotate the screen. I didn't want to use a viewModel because it would make things really complex. If anyone has any idea how to fix this then it would be highly appreciated!

semenvav:
@Programmer-0-0 You was on the right way. If we do it like this then I think everything will work.

```
    val categoryExpandedMapSaver: Saver<MutableMap<String, Boolean>, *> = Saver(
        save = { map -> map.toMap() },
        restore = { map -> mutableStateMapOf(*map.toList().toTypedArray()) },
    )

    val expanded= rememberSaveable(
        saver = categoryExpandedMapSaver,
        key = "CategoryExpandedMap",
        init = { mutableStateMapOf(*categoryMap.keys.toList().map { it to false }.toTypedArray()) },
    )
```
It fix the problem you described.
Thank you, you find things I didn't even think about )

app/src/main/java/eu/kanade/presentation/more/download/DownloadStatsScreenModel.kt
```diff
@@ -0,0 +1,357 @@
...
+        return size
+    }
+
+    private fun getFileSize(file: File): Long {
+        val archiveFormats = setOf(".zip", ".cbz", ".rar", ".cbr")
+        return if (file.isDirectory) {
+            getFolderSize(file.path)
+        } else if (file.isFile) {
+            file.length()
+        } else if (file.extension.lowercase() in archiveFormats) {
```
Mekanik, line 107: Isn't `file.isFile` condition above already handles archive files? In that case code above is never executed.
Also, for device storage management file size on disk is probably more important than size of files in an archive.

semenvav: You absolutely right, file.isFile handle archive files, my mistake

semenvav:
Based on the feedback i made many changes, there are concise changes
What has been done
- I want to make a system that can track downloaded/deleted chapters
- Made the screen divided into tabs, all contents of previous work are now on the entry tab
- The overall tab provides overview statistics about downloads and memory usage
- Improvement of the entry tab
- Fix bugs with the selection
- Fix some troubles with different libraryManga with the same manga
- A category element provides more information about its contents
- Add sort mode by downloaded chapter count
- added an option to show entries that have records of downloaded/deleted chapters but currently do not have downloaded chapters
- Replaced settings dropdown menus to one settings sheet
- On entry click open list that includes information about deleted/downloaded chapters for manga and its graph

semenvav:
Some of my thoughts on further development
- @Osyx comment directed me to create the overall tab, and I was thinking about how to group my statistics screen with the existing one, but on the way I met some problems: with a large weight of downloads, loading of the screen may take several seconds, I don't want to keep people waiting who only need ordinary statistics, so I'm still not sure what's the best way to do this
- I really like graphs, and I would like to make more graphs for statistics, such as the amount of reading by day
- it would be nice to take into account the orphans removed from the manga entries

semenvav:
Resolved conflicts with main. I also changed the formation of a size string to formatFileSize, for consistency with how @arkon made it [here](https://github.com/semenvav/tachiyomi/commit/cb8ea5eab0ec68ea02e8dc98c0a52771c460d5ab), after that on some devices the size of the files on the screen and in the explorer may differ and this is not a bug

arkon:
I haven't actually looked through this PR much yet, but I'd recommend:

- Putting the link to the download stats screen in the Data and Storage screen
- Splitting this into smaller PRs, starting with the new screen and some basic stats first; it's hard to review otherwise

semenvav:
ok, I moved it, I’ll try to divide it into parts in the next few days

PR #9790, opened 2023-08-02
Add Tachiyomi Cross Device Google Drive service.

KaiserBh:
Requires #9675.

I think everything is done and I fixed the conflict in #9675 do let me know if there is any feedback.


alessandrojean:
Just a curiosity question: Wouldn't it be more interesting to make usage of the [application data folder](https://developers.google.com/drive/api/guides/appdata) (with the `drive.appdata` scope) as some other apps uses to backup data (like Meta's WhatsApp)?

KaiserBh:
> Just a curiosity question: Wouldn't it be more interesting to make usage of the [application data folder](https://developers.google.com/drive/api/guides/appdata) (with the `drive.appdata` scope) as some other apps uses to backup data (like Meta's WhatsApp)?
> 
> Using a special app folder could prevent some issues like the users manually changing/corrupting the backup file (since the folder is hidden) and would allow seeing how much space Tachiyomi is using in a proper statistics screen in the Google Drive app/site.
> 
> Also, good to remember this:
> 
> > Import to note, we don't want to use the GMS libraries because of their licensing. We'd want to actually implement Google Drive integration using their REST API.
> > -- [#9326 (comment)](https://github.com/tachiyomiorg/tachiyomi/pull/9326#issuecomment-1518674123)

We could use the app folder there shouldn't be an issue with that I think.

Also regarding the GMS I am sure I removed them all.

The only two thing we are using from google are:
[google-api-services-drive](https://mvnrepository.com/artifact/com.google.apis/google-api-services-drive/v3-rev197-1.25.0)
[google-api-client-oauth](https://mvnrepository.com/artifact/com.google.oauth-client/google-oauth-client/1.34.1)

```prolog
google-api-services-drive = "com.google.apis:google-api-services-drive:v3-rev197-1.25.0"
google-api-client-oauth = "com.google.oauth-client:google-oauth-client:1.34.1"
```
Which are under Apache 2.0.

Turhvjbufv:
Not encrypted right?

KaiserBh:
> Not encrypted right? 

Yes, that's right, it's not encrypted.

AntsyLich:
I still think we should prioritize drive sync before the server one

KaiserBh:
> I still think we should prioritize drive sync before the server one

I want to clarify that both the drive sync and self-host options share the same internal logic for syncing. In essence, they are one and the same in terms of implementation.

Given this, I would propose focusing on this PR, as it encompasses both functionalities and is a more streamlined version compared to the original, extensive PR. If we all agree, we can close #9675 and concentrate our efforts on this more concise and efficient PR.

AntsyLich:
My thought process is more people will be able to test drive sync so we should implement that first. Take in people's suggestions and whatnot and after that add more sync methods like server, webdav etc.

app/src/main/java/eu/kanade/tachiyomi/data/sync/service/GoogleDriveSyncService.kt
```diff
@@ -0,0 +1,332 @@
...
+            InputStreamReader(context.assets.open("client_secrets.json")),
+        )
+
+        val tokenResponse: GoogleTokenResponse = GoogleAuthorizationCodeTokenRequest(
+            NetHttpTransport(),
+            jsonFactory,
+            secrets.installed.clientId,
+            secrets.installed.clientSecret,
+            authorizationCode,
+            "eu.kanade.google.oauth:/oauth2redirect",
```
undefiened, line 308: Sorry, I don't know how to create a PR to a PR, so I will just write it as a comment. Would it make sense to put the redirect URL in some constant in this class since it is repeated twice anyway (lines 211 and 308)? This is just a minor comment, please ignore it if it doesn't make any sense.

undefiened:
I built the branch and registered my own Google application since the provided client_secret is not working for the public yet without manual approval from your side (by the way, will Google ever approve this app for public use?).

I now wonder if it makes any sense to raise discussions here on usability, bugs, and minor things, or should I wait until this PR is merged?

For those as impatient as me, you can register your own Google app at `https://console.cloud.google.com/apis/dashboard` and generate your own Android OAuth 2.0 Client ID. Then you need to replace the `client_secret.json` with whatever Google generated and replace the redirect URLs (`eu.kanade.google.oauth:/oauth2redirect` in `/tachiyomi/data/sync/service/GoogleDriveSyncService.kt`) with the inverted client secret, something like `com.googleusercontent.apps.1234566MANYLETTERSANDDIGITS:/oauth2redirect`. You will also need to sign the APKs and give the key to Google in one of the app registration steps. If you did everything correctly, and deployed the app to your device, when you sign in to Google Drive, after a couple of warnings you should be redirected back into the app and you should see a "toast" notification telling you that you successfully authorized.

app/src/main/java/eu/kanade/tachiyomi/data/sync/service/GoogleDriveSyncService.kt
undefiened: Never mind, I think I figured out how to create PRs to PRs, so I opened https://github.com/KaiserBh/tachiyomi/pull/5 which also should fix the bug with the sync process reporting success when it actually should have failed.

preeefix:
2c. Given that the syncing behavior is fairly standardized on the Tachiyomi side, would it not make sense to implement the backends as different "extensions" (apks) like we do for sources and for "enhanced" trackers?

This allows anyone to be able to build a sync provider (GDrive, OneDrive, SyncYomi, etc.) without having to upstream the code directly into the core app. This also means that the sync provider can be built to have a UI Activity that allows things like reconfiguring the keys used for accessing the various APIs.

Yes, it'll be more work here on the front end, but in the long term it fits with the pre-existing model.

undefiened:
> 2c. Given that the syncing behavior is fairly standardized on the Tachiyomi side, would it not make sense to implement the backends as different "extensions" (apks) like we do for sources and for "enhanced" trackers?
> 
> This allows anyone to be able to build a sync provider (GDrive, OneDrive, SyncYomi, etc.) without having to upstream the code directly into the core app. This also means that the sync provider can be built to have a UI Activity that allows things like reconfiguring the keys used for accessing the various APIs.
> 
> Yes, it'll be more work here on the front end, but in the long term it fits with the pre-existing model.

I am not very knowledgeable about the security issues, but I personally would be more scared about giving access to my GDrive, OneDrive, or whatever cloud service to an external apk. There is some degree of trust in Tachiyomi, but a much lesser degree of trust in external people. The sources apks have no access to anything on my phone aside from doing whatever they are doing. I am not sure what enhanced trackers are, but I still would argue that losing access to animelist or something is not as serious as to my cloud. But this is just my 5 cents.

app/src/main/java/eu/kanade/presentation/more/MoreScreen.kt
```diff
@@ -144,7 +144,7 @@ fun MoreScreen(
             }
             item {
                 TextPreferenceWidget(
-                    title = stringResource(MR.strings.label_data_storage),
+                    title = stringResource(MR.strings.label_backup_and_sync),
```
arkon, line 147: Let's stick with the newer "Data and storage" naming.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDataScreen.kt
```diff
@@ -59,18 +65,37 @@ object SettingsDataScreen : SearchableSettings {
 
     @ReadOnlyComposable
     @Composable
-    override fun getTitleRes() = MR.strings.label_data_storage
+    override fun getTitleRes() = MR.strings.label_backup_and_sync
```
arkon, line 68: Same here re: naming.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDataScreen.kt
```diff
@@ -59,18 +65,37 @@ object SettingsDataScreen : SearchableSettings {
...
         PermissionRequestHelper.requestStoragePermission()
 
+        val syncPreferences = remember { Injekt.get<SyncPreferences>() }
+        val syncService by syncPreferences.syncService().collectAsState()
+
         return listOf(
             getBackupAndRestoreGroup(backupPreferences = backupPreferences),
             getDataGroup(),
-        )
+        ) + listOf(
```
arkon, line 82: Move this to a separate function.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDataScreen.kt
```diff
@@ -319,6 +344,212 @@ object SettingsDataScreen : SearchableSettings {
     }
 }
 
+@Composable
+private fun getSyncServicePreferences(syncPreferences: SyncPreferences, syncService: Int): List<Preference> {
+    val syncServiceType = SyncManager.SyncService.fromInt(syncService)
+    return when (syncServiceType) {
+        SyncManager.SyncService.NONE -> emptyList()
```
arkon, line 351: Just do an early return rather than this `emptyList() + emptyList()`.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDataScreen.kt
```diff
@@ -319,6 +344,212 @@ object SettingsDataScreen : SearchableSettings {
...
+            },
+        ),
+        getGoogleDrivePurge(),
+    )
+}
+
+@Composable
+private fun getGoogleDrivePurge(): Preference.PreferenceItem.TextPreference {
+    val scope = rememberCoroutineScope()
+    val showPurgeDialog = remember { mutableStateOf(false) }
```
arkon, line 381: Use `by remember`

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDataScreen.kt
```diff
@@ -319,6 +344,212 @@ object SettingsDataScreen : SearchableSettings {
...
+    }
+
+    return Preference.PreferenceItem.TextPreference(
+        title = stringResource(MR.strings.pref_google_drive_purge_sync_data),
+        onClick = { showPurgeDialog.value = true },
+    )
+}
+
+@Composable
+fun PurgeConfirmationDialog(
```
arkon, line 415: `private fun`

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDataScreen.kt
```diff
@@ -319,6 +344,212 @@ object SettingsDataScreen : SearchableSettings {
...
+            ),
+            Preference.PreferenceItem.InfoPreference(
+                stringResource(MR.strings.last_synchronization, relativeTimeSpanString(lastSync)),
+            ),
+        ),
+    )
+}
+
+@Composable
+fun SyncConfirmationDialog(
```
arkon, line 532: `private fun`

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsMainScreen.kt
```diff
@@ -209,7 +209,7 @@ object SettingsMainScreen : Screen() {
             screen = SettingsBrowseScreen,
         ),
         Item(
-            titleRes = MR.strings.label_data_storage,
+            titleRes = MR.strings.label_backup_and_sync,
```
arkon, line 212: Same re: naming

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupRestorer.kt
```diff
@@ -321,15 +317,17 @@ class BackupRestorer(
                     .copy(
                         id = dbChapter.id,
                         bookmark = updatedChapter.bookmark || dbChapter.bookmark,
+                        // Overwrite read status with the backup's status
+                        read = updatedChapter.read,
```
arkon, line 321: Wouldn't this change the existing backup behaviour where locally read content will become unread?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncDataJob.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        fun isAnyJobRunning(context: Context): Boolean {
+            return jobTagList.any { context.workManager.isRunning(it) }
+        }
+
+        fun setupTask(context: Context, prefInterval: Int? = null) {
+            val syncPreferences = Injekt.get<SyncPreferences>()
+            val interval = prefInterval ?: syncPreferences.syncInterval().get()
+            if (interval > 0) {
+                // Generate a random delay in minutes (e.g., between 0 and 15 minutes) to avoid conflicts.
```
arkon, line 69: What's the chances of conflicts? These jobs' timings eventually drift anyway.

In the case of a conflict, could it just retry (up to _n_ times)?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,262 @@
...
+ * A manager to handle synchronization tasks in the app, such as updating
+ * sync preferences and performing synchronization with a remote server.
+ *
+ * @property context The application context.
+ */
+class SyncManager(
+    private val context: Context,
+    private val handler: DatabaseHandler = Injekt.get(),
+    private val syncPreferences: SyncPreferences = Injekt.get(),
+    private var json: Json = Json {
```
arkon, line 45: Why a custom instance and not injecting the existing one?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,262 @@
...
+
+        if (remoteBackup != null) {
+            val (filteredFavorites, nonFavorites) = filterFavoritesAndNonFavorites(remoteBackup)
+            updateNonFavorites(nonFavorites)
+
+            val newSyncData = backup.copy(
+                backupManga = filteredFavorites,
+                backupCategories = remoteBackup.backupCategories,
+                backupSources = remoteBackup.backupSources,
+                backupBrokenSources = remoteBackup.backupBrokenSources,
```
arkon, line 121: Not sure we really need these old broken ones...

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,262 @@
...
+            if (backupUri != null) {
+                BackupRestoreJob.start(context, backupUri, sync = true)
+            } else {
+                logcat(LogPriority.ERROR) { "Failed to write sync data to file" }
+            }
+        }
+    }
+
+    private fun writeSyncDataToFile(context: Context, backup: Backup): Uri? {
+        val file = File(context.filesDir, "tachiyomi_sync_data.proto.gz")
```
arkon, line 142: Cache might be more appropriate?

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
```diff
@@ -0,0 +1,262 @@
...
+        return handler.awaitList { mangasQueries.getAllManga(::mapManga) }
+    }
+
+    /**
+     * Compares two Manga objects (one from the local database and one from the backup) to check if they are different.
+     * @param localManga the Manga object from the local database.
+     * @param remoteManga the BackupManga object from the backup.
+     * @return true if the Manga objects are different, otherwise false.
+     */
+    private suspend fun isMangaDifferent(localManga: Manga, remoteManga: BackupManga): Boolean {
```
arkon, line 169: Could maybe convert the `BackupManga` to a `Manga` and just do a simple equality check, unless there's some fields that are always different?

app/src/main/java/eu/kanade/tachiyomi/ui/setting/track/GoogleDriveLoginActivity.kt
```diff
@@ -0,0 +1,52 @@
...
+        val error = data?.getQueryParameter("error")
+        if (code != null) {
+            lifecycleScope.launchIO {
+                googleDriveService.handleAuthorizationCode(
+                    code,
+                    this@GoogleDriveLoginActivity,
+                    onSuccess = {
+                        Toast.makeText(
+                            this@GoogleDriveLoginActivity,
+                            "Authorization successful.",
```
arkon, line 24: These need to be localized.

domain/src/main/java/tachiyomi/domain/sync/SyncPreferences.kt
```diff
@@ -0,0 +1,38 @@
...
+    fun syncInterval() = preferenceStore.getInt("sync_interval", 0)
+
+    fun syncService() = preferenceStore.getInt("sync_service", 0)
+
+    private fun googleDriveAccessToken() = preferenceStore.getString(
+        Preference.appStateKey("google_drive_access_token"),
+        "",
+    )
+
+    fun setGoogleDriveAccessToken(accessToken: String) {
```
arkon, line 22: Aren't these setters a bit redundant?

gradle/libs.versions.toml
```diff
@@ -95,6 +95,10 @@ voyager-tab-navigator = { module = "cafe.adriel.voyager:voyager-tab-navigator",
 voyager-transitions = { module = "cafe.adriel.voyager:voyager-transitions", version.ref = "voyager" }
 
 ktlint = "org.jlleitschuh.gradle:ktlint-gradle:11.6.1"
+google-api-services-drive = "com.google.apis:google-api-services-drive:v3-rev197-1.25.0"
+google-api-client-oauth = "com.google.oauth-client:google-oauth-client:1.34.1"
+
+kotlinter = "org.jmailen.gradle:kotlinter-gradle:3.13.0"
```
arkon, line 101: Remove this.

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupRestorer.kt
KaiserBh: > Wouldn't this change the existing backup behaviour where locally read content will become unread?

Oh yea You're right, the way it's set up now could mess with the read status of chapters that people have already gone through locally pretty sure.

I was thinking the backup should always be the boss, but I get your concern.

So, here's a thought: I could fix this maybe something like adding a check if it's sync then  do what it currently does other wise just stick how the old version is where we prioritise the local db, I personally think that the backup should be prioritised anyway it's your call. It might make the code a bit more complex, but it could be a good middle ground.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
KaiserBh: > Why a custom instance and not injecting the existing one?

I need the encodeDefaults and ignoreUknownKeys configuration which by default is false, otherwise I keep getting JSON decode failed.

```prolog
7:15.419 12112-12177 SyncDataJob             eu.kanade.tachiyomi.debug            E  kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 2037: Expected start of the array '[', but had 'n' instead at path: $.backup.backupManga[0].history
2023-11-20 06:37:15.419 12112-12177 SyncDataJob             eu.kanade.tachiyomi.debug            E  JSON input: .....1],"viewer_flags":0,"history":null,"lastModifiedAt":17004190.....
2023-11-20 06:37:15.419 12112-12177 SyncDataJob             eu.kanade.tachiyomi.debug            E  	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:24)
2023-11-20 06:37:15.419 12112-12177 SyncDataJob             eu.kanade.tachiyomi.debug            E  	at kotlinx.serialization.json.internal.JsonExceptionsKt.JsonDecodingException(JsonExceptions.kt:32)```

It happens when I use  ` private val json : Json by injectLazy()`

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
KaiserBh: > Could maybe convert the `BackupManga` to a `Manga` and just do a simple equality check, unless there's some fields that are always different?

I think mainly the `last modified at` will always be different. When I made these checks, is to make sure to ignore manga that only have last_modified_at changed this way we are not restoring or going through unnecessary data which makes the sync faster.

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupRestorer.kt
KaiserBh: update: went with the solution I had, if it's sync then uses what I had before if it's not keep current logic.

app/src/main/java/eu/kanade/tachiyomi/data/sync/SyncManager.kt
KaiserBh: update: I tried simple equality check but it increases the sync time by a lot syncing/restoring unnecessary data, so current approach is optimised and we can exclude more fields if needed.

PR #9791, opened 2023-08-02
Screen stabilization

bringonthepretty:
Added screen stabilization feature for ReaderActivity. 
Corresponding settings entry can be found under Settings -> Reader -> Display.
Setting strength to 0 turns stabilization off.

### Images
| Image 1 | Image 2 | Image 3 |
| ------- | ------- | ------- |
|  |  |  |


arkon:
Interesting idea, but quite awkward in how it cuts off content to compensate, which isn't exactly desirable when you're trying to view the content.

bringonthepretty:
@arkon It looks awkward because i set strength to max to show what it does. 
Screen will return to default position quickly
Real experience looks like this: 
|https://github.com/tachiyomiorg/tachiyomi/assets/53440846/9a5676e5-45b7-4ced-bc46-944a267c9691

arkon:
I don't think any screen recording is really going to convince me that it's a nice feature given I can't tell if it's actually compensating actual movements well, especially when a lot of devices have not great accelerometers.

I can see this being useful for people with disabilities, but in general it seems too niche to incorporate.

PR #9797, opened 2023-08-03
Add support to kotlin.time APIs in the rate limit interceptor

alessandrojean:
I kept the old legacy method to maintain compatibility with the existing codebase for the extensions, but under the hoods it is already using the `kotlin.time` APIs by converting the types.

The existing code can be migrated after some time and the legacy method marked as obsolete.

This will allow the usage of the methods like so:

```kotlin
client.newBuilder()
    .rateLimit(1, 2.seconds)
    .rateLimitHost(baseUrl.toHttpUrl(), 1, 3.seconds)
    .build()
```

Extensions-lib PR: tachiyomiorg/extensions-lib#10.


AntsyLich:
Imo should put the new `rateLimitHost` inside `RateLimitInterceptor.kt` so we can remove the `SpecificHostRateLimitInterceptor.kt` file later

AntsyLich:
Also don't forget a pr against https://github.com/tachiyomiorg/extensions-lib

core/src/main/java/eu/kanade/tachiyomi/network/interceptor/SpecificHostRateLimitInterceptor.kt
```diff
@@ -1,27 +0,0 @@
...
- * httpUrl = "imagecdn.manga.com".toHttpUrlOrNull(), permits = 10, period = 2, unit = minutes  =>  10 requests per 2 minutes to imagecdn.manga.com
- *
- * @since extension-lib 1.3
- *
- * @param httpUrl {HttpUrl} The url host that this interceptor should handle. Will get url's host by using HttpUrl.host()
- * @param permits {Int}   Number of requests allowed within a period of units.
- * @param period {Long}   The limiting duration. Defaults to 1.
- * @param unit {TimeUnit} The unit of time for the period. Defaults to seconds.
- */
-fun OkHttpClient.Builder.rateLimitHost(
```
arkon, line 22: I think this actually needs to stay in this file for the stuff to match up.

arkon (changes requested):
Mark the old ones as deprecated too, and we can just outright remove them from extension-lib 1.5 and update everything at once so we can drop them from the codebase soon thereafter.

core/src/main/java/eu/kanade/tachiyomi/network/interceptor/SpecificHostRateLimitInterceptor.kt
AntsyLich: Yeah the old one needs to be kept in this file otherwise extensions will break. @stevenyomi tested this during the intercept rewrite.

Issue #6284, opened 2021-11-27
Ability to have colored bookmarks


AwayJob:
### Describe your suggested feature

So I loved manga/manhwa progress their story. I bookmark the sexy scene and a good plot twist. But sometimes I want to go back to that sexy scene or to see a really good drawing of that page. Since I have a bad memory of telling which is which.

In the list of chapters while in Select Mode. It should work like if you press the Bookmark Symbol long enough, it will show a variety of colors (I think it's a hassle to have a customizable color so just include the ROYGBV colors and it would be fine by me.) and you can choose the color of your choice. 
![concept](https://user-images.githubusercontent.com/66241829/143668264-5aa69e6f-45e1-4179-a2f3-18fdab334e79.png)
So this is my concept. I hope you guys approve. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have updated the app to version **[0.12.3](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

shawkik:
Also want to add....
Filter the chapters according to their color

AwayJob:
> Also want to add: **Filter** the chapters according to their color.

I like your idea too ngl

ShockedCoder:
I'd also like to add to this: ability to have notes in the bookmarks

For example: the red bookmarks are fight chapters(or chapters that have them) and there's a really nice song that goes well with the scene and your reading speed

Also also, the ability to choose any color you want instead of a predefined one.
What about bookmark groups? Like chapter categories there could be a "bookmarks category" where different groups of bookmarks have names

Issue #5046, opened 2021-05-08
[Feature Request] Bookmark tab similar to Updates tab


paulo27ms:
## Why/User Benefit/User Problem
Having a tab that'd make a list of all your bookmarked chapters would be a nice addition.

## What/Requirements
I wouldn't want this to be a tab next to Updates in the navbar so perhaps a subcategory/menu somewhere else would be better. Perhaps just have it be a menu in the 3 dot menu in the library tab.
This would basically work the same way as the Updates tab but would only display chapters you have bookmarked in the library.

MaxJohnson233:
I have a lot of bookmarked chapters, for art, humor or just something that was said; however to find bookmarks I'd have to go to the specific manga and look for bookmarks. I don't leave bookmarks on most series I read so a category or even section under the recents page which shows bookmarked chapters would be nice.

Issue #5043, opened 2021-05-08
[Feature Request] Namable bookmarks


AlbedoSuzuki:
## Why/User Benefit
This function would allow many users to know, for example, why they have bookmarked this particular chapter, as well as make it possible to find desired chapters more quickly.

## What/Requirements
1. it should be possible to name bookmarks.
2. since maybe not everyone wants to have this function, it should be possible to disable it in the settings.
3. the design should perhaps also be a bit variable for example:

[chapter name][bookmark name]

[bookmark name][chapter name]

or whatever you can think of.

disori99:
+1 this feature. Really hope this gets added.

DelgadoJosh:
+1 this feature as well, it would be great to have a way to take brief notes on chapters. It's difficult to remember what one of your favorite chapters are when they're labeled as "Chapter X" without any note.

Issue #565, opened 2016-12-05
Add ability to star pages and chapters


ywecur:
Right now there's no way to do this. It's useful if it's a point in the story you might want to come back to without having to manually write down the page number. 

Similar to adding a bookmark.

inorichi:
Sorry but I don't want to maintain too many features and clutter the UI. You can already save a single image in a folder, and IMO this is enough.

Maybe at some point, but not yet.

ywecur:
Look at how Play Books does it. You just tap the corner and a bookmark appears, no need for additional UI elements. 

![Before pressing upper right corner](http://i.imgur.com/vrTm5Hl.png)
![After pressing it](http://i.imgur.com/P2MUsL0.jpg)

Notice: No UI clutter at all, just make pressing that particular point equal to setting a bookmark.

inorichi:
But there has to be an UI section to browse bookmarked pages, no?

Only showing a bookmark icon in the reader is the same as not having anything IMO.

ywecur:
@inorichi That could be added to the right of the chapter section. This is how it's done in Play Books, which is very intuitive. 

![](http://i.imgur.com/Egj3qfD.jpg)

Having the bookmarks menu to the right won't add any clutter, and makes sense as they to are sorted in chronological order. 

![](http://i.imgur.com/8EwQzeJ.jpg)

inorichi:
Ok, I'll reopen this. Chapter bookmarking is already implemented though.

ywecur:
I understand. But being able to bookmark every page is much more useful, IMO. Many mangas have chapters with more than 70 pages, which makes the chapter bookmark pretty useless.

avalonv:
Any update on this feature? This would be incredibly useful.

SKY42000:
Will this page bookmarking feature be implemented in the near future? Many thanks.

Gaibuh:
Would also love this feature. Saving images is very nice but being able to go back and read from a certain point is also very nice

Mekanik:
I'm working on this feature -- page bookmarks with an optional note and a dedicated view to see all bookmarks and navigate directly to a bookmarked page.

If completed, it should address two other feature requests #5046 and #5043, if this per-page bookmarks merged with currently available chapter bookmarks.

I have a working prototype that adds page bookmarks as a separate feature without trying to merge with chapter bookmarks.

New bookmark page action.
[...]

PR #9798, opened 2023-08-04
Add per-page bookmarks with optional notes and new view to see bookmarked pages.

Mekanik:
1. Add ability to set/update/delete a page bookmark with optional text note.
Dialog is invoked in reader long tap menu (the one with “set as cover”, “share”, “save image”)

2. Add new view that displays all bookmarks for library items, both chapter and page, “More” -> “Bookmarks”.
    Implemented as two level navigation:
    - Top level view shows all manga with bookmarks, number of bookmarks and time of last update.
    - Click on a manga opens list of all bookmarks sorted by `chapter ASC, page ASC` with chapter, page and optional note
        - Click on a bookmark opens a reader activity on a bookmark page or on default page for chapter bookmark
        - Show toolbar icon to delete all bookmarks for a manga (with confirmation)

### Related feature requests
#565 (completed)
#5046 (completed)
#5043 (done for page bookmarks, notes for chapter bookmarks are in next iteration)
Could probably make multicolored bookmark request obsolete: #6284

### Implementation details
Bookmarks are stored in a new table.

On DB migration for the change a new record is added for all existing chapter bookmarks to the table (along with chapter boolean bookmark flag).
Same when an older format backup is restored when there are bookmarked chapters but no bookmark records.

Migrating  of bookmarks is changed - now chapter-level bookmarks are not merged unless “Bookmarks” is selected in merging dialog.
Selecting “Bookmarks” merges by inserting and updating, but never deleting existing page or chapters bookmarks. This is different from previous behavior when bookmarks were removed. When chapters merge is selected and source has no bookmarks.

Other features dependent on chapter bookmarks are not changed or affected by page bookmarks (filters, deletion).

Chapter-level bookmarks UI and data are not changed - still in chapters table.

When chapter bookmarks are set or removed, a correspondent record in a `bookmarks` table is inserted or deleted.

### Images
| Long tap menu new action "Bookmark page" | Dialog to add bookmark |
| ------- | ------- |
| image | image |

| Edit an existing page bookmark | Delete confirmation |
| ------- | ------- |
| image | image |

| New "Bookmarks" view | Top level aggregated by manga |
| ------- | ------- |
| image | image |

| View for a manga | Delete all for a manga |
| ------- | ------- |
| image | image |

| Create Backup | Migrate |
| ------- | ------- |
| image | image |


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderViewModel.kt
```diff
@@ -247,7 +259,15 @@ class ReaderViewModel(
                     val source = sourceManager.getOrStub(manga.source)
                     loader = ChapterLoader(context, downloadManager, downloadProvider, manga, source)
 
-                    loadChapter(loader!!, chapterList.first { chapterId == it.chapter.id })
+                    val chapter = chapterList.first { chapterId == it.chapter.id }
+                    // TODO: this is hacky solution, what is proper?
+                    //   Don't update requestedPage if it's already >= 0, as initialized as -1.
+                    //   But maybe it's sometimes not -1 but expected to be updated?
+                    if (pageIndex != null && pageIndex >= 0) {
```
Mekanik, line 266: I need help with a proper approach suggestion for this.

specterflare:
IMO calling them "bookmarks" could lead to some confusion if users conflate them with regular, whole-chapter bookmarks, especially considering you have them using the same icon (plus I just feel like it's bad design practice to present two totally different features identically). I would suggest calling them something else, like "tags", and using a different icon, maybe [sell](https://fonts.google.com/icons?selected=Material+Icons+Outlined:sell:&icon.query=tag&icon.set=Material+Icons)?

app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
```diff
@@ -150,6 +155,12 @@ class SyncChaptersWithSource(
         }
 
         val reAdded = mutableListOf<Chapter>()
+        val reAddedBookmarks = mutableListOf<BookmarkWithChapterNumber>()
+        val bookmarksByChapterNumber = if (toAdd.isEmpty()) {
+            emptyMap()
+        } else {
+            getBookmarks.awaitWithChapterNumbers(manga.id).groupBy { it.chapterNumber }
+        }
```
AntsyLich, line 163: Re uploads most of the time has different page count (scans vs official) using the old bookmarks will just have the wrong pages bookmarked

Mekanik: Correct, having bookmarks moved by couple of pages is very probable. Current implementation is best effort match.
What could be an acceptable workaround (assuming that the only solution is to do image-similarity based match and is unfeasible)?
Some options I can think about:
- Do a validation pass before actual changes and warn about any chapters with bookmarks that have different number of pages with options to continue or not.
- Do migration but show warnings about lost chapters, bookmarks, different page numbers in chapters, etc.

In general, current migration lacks validation, it's possible to migrate to a 0 chapters entity or to something completely different. Sounds like a complete pre-validation is a good feature to have to highlight possible issues, but is probably outside of scope for this.

Mekanik:
> Thanks so much for doing this! This has been a long desired feature of mine.
> 
> IMO calling them "bookmarks" could lead to some confusion if users conflate them with regular, whole-chapter bookmarks, especially considering you have them using the same icon (plus I just feel like it's bad design practice to present two totally different features identically). I would suggest calling them something else, like "tags", and using a different icon, maybe [sell](https://fonts.google.com/icons?selected=Material+Icons+Outlined:sell:&icon.query=tag&icon.set=Material+Icons)? I guess that would also mean changing the view panel to "Tags and bookmarks". IDK, just my two cents.

Thanks for providing a comment!
I'm mixed on this. I also thought about making it a **page tags** or **page notes** feature.
I think bookmark as a concept still works for page-level. Chapter bookmarks are not very usefull for cases when you have, for example, 4 volumes with 400+ pages each.
I really want both page and chapter bookmarks to become a similar feature not completely different, but this first version keeps them more separated trying to not to affect existing behavior.
Next, I’m planning to try the same dialog as for page bookmarks when long-tapped (maybe) on individual chapter bookmarks to unify them even more.

Pros of having them as "page tags":
- No confusion with existing chapter bookmarks.
- Simpler code logic in backups, migrations, batch bookmark changes for chapters.

Downsides:
- Doesn’t address “named bookmarks” feature request.
- Either need two views to address “separate view for all bookmarks” or a view that will mix two different types of bookmarks which will add confusion.

Thoughts on reducing confusion (more like a brainstorm dump):

- Show chapter as bookmarked if there are any page bookmarks in it (but what to do on unbookmark?).
- Make toolbar bookmark action by default show same page bookmark dialog but allow checkbox to do a chapter level (complicates existing UX flow, more clicks to bookmark).
- Make “bookmark” to be used only as page-level bookmark (you can mark the first page if want existing chapter-like behavior); then rename current chapter-level bookmarks to “favorite chapters” with “star” as an icon (not good at all for existing users).

Lzmxya:
> * Make “bookmark” to be used only as page-level bookmark (you can mark the first page if want existing chapter-like behavior); then rename current chapter-level bookmarks to “favorite chapters” with “star” as an icon (not good at all for existing users).

Usually the term “bookmark” refers to marking pages in a book. I think this solution (renaming the existing chapter-based bookmarking feature to “star”) is a better way to go.

arkon:
Changing existing chapter-level bookmarks to be called stars seems reasonable to me. That should be a separate change though.

PR #9805, opened 2023-08-05
Fix migration flags usage (incorrect defaults and copy mode)

Mekanik:
Two bugs in migrate dialog related to "data to include" flags.
1. “Copy” under migration ignores current selection of “data to include” and uses value from a previous “Migrate”.
2. New migration incorrectly restores previous selections as relies on indexes in array ignoring custom options per manga.

**Fixed by:**
- Introduce a data class to use as part of a state.
- Remove reliance on global variables of MigrationFlags and make it stateless.
- Use preference as auxiliary, and pass selected flags to migrate method explicitly.

### Images
| Delete Downloaded is not selected | Delete is selected as on a same position as cover from a previous migrate |
| ------- | ------- |
| Pasted Graphic | Pasted Graphic 1 |

There could be more improvements, defaults selection bitmap for options that are not presented for a current migration should not even be updated. But not the point of this PR.


arkon:
Thanks!

Issue #6779, opened 2022-03-14
Some Exception strings are hardcoded and not translatable


alessandrojean:
### Describe your suggested feature

Currently, the app still have some strings that are hardcoded in the code. It would be good if they're translatable.

https://github.com/tachiyomiorg/tachiyomi/blob/95b253db0984c87750b1cb284007754bf6c31880/app/src/main/java/eu/kanade/tachiyomi/network/OkHttpExtensions.kt#L127

https://github.com/tachiyomiorg/tachiyomi/blob/2e127dff1f0b00f6a92359a07132c6016db6ea36/app/src/main/java/eu/kanade/tachiyomi/extension/installer/ShizukuInstaller.kt#L61

https://github.com/tachiyomiorg/tachiyomi/blob/2e127dff1f0b00f6a92359a07132c6016db6ea36/app/src/main/java/eu/kanade/tachiyomi/extension/installer/ShizukuInstaller.kt#L65

https://github.com/tachiyomiorg/tachiyomi/blob/2e127dff1f0b00f6a92359a07132c6016db6ea36/app/src/main/java/eu/kanade/tachiyomi/extension/installer/ShizukuInstaller.kt#L70

https://github.com/tachiyomiorg/tachiyomi/blob/3dc4fd8dd1c121080eedfd11e0d693e2bd1c3d60/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt#L93

https://github.com/tachiyomiorg/tachiyomi/blob/3dc4fd8dd1c121080eedfd11e0d693e2bd1c3d60/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt#L145

https://github.com/tachiyomiorg/tachiyomi/blob/0cb4094dd9bbc9f3c9f493bb8b1d01dd4b87b1c8/app/src/main/java/eu/kanade/tachiyomi/source/online/HttpSource.kt#L210

https://github.com/tachiyomiorg/tachiyomi/blob/ff369010074b058bb734ce24c66508300e6e9ac6/app/src/main/java/eu/kanade/tachiyomi/data/track/shikimori/ShikimoriInterceptor.kt#L21

https://github.com/tachiyomiorg/tachiyomi/blob/ff369010074b058bb734ce24c66508300e6e9ac6/app/src/main/java/eu/kanade/tachiyomi/data/track/kitsu/KitsuInterceptor.kt#L21

[...]

PR #9811, opened 2023-08-05
Make some error messages localizable

Mekanik:
Allow translation of 2 exception messages in BackupManager and "Licensed - No chapters to show".

Feature request: #6779
There are other non-localized exceptions in the system and mentioned in a feature request, but at least some are not meant to be shown to a user. 

image


source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/HttpSource.kt
```diff
@@ -404,3 +404,5 @@ abstract class HttpSource : CatalogueSource {
      */
     override fun getFilterList() = FilterList()
 }
+
+class LicensedMangaChaptersException : Exception("Licensed - No chapters to show")
```
curche, line 408: Shouldn't this use the newly added `licensed_manga_chapters_error` from `strings.xml`

Mekanik: Resolution into a string needs a context which is not available here. This string is left as a fallback so the message will appear in logs and in case exception is not handled by formattedMessage but shown to a user, there's still some text.

app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupManager.kt
```diff
@@ -106,10 +106,10 @@ class BackupManager(
                     UniFile.fromUri(context, uri)
                 }
                 )
-                ?: throw Exception("Couldn't create backup file")
+                ?: throw Exception(context.getString(R.string.create_backup_file_error))
 
             if (!file.isFile) {
-                throw IllegalStateException("Failed to get handle on file")
+                throw IllegalStateException(context.getString(R.string.get_backup_file_handle_error))
```
arkon, line 112: I'm not even really sure this ever happens in a user-facing fashion.

Mekanik: Let me remove this `!file.isFile` from `R.string.` then to not to spend translators time on this.

PR #9831, opened 2023-08-10
Add local chapter

semenvav:
Continuation of the work to eliminate orphans started in #9621. This is an implementation of my vision of Arkon's proposal.

The main idea is that when migrating a manga with download chapters and turning off the flag for deleting downloaded, they are moved to the new manga.
In this case, I want to move the chapter files to the new manga folder and create new entries in the database for them. The main features of these chapters are that they should be marked, ignored by any means of auto-deletion, not conflict with chapters from the new manga with the same names, and have the ability to be completely deleted by the user.
To do this, I added a new field to the chapter, localChapter, which is needed to mark chapters that are not related to the source, and came there during the migration or other user actions (there are still cases where this can be applied), they are marked with an Indicator on the right, replacing the usual download indicator, they are not deleting by manga sync and any auto-delete settings, and when deleted by the user, a modified alert appears, in case of deletion, complete deletion chapters, both file and entry from the database. If the user has deleted the chapter manually, then when trying to open it, he will receive a notification and it will be removed from the list when updating

| Migrated chapters| delete alert |
| ------- | ------- |
| migrated | alert |

In general, this pr is needed to discuss the idea and show how I see it. If such a solution is acceptable, then it can also be used for other cases with orphans, for example, when synchronizing with the source, the downloaded chapter can be deleted from the database.

And a few more thoughts on the topic that I have been working on but am not yet sure of their appropriateness and would like to discuss:
1. How do you like the idea of making the ability to manage local chapters in the application? I was working on additional features for local chapters: editing and moving, I wanted to make it possible to edit chapter parameters and shuffle chapters between mangas, but I ended up putting it on pause. This functionality was made with an eye to the next item
2. I wanted to make it possible to manage chapters in local mangas directly in the application, I mean delete/move/change them and from here I have two questions: how about using the localChapter system for local manga chapters? What is the idea of a separate folder for the local source rather than making it a folder like all the other sources in the downloads folder, I understand that the current option seems more intuitive, but it would make it possible to use existing methods to manage chapters.

### DIFF
--- a/app/src/main/java/eu/kanade/domain/chapter/interactor/SetReadStatus.kt
+++ b/app/src/main/java/eu/kanade/domain/chapter/interactor/SetReadStatus.kt
@@ -48,6 +48,7 @@ class SetReadStatus(
 
         if (read && downloadPreferences.removeAfterMarkedAsRead().get()) {
             chaptersToUpdate
+                .filterNot { it.localChapter }
                 .groupBy { it.mangaId }
                 .forEach { (mangaId, chapters) ->
                     deleteDownload.awaitAll(
--- a/app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
+++ b/app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
@@ -80,7 +80,7 @@ class SyncChaptersWithSource(
         val toDelete = dbChapters.filterNot { dbChapter ->
             sourceChapters.any { sourceChapter ->
                 dbChapter.url == sourceChapter.url
-            }
+            } || dbChapter.localChapter
         }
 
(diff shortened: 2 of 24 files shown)


Mekanik:
1. I think, this type of functionality could be better to have under a separate flag option in migration dialog.
2. Is common scenario to merge with itself to convert all to local or to merge different entities to have combined chapters?
If the first, after migration with "migrate chapters" + "move to local" both enabled, will you end up with duplicate local & non-local chapters? 
Could it be more convenient to have an explicit option to "migrate to local" instead of combining with other options of migration UI, as it could be hard to reason about what will happen for different combinations of checkboxes?
Migrated manga with local chapters then could be migrated to other mangas if needed following the logic implemented in the PR, but act of converting to local will be separated.

PR #9853, opened 2023-08-19
Feature subcategory

### LINKED ISSUE #1803: [Request] Subcategory
**App version:**
0.8.0

**Issue/Request:**
    I would like to make a suggestion: to be able to add subcategories. Ex: when I create a category "Mangas", could have as add subcategories as the genres of manga.

Gfadebayo:
This PR works on and should eventually close #1803.
Like the title says, it adds subcategories to the app
Some notes about it includes:
- A subcategory is still a category, every feature and use case possible with a category is available in a subcategory.
- A subcategory is also a library item, so some features available to them is also available to it.
- Subcategory is implemented with many to many relations in mind, that is, 1 parent to many children and 1 child to many parents. This allows moving (or copying) a subcategory to different parents
- A subcategory can also have it's own subcategories

This PR isn't complete, some features are not implemented yet and it also has issues in certain areas:
- On clicking a subcategory, the selected tab uses the index of the parent instead of the index of the selected subcategory
- I have no idea on how the List layout should look like
- Displaying covers in the subcategory layout uses MangaCover.Book but trying to properly constraint a cover breaks the layout
- Edit categories screen shows all categories when it should instead only show the main categories but allow clicking to Edit categories under the clicked category
- Change category dialog used to copy mangas to different categories should be used for subcategories too, however the simple list dialog handling both cases can lead to a confusing experience (rules also need to be defined on which categories a subcategory can be copied to)

https://github.com/tachiyomiorg/tachiyomi/assets/43231926/d1e6ca42-c4b1-426f-bdee-6504b8a225b8

| Badge shows remaining number of items inside a subcategory and tab pill shows subcategory count and manga count | 
| ---------------------------- |

Screenshot_20230819_034324


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -644,23 +664,84 @@ class LibraryScreenModel(
...
+    }
+
+    fun enterSubcategoryForItem(item: LibraryManga): Boolean {
+        return if (state.value.categories.fastAny { it.id == item.category }) {
+            false
+        } else {
+            mutableState.update {
+                val newState = it.copy(parentCategory = item.parentCategory)
+
+                activeCategoryIndex = newState.categories.indexOfFirst { it.id == item.category }
```
Gfadebayo, line 716: activeCategoryIndex doesn't change here, maybe cause the value isn't committed immediately?. Maybe a local variable not attached to the preference except on initial read can be used instead

PR #9795, opened 2023-08-03
Added In-app list for failed library updates

Programmer-0-0:
This should finally (and hopefully) close #4616. Here are some images about the newly added feature:

<details>
<summary>Details</summary>

| The warning icon is disabled and not visible by default but when an update job is run, and an error is detected it will become enabled| For now I only did it as a list (no grid view) and the migrate button has the same function as going to the migrate source screen|
| -------------   | ------------- |
WhatsApp Image 2023-08-02 at 22 24 46|                    WhatsApp Image 2023-08-02 at 22 24 46
| I took @T3sT3ro advice and made the Pills to display the error count along side with the total count of mangas in that source| In the bottom bar there is a delete button as well as a dismiss button|
WhatsApp Image 2023-08-02 at 22 24 48 |        WhatsApp Image 2023-08-02 at 22 24 50    
| There is also an option to group by category but I couldn't find a way to efficiently get the total manga count in that category   |I edited the EmptyScreen function to display happy faces when you dismiss or ignore all the errors|
WhatsApp Image 2023-08-02 at 22 34 39 | WhatsApp Image 2023-08-02 at 22 34 39

</details>

<details>
<summary>Extra Image</summary>

WhatsApp Image 2023-08-02 at 22 24 48

</details>

As I'm still a university student, I don't have much experience with Koltin, so I'd like to give most of the credit of this PR to @semenvav as I took his code from #9737 as reference to create this feature. 

If you notice any mistakes in the code or if you have a more efficient way of writing it, then please notify me so I can fix it.

PR #9883, opened 2023-08-27
Added more common exceptions

Programmer-0-0:
Added more common exceptions that will be used in #9795


app/src/main/java/eu/kanade/presentation/util/ExceptionFormatter.kt
```diff
@@ -27,7 +27,44 @@ val Throwable.formattedMessage: String
...
+            "Exception" -> message ?: className
+            // General networking exceptions
+            "SocketException" -> getString(R.string.exception_socket_error)
+            "ConnectException" -> getString(R.string.exception_connect)
+            "BindException" -> getString(R.string.exception_bind_port)
+            "InterruptedIOException" -> getString(R.string.exception_io_interrupted)
+            "HttpRetryException" -> getString(R.string.exception_http_retry)
+            "PortUnreachableException" -> getString(R.string.exception_port_unreachable)
+            // General IO-related exceptions
+            "IOException" -> if (isOnline()) getString(R.string.exception_io_error) else getString(R.string.exception_offline)
```
alessandrojean, line 39: IOException is used inside the extensions interceptors to throw custom errors, so needs to be kept with the Exception guard as it was before.

alessandrojean (changes requested):

app/src/main/java/eu/kanade/presentation/util/ExceptionFormatter.kt
```diff
@@ -27,7 +27,44 @@ val Throwable.formattedMessage: String
...
+            "NoRouteToHostException" -> getString(R.string.exception_route_to_host)
+            // URL & URI related
+            "URISyntaxException" -> getString(R.string.exception_uri_syntax)
+            "MalformedURLException" -> getString(R.string.exception_malformed_url)
+            // Authentication & Proxy
+            "ProtocolException" -> getString(R.string.exception_protocol_proxy_type)
+            // Concurrency & Operation-related
+            "CancellationException" -> getString(R.string.exception_cancelled)
+            "InterruptedException" -> getString(R.string.exception_interrupted)
+            "IllegalStateException" -> getString(R.string.exception_unexpected_state)
```
alessandrojean, line 64: Some extensions also throws this with custom error messages while using `checkNotNull`, so needs to be removed as well.

alessandrojean:
I think a good start would be addressing the fetching errors first, which are thrown by OkHttp and can't be manually thrown from inside the extensions.

stevenyomi:
I think most of these strings are useless. Most of the exceptions are rarely seen, and explaining them doesn't enable users to troubleshoot and solve the issue themselves.

PR #9918, opened 2023-09-09
[Download Queue] Move series to bottom

joshuaxu71:
Issue: https://github.com/tachiyomiorg/tachiyomi/issues/9816
This PR adds an option for user to move series to bottom in the download queue page. I've added text for English only since I'm assuming the translation for other languages will be done with Weblate because of the conversation in this PR: https://github.com/tachiyomiorg/tachiyomi/pull/6794

image


arkon:
> I'm assuming the translation for other languages will be done with Weblate 

Yup.

PR #9937, opened 2023-09-17
Run Netlify Build Hook after Release

Soitora:
Requires adding `NETLIFY_HOOK_RELEASE` as Secret ( `${{ secrets.NETLIFY_HOOK_RELEASE }}` ).

Testing was first done on Preview tachiyomiorg/tachiyomi-preview#10
Result seemed good enough

image


arkon (changes requested):
This will happen on every push. You should add a condition like the other steps that only occur for tagged releases (`if: startsWith(github.ref, 'refs/tags/') && github.repository == 'tachiyomiorg/tachiyomi'`)

Soitora:
> This will happen on every push. You should add a condition like the other steps that only occur for tagged releases (`if: startsWith(github.ref, 'refs/tags/') && github.repository == 'tachiyomiorg/tachiyomi'`)

Fixed it 👍

Is there a reason you haven't dropped the build part to another job and have the entire job be behind the if statement?

arkon:
Mostly just so I don't need to deal with moving artifacts between jobs.

PR #9941, opened 2023-09-17
Use access token instead for Bangumi tracker

KayZhang52:
Changed the bangumi tracker's authentication to a dialog. User will have to go to the new token site, generate the token and add it here with the expiry date. Does solve the 7 day problem though.
  

screenshot


app/build.gradle.kts
```diff
@@ -20,7 +20,7 @@ android {
     namespace = "eu.kanade.tachiyomi"
 
     defaultConfig {
-        applicationId = "eu.kanade.tachiyomi"
+        applicationId = "eu.kanade.tachiyomi_bangumi_tracker_expiry"
```
alessandrojean, line 23: This shouldn't be commited.

app/src/main/AndroidManifest.xml
```diff
@@ -35,7 +35,7 @@
         android:enableOnBackInvokedCallback="true"
         android:hardwareAccelerated="true"
         android:icon="@mipmap/ic_launcher"
-        android:label="@string/app_name"
+        android:label="Tachiyomi_track_bangumi_fix"
```
alessandrojean, line 38: Neither this.

alessandrojean (changes requested):

mutsumi63:
> Changed the bangumi tracker's authentication to a dialog. User will have to go to the new token site, generate the token and add it here with the expiry date. Does solve the 7 day problem though.
> 
> screenshot

In the dialog box, there may be a need for some explanations and URL link.

app/src/main/AndroidManifest.xml
KayZhang52: > Neither this.

Removed. My bad.

mutsumi63:
image

image

Click the tracker again will reset the local records to empty/0.

Click rating button causes a crash.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsTrackingScreen.kt
```diff
@@ -281,6 +290,102 @@ object SettingsTrackingScreen : SearchableSettings {
...
+        var processing by remember { mutableStateOf(false) }
+        var inputError by remember { mutableStateOf(false) }
+
+
+        AlertDialog(
+            onDismissRequest = onDismissRequest,
+            title = {
+                Row(verticalAlignment = Alignment.CenterVertically) {
+                    Text(
+                        text = "Enter access token for Bangumi",
```
arkon, line 318: Any text in the UI should be added to `strings.xml` instead.

app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsTrackingScreen.kt
```diff
@@ -281,6 +290,102 @@ object SettingsTrackingScreen : SearchableSettings {
...
+                        label = { Text(text = "access token") },
+                        visualTransformation = VisualTransformation.None,
+                        keyboardOptions = KeyboardOptions(
+                            keyboardType = KeyboardType.Text,
+                            imeAction = ImeAction.Done,
+                        ),
+                        singleLine = true,
+                        isError = inputError && !processing,
+                    )
+                    OutlinedTextField(
```
arkon, line 344: Why does the user need to specify this?

gradle/kotlinx.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 kotlin_version = "1.9.10"
 serialization_version = "1.6.0"
-xml_serialization_version = "0.86.2"
+xml_serialization_version = "0.86.1"
```
arkon, line 4: These dependency downgrades should be removed.

arkon (changes requested):
Seems like @mutsumi63 pointed out some functional issues too?

mutsumi63:
> Seems like @mutsumi63 pointed out some functional issues too?

I added some details.

arkon:
Assuming this is abandoned.

PR #9985, opened 2023-10-06
Reduce recomposition of MangaHeader

Iamlooker:
There was a flawed use of `derivedStateOf` function in `MangaScreen` [here](https://github.com/tachiyomiorg/tachiyomi/blob/master/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt#L283)

Current implementation is almost like use of `mutableStateOf`
By replacing it with new implementation we can reduce calculation on scroll.

I also added animation **labels**, this will help when spectating animation in _Android Studio_


Iamlooker:
I suppose the linting issue is fixed and also I have added some more optimizations.

Iamlooker:
Btw `fastAny` or `fastAll` is still unsafe to call in compose code.

If you want I can fix that in this same PR

It is mainly present in [MangaScreen](https://github.com/Iamlooker/tachiyomi/blob/6573610e815e780fc718123ca49c8efe3cc12568/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt) only

arkon:
Sure

Iamlooker:
**Reason to use `derivedStateOf` in `isAnySelected`:**

There is a chance that we select one chapter and then select more, which will cause recalculation of the variable, and is avoided using the side-effect.

But...
There is a issue which is actually pretty deep-rooted, Selection state of the chapter is part of the parent data class. This will cause recomposition either way.

So in future if you plan to refactor the `Chapter` data class you will not need to update this logic.

Iamlooker:
I believe its ready to merge. If you have any suggestions you can say

arkon:
Thanks!

PR #9996, opened 2023-10-08
Allow extensions to open manga or chapter by URL

VIPlearner:
Added functions to ResolvableSource to cater for getting chapters via links 
Automatically load MangaScreen and ReaderActivity from links

Attached is the issue link
https://github.com/tachiyomiorg/tachiyomi/issues/9188

https://github.com/tachiyomiorg/tachiyomi/assets/63104739/54f071a6-a3ba-47c1-afbe-44e2b8577391

ezgif com-resize

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
@@ -11,10 +11,10 @@ import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.source.model.SChapter
 import eu.kanade.tachiyomi.source.model.SManga
 import eu.kanade.tachiyomi.source.online.ResolvableSource
+import eu.kanade.tachiyomi.source.online.UriType
 import kotlinx.coroutines.flow.update
-import logcat.LogPriority
-import logcat.logcat
 import tachiyomi.core.util.lang.launchIO
+import tachiyomi.core.util.system.logcat
 import tachiyomi.domain.chapter.interactor.GetChapterByUrlAndMangaId
 import tachiyomi.domain.chapter.model.Chapter
 import tachiyomi.domain.manga.interactor.GetMangaByUrlAndSourceId
@@ -31,29 +31,33 @@ class DeepLinkScreenModel(
     private val networkToLocalManga: NetworkToLocalManga = Injekt.get(),
     private val getChapterByUrlAndMangaId: GetChapterByUrlAndMangaId = Injekt.get(),
     private val getMangaByUrlAndSourceId: GetMangaByUrlAndSourceId = Injekt.get(),
-    private val syncChaptersWithSource: SyncChaptersWithSource = Injekt.get()
+    private val syncChaptersWithSource: SyncChaptersWithSource = Injekt.get(),
 ) : StateScreenModel<DeepLinkScreenModel.State>(State.Loading) {
-
 
     init {
         coroutineScope.launchIO {
             val source = sourceManager.getCatalogueSources()
                 .filterIsInstance<ResolvableSource>()
-                .filter{ it.id.toString() !in sourcePreferences.disabledSources().get() }
-                .firstOrNull { it.canResolveUri(query) }
+                .filter { it.id.toString() !in sourcePreferences.disabledSources().get() }
+                .firstOrNull { it.getUriType(query) != UriType.Unknown }
 
-            val manga = source?.getManga(query)?.let { getMangaFromSManga(it, source.id) }
+            val manga = source?.getManga(query)?.let {
+                getMangaFromSManga(it, source.id)
+            }
 
-            val chapter = if(source?.isChapterUri(query) == true && manga != null){
+            val chapter = if (source?.getUriType(query) == UriType.Chapter && manga != null) {
                 source.getChapter(query)?.let { getChapterFromSChapter(it, manga, source) }
-            }else null
+            } else {
+                null
+            }
 
             mutableState.update {
                 if (manga == null) {
                     State.NoResults
                 } else {
-                    if(chapter == null) State.Result(manga)
-                    else {
+                    if (chapter == null) {
+                        State.Result(manga)
+                    } else {
                         State.Result(manga, chapter.id)
                     }
                 }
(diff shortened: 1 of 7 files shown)


source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/ResolvableSource.kt
```diff
@@ -17,10 +18,31 @@ interface ResolvableSource : Source {
      */
     fun canResolveUri(uri: String): Boolean
```
arkon, line 19: Maybe we can have a sealed interface that captures chapter or manga or neither, which would eliminate the need for `isChapterUri`/`isMangaUri`?

app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
```diff
@@ -15,25 +27,65 @@ import uy.kohesive.injekt.api.get
...
 
+
     init {
         coroutineScope.launchIO {
-            val manga = sourceManager.getCatalogueSources()
+            val source = sourceManager.getCatalogueSources()
                 .filterIsInstance<ResolvableSource>()
-                .filter { it.canResolveUri(query) }
-                .firstNotNullOfOrNull { it.getManga(query)?.toDomainManga(it.id) }
+                .filter{ it.id.toString() !in sourcePreferences.disabledSources().get() }
```
arkon, line 42: Why?

app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
```diff
@@ -15,25 +27,65 @@ import uy.kohesive.injekt.api.get
...
+                .firstOrNull { it.canResolveUri(query) }
+
+            val manga = source?.getManga(query)?.let { getMangaFromSManga(it, source.id) }
+
+            logcat{ "isChapterUri ${source?.isChapterUri(query)}" }
```
arkon, line 47: Can remove the log statements.

app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
```diff
@@ -15,25 +27,65 @@ import uy.kohesive.injekt.api.get
...
+            }else null
+
+            logcat{ "chapter $chapter" }
 
             mutableState.update {
                 if (manga == null) {
                     State.NoResults
                 } else {
-                    State.Result(manga)
+                    val localManga = networkToLocalManga.await(manga)
```
arkon, line 59: Is this really necessary now?

app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
```diff
@@ -15,25 +27,65 @@ import uy.kohesive.injekt.api.get
...
+                        State.Result(manga, chapter.id)
+                    }
                 }
             }
         }
     }
 
+    private suspend fun getChapterFromSChapter(sChapter: SChapter, manga: Manga, source: Source): Chapter? {
+        val localChapter = try{
+            getChapterByUrlAndMangaId.await(sChapter.url, manga.id)!!
```
arkon, line 71: The non-null assertion seems unnecessary considering we handle nulls right after.

source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/ResolvableSource.kt
VIPlearner: Hmmm...Okay
Are you saying we should change canResolveUri or we should have another function that does this check
What I'm saying is Do we want canResolveUri and this new function or this new function should replace canResolveUri totally

app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
VIPlearner: Hi @arkon I'm not sure exactly what you're referencing here
I added the disabledSources because while testing (with the MangaPark Extension) it kept returning a source with a different language than I actually wanted so the only way to stop that was to disable the other languages and add this check

app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
VIPlearner: Removed

app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
VIPlearner: No it isn't 
removing now

source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/ResolvableSource.kt
arkon: Considering ext-lib 1.5 hasn't launched, we can change `canResolveUri`.

arkon (commented):
You'll also want to run `./gradlew ktlintFormat` at some point.

source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/ResolvableSource.kt
VIPlearner: Making the changes

VIPlearner: Made the changes You can check them out

app/src/main/java/eu/kanade/tachiyomi/ui/deeplink/DeepLinkScreenModel.kt
VIPlearner: Hi @arkon do you think we should leave this in
The issue with this is that there could be multiple sources that can resolve a URI especially with extensions that have multiple sources with different languages and same uri so How do we determine the source to use 
because using the first one could be very arbitrary and random

arkon: Yeah some other people have brought that up before too. We'd realistically probably want to end up with something like the Global Search UI with every source that says it can resolve it, but that change can happen independently.

VIPlearner: Okay understood
Will remove it then

VIPlearner:
> You'll also want to run `./gradlew ktlintFormat` at some point.

Done

PR #9999, opened 2023-10-09
Remove too many mistakes from strings.xml 

TarunRaghav3699:


arkon:
Thanks, but translations happen on Weblate: https://github.com/tachiyomiorg/tachiyomi/blob/master/CONTRIBUTING.md#translations

TarunRaghav3699:
But you can use them on weblate also

arkon:
Modifying the translations directly tends to break the Git integration that's set up with Weblate. Please just contribute the translations on Weblate directly and they'll make it back into this repo via the normal process.

PR #10005, opened 2023-10-12
Added library sort by mean Tracker score

CalebMorris:
## What did I change?

I've added the ability to sort by tracker score

### UI

When you have at least 1 tracker signed in, you see a new item in `Library` tab's `Sort` configuration

### Functionality

Sorting the library by `Tracker score` uses the mean of normalized scores across all the tracker values to construct the order.

Originally, I wanted to create sort-by selection for each logged-in tracker much like the filter listing, but upon diving into the `flag` system for handling sorting, there doesn't appear to be a simple way to generating a dynamic set of additional sort items and have them preserve correctly with the current system (it would need some additional data passed along). Honestly, I like having an average after some reflection.

## Demo Image

Tachiyomi Demo Screenshot - Tracker Score Sort

## Testing

* Sort option is not present when no tracker is logged in
* Sort option is present when 1 or 2 trackers are logged in
* Sorted ascended with 1 track and confirmed order matches expectations
* Sorted descended with 1 track and confirmed order matches expectations
* Sorted with 2 tracker scores on items and confirmed order
** Extension - used 2 trackers with divergent value ranges (Anilist & MangaUpdates)

#### Removing Tracker Doesn't Clean Up

1. Added tracking for Anilist & MangaUpdates to a manga
2. Sign out of MangaUpdates
3. Debugging when sorting by tracking-mean shows the MangaUpdates tracking is still present

I'm honestly not sure if this is a bug since I'm not familiar enough with how clean up mechanisms work in the app


CalebMorris:
Ah, missed the unit test to confirm that the number of flags hasn't changed

arkon:
> I'm honestly not sure if this is a bug since I'm not familiar enough with how clean up mechanisms work in the app

Yeah, bound tracks aren't really removed, they're just no longer attempt to be updated. If you re-add the tracking service later on, the bound tracks become useful again.

PR #10012, opened 2023-10-14
[dev QoL] Added AndroidStudio previews for [presentation.history] namespace

CalebMorris:
## What?

Android Studio will now render previews for each component within the namespace (`HistoryDialogs`, `HistoryItem`, and `HistoryScreen`). This makes it faster to change the UI than having to deploy to device.

### Change to Actual Production Code

The `HistoryScreen` class had an inner IoC injection for UIPreferences which I moved to up to the public outer class so it can be overwritten in the preview. The preview overrides with an in-memory preferences construct.

### HistoryDialogs

Screenshot 2023-10-13 at 9 04 20 PM

### HistoryItem

Screenshot 2023-10-13 at 9 04 32 PM

### HistoryScreen

Screenshot 2023-10-13 at 9 04 05 PM


arkon (commented):
You'll want to run `./gradlew ktlintFormat` to auto-fix some linting errors.

PR #10022, opened 2023-10-15
[dev QoL] Added AndroidStudio previews for [presentation.track] namespace

CalebMorris:
## What?

Android Studio will now render previews for each component within the `eu.kanade.presentation.track` namespace. This makes it faster to change the UI than having to deploy to device.

#### Tracker

Changed `Tracker` to `BaseTracker` and abstracted all `val`s and `fun`s to an interface named `Tracker`. The original abstract class tracker was coupled to the IoC injection which make it difficult to add preview providers. Creating this interface allows for much easier stubbing without any breaking changes.

#### TrackInfoDialogHome Color

The preview system with Compose likes to assume the color theme configuration for some elements, but not others. This results in the dark-preview to have black text on a dark background. The `TrackInfoDialogHome` component wasn't being assumed, so I've added the theme `MaterialTheme.colorScheme.onSurface` which matches the inherited coloration as far as I can tell.

### TrackLogoIcon

TrackLogoIconPreviews

### TrackerSearch

TrackerSearchPreviews

### TrackInfoDialogHome

Screenshot 2023-10-14 at 11 30 20 PM

### TrackStatusSelector

TrackStatusSelectorPreviews


app/src/main/java/eu/kanade/presentation/track/TrackInfoDialogSelector.kt
```diff
@@ -218,3 +220,25 @@ private fun BaseSelector(
         },
     )
 }
+
+@ThemePreviews
+@Composable
+internal fun TrackStatusSelectorPreviews() {
```
arkon, line 226: Can you make these `private`?

arkon (changes requested):
Need to run the `ktlintFormat` task.

app/src/main/java/eu/kanade/presentation/track/TrackInfoDialogSelector.kt
CalebMorris: Apologies, I'd originally thought that private doesn't work with these, but it looks like they do. I've updated all the ones specific to this PR

PR #10042, opened 2023-10-21
Decoupled Tracker Interface

CalebMorris:
# What?

Split out `Tracker` into an interface while renaming the original abstract class `Tracker` to `BaseTracker`.

No actual logic or behavior changes.

## Why?

The original `Tracker` was coupled to the IoC field injection which made any kind of pure abstractions such as those needed for presentation previews very difficult. This also creates a clear contract for the Tracker to make future changes simpler.

## Note

This is a split out from #10022 to break up the PR into more easily reviewable chunks


arkon (approved):
Thanks; the tracking stuff definitely needs some love 😅

PR #10051, opened 2023-10-22
Detached permission request from DiskUtil

CalebMorris:
# What did I change?

Detached permission request from `DiskUtil`

## Why?

Code cleanup.

Being an extension on the `DiskUtil` couples to a class at a different abstraction layer without really needing to. Created `PermissionRequestHelper` as a place to put similar requests if needed in the future.

## Testing

Confirmed that permissions request shows up as expected when opening backup screen


arkon (commented):
FYI, I do have plans on unifying the storage locations (i.e. for downloads, automatic backups, and local source) for use with scoped storage, so this is going to be irrelevant at that point.

PR #10068, opened 2023-10-26
Continue Reading Button size/style normalization

e-shl:
Goal of the pull is to make style the ContinueReadingButton less knockout-annoying (balanced with the surrounding elements)

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/library/components/CommonMangaItem.kt
+++ b/app/src/main/java/eu/kanade/presentation/library/components/CommonMangaItem.kt
@@ -127,13 +127,14 @@ private fun BoxScope.CoverTextOverlay(
             .align(Alignment.BottomCenter),
     )
     Row(
-        modifier = Modifier.align(Alignment.BottomStart),
+        modifier = Modifier.align(Alignment.BottomStart)
+            .height(ContinueReadingButtonSize * 2),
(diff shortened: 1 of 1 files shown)


CarlosEsco:
The fab is this way because it follows material 3 specs

arkon:
Not to mention that the sizing in the specs is intended such that it's actually easy to tap.

If people think it's out of place, they could just disable it entirely; I'm not sure tweaking it really makes it any less out of place.

e-shl:
Alternative size 24

arkon:
> Alternative size 24

It's better than the original proposal since it's at least consistent with the FAB in the manga details screen, but it feels too small to me relative to the 2 lines of text in the compact grid layout.

e-shl:
> feels too small to me relative to the 2 lines of text in the compact grid layout.



it seems to have achieved the effect I wanted, but the one-liners go down

e-shl:
fixed it

arkon:
I didn't actually mean to move the single-line text up; that would look weird normally without the buttons.

e-shl:
> I didn't actually mean to move the single-line text up; that would look weird normally without the buttons.

I perceive this as alignment of the title relative to entire library

arkon:
I'll just slightly shrink it a bit (0696e4bce0e61c69e76bc8e02b49ad40ecf53996).

PR #10079, opened 2023-10-28
Add icons for author and artist in MangaInfoHeader

e-shl:
Separation icon of author and artist


e-shl:
@arkon what hides these icons in tablet mode?

arkon:
Ah good catch, you need to update `MangaAndSourceTitlesLarge` too.

PR #10095, opened 2023-10-31
Text on tabs Overflow Ellipsis

e-shl:
Text on tabs Overflow Ellipsis analog navigation bar

fixes this 

I do not know if there are examples of translations from the basic tachi, but I prefer to fix it for the future in primary source code.

UPD


arkon:
I'm confused by the PR title and description. Are they actually corresponding to the code change?

presentation-core/src/main/java/tachiyomi/presentation/core/components/material/Tabs.kt
```diff
@@ -62,7 +63,11 @@ fun TabText(text: String, badgeCount: Int? = null) {
     Row(
         verticalAlignment = Alignment.CenterVertically,
     ) {
-        Text(text = text)
+        Text(
+            text = text,
+            maxLines = 1,
```
arkon, line 68: Labels are allowed to have 2 lines: https://m3.material.io/components/tabs/guidelines#e8be085b-7b39-4dbe-b8ec-cfa15c8d809a

That seems reasonable given that longer text on smaller screens will always end up getting cut off to a very shortened version if we limit it to 1 line.

e-shl:
> I'm confused by the PR title and description. Are they actually corresponding to the code change?

Replaced with the main meaning.

I wanted to create something in between scrolltab and tabrow. So that tabs's width of rests first on the text, then it fills up the space. 
But I need to more understand layout android.
The current solution is simplest at moment for minimizing labels situations.

presentation-core/src/main/java/tachiyomi/presentation/core/components/material/Tabs.kt
arkon: I'm still confused as to why you're showing screenshots of the bottom navigation bar; those aren't tabs and are supposed to have shorter labels in the first place.

e-shl: > I'm still confused as to why you're showing screenshots of the bottom navigation bar; those aren't tabs and are supposed to have shorter labels in the first place.

Because for me the essence is the same - it's navigation, even without icons. And the long text goes beyond the same.

+ in basic tachi, navbar stands out more clearly because there are already many tabs here

PR #10101, opened 2023-11-04
DownloadStatsScreen DB changes and base stat

semenvav:
Part of the #9737 
As @arkon suggested I split the PR for review.
This PR contains only DB for recording downloads/deletions operations and a simple screen with totally basic info


data/src/main/sqldelight/tachiyomi/data/download_stat.sq
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE download_stat(
+    _id INTEGER NOT NULL PRIMARY KEY,
+    manga_id INTEGER,
+    date INTEGER NOT NULL DEFAULT 0,
```
arkon, line 4: What's the date for?

data/src/main/sqldelight/tachiyomi/data/download_stat.sq
```diff
@@ -0,0 +1,18 @@
...
+    size INTEGER NOT NULL,
+    units INTEGER NOT NULL,
```
arkon, line 6: Why units? Wouldn't it just always be bytes?

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -406,6 +409,22 @@ class Downloader(
 
             DiskUtil.createNoMediaFile(tmpDir, context)
 
+            addDownloadStatOperation.await(
```
arkon, line 412: This ideally shouldn't be blocking download progress. Same for the things in `DownloadManager`.

data/src/main/sqldelight/tachiyomi/data/download_stat.sq
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE download_stat(
```
arkon, line 1: I'm hesitant to even persist this data, why not just compute lazily when needed?

arkon (commented):
I ended up rebasing/cleaning up this PR for you, but I haven't looked too deeply into the actual statistics reading/writing parts.

data/src/main/sqldelight/tachiyomi/data/download_stat.sq
semenvav: “units” here doesn't mean memory units, but the number of chapters deleted in one operation, I probably named this field incorrectly. Memory is always in bytes.

data/src/main/sqldelight/tachiyomi/data/download_stat.sq
semenvav: Yes, the content in this PR doesn't make much sense for this. Saving the history in this form is necessary for its further display in the manga on the next tab, and on the graph.
image
image
I planned to do all this in 2 more separate PRs. I wanted to make a kind of analog of history tab and translate it into a graphical form. The problem is that if I include these functions here, I will end up with a complete previous pr

Issue #9157, opened 2023-02-27
Plural formulas are wrong for some languages in Weblate


Ereza:
### Steps to reproduce

1. Open Hosted Weblate.
2. Open Tachiyomi 0.x project.
3. Open the Catalan translation.
4. There are 14 strings with "Failing check: Missing plural".
5. If you open the list, all of them are missing the plural form "Others". However, these have been translated several times, and they keep appearing as "Missing plural" every time a new string is added to the project.

### Expected behavior

There should be 0 strings with "Failing check: Missing plural" after translating them.
The plural formula for Catalan (and other languages like Spanish) should be `n != 1`.

### Actual behavior

There are 14 strings with "Failing check: Missing plural" even after translating them. They appear as missing plural every time a new Tachiyomi string is available for translation.
The plural formula for Catalan (and other languages like Spanish) is wrong: `(n == 1) ? 0 : ((n != 0 && n % 1000000 == 0) ? 1 : 2)`.

### Tachiyomi version

N/A

### Android version

N/A

### Device

N/A

### Other details

Every time the Weblate project for Tachiyomi is updated, there are some strings with plurals that keep popping up as untranslated. For example, `lock_after_mins`. I maintain the Catalan translation and the plural formula is wrong. It should be just like in English (`n != 1`), but it is `(n == 1) ? 0 : ((n != 0 && n % 1000000 == 0) ? 1 : 2)`.

My guess is that this has to do with some change made some time ago to use different strings for things exceeding one million, but it clearly generates some issues when syncing between the GitHub repository and Weblate. For example, in Catalan I have translated the `lock_after_mins` "other" plural string many times, but it keeps disappearing each time there is a new string.

I have checked other languages and at least Spanish is also affected and has the same wrong formula. English has the correct formula.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have gone through the [FAQ](https://tachiyomi.org/help/faq/) and [troubleshooting guide](https://tachiyomi.org/help/guides/troubleshooting/).
- [X] I have updated the app to version **[0.14.5](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
[...]

PR #10133, opened 2023-11-13
Fix Catalan plurals manually

Ereza:
This PR manually changes some Catalan language strings that keep popping up as «Failing checks» on Weblate due to issues in how Weblate handles plural forms.

As I previously reported on #9157 (although back then I wrongly assumed that it was a config error), there are several pluralized strings that, when translated on Weblate, are never propagated to Tachiyomi's repository. Each time a new strings sync lands on Weblate, all those strings lose one of the translated plurals and are flagged as «Failing checks». This PR changes those strings manually by adding the missing plural form.

Seems related to #2714 which was fixed in the same way.


AntsyLich:
Can you also open issue on weblate [repo](https://github.com/WeblateOrg/weblate).

arkon:
I don't know if you ever reported the issue to Weblate or not, but it seems like the plural formulas are still the same. I'll merge this, but I'm not really sure if it's just going to get overwritten again by Weblate...

Ereza:
@arkon Thanks. The plural forms are OK (they were changed at some point in Unicode's CLDR from one/other to one/many/other and I wasn't aware of that when opening #9157), but the issue now is that the «many» form is always lost when the strings are changed in Weblate. I was hoping that adding the missing forms manually would fix the Weblate issue, but it hasn't. I just opened WeblateOrg/weblate#10398 hoping that someone will take a look into this. Thanks again for your help.

PR #10154, opened 2023-11-19
5 is good

AntsyLich:
Five is inherently good due to its ubiquity and significance across various disciplines. In mathematics, it forms the basis of many calculations and patterns. In nature, it resonates in the five senses and elements. Culturally, it symbolizes balance and harmony, making it a universally positive and essential number.


arkon:
Nah dog, backups take up storage space so 4's good enough for redundancy purposes.

AntsyLich:
I'm forking up smh

PR #10211, opened 2023-12-03
support ignore download error

PinkD:
fix https://github.com/tachiyomiorg/tachiyomi/issues/3957

### Images
| Settings | Donload error | Download finish |
| ------- | ------- | ------- |
| image | image | image |


arkon:
That, or it should be an explicit action taken in the download queue per failed chapter to acknowledge which pages are missing and continue anyway.

PR #10291, opened 2024-01-06
fix Can't scroll down to the bottom of the webtoon on the last chapter

### LINKED ISSUE #3052: [Bug] Can't scroll down to the bottom of the webtoon on the last chapter
### Device information
* Tachiyomi version: R2252
* Android version: Android 9
* Device: Oneplus 5T

## Steps to reproduce
1. Open the last chapter of a webtoon (so webtoon reader mode)
2. Zoom out
3. Now zoom in again
4. Scroll down untill you can't anymore
5. Now you can't see the last part of the chapter and you won't see 'There's no next chapter'

### Expected behavior
Seeing the last part of the chapter

### Actual behavior
Not seeing it untill you close and open the reader

Naputt1:
fix #3052
I think the problem come from scaling, where the layout height is updated. However, when double-clicking, the layout height wasn't reset to original height.


arkon:
Thanks!

PR #10375, opened 2024-01-08
Use transaction on restore to go brr.

KaiserBh:
By wrapping the restoration steps within a single database transaction, we've achieved a considerable reduction in the overall time required to restore large dataset of manga entries into both fresh and non-fresh databases.

# Changes
- Wrapped the entire `restoreManga` function within a transaction block using `handler.await(inTransaction = true)`.

# Performance Gains
- Fresh DB restores have shown a 57.31% speed increase, reducing the time from 14 minutes to approximately 6 minutes.

- Non-fresh DB restores have shown an even more impressive 62.57% speed increase, slashing the time from 27 minutes to around 10 minutes.

# Scope of Optimization
- The optimization was tested on a dataset containing approximately 3,053 manga entries, each with a varying number of chapters.

edit: added images.

### With transaction (fresh db)
image

### Without transaction (fresh db)
image

### With transaction (not fresh db)
image

### Without transaction (not fresh db)
image

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt
@@ -57,22 +57,24 @@ class MangaRestorer(
         backupManga: BackupManga,
         backupCategories: List<BackupCategory>,
     ) {
-        val dbManga = findExistingManga(backupManga)
-        val manga = backupManga.getMangaImpl()
-        val restoredManga = if (dbManga == null) {
-            restoreNewManga(manga)
-        } else {
-            restoreExistingManga(manga, dbManga)
-        }
+        handler.await(inTransaction = true) {
+            val dbManga = findExistingManga(backupManga)
+            val manga = backupManga.getMangaImpl()
+            val restoredManga = if (dbManga == null) {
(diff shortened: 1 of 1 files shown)


arkon:
What happens if you try to perform database interactions in other parts of the app while the restore's in progress?

KaiserBh:
> What happens if you try to perform database interactions in other parts of the app while the restore's in progress?

It still works fine, no crashes, but that being said I am going to test it a bit more and see if everything is working as intended while the restore is happening.

KaiserBh:
What if we only use transaction on a fresh db would that be better? and leave the rest as it is. 
@arkon

KaiserBh:
> The user can still browse which also uses the db

Yea makes sense but wait a second, 

"Update:
Okay so found some issue, when it's in transaction if we mark a whole manga as a read it will disappear."

I think this bug has nothing to do with the transaction, because it's happening even without using the transaction. So while the backup is restoring and if we mark a manga as a read it will basically disappear.

KaiserBh:
Quick update this was tested on a real device, and this is for S20 Ultra.

## With transaction


## Without transaction
TBD (But it should be around 86 minutes and 54 seconds based on the previous data, which was 57.31% speed increase)

KaiserBh:
> For the same set of data I'm assuming?

Yes, that's correct. 3053 manga, I believe.

Issue #10004, opened 2023-10-11
Viewer navigation taps have strange effects or miss completely when window is not at the top left


ackitalk:
### Steps to reproduce

1. In some sort of multi window mode (dex, WSA, pop out or split screen view on your phone or tablet), open tachiyomi and snap its window to the right side
2. Open a chapter and try to use the navigation tap zones

### Expected behavior

The navigation tap zones should work

### Actual behavior

The navigation tap zones behave strangely or not at all

### Tachiyomi version

0.14.6

### Android version

Android 13

### Device

Galaxy Fold 3

### Other details

It looks like getrawx and getrawy in [here](https://github.com/tachiyomiorg/tachiyomi/blob/79b37df647049df90921cae1fabdca2312404784/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/pager/PagerViewer.kt#L105) (?) returns screen coordinates and not window coordinates, which makes the coordinates inaccurate if tachiyomi's window is not located at 0,0

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/tachiyomi-extensions/issues/new/choose).
- [X] I have gone through the [FAQ](https://tachiyomi.org/docs/faq/general) and [troubleshooting guide](https://tachiyomi.org/docs/guides/troubleshooting/).
- [X] I have updated the app to version **[0.14.6](https://github.com/tachiyomiorg/tachiyomi/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

PR #10378, opened 2024-01-08
Fixed tap controls not working when zoomed in

Paloys:
Changed the tap event to reflect absolute position of the `MotionEvent` instead of position relative to the `WebtoonRecyclerView`/`Pager`
Tested it with several resolutions, it seems to be working.

Previously, a problem existed in which, when one zoomed in and clicked on a specific section of the screen (e.g., the center), the Navigation's response was determined by the click's position in relation to the view. This behavior was contradictory because, if zoomed in on the side of the view and clicking the center, it would not summon the menu (assuming `LNavigation` was selected) but would instead cause vertical scrolling, which is counter-intuitive.
With these changes, any area of the screen now results in the intended action specific to that region.

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/components/TabbedScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/components/TabbedScreen.kt
@@ -101,6 +101,6 @@ data class TabContent(
     val titleRes: StringResource,
     val badgeNumber: Int? = null,
     val searchEnabled: Boolean = false,
-    val actions: ImmutableList<AppBar.Action> = persistentListOf(),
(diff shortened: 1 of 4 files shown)


arkon:
This actually contradicts the fix for #10004 :\

Paloys:
Hmmm, indeed...
Is there a way we could reach some compromise between those two issues or should I just close the pull request?

Paloys:
Although after thinking about it, there's probably a way to make it work in both situation by checking if the app is taking the full screen or not. I'll try to look into it in the following days.

arkon:
Whoops I messed that up. Merged as c65a9aecf5ed228550fd994e62a16fc2300c344d

PR #10385, opened 2024-01-09
Fixed dev UI preview

CalebMorris:
The TachiyomiTheme has a dependency-injection construct that doesn't have any registered elements at the time of rendering previews, so I've changed the preview function to use a preview-specific version of the theme that has declarative configuration with defaults over dependency injection.

For `AppThemesListPreview`, where we need to demonstrate multiple constructs with the production theme, I instead register a configuration in the DI.


CalebMorris:
Fixed build failures - 2 unused imports left over

